Issue #18, opened 2023-01-24
Parameterize Devices


callumforrester:
@DominicOram and @noemifrisina

Relates to https://github.com/DiamondLightSource/python-dodal/issues/6, https://github.com/DiamondLightSource/python-dodal/issues/3 and https://github.com/DiamondLightSource/python-artemis/issues/200

It seems, to avoid verbosity, we want module level devices so we can say
```python
from i03 import dcm
```
and immediately use `dcm` as oppose to having a `make_dcm()` function that only instantiates the object when called.

However, we may want the device to do slightly different things, depending on context, when we import it, for example:
* Use a different PV prefix, for example, if it's a simulated beamline vs. real (https://github.com/DiamondLightSource/python-dodal/issues/6) or if there are macros in the IOCs themselves that change the prefix (e.g. for the soft motors available in Diamond).
* Test that the devices connect on import (in Ophyd v1, this is done by calling `wait_for_connection()`) so that it doesn't unexpectedly cause an error when first used
* Explicitly _not_ test that devices connect on import, for example because of long startup times, or because only a subset of devices are active on the beamline at any one time
* Run custom checks on the devices (e.g. `detector.hdf5.warmup()`) or not, or sometimes, as above

If we take parameterization as a possible solution for these use cases, we should think about how to achieve it. We want to maximize granularity and flexibility while minimizing boilerplate. The two are likely to work in mutual contravention so we would need to find a good compromise. Also, depending on exactly how much flexibility we feel we need, we should keep in mind the possibility that the `make_dcm()` function above may actually be the least worst option.

callumforrester:
One immediate thought I had (and I'm not sure I like it) was using environment variables. Something like:

```python
# <beamline>.py
import socket

from pydantic import BaseSettings

from .devices.simpledet import SimpleAreaDetector
from .devices.simstage import SimStage


class Settings(BaseSettings):
   # $HOSTNAME
    hostname: str = socket.gethostname().split(".")[0]
   # $CHECK_CONNECTION
    check_connection: bool = True
   # $WARMUP_DETECTOR
    warmup_detector: bool = True


stage = SimStage(name="sim_stage", prefix=f"{Settings.hostname}-MO-SIM-01:")
det = SimpleAreaDetector(name="adsim", prefix=f"{Settings.hostname}-AD-SIM-01:")

if Settings.check_connection:
    stage.wait_for_connection()
    det.wait_for_connection()

if Settings.warmup_detector:
[...]

Issue #3, opened 2022-08-30
Add a system tests that confirms all of i03/s03 is connected


DominicOram:
As a developer it would be good to know ASAP if there have been changes to i03 that cause my ophyd devices to no longer be able to connect. As such it would be good to have a system test to check this. The test will only be able to be run on DLS network and getting it run in CI will be hard but just having it would be a good start.

dperl-dls:
These have existed for some time now

PR #21, opened 2023-02-16
Add configs for P45 and simulated AreaDetector

callumforrester:
PR based on discussion around https://github.com/DiamondLightSource/dodal/issues/18 and similar issues. It may be a good idea to split it into 2-3 PRs. Mainly interested in opinions for now.

Changes:
* Add beamline module for P45
* Add beamline module for DLS simulated AreaDetector
* Add Ophyd devices for simulated AreaDetector and Aravis GigE camera
* Add utils module for instantiating all devices in a beamline module

### DIFF
--- a/src/dodal/devices/areadetector/adaravis.py
+++ b/src/dodal/devices/areadetector/adaravis.py
@@ -4,13 +4,13 @@ from ophyd import DetectorBase
 from ophyd.areadetector.base import ADComponent as Cpt
 from ophyd.areadetector.detectors import DetectorBase
 
-from .adutils import Hdf5Writer, SyncrhonisedAdDriverBase
+from .adutils import Hdf5Writer, SynchronisedAdDriverBase
 
 _ACQUIRE_BUFFER_PERIOD = 0.2
 
 
 class AdAravisDetector(SingleTriggerV33, DetectorBase):
-    cam: SyncrhonisedAdDriverBase = Cpt(SyncrhonisedAdDriverBase, suffix="DET:")
+    cam: SynchronisedAdDriverBase = Cpt(SynchronisedAdDriverBase, suffix="DET:")
     hdf: Hdf5Writer = Cpt(
         Hdf5Writer,
         suffix="HDF5:",
--- a/src/dodal/devices/areadetector/adaravis.py
+++ b/src/dodal/devices/areadetector/adaravis.py
@@ -1,0 +1,51 @@
+from nslsii.ad33 import SingleTriggerV33
+from ophyd import Component as Cpt
+from ophyd import DetectorBase
+from ophyd.areadetector.base import ADComponent as Cpt
+from ophyd.areadetector.detectors import DetectorBase
+
+from .adutils import Hdf5Writer, SyncrhonisedAdDriverBase
+
+_ACQUIRE_BUFFER_PERIOD = 0.2
+
+
+class AdAravisDetector(SingleTriggerV33, DetectorBase):
+    cam: SyncrhonisedAdDriverBase = Cpt(SyncrhonisedAdDriverBase, suffix="DET:")
+    hdf: Hdf5Writer = Cpt(
+        Hdf5Writer,
+        suffix="HDF5:",
+        root="",
+        write_path_template="",
+    )
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.hdf.kind = "normal"
+
+        self.stage_sigs = {
+            # Get stage to wire up the plugins
+            self.hdf.nd_array_port: self.cam.port_name.get(),
+            # Reset array counter on stage
+            self.cam.array_counter: 0,
+            # Set image mode to multiple on stage so we have the option, can still
+            # set num_images to 1
+            self.cam.image_mode: "Multiple",
+            # For now, this Ophyd device does not support hardware
+            # triggered scanning, disable on stage
+            self.cam.trigger_mode: "Off",
+            **self.stage_sigs,  # type: ignore
+        }
+
+    def stage(self, *args, **kwargs):
+        # We have to manually set the acquire period bcause the EPICS driver
+        # doesn't do it for us. If acquire time is a staged signal, we use the
+        # stage value to calculate the acquire period, otherwise we perform
+        # a caget and use the current acquire time.
+        if self.cam.acquire_time in self.stage_sigs:
+            acquire_time = self.stage_sigs[self.cam.acquire_time]
+        else:
+            acquire_time = self.cam.acquire_time.get()
+        self.stage_sigs[self.cam.acquire_period] = acquire_time + _ACQUIRE_BUFFER_PERIOD
+
+        # Now calling the super method should set the acquire period
+        super(AdAravisDetector, self).stage(*args, **kwargs)
--- a/src/dodal/devices/areadetector/adaravis.py
+++ b/src/dodal/devices/areadetector/adaravis.py
@@ -1,10 +1,9 @@
-from nslsii.ad33 import SingleTriggerV33
 from ophyd import Component as Cpt
 from ophyd import DetectorBase
 from ophyd.areadetector.base import ADComponent as Cpt
 from ophyd.areadetector.detectors import DetectorBase
 
-from .adutils import Hdf5Writer, SynchronisedAdDriverBase
+from .adutils import Hdf5Writer, SingleTriggerV33, SynchronisedAdDriverBase
 
 _ACQUIRE_BUFFER_PERIOD = 0.2
 
--- a/src/dodal/devices/areadetector/adutils.py
+++ b/src/dodal/devices/areadetector/adutils.py
@@ -1,20 +1,34 @@
-from nslsii.ad33 import CamV33Mixin
+from ophyd import Component as Cpt
+from ophyd import EpicsSignal, EpicsSignalRO
 from ophyd.areadetector.cam import AreaDetectorCam
 from ophyd.areadetector.filestore_mixins import FileStoreHDF5, FileStoreIterativeWrite
 from ophyd.areadetector.plugins import HDF5Plugin
 
 
-class SynchronisedAdDriverBase(AreaDetectorCam, CamV33Mixin):
+class SynchronisedAdDriverBase(AreaDetectorCam):
     """
     Base Ophyd device to control an AreaDetector driver and
     syncrhonise it on other AreaDetector plugins, even non-blocking ones.
     """
+
+    adcore_version = Cpt(EpicsSignalRO, "ADCoreVersion_RBV", string=True, kind="config")
+    driver_version = Cpt(EpicsSignalRO, "DriverVersion_RBV", string=True, kind="config")
+    wait_for_plugins = Cpt(EpicsSignal, "WaitForPlugins", string=True, kind="config")
 
     def stage(self, *args, **kwargs):
         # Makes the detector allow non-blocking AD plugins but makes Ophyd use
         # the AcquireBusy PV to determine when an acquisition is complete
         self.ensure_nonblocking()
         return super().stage(*args, **kwargs)
+
+    def ensure_nonblocking(self):
+        self.stage_sigs["wait_for_plugins"] = "Yes"
+        for c in self.parent.component_names:
+            cpt = getattr(self.parent, c)
+            if cpt is self:
+                continue
+            if hasattr(cpt, "ensure_nonblocking"):
+                cpt.ensure_nonblocking()
 
 
 class Hdf5Writer(HDF5Plugin, FileStoreHDF5, FileStoreIterativeWrite):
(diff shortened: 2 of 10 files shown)


src/dodal/devices/areadetector/adaravis.py
```diff
@@ -0,0 +1,51 @@
...
+from ophyd import DetectorBase
+from ophyd.areadetector.base import ADComponent as Cpt
+from ophyd.areadetector.detectors import DetectorBase
+
+from .adutils import Hdf5Writer, SyncrhonisedAdDriverBase
+
+_ACQUIRE_BUFFER_PERIOD = 0.2
+
+
+class AdAravisDetector(SingleTriggerV33, DetectorBase):
```
callumforrester, line 12: TODO: Consider better ways to share functionality with `AdSimDetector` without resorting to mixins.

src/dodal/devices/areadetector/adutils.py
```diff
@@ -0,0 +1,28 @@
+from nslsii.ad33 import CamV33Mixin
```
callumforrester, line 1: `CamV33Mixin` is very useful for synchronising the AD plugins, but it is not ideal to depend on the `nslsii` package, which brings in a number of dependencies we don't need. It may be worth broaching the idea of moving that functionality into ophyd.

src/dodal/util.py
```diff
@@ -0,0 +1,16 @@
+from importlib import import_module
+from types import ModuleType
+from typing import Any, Callable, Dict, Iterable, Union
+
+
+def make_all_devices(module: Union[str, ModuleType, None] = None) -> Dict[str, Any]:
```
callumforrester, line 6: This is just an idea but I think it would be useful to have _some_ way of creating all the devices in a beamline module. The approach here is just to take all functions in a module whose names begin with `make_` (inspired by various test frameworks). An alternative is some sort of `@device` decorator to mark functions to be called. Either way would also have to decide whether to allow functions that take non-default arguments and what to do with them if so.

src/dodal/p45.py
```diff
@@ -0,0 +1,73 @@
...
+from ophyd import EpicsMotor, MotorBundle
+from ophyd.areadetector.base import ADComponent as Cpt
+from pydantic import BaseSettings
+
+from dodal.devices.areadetector import AdAravisDetector
+
+
+# Settings can be customized via environment variables
+class Settings(BaseSettings):
+    pv_prefix: str = "BL45P"
```
DominicOram, line 11: Nit: To override this we would need to have the `PV_PREFIX` environment variable set, right? I was thinking we could instead look at the `BEAMLINE` environment variable, which we already have on beamlines, and create the prefix based on that or use a simulator if it doesn't exist. e.g. like https://github.com/DiamondLightSource/python-artemis/pull/522/files#diff-2ee96bef30cad81682a38f7f9c30ffb387bc82f29e5f09393b7a3b5f2c88af7e. This would mean it'll work seamlessly on beamlines and dev machines as they are currently

src/dodal/p45.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class TomoStageWithStretchAndSkew(MotorBundle):
+    """
+    Grouping of motors for the P45 tomography stage
+    """
+
+    x: EpicsMotor = Cpt(EpicsMotor, "X")
+    y: SampleY = Cpt(SampleY, "")
+    theta: SampleTheta = Cpt(SampleTheta, "")
```
DominicOram, line 46: Nit: Why not do these as ophyd v2 devices? I believe motors are basically done?

DominicOram: Nit: Given this file is basically the interface to the user writing plans I was thinking it would be better to just have the `make_` functions here. It avoids a user accidentally writing:
```
from p45 import SampleY
my_stage = SampleY()
...
```
and wondering why everything is broken

src/dodal/p45.py
```diff
@@ -0,0 +1,73 @@
...
+def make_choppers(name: str = "chopper") -> Choppers:
+    return Choppers(name=name, prefix=f"{_settings.pv_prefix}-MO-CHOP-01:")
+
+
+def make_det(name: str = "det") -> AdAravisDetector:
+    return AdAravisDetector(name=name, prefix=f"{_settings.pv_prefix}-EA-MAP-01:")
+
+
+def make_diff(name: str = "diff") -> AdAravisDetector:
+    return AdAravisDetector(name=name, prefix=f"{_settings.pv_prefix}-EA-DIFF-01:")
```
DominicOram, line 73: Nit: I'm not that adverse to having `make_` precede all of these but I wonder if users would rather just be able to call `det()` to get the detector?

src/dodal/p45.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+
+    x: EpicsMotor = Cpt(EpicsMotor, "ENDAT")
+    y: EpicsMotor = Cpt(EpicsMotor, "BISS")
+
+
+def make_sample(name: str = "sample") -> TomoStageWithStretchAndSkew:
+    return TomoStageWithStretchAndSkew(
+        name=name, prefix=f"{_settings.pv_prefix}-MO-STAGE-01:"
+    )
```
DominicOram, line 61: Should: `def make_sample_stage(name: str = "sample_stage")` seems clearer?

src/dodal/util.py
DominicOram: I agree that something like that is nice, makes https://github.com/DiamondLightSource/dodal/issues/3 easier too. As above, I'm not sure if the `make_` convention is 100% correct anyway though. Could we inspect if the return type is a device and use that?

src/dodal/devices/areadetector/adutils.py
```diff
@@ -0,0 +1,28 @@
...
+from ophyd.areadetector.cam import AreaDetectorCam
+from ophyd.areadetector.filestore_mixins import FileStoreHDF5, FileStoreIterativeWrite
+from ophyd.areadetector.plugins import HDF5Plugin
+
+
+class SyncrhonisedAdDriverBase(AreaDetectorCam, CamV33Mixin):
+    """
+    Base Ophyd device to control an AreaDetector driver and
+    syncrhonise it on other AreaDetector plugins, even non-blocking ones.
```
DominicOram, line 10: Should: Spelling, `synchronise`

DominicOram (approved):
Minor comments in code, I don't think any are blocking merge though

src/dodal/p45.py
callumforrester: I did think about that, my main issue was that `$BEAMLINE` takes the form `Ixx` where the prefix is `BLxxI`. As you say, we could write some logic to convert between the two but then we bake in the assumption of those naming conventions. @coretl, how do controls feel, having lived with those conventions for 20 years? Is is something we'd be happy to keep going into the D-II era or should we be thinking about a more flexible approach?

src/dodal/p45.py
callumforrester: Makes sense, but then where should be put very-beamline-specific devices? `dodal/devices/<beamline>.py`?

src/dodal/util.py
callumforrester: Ooh, that's an idea! It does require the user to remember to include a return type in the function signature. Do we prefer that to the generator? I guess it's not _added_ boilerplate per-say as we generally expect annotated code in these repos. Some boilerplate is required for all 3 options so relative pros are:

* `make_`: similar to pytest et al. so may be familiar to the user
* Inspect return type: the user was probably going to include that anyway
* Decorator: explicit, makes it very obvious to the user which functions will get picked up

I have no strong attachment to any of these, I'm all ears for opinions.

src/dodal/p45.py
coretl: There are some annoying inconsistencies about converting from `Ixx` to `BLxxI` especially for some branchlines, so I think the pv_prefix is safer...

src/dodal/devices/areadetector/adutils.py
coretl: I'm hesitant to add anything to `ophyd` as we're about to create all the v2 devices in `ophyd-epics-devices`...

src/dodal/p45.py
callumforrester: RE: Ophyd v2, probably a good idea to move to it sooner rather than later, but we want to prove the initial stack with Ophyd v1 first. @RAYemelyanova has a ticket to make some Ophyd v2 devices for P45 so those should go in here whenever they are ready (not urgent, Rose, just thought it was worth letting you know).

src/dodal/util.py
coretl: For v2, we could put all this in the `__init__` function of the compound devices

```python
class Choppers(StandardReadable):
    """
    Grouping for the P45 chopper motors
    """
    def __init__(self, name="choppers", prefix=f"{_settings.pv_prefix}-MO-CHOP-01:"):
        self.x = EpicsMotor("ENDAT")
        self.y = EpicsMotor("BISS")
```

Although I'm not convinced about the `make_` functions in general, in my experiences beamlines all have subtle differences, and I prefer the readability of instantiating all the devices once per beamline to the DRY of some `make_` functions:

```python
with DeviceCollector(set_name=True):
    choppers = p45.Choppers()
    det = p45.AdAravisDetector()
```

src/dodal/devices/areadetector/adutils.py
callumforrester: Reasonable, it's not a big class, I wonder how they'd feel about us duplicating it in Dodal

src/dodal/util.py
callumforrester: Ultimately, this approach comes from offline discussions in which we agreed we were against imports with side effects and nothing should be created just by importing the module. Is it possible to make an approach like this work with that constraint?

callumforrester: (wrote a version that inspects the return type, just to see how it looks)

coretl: I'm not suggesting making a module with side effects, I'm suggesting putting all the device creation in a beamline startup script, and if there is duplication between beamlines then so be it. However I am aware this comes from a non-MX experience where every beamline is a bit different, so maybe the MX use-case is strong enough for `make_` functions. In this case though, could you just have an MX specific `make_` function rather than requiring it as an approach for everything?

src/dodal/p45.py
DominicOram: I don't really like that this means for a developer I need to make sure to have `PV_PREFIX=S03` in my environment, then if I want to switch working to i04 I would need to switch to `PV_PREFIX=S04` before starting. It seems we need an environment variable that defines if you are on a live machine and reverts to the sim PVs if not. We could add `PV_PREFIX` to the beamline machines?

src/dodal/p45.py
DominicOram: I think `dodal/devices/<beamline>.py` is fine, yh

src/dodal/util.py
DominicOram: I don't think MX are completely tied to the factory functions format. I'm not sure I can picture the alternative you're suggesting though, maybe worth an offline chat?

src/dodal/devices/areadetector/adutils.py
danielballan: I think it makes sense to copy it for now and then refactor to use a shared version into `ophyd-epics-devices` when we are ready. The goal with `nslsii` is to hold devices (and other things) that are not useful outside NSLS2 or that are not yet at the standard (tests, etc.) of ophyd, so in general our intent is to hoist things like this upstream when they are broadly useful, rather than introduce a dependency on `nslsii`.

src/dodal/p45.py
callumforrester: Hmm, my first thought was that the adsim case (hostname based) and the s03 case (static prefix based) are different and require different solutions.

```python
# adsim
class Settings(BaseSettings):
    pv_prefix: str = hostname

_settings = Settings()

# i03
class Settings(BaseSettings):
    simulation: bool = False

_settings = Settings()
_pv_prefix = "BL03S" if _settings.simulation else "BL03I"
```

However, I also remember we've previously discussed the possibility of prefixing the S03 PVs with hostnames or other differentiators so that multiple instances can be run on the same network, do you have any more thoughts in that direction?

coretl: > There are some annoying inconsistencies about converting from `Ixx` to `BLxxI` especially for some branchlines, so I think the pv_prefix is safer...

Although this might be a sensible thing to do if we can override the PV prefixes with some other mechanism

callumforrester:
@DominicOram Have fixed the build issues upstream. Shall we continue the factory format discussion offline? Also shall we merge this in the interim or leave it until decided?

DominicOram:
> @DominicOram Have fixed the build issues upstream. Shall we continue the factory format discussion offline? Also shall we merge this in the interim or leave it until decided?

Yes, let's get this merged

PR #23, opened 2023-03-08
Tidy up util module(s) and tests

callumforrester:
Changes:
* Unify `util` and `utils` modules into a single file (`utils.py`)
* Add tests for the functions formerly in util module
* Make use of beamline prefix code in util module in P45
* Correct wrongly-named device in P45
* Remove pydantic as a dependency, it is no longer used
* Fix scatterguard test incompatibility with mypy 1.1.1


DominicOram (approved):
Great, thank you!

PR #24, opened 2023-03-09
Establish dodal as primary project name over python-dodal

callumforrester:
Rename instances of python-dodal to dodal in docs and config following renaming of repo. This should ensure smooth publication to pypi under the correct name.


DominicOram (approved):
LGTM, confirmed no more incidents of `python-dodal` remain

PR #26, opened 2023-03-16
Add a base class for experiment parameter types to ensure e.g. number of images is always available (Artemis570)

dperl-dls:
Goes with changes in https://github.com/DiamondLightSource/python-artemis/pull/585


DominicOram:
Can you make the PR title a bit more descriptive, dodal will use it for release note generation so should probably use something that makes sense to general DLS devs

PR #28, opened 2023-03-21
Artemis590 - Add a number of triggers param to the eiger parameters

dperl-dls:
Previously, Eiger has been used in only one mode - with a trigger per images. to support rotation scans we need it to also be able to collect many images for each (could be just one) trigger


src/dodal/devices/eiger.py
```diff
@@ -172,10 +172,16 @@ def set_detector_threshold(self, energy: float, tolerance: float = 0.1) -> Statu
...
+        trigger.
+        """
         assert self.detector_params is not None
-        status = self.cam.num_images.set(1)
-        status &= self.cam.num_triggers.set(self.detector_params.num_images)
-        status &= self.odin.file_writer.num_capture.set(self.detector_params.num_images)
+        status = self.cam.num_images.set(self.detector_params.num_images_per_trigger)
+        status &= self.cam.num_triggers.set(self.detector_params.num_triggers)
+        status &= self.odin.file_writer.num_capture.set(
+            self.detector_params.num_triggers
```
DominicOram, line 183: Must: This should presumably be `num_triggers * num_images_per_trigger`?

DominicOram (approved):
Thanks!

PR #31, opened 2023-03-27
Artemis 561 588 - add device management tools to i03.py

### LINKED ISSUE #27: Make beamline modules ensure no dduplicate devices
See https://github.com/DiamondLightSource/python-artemis/issues/588

dperl-dls:
Fixes #27 

Fixes some errors in i03.py and adds some tools to manage device creation, also adds a few unit tests

To test run tests, here and in artemis


src/dodal/i03.py
```diff
@@ -1,43 +1,236 @@
...
+
+def clear_devices():
+    global ACTIVE_DEVICES
+    for d in list(ACTIVE_DEVICES):
+        del ACTIVE_DEVICES[d]
 
-def dcm() -> DCM:
-    return DCM(f"{BeamlinePrefix(BL).beamline_prefix}")
 
+def device_instantiation(
```
DominicOram, line 31: Should: I think some specific tests around this function would be good

DominicOram (changes requested):
@callumforrester what do you think about this? If we're happy with it we should move it somewhere more central and have P45 use the same

src/dodal/i03.py
```diff
@@ -1,43 +1,236 @@
...
-        prefix=f"{BeamlinePrefix(BL).beamline_prefix}",
+
+def dcm(wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False) -> DCM:
+    """Get the i03 DCM device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i03, it will return the existing object.
+    """
+    return device_instantiation(
+        device=DCM,
+        name="dcm",
+        prefix="",
```
DominicOram, line 68: Must: This (and other devices in this file) need the beamline prefix

DominicOram (approved):
I did a minor refactor to make use of the `make_all_devices` util, otherwise looks good, thanks!

PR #36, opened 2023-04-17
Added smoke system tests against s03 and i03

DominicOram:
To test:
* Get a running S03
* Run the new system tests (note the s03 one is run using `pytest -m s03` but the other one must be run with `python tests/system_tests/test_i03_system.py`)

Note there are currently issues with `BL03I-EA-BL-01:CTRL`


dperl-dls (approved):
always good to have more tests

PR #38, opened 2023-04-25
Improve logging capabilities and add associated docs

### LINKED ISSUE #9: Add generic logging function
We should make it as easy as possible for scientists to get their code logging to file/graylog. As such there should be helper functions to initialise this and documentation on how to use it.

DominicOram:
Fixes #9 

Also see https://github.com/DiamondLightSource/python-artemis/issues/569

### DIFF
--- a/src/dodal/log.py
+++ b/src/dodal/log.py
@@ -3,32 +3,59 @@ from __future__ import annotations
 import logging
 from os import environ
 from pathlib import Path
-from typing import List, Optional
+from typing import List, Optional, Tuple, Union
 
 from bluesky.log import config_bluesky_logging
 from bluesky.log import logger as bluesky_logger
+from graypy import GELFTCPHandler
 from ophyd.log import config_ophyd_logging
 from ophyd.log import logger as ophyd_logger
 
 LOGGER = logging.getLogger("Dodal")
-LOGGER.setLevel(logging.DEBUG)  # default logger to log everything
+LOGGER.setLevel(logging.DEBUG)
 ophyd_logger.parent = LOGGER
 bluesky_logger.parent = LOGGER
 
 
+class BeamlineFilter(logging.Filter):
+    beamline: Union[str, None] = environ.get("BEAMLINE")
+
+    def filter(self, record):
+        record.beamline = self.beamline if self.beamline else "dev"
+        return True
+
+
+beamline_filter = BeamlineFilter()
+
+
+def set_beamline(beamline_name: str):
+    """Set the beamline on all subsequent log messages."""
+    beamline_filter.beamline = beamline_name
+
+
 def set_up_logging_handlers(
-    logging_level: Optional[str] = "INFO", dev_mode: bool = False
+    logging_level: Optional[str] = "INFO",
+    dev_mode: bool = False,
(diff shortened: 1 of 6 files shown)


DominicOram:
Test is intermittently flaky, unrelated to this PR

callumforrester:
> Test is intermittently flaky, unrelated to this PR

Is there a ticket relating to the test flakiness?

src/dodal/log.py
```diff
@@ -3,32 +3,59 @@
...
+    dev_mode: bool = False,
+    logging_path: Optional[Path] = None,
 ) -> List[logging.Handler]:
     """Set up the logging level and instances for user chosen level of logging.
-
-    Mode defaults to production and can be switched to dev with the --dev flag on run.
+    Args:
+        logging_level: The level of logs that should be saved to file/graylog. Defaults to INFO.
+        dev_mode: True if in dev mode, will not log to graylog in dev. Defaults to False.
+        logging_path: The location to store log files, if left as None then puts them in the default location.
```
callumforrester, line 45: Would it be reasonable to enable one or the other out of file logging and graylog?
It feels more elegant to have a series of optional routines for setting up various logging handlers and call each of them (or not) according to config/args

src/dodal/log.py
```diff
@@ -3,32 +3,59 @@
...
 
 LOGGER = logging.getLogger("Dodal")
-LOGGER.setLevel(logging.DEBUG)  # default logger to log everything
+LOGGER.setLevel(logging.DEBUG)
 ophyd_logger.parent = LOGGER
 bluesky_logger.parent = LOGGER
 
 
+class BeamlineFilter(logging.Filter):
+    beamline: Union[str, None] = environ.get("BEAMLINE")
```
callumforrester, line 21: Nit:
```suggestion
    beamline: Optional[str] = environ.get("BEAMLINE")
```

callumforrester (changes requested):
Approach makes sense, and can see us using something similar elsewhere. See my comment on `set_up_logging_handlers`

src/dodal/log.py
DominicOram: I can't see a use case for only using one at the moment and I like the idea of a default simple function that gives you broad logging into a standard place. I'm happy to split it up a bit though so if people want to customise it they can call individual functions.

DominicOram:
> Approach makes sense, and can see us using something similar elsewhere.

This is also very similar to the approach taken in zocalo by @rjgildea. Do we think there's value in a generic diamond utils module that contains common things like logging, monitoring?

DominicOram:
> > 
> 
> Is there a ticket relating to the test flakiness?

Yep, was in Artemis, I've moved it across. Probably should go through and move some others too

src/dodal/log.py
callumforrester: Sounds good, a use case I had in mind was running inside a container. Logging to a file requires setting up a mount to somewhere persistent, which is enough effort that you may as well just use graylog.

DominicOram:
This is needed on the beamline so going to make a release off this branch

olliesilvester (approved):

PR #40, opened 2023-04-27
Artemis 191 fix image metadata

noemifrisina:
Redefine end position of grid axis to work with scanspec in artemis


dperl-dls (approved):
great

PR #43, opened 2023-04-28
Artemis 262 move arming out of stage

olliesilvester:
Fixes Artemis issue [#262](https://github.com/DiamondLightSource/python-artemis/issues/262)


src/dodal/devices/eiger.py
```diff
@@ -24,6 +25,10 @@ class EigerTriggerNumber(str, Enum):
 
 
 class EigerDetector(Device):
+    class ArmingSignal(Signal):
+        def set(self, value, *, timeout=None, settle_time=None, **kwargs):
+            return self.parent.stage()
```
DominicOram, line 30: Must: This creates the class but you don't actually have an instance of the signal

src/dodal/devices/eiger.py
```diff
@@ -85,7 +91,7 @@ def stage(self):
         LOGGER.info("Waiting on parameter callbacks")
         status.wait(self.STALE_PARAMS_TIMEOUT)
 
-        self.arm_detector()
+        return self.arm_detector()
```
DominicOram, line 94: Must: With this now being asynchronous it no longer conforms to the interface for `stage` so we should rename it.

src/dodal/devices/eiger.py
```diff
@@ -190,29 +201,46 @@ def set_num_triggers_and_captures(self) -> Status:
...
+        statuses[3] = await_value(self.odin.fan.ready, 1)  # timeout 10?
+        pass
 
     def forward_bit_depth_to_filewriter(self):
         bit_depth = self.bit_depth.get()
         self.odin.file_writer.data_type.put(f"UInt{bit_depth}")
 
-    def arm_detector(self):
+    def arm_detector(self) -> Status:
+        statuses = [Status(1), Status(2), Status(3), Status(4)]
```
DominicOram, line 225: Should: Each function that we pass these to is overwriting the entry so it's a bit confusing having these here. Might be better to start with an empty list then each function can just append to the list.

DominicOram (changes requested):
Thanks! With the requested changes it looks like a good start that we could at least test on the beamline

DominicOram (changes requested):

DominicOram:
@callumforrester and @coretl, you might have some opinions on this. Callum, I know you weren't keen on this approach but I think it makes the plan side quite elegant (https://github.com/DiamondLightSource/python-artemis/pull/631/files). I'm particularly interested in:
* It feels like there might be a cleaner way of writing `wrap_and_do_funcs`?
* Do we think `wrap_and_do_funcs` is worth pushing into core ophyd? Seems generically useful

src/dodal/devices/utils.py
```diff
@@ -12,3 +18,71 @@ def epics_signal_put_wait(pv_name: str, wait: float = 1.0) -> EpicsSignal:
         EpicsSignal: An EpicsSignal that will wait for a callback.
     """
     return Component(EpicsSignal, pv_name, put_complete=True, write_timeout=wait)
+
+
+def wrap_and_do_funcs(unwrapped_funcs, timeout=60):
```
callumforrester, line 23: Type annotations strongly recommended here, makes IDE code completion much nicer
```suggestion
def wrap_and_do_funcs(
    unwrapped_funcs: list[Callable[[], Status]],
    timeout: float = 60.0,
) -> Status:
```

callumforrester, line 23: @DominicOram Have been thinking about a cleaner implementation, how about something like this?
It's not completely right because some of the tests don't like it, but I think recursion may be the way to go:

```python
def wrap_and_do_funcs(
    unwrapped_funcs: list[Callable[[], Status]],
    timeout: float = 60.0,
    previous_status: Optional[Status] = None,
) -> Status:
    func, other_funcs = unwrapped_funcs[-1], unwrapped_funcs[1:-1]

    def wrapper(previous_status: Optional[Status] = None) -> Status:
        if previous_status is not None and previous_status.exception() is not None:
            # If there is an exception, return the status immediately, ends recursion
            # so no further statuses are evaluated
            LOGGER.error(
                f"{previous_status} has failed with error {previous_status.exception()}"
            )
            return previous_status

        LOGGER.info(f"Doing {func.__name__}")
        next_status = func()
        if len(other_funcs) > 0:
            # Queue up the next function to run when this status completes.
            # Do not queue up more functions if there aren't any, ends recursion
            next_status.add_callback(lambda: wrap_and_do_funcs(other_funcs, timeout))
        return next_status

    return wrapper(previous_status)

```

callumforrester (commented):
@DominicOram I can see the elegance of the plan side, but you've actually done two things:
1. Move arming out of stage
2. Remove automatic staging/unstaging from the plan and do it manually
Can we not achieve the same just by doing 2.?

i.e.

```python
yield from bps.stage(eiger)
...
def do_fgs():
    try:
        yield from bps.wait()  # Wait for all moves to complete
        yield from bps.kickoff(fgs_motors)
        yield from bps.complete(fgs_motors, wait=True)
    finally:
        yield from bps.unstage(fgs_composite.eiger)
```

Regardless of that, `wrap_and_do_funcs` does look like a useful addition, see my other comment for musings on implementation.

DominicOram:
> @DominicOram I can see the elegance of the plan side, but you've actually done two things:
> 
>     1. Move arming out of stage
> 
>     2. Remove automatic staging/unstaging from the plan and do it manually
>        Can we not achieve the same just by doing 2.?
> 
> 
> i.e.
> 
> ```python
> yield from bps.stage(eiger)
> ...
> def do_fgs():
>     try:
>         yield from bps.wait()  # Wait for all moves to complete
>         yield from bps.kickoff(fgs_motors)
>         yield from bps.complete(fgs_motors, wait=True)
>     finally:
>         yield from bps.unstage(fgs_composite.eiger)
> ```
> 
> Regardless of that, `wrap_and_do_funcs` does look like a useful addition, see my other comment for musings on implementation.

Can we? I'm not sure if we can if `stage` doesn't return a status? We need something to wait on to make sure the stage has completed?

callumforrester:
In this case, don't we just make stage block on the status?

src/dodal/devices/utils.py
DominicOram: Yes, I feel like we can do it nicer with recursion. I think that might require a bit more thought so let's do a beamline test first @olliesilvester then we'll try and refactor to use recursion

callumforrester:
Ah, didn't see that you were waiting for a group, nevermind

DominicOram:
As discussed offline with @callumforrester, we should tidy up the recursion then put in a PR to ophyd

src/dodal/devices/eiger.py
```diff
@@ -192,43 +208,88 @@ def set_num_triggers_and_captures(self) -> Status:
...
+        if detector_params.use_roi_mode:
+            unwrapped_funcs.append(lambda: self.change_roi_mode(enable=True))
+
+        unwrapped_funcs.extend(
+            [
+                lambda: self.set_detector_threshold(
+                    energy=detector_params.current_energy
+                ),
+                self.set_cam_pvs,
+                self.set_odin_pvs_after_file_writer_set,
```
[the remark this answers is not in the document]
olliesilvester: Thanks I missed this - the function is back in and renamed now

DominicOram:
Testing on the beamline went well, couple of issues:
* https://github.com/DiamondLightSource/dodal/pull/43#discussion_r1200259833 caught us up and caused things to fail
* The logging isn't quite right. The lambda's don't produce useful logs and the logs look to be coming from `utils` I think move the `Doing x` log out of the util function and put it in a couple of the Eiger functions that are being called
* After a plan we get `ophyd status WARNING: Status(obj=None, done=False, success=False) has timed out` I think this may be the unused `arming_status` hanging around?

src/dodal/devices/eiger.py
```diff
@@ -98,74 +98,91 @@ def unstage(self) -> bool:
...
         status_ok = self.odin.check_odin_state()
         self.disable_roi_mode()
         return status_ok
 
+    # this is no longer used TODO clean up
     def enable_roi_mode(self):
         self.change_roi_mode(True)
 
     def disable_roi_mode(self):
         self.change_roi_mode(False)
```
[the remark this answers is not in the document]
DominicOram: Comment still stands

src/dodal/devices/eiger.py
```diff
@@ -98,74 +96,109 @@ def unstage(self) -> bool:
...
+        )
+        status &= self.odin.file_writer.image_width.set(
+            detector_dimensions.width, timeout=self.GENERAL_STATUS_TIMEOUT
+        )
+        status &= self.odin.file_writer.num_row_chunks.set(
+            detector_dimensions.height, timeout=self.GENERAL_STATUS_TIMEOUT
+        )
+        status &= self.odin.file_writer.num_col_chunks.set(
+            detector_dimensions.width, timeout=self.GENERAL_STATUS_TIMEOUT
+        )
```
DominicOram, line 136: Could: You're right that these should all have a timeout but it's getting messy with them set so many times here. What about something like:
```python
status = Status(timeout=self.GENERAL_STATUS_TIMEOUT) # Overall status to track timeout for the whole operation

status &= self.cam.roi_mode.set(1 if enable else 0)
...
```

DominicOram: And we could repeat the pattern elsewhere

src/dodal/devices/eiger.py
```diff
@@ -192,43 +208,88 @@ def set_num_triggers_and_captures(self) -> Status:
...
+        self.odin.file_writer.data_type.put(f"UInt{bit_depth}")
 
-        await_value(self.odin.fan.ready, 1).wait(10)
+        return self.arming_status
 
     def disarm_detector(self):
         self.cam.acquire.put(0)
+        self.armed = False
+
+    def make_chained_functions(self) -> Status:
```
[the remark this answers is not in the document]
DominicOram: Still valid comment

src/dodal/devices/eiger.py
```diff
@@ -192,43 +226,85 @@ def set_num_triggers_and_captures(self) -> Status:
...
+        )
+        LOGGER.info("Eiger staging: awaiting odin metadata")
+        this_status &= await_value(self.odin.meta.ready, 1, self.GENERAL_STATUS_TIMEOUT)
+        return this_status
+
+    def _wait_fan_ready(self) -> Status:
+        self.filewriters_finished = self.odin.create_finished_status()
+        LOGGER.info("Eiger staging: awaiting odin fan ready")
+        this_status = await_value(self.odin.fan.ready, 1, self.GENERAL_STATUS_TIMEOUT)
+        return this_status
```
DominicOram, line 268: Nit: Can do this in one line

src/dodal/devices/eiger.py
```diff
@@ -192,43 +226,85 @@ def set_num_triggers_and_captures(self) -> Status:
...
+                self.set_cam_pvs,
+                self.set_odin_number_of_frame_chunks,
+                self.set_odin_pvs,
+                self.set_mx_settings_pvs,
+                self.set_num_triggers_and_captures,
+            ]
+        )
 
-        await_value(self.odin.fan.ready, 1).wait(10)
+        unwrapped_funcs.extend(
```
DominicOram, line 300: Should: We can just extend it once

src/dodal/devices/utils.py
```diff
@@ -12,3 +18,84 @@ def epics_signal_put_wait(pv_name: str, wait: float = 1.0) -> EpicsSignal:
         EpicsSignal: An EpicsSignal that will wait for a callback.
     """
     return Component(EpicsSignal, pv_name, put_complete=True, write_timeout=wait)
+
+
+class StatusException(Exception):
+    """For general status failures"""
+
+    pass
```
DominicOram, line 26: Should: This is no longer used, we can remove it

src/dodal/devices/utils.py
```diff
@@ -12,3 +18,84 @@ def epics_signal_put_wait(pv_name: str, wait: float = 1.0) -> EpicsSignal:
...
+
+def wrap_and_do_funcs(
+    unwrapped_funcs: list[Callable[[], StatusBase]],
+    timeout: float = 60.0,
+) -> Status:
+    """Creates and initiates an asynchronous chaining of functions which return a status.
+
+    Usage:
+    This function can be used to convert a series of blocking, status-returning functions to a series of sequential, asynchronous,
+    status-returning functions by making use of callbacks. It also checks for exceptions on each returned status
```
DominicOram, line 37: Nit: Still doesn't sound quite right.
```suggestion
    This function can be used to take a series of status-returning functions and run them all sequentially and in the background by making use of callbacks. It also ensures exceptions on each returned status are propagated
```

src/dodal/devices/utils.py
```diff
@@ -12,3 +18,84 @@ def epics_signal_put_wait(pv_name: str, wait: float = 1.0) -> EpicsSignal:
...
     return Component(EpicsSignal, pv_name, put_complete=True, write_timeout=wait)
+
+
+class StatusException(Exception):
+    """For general status failures"""
+
+    pass
+
+
+def wrap_and_do_funcs(
```
DominicOram, line 29: Nit: How about something like `chain_and_do_funcs`

src/dodal/devices/utils.py
```diff
@@ -12,3 +18,71 @@ def epics_signal_put_wait(pv_name: str, wait: float = 1.0) -> EpicsSignal:
...
+        )
+    )
+
+    # Wrap each function in reverse
+    for num, func in enumerate(list(reversed(unwrapped_funcs))[1:-1]):
+        wrapped_funcs.append(
+            partial(
+                wrap_func,
+                current_func=func,
+                next_func=wrapped_funcs[-1],
```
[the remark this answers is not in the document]
DominicOram: Comment still stands

tests/devices/unit_tests/test_eiger.py
```diff
@@ -53,6 +56,30 @@ def fake_eiger():
     return fake_eiger
 
 
+@pytest.fixture
+def finished_status():
+    status = Status()
+    status.set_finished()
+    return status
```
DominicOram, line 63: Nit: I wouldn't make this a fixture, it can just be a function you call to get a status when you need one

src/dodal/devices/utils.py
```diff
@@ -12,3 +18,84 @@ def epics_signal_put_wait(pv_name: str, wait: float = 1.0) -> EpicsSignal:
...
+
+
+class StatusException(Exception):
+    """For general status failures"""
+
+    pass
+
+
+def wrap_and_do_funcs(
+    unwrapped_funcs: list[Callable[[], StatusBase]],
```
DominicOram, line 30: Nit: `functions_to_chain` rather than `unwrapped_funcs` maybe?

src/dodal/devices/eiger.py
```diff
@@ -192,43 +226,85 @@ def set_num_triggers_and_captures(self) -> Status:
...
-        odin_status.wait(10)
-
-        LOGGER.info("Setting aquire")
-        self.cam.acquire.set(1).wait(timeout=10)
+    def disarm_detector(self):
+        self.cam.acquire.put(0)
 
-        self.filewriters_finished = self.odin.create_finished_status()
+    def make_chained_functions(self) -> Status:
+        unwrapped_funcs = list()
```
DominicOram, line 282: Should: `unwrapped_funcs` is too generic, something like `functions_to_do_arm`

tests/devices/unit_tests/test_eiger.py
```diff
@@ -282,34 +373,132 @@ def wait_on_staging():
...
+        ("fake_eiger.set_odin_pvs"),
+        ("fake_eiger.set_mx_settings_pvs"),
+        ("fake_eiger.set_num_triggers_and_captures"),
+        ("fake_eiger._wait_for_odin_status"),
+        ("lambda: fake_eiger.cam.acquire.set(1, timeout=10)"),
+        ("fake_eiger._wait_fan_ready"),
+        ("fake_eiger._finish_arm"),
+    ],
+)
+def test_check_callback_error(fake_eiger: EigerDetector, func):
```
DominicOram, line 462: Should: I think a more useful test here would be to keep the same list as defined in the `eiger` but change which one throws an error

tests/devices/unit_tests/test_eiger.py
```diff
@@ -282,34 +373,132 @@ def wait_on_staging():
...
-    happy_status.set_finished()
-    fake_eiger.filewriters_finished = happy_status
+    fake_eiger.filewriters_finished = finished_status
 
     fake_eiger.odin.file_writer.capture.sim_put(1)
     fake_eiger.odin.file_writer.num_captured.sim_put(2000)
     fake_eiger.odin.check_odin_state = MagicMock(return_value=True)
 
     fake_eiger.detector_params.trigger_mode = TriggerMode.FREE_RUN
+    fake_eiger.armed = True
```
DominicOram, line 501: Should: No longer used, remove

src/dodal/devices/eiger.py
olliesilvester: I think if we did this then the first part of the status would never automatically be set as finished. Would attaching a timeout to just one of these statuses give the same effect?

olliesilvester: As discussed we'll leave this as it is for now

DominicOram (approved):
Much nicer naming, thank you!

PR #46, opened 2023-05-04
Remove EigerTriggerNumber as it's unused

### LINKED ISSUE #45: Remove EigerTriggerNumber
This isn't actually used anywhere, instead we use the `num_triggers` and `num_images_per_trigger` variables for this.

DominicOram:
Fixes #45


dperl-dls (approved):
Thanks

PR #48, opened 2023-05-04
Allow eiger to be armed without setting number of triggers

DominicOram:
See DiamondLightSource/python-artemis#560

Requires the following so do not merge until these are:
* https://github.com/DiamondLightSource/dodal/pull/46
* https://github.com/DiamondLightSource/dodal/pull/47


dperl-dls (approved):
great!

PR #49, opened 2023-05-09
Artemis 526 - changes for rotation scans

dperl-dls:


DominicOram (approved):
Current system tests are still fine, these exist on the simulator

src/dodal/devices/zebra.py
```diff
@@ -223,7 +223,15 @@ def __str__(self) -> str:
         return ", ".join(input_strings)
 
 
+class SoftInputs(Device):
+    soft_in_1: EpicsSignal = Component(EpicsSignal, "SOFT_IN:B0")
+    soft_in_2: EpicsSignal = Component(EpicsSignal, "SOFT_IN:B1")
+    soft_in_3: EpicsSignal = Component(EpicsSignal, "SOFT_IN:B2")
+    soft_in_4: EpicsSignal = Component(EpicsSignal, "SOFT_IN:B3")
```
DominicOram, line 230: Could: These are binary PVs, could we have an enum of something like `ON`/`OFF` to set them?

DominicOram (changes requested):
Minor comment in code after looking at Artemis side

DominicOram (approved):
Great, thanks!

PR #55, opened 2023-05-15
Getting an aperture scatterguard twice with no positions is valid if positions already specified

### LINKED ISSUE #51: Aperture scatterguard can end up with no positions
If `i03.aperture_scatterguard` is called multiple times, sometimes with the `AperturePositions` and sometimes without you can end up with an instance that has no set positions and so is unusable

DominicOram:
Fixes #51


dperl-dls (approved):
looks good, thanks!

PR #59, opened 2023-05-17
I23 grid scans

### LINKED ISSUE #578: Add a flowchart on the decisions to make a new ophyd device for DLS
to simplify and standardize the development, approval and testing processes involved in getting a new device

based on ophyd-async [PR 323]( https://github.com/bluesky/ophyd-async/pull/323)

NeilSmithDLS:
Fixes #578 and #574

This is simply a copy and rename of the fast grid scan code. It will need updating to make it specific for stepped grid scans once the list of arguments is know


src/dodal/devices/stepped_grid_scan.py
```diff
@@ -0,0 +1,285 @@
...
+                raise IndexError(f"{grid_position} is outside the bounds of the grid")
+
+        return Point3D(
+            self.x_axis.steps_to_motor_position(grid_position.x),
+            self.y_axis.steps_to_motor_position(grid_position.y),
+            self.z_axis.steps_to_motor_position(grid_position.z),
+        )
+
+
+class GridScanCompleteStatus(DeviceStatus):
```
DominicOram, line 123: Should: This is relying on the hardware to work out if we're finished, we may have something similar for stepped scans eventually but I don't think it's similar enough for this to be useful now

src/dodal/devices/stepped_grid_scan.py
```diff
@@ -0,0 +1,285 @@
...
+    @property
+    def end(self):
+        return self.steps_to_motor_position(self.full_steps)
+
+    def is_within(self, steps):
+        return 0 <= steps <= self.full_steps
+
+
+@dataclass
+class SteppedGridScanParams(DataClassJsonMixin, AbstractExperimentParameterBase):
```
DominicOram, line 41: Should: I think it makes sense for these parameters to be on the dodal side for the fast grid scan as they directly relate to the hardware but for the stepped scan the parameters should probably be on the artemis side (see https://github.com/DiamondLightSource/python-artemis/blob/main/src/artemis/parameters/internal_parameters/plan_specific/rotation_scan_internal_params.py)

DominicOram (changes requested):
Thanks, see comments in code

DominicOram (approved):

PR #60, opened 2023-05-17
Get logs to rollover after a certain time/size

DominicOram:
Fixes DiamondLightSource/python-artemis#482

## To test
* Run for sometime and confirm logs roll over

### DIFF
--- a/src/dodal/log.py
+++ b/src/dodal/log.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from logging.handlers import TimedRotatingFileHandler
 from os import environ
 from pathlib import Path
 from typing import List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


src/dodal/log.py
```diff
@@ -21,6 +21,49 @@
...
+                self.stream.seek(0, 2)  # due to non-posix-compliant Windows feature
+                if self.stream.tell() + len(msg) >= self.maxBytes:
+                    return 1
+        return 0
+
+    def getFilesToDelete(self):
+        """
+        Override method to do nothing - we don't want logs to be deleted
+        """
+        return []
```
DominicOram, line 64: It looks like just setting `backupCount` to 0 will cause no files to ever be deleted. We should do this and remove the override

src/dodal/log.py
```diff
@@ -21,6 +21,49 @@
...
+            return 1
+        else:
+            if self.stream is None:  # delay was set...
+                self.stream = self._open()
+            if self.maxBytes > 0:  # are we rolling over?
+                msg = "%s\n" % self.format(record)
+                self.stream.seek(0, 2)  # due to non-posix-compliant Windows feature
+                if self.stream.tell() + len(msg) >= self.maxBytes:
+                    return 1
+        return 0
```
DominicOram, line 58: We can simplify this a bit by doing our check first then just returning the result of the super check

src/dodal/log.py
```diff
@@ -76,7 +119,13 @@ def set_up_file_handler(
...
         logging_path = _get_logging_file_path()
         print(f"Logging to {logging_path}")
-    file_handler = logging.FileHandler(filename=logging_path)
+    file_handler = EnhancedRollingFileHandler(
+        filename=logging_path,
+        when="D",
+        interval=1,
+        backupCount=10,
+        maxBytes=1e8,
+    )
```
DominicOram, line 128: Most of these are the defaults anyway so we can just call it with default args

olliesilvester (approved):
Looks good - I just changed the name for the test for logs not rolling over

PR #61, opened 2023-05-17
Allow aliased or extended builtin imports in device modules

DiamondJoseph:
- Fixes #56
- e.g. a = dict
- e.g. from ophyd.utils import DisconnectedError

PR #62, opened 2023-05-18
Allow creating all devices when device instantiation(s) raise an exception

### LINKED ISSUE #58: Handle exceptions in make_all_devices
(As I understand it): dodal in Artemis/Hyperion lazily instantiates devices when plans that use the devices are required, so as to allow those plans to run regardless of the state of other devices on the beamline, and in this case if a device fails to instantiate the exception should be thrown.

make_all_devices is called by tests and when dodal is imported into Blueapi, which wants to know about all devices and register them to its context eagerly: in an analogous way, Blueapi would like to successfully start but not allow any devices that failed to be instantiated to be used as arguments for plans.

Tangentially related to but potentially opposed to desired behaviour of #3 (Presumably will call make_all_devices, but would be happy allowing an exception to be fatal in this case?)

DiamondJoseph:
Closes #58 but as an optional behaviour. Changes the behaviour of make_all_devices to not fail at the first exception, but to collect all exceptions from devices that failed to instantiate and raise an ExceptionBundle with them all (unless there is only 1 device, with no dependencies that fails to instantiate, then the raise is just delayed until all other devices are tried)

Requires #61 (to allow importing an Exception that is thrown by the device instantiation)

### DIFF
--- a/src/dodal/utils.py
+++ b/src/dodal/utils.py
@@ -58,6 +58,13 @@ BLUESKY_PROTOCOLS = [
 DeviceDict = Dict[str, HasName]
 
 
+class DependentDeviceInstantiationException(Exception):
+    """One or more exceptions was raised while instantiating devices this device depends on"""
+
+    def __init__(self, exceptions: Mapping[str, Exception]):
+        self.exceptions = exceptions
+
+
 @dataclass
 class ExceptionInformation:
     exception: Exception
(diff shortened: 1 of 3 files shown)


DominicOram:
@DiamondJoseph what is the state of this? Anything we can help with to get it merged?

PR #63, opened 2023-05-18
Adds some tolerance in the aperture/scatterguard code

### LINKED ISSUE #52: ApertureScatterguard should allow movement when sufficiently close in Z
Currently the ApertureScatterguard says it's not allowed to move when it is sufficiently close

DominicOram:
Fixes #52 

## To test:
* Confirm new tests are sensible and that they pass
* Run `test_i03_system` and confirm PVs still connect

### DIFF
--- a/tests/devices/unit_tests/test_aperture_scatterguard.py
+++ b/tests/devices/unit_tests/test_aperture_scatterguard.py
@@ -1,0 +1,52 @@
+from unittest.mock import MagicMock
+
+import pytest
+from ophyd.sim import make_fake_device
+from ophyd.status import Status, StatusBase
+
+from dodal.devices.aperturescatterguard import ApertureScatterguard, InvalidApertureMove
+
+
+@pytest.fixture
+def fake_aperture_scatterguard():
+    FakeApertureScatterguard = make_fake_device(ApertureScatterguard)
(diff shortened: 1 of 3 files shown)


tests/devices/unit_tests/test_aperture_scatterguard.py
```diff
@@ -0,0 +1,52 @@
...
+    fake_aperture_scatterguard: ApertureScatterguard,
+):
+    fake_aperture_scatterguard.aperture.z.motor_resolution.sim_put(0.001)
+    fake_aperture_scatterguard.aperture.z.user_setpoint.sim_put(1)
+    fake_aperture_scatterguard.aperture.z.motor_done_move.sim_put(1)
+
+    completed_status = Status()
+    completed_status.set_finished()
+
+    mock_set = MagicMock(return_value=Status(done=True, success=True))
```
olliesilvester, line 40: Nit: I think we can just use completed_status as the return value here
`mock_set = MagicMock(return_value=completed_status`)

olliesilvester (commented):
Two tiny suggestions but this looks good

tests/devices/unit_tests/test_aperture_scatterguard.py
DominicOram: Oops, my bad! I think the better solution is to not have `completed_status` at all, will do that

PR #66, opened 2023-05-22
make get beamline name return unchanged beamline

dperl-dls:


DominicOram (approved):
I think `i03` should still give you `s03` if you have no BEAMLINE, I have put this in

PR #68, opened 2023-05-23
Artemis 678 pydantic internal params

dperl-dls:
Makes some dataclasses into pydantic models to support easier serialisation of parameters in artemis


noemifrisina (approved):
System tests now okay

PR #70, opened 2023-05-24
reset values after tests so following pass

dperl-dls:
some dependence between tests in test_beam_converter means they are intermittently failing. test values are now reset at the end of the test responsible, although this is not a satisfying solution because `test_interpolate_beam_xy_from_det_distance` should not be reusing these anyway, but parametrize makes that difficult to debug and inspect


DominicOram (approved):
Looks good, I updated it to be a bit clearer what we're doing

PR #72, opened 2023-05-26
Rename project on pypi to dls-dodal

callumforrester:
The dodal name conflict on pypi is unlikely to be resolved soon. I suggest releasing as dodal-dls for the time being. If the dodal name is released to us we can deprecate and move.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -176,6 +176,10 @@ jobs:
     env:
       HAS_PYPI_TOKEN: ${{ secrets.PYPI_TOKEN != '' }}
 
+    environment:
+      name: pypi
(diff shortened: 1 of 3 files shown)


callumforrester:
First of all, @gilesknap, do you think this change to the skeleton action will work? It's hard to test so I could do with a second opinion.

graeme-winter:
Is that preferable to `dls-dodal` which feels more like a thing in the dls namespace?

gilesknap:
@callumforrester my reading of the docs for https://github.com/pypa/gh-action-pypi-publish agrees with your change.

I'd just try it as its not going to do anything catastrophic. If it does fail you can delete the release and tag, fix and try again.

Issue #69, opened 2023-05-24
make an i24.py


dperl-dls:
to support jungfrau work and ssx

PR #79, opened 2023-05-31
call precondition

dperl-dls:
`skip_device` was checking for the existence of the precondition rather than the return value


dperl-dls:
was included in #69

PR #86, opened 2023-06-01
Add specific pin tip detection device

DominicOram:
See https://github.com/DiamondLightSource/python-artemis/issues/683


dperl-dls (approved):
seems useful, thanks!

PR #87, opened 2023-06-02
update current_energy with units

dperl-dls:
supports https://github.com/DiamondLightSource/python-artemis/pull/697
GDA change at https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/39308


DominicOram (dismissed):
Looks good, we need to bump the param version and a corresponding GDA change (to update version number and the key that's passed in) before I can merge

DominicOram:
I won't get time to review this @noemifrisina can you take a look that this makes sense with the GDA change too?

noemifrisina (approved):
Looks okay, tests pass.

GDA seems to fail when trying to connect to some of the s03 devices but that happens after the current change

PR #89, opened 2023-06-09
exclude axes from gridscanparams model

dperl-dls:
to support https://github.com/DiamondLightSource/python-artemis/pull/696


DominicOram (approved):
Great, thanks

PR #90, opened 2023-06-12
Add 16 filter attenuator, with set transmission function

### LINKED ISSUE #82: Implement attenuator as Ophyd device
This will be used along with the device added in #71 to run attenuation optimisation for the Xspress3MiniDetector within Dodal

olliesilvester:
Fixes #82 

To test:
Check new unit test
Confirm that the set_transmission() provides the same functionality is the same function in GDA (in EPICSAttenuatorBase.java)


src/dodal/devices/attenuator/attenuator.py
```diff
@@ -0,0 +1,136 @@
...
+from dodal.devices.status import await_value
+from dodal.log import LOGGER
+
+
+class Attenuator(Device):
+    class TransmissionSignal(Signal):
+        def set(self, value, *, timeout=None, settle_time=None, **kwargs):
+            return self.parent.set_transmission()
+
+    do_set_transmission: TransmissionSignal = Component(TransmissionSignal)
```
DominicOram, line 17: Should: I think we can just call this transmission

DominicOram: @neil-i03 how often do you want to touch any of the underlying PVs on the attenuator? I assume 99% of the time you're just setting transmission?

src/dodal/devices/attenuator/filter.py
```diff
@@ -0,0 +1,5 @@
+from ophyd import Component, Device, EpicsSignalRO
+
+
+class AtteunatorFilter(Device):
```
DominicOram, line 4: Should: This is so small I would actually just put it in the attenuator file then move them out of the folder

tests/devices/unit_tests/test_attenuator.py
```diff
@@ -0,0 +1,34 @@
+from unittest.mock import MagicMock
+
+import pytest
+from ophyd.sim import make_fake_device
+from ophyd.status import Status
+
+from dodal.devices.attenuator.attenuator import Attenuator
+
+CALCULATED_VALUE = 10
```
DominicOram, line 9: Could: I would use a range or something here, just so that we have some variation that we can test against

src/dodal/devices/attenuator/attenuator.py
DominicOram: I also think it needs a comment that it is in the range 0-1, people may expect it to be a percentage

DominicOram (changes requested):
Pretty good, some comments but generally should work

src/dodal/devices/attenuator/filter.py
olliesilvester: There are other at least two other PVs which use :FILTER{number} which I haven't needed to access yet. Better to leave the class in for if these PVs are added? Or remove it now and maybe readd later?

tests/devices/unit_tests/test_attenuator.py
```diff
@@ -0,0 +1,35 @@
+from unittest.mock import MagicMock
+
+import numpy as np
+import pytest
+from ophyd.sim import make_fake_device
+from ophyd.status import Status
+
+from dodal.devices.attenuator import Attenuator
+
+CALCULATED_VALUE = np.random.randint(0, 10, size=16)
```
DominicOram, line 10: Should: Sorry, we shouldn't use random numbers here. Randomness in unit tests is generally bad as it causes repeatability issues

tests/devices/unit_tests/test_attenuator.py
```diff
@@ -0,0 +1,35 @@
...
+        return Status(done=True, success=True)
+
+    fake_attenuator.change.set = MagicMock(side_effect=mock_apply_values)
+
+    return fake_attenuator
+
+
+def test_set_transmission_success(fake_attenuator: Attenuator):
+    fake_attenuator.calulated_filter_state_10.sim_put(1)
+    fake_attenuator.set_transmission(1.0).wait(10)
```
DominicOram, line 35: Should: Can you add another test that checks setting the transmission signal through a run engine, will improve coverage a bit

DominicOram (changes requested):
Apologies, couple more small bits

src/dodal/devices/attenuator/attenuator.py
DominicOram: As discussed, I think we can move the set override to the Attenuator itself so we can do something like `attenuator.set(desired_transmission)`

olliesilvester: This is changed now

PR #91, opened 2023-06-13
Snapshot paths in grid overlay

noemifrisina:
Changes to devices needed for Artemis 671 (use callbacks in grid detection)


DominicOram (approved):
Looks good, thanks.

Issue #76, opened 2023-05-30
Make stage synchronous if arming not yet started


DominicOram:
With https://github.com/DiamondLightSource/dodal/pull/43 the `stage` of the eiger no longer exists. We should reinstate it so that:
* If the Eiger has already been armed it does nothing
* If the `async_stage` has started it waits on it to complete
* If the `async_stage` has not started it starts it and waits on it to complete

This may be the cause of https://github.com/DiamondLightSource/python-artemis/issues/686, we should confirm if this is fixed after we do this

PR #92, opened 2023-06-14
Make stage synchronous if arming not started

olliesilvester:
Fixes https://github.com/DiamondLightSource/dodal/issues/76

This should also fix the very slow unit tests introduced in #43 

To test:
Run new tests and check they are sensible
Check changes satisfy the criteria
See if this fixes https://github.com/DiamondLightSource/python-artemis/issues/686

### DIFF
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -122,10 +122,6 @@ class EigerDetector(Device):
         self.disable_roi_mode()
         return status_ok
 
-    # This is no longer used by staging
-    def enable_roi_mode(self):
-        self.change_roi_mode(True)
-
     def disable_roi_mode(self):
         self.change_roi_mode(False)
 
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -85,14 +85,19 @@ class EigerDetector(Device):
         self.arming_status = self.do_arming_chain()
         return self.arming_status
 
+    def is_armed(self):
+        if self.odin.fan.ready.get() == 1 and self.cam.acquire.get() == 1:
+            return True
+        else:
+            return False
+
     def stage(self):
         if self.arming_status.done is False:
             # Arming has started so wait for it to finish
             self.arming_status.wait(60)
-        else:
-            if self.odin.fan.ready.get() != 1:
-                # Arming hasn't started, do it asynchronously
-                self.async_stage()
+        elif not self.is_armed():
+            # Arming hasn't started, do it asynchronously
+            self.async_stage().wait(timeout=self.GENERAL_STATUS_TIMEOUT)
 
     def unstage(self) -> bool:
         assert self.detector_params is not None
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -26,6 +26,13 @@ class EigerDetector(Device):
         def set(self, value, *, timeout=None, settle_time=None, **kwargs):
             return self.parent.async_stage()
 
+    class ArmedState(Enum):
+        UNARMED = "unarmed"
+        ARMING = "arming"
+        ARMED = "armed"
+
+    armed_state = ArmedState.UNARMED
+
     do_arm: ArmingSignal = Component(ArmingSignal)
     cam: EigerDetectorCam = Component(EigerDetectorCam, "CAM:")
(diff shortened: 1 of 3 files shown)


src/dodal/devices/eiger.py
```diff
@@ -79,8 +86,19 @@ def async_stage(self):
...
+        self.arming_status = self.do_arming_chain()
+        return self.arming_status
+
+    def stage(self):
+        if self.armed_state.value == "unarmed":
+            self.async_stage().wait(60)
 
-        return self.do_arming_chain()
+        elif self.armed_state.value == "arming":
+            self.arming_status.wait(60)
```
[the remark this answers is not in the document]
DominicOram: I think this also means we can remove the `None` return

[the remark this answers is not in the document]
olliesilvester: The last part of staging is awaiting for self.odin.fan.ready to be set to 1, shouldn't we check this value rather than self.cam.acquire?

src/dodal/devices/eiger.py
```diff
@@ -290,7 +306,7 @@ def do_arming_chain(self) -> Status:
                 self.set_odin_pvs,
                 self.set_mx_settings_pvs,
                 self.set_num_triggers_and_captures,
-                lambda: await_value(self.STALE_PARAMS_TIMEOUT, 0, 60),
+                lambda: await_value(self.stale_params, 0, 60),
```
olliesilvester, line 309: This typo looks like it should have broken the arming, not sure how it was working before if this was in there

src/dodal/devices/eiger.py
```diff
@@ -79,8 +82,17 @@ def async_stage(self):
         status_ok, error_message = self.odin.check_odin_initialised()
         if not status_ok:
             raise Exception(f"Odin not initialised: {error_message}")
+        self.arming_status = self.do_arming_chain()
+        return self.arming_status
 
-        return self.do_arming_chain()
+    def stage(self):
+        if self.arming_status.done is False:
```
DominicOram, line 89: Nit: 
```suggestion
        if not self.arming_status.done:
```

src/dodal/devices/eiger.py
DominicOram: Yh, it's probably just not covered well in the tests

src/dodal/devices/eiger.py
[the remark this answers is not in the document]
DominicOram: Good spot, I think we should confirm `acquire` and `fan.ready`. Can you make a new function called something like `is_armed` that we use to check both of these?

src/dodal/devices/eiger.py
```diff
@@ -79,8 +82,17 @@ def async_stage(self):
...
+            # Arming has started so wait for it to finish
+            self.arming_status.wait(60)
+        else:
+            if self.odin.fan.ready.get() != 1:
+                # Arming hasn't started, do it asynchronously
+                self.async_stage()
```
DominicOram, line 95: Must: I think in this case we need to be synchronous so we should have a `wait` here

src/dodal/devices/eiger.py
```diff
@@ -79,8 +82,17 @@ def async_stage(self):
...
+        self.arming_status = self.do_arming_chain()
+        return self.arming_status
 
-        return self.do_arming_chain()
+    def stage(self):
+        if self.arming_status.done is False:
+            # Arming has started so wait for it to finish
+            self.arming_status.wait(60)
+        else:
+            if self.odin.fan.ready.get() != 1:
```
DominicOram, line 93: Nit: This can be an `elif`

DominicOram (changes requested):
Good work, thanks. Couple of comment

src/dodal/devices/eiger.py
```diff
@@ -85,14 +85,19 @@ def async_stage(self):
         self.arming_status = self.do_arming_chain()
         return self.arming_status
 
+    def is_armed(self):
+        if self.odin.fan.ready.get() == 1 and self.cam.acquire.get() == 1:
+            return True
+        else:
+            return False
```
DominicOram, line 92: Nit:
```suggestion
        return self.odin.fan.ready.get() == 1 and self.cam.acquire.get() == 1
```

DominicOram (approved):
Minor nit change, otherwise good, thanks!

PR #94, opened 2023-06-14
Add xspress3mini arming

### LINKED ISSUE #93: Add arming to Xspress3Mini

olliesilvester:
Fixes #93

To test:
Check and run new unit test
Confirm that the logic of arming the xspress3mini matches the function in gda: armXspress3MiniFluo_detector in mx-config/scripts/zebraBox


src/dodal/devices/xspress3_mini/xspress3_mini.py
```diff
@@ -1,23 +1,97 @@
...
+
+    detector_busy_states = [
+        DetectorState.ACQUIRE.value,
+        DetectorState.CORRECT.value,
+        DetectorState.ABORTING.value,
+    ]
+
+    def do_start(self) -> Status:
+        self.erase.put(EraseState.ERASE.value)
+        status = self.channel_1.sca5_update_arrays_mini.set(AcquireState.DONE.value)
```
DominicOram, line 88: Should: The name of this signal isn't great, I think we can remove `sca5` and `mini` from it

src/dodal/devices/xspress3_mini/xspress3_mini.py
```diff
@@ -1,23 +1,97 @@
...
+        status &= self.acquire.set(AcquireState.ACQUIRE.value)
```
DominicOram, line 89: Must: We need to be careful here, from what the GDA comments say this PV will stay high whilst data is being is taken so we shouldn't put it in with this status. I think instead we should make a new status that the object holds that is something like `acquiring`

tests/devices/unit_tests/test_xspress3mini.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.fixture
+def fake_xspress3mini():
+    FakeXspress3Mini: Xspress3Mini = make_fake_device(Xspress3Mini)
+    fake_xspress3mini: Xspress3Mini = FakeXspress3Mini(name="xspress3mini")
+    return fake_xspress3mini
+
+
+def test_arm_success_on_busy_state(fake_xspress3mini):
```
DominicOram, line 17: Nice test, thanks

DominicOram (changes requested):
Sorry, meant to put request changes, few minor issues to clean up in code

tests/devices/unit_tests/test_status.py
```diff
@@ -0,0 +1,29 @@
...
+    return fake_device
+
+
+@pytest.mark.parametrize("awaited_value", [(1), (5.3), (False)])
+def test_await_value_in_list_with_no_list_parameter_fails(awaited_value, fake_device):
+    with pytest.raises(TypeError):
+        await_value_in_list(fake_device.pv, awaited_value)
+
+
+def test_await_value_in_list_success(fake_device):
```
DominicOram, line 25: Nice tests, thank you!

DominicOram (changes requested):
Good work, thanks! I think just the one comment now.

DominicOram (approved):
Great, thank you

PR #99, opened 2023-06-16
Add initial check to await_value

### LINKED ISSUE #96: Get await_value working for cases where the PV doesn't change.
await_value() in dodal/devices/status.py returns an unsuccessful status when the subscribable PV never gets changed. If the PV is already at the correct value but doesn't get modified, the function returns an unsuccessful status.

olliesilvester:
Fixes #96

To test:
Check addition to await_value() fixes described issue
Run new unit test


olliesilvester:
Closing this for now since I can't seem to replicate the issue

PR #102, opened 2023-06-22
Artemis 754 add deadtime optimisation

olliesilvester:
Fixes some pv names which are needed for https://github.com/DiamondLightSource/python-artemis/pull/755

To test:
Check new PV names exist


src/dodal/devices/attenuator.py
```diff
@@ -17,7 +17,7 @@ class Attenuator(Device):
     def set(self, transmission) -> SubscriptionStatus:
         """Get desired states and calculated states, return a status which is complete once they are equal"""
 
-        LOGGER.info("Using current energy")
+        LOGGER.info("Setting current energy ")
```
DominicOram, line 20: Should: I think the previous was more correct. "Setting current energy" makes it sound like we're changing energy of the beamline

DominicOram (changes requested):
I tested this by running `python tests/system_tests/test_i03_system.py`. It looks like the PVs you changed are correct but there is a missing G in `BL03I-EA-ATTN-01:E2WL:USECURRENTENERY.PROC`, could you fix that as part of this PR too?

src/dodal/devices/attenuator.py
olliesilvester: Thanks, I've changed these now

DominicOram (changes requested):
Thanks. More issues when running `python tests/system_tests/test_i03_system.py` though:
```
TimeoutError: Failed to connect to all signals: xspress3mini.detector_state (BL03I-EA-XSP3-01::DetectorState_RBV), xspress3mini.dt_corrected_latest_mca (BL03I-EA-XSP3-01::ARR1:ArrayData), xspress3mini.set_num_images (BL03I-EA-XSP3-01::NumImages)
```
Can you use this PR to fix them all up until the system test runs?

DominicOram (approved):
Great, PVs now all connect!

PR #104, opened 2023-06-26
Apply flatfield on zoom level change by default

### LINKED ISSUE #103: OAV device should apply flatfield on zoom set to 1
In 99% of cases (on i03) we want to apply the flatfield to the OAV when we set the zoom level to 1x. Given this is so common we should do it device side and create a combined OAV with zoom device.

DominicOram:
Fixes #103 

To test:
* Confirm tests are sensible and pass

### DIFF
--- a/tests/devices/unit_tests/test_oav.py
+++ b/tests/devices/unit_tests/test_oav.py
@@ -124,6 +124,8 @@ def test_when_zoom_not_1_then_flat_field_removed(fake_oav: OAV):
 def test_when_zoom_is_externally_changed_to_1_then_flat_field_not_changed(
     fake_oav: OAV,
 ):
+    """This test is required to ensure that Artemis doesn't cause unexpected behaviour
+    e.g. change the flatfield when the zoom level is changed through the synoptic"""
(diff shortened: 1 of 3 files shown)


noemifrisina (approved):
Tests look okay and still pass.

tests/devices/unit_tests/test_oav.py
```diff
@@ -99,3 +105,28 @@ def test_bottom_right_from_top_left():
...
+
+
+def test_when_zoom_not_1_then_flat_field_removed(fake_oav: OAV):
+    RE = RunEngine()
+    RE(bps.abs_set(fake_oav.zoom_controller, "10.0x"))
+    assert fake_oav.mxsc.input_plugin.get() == "CAM"
+    assert fake_oav.snapshot.input_plugin.get() == "CAM"
+
+
+def test_when_zoom_is_externally_changed_to_1_then_flat_field_not_changed(
```
DominicOram, line 124: Should: Comment to explain why this is a requirement

noemifrisina: The comment makes it more clear, thanks.

noemifrisina (approved):

PR #107, opened 2023-06-30
Fix the indexing on the grid positions

DominicOram:
See https://github.com/DiamondLightSource/python-artemis/issues/767

To test:
* Confirm co-ordnate system makes sense compared to motion program in `/dls_sw/work/motion/BL03I/Settings/BL03I-MO-PMAC-03/PLCs/11_fast_grid_scans.pmc`
* Confirm comments make sense to explain what is going on
* Confirm tests make more sense now

### DIFF
--- a/src/dodal/devices/fast_grid_scan.py
+++ b/src/dodal/devices/fast_grid_scan.py
@@ -35,7 +35,7 @@ class GridAxis:
     @property
     def end(self):
         """Gives the point where the final frame is taken"""
-        # Not that full_steps is one indexed e.g. if there is one step then the end is
+        # Note that full_steps is one indexed e.g. if there is one step then the end is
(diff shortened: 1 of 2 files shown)


noemifrisina (approved):
The coordinate system and tests both look okay.

Just a small nit: It may be nice to have the comment in `GridAxis.end` about `full_steps` being 1 indexed into the docstring

DominicOram:
> The coordinate system and tests both look okay.
> 
> Just a small nit: It may be nice to have the comment in `GridAxis.end` about `full_steps` being 1 indexed into the docstring

Actually, on further thought I disagree, the docstring is what a user will see when they call `help(axis.end)` the `full_steps` issue is an implementation detail that I don't think a person just calling `end` will care about.

PR #109, opened 2023-06-30
Make unstage eiger more robust

### LINKED ISSUE #108: Unstaging the eiger should be more robust
Given that staging is now asynchronous if it's still in progress when we unstage we should wait for it to finish (most likely this will only happen in an error state). We should also be more robust when unstaging, ensuring that if it fails at earlier points the detector still gets disarmed. There should also be tests on this

DominicOram:
Fixes #108


noemifrisina (approved):
Looks good

PR #113, opened 2023-07-10
Artemis 632 get flux from pv

noemifrisina:
Set up device for flux reading (PV `BL03I-MO-MAPT-01:Y:FLUX`)

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -263,7 +263,7 @@ def flux(wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False) ->
     return device_instantiation(
         Flux,
         "flux",
-        "",
+        "-MO-MAPT-01:Y:",
         wait_for_connection,
         fake_with_ophyd_sim,
     )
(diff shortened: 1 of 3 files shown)


src/dodal/beamlines/i03.py
```diff
@@ -252,3 +253,17 @@ def attenuator(
         wait_for_connection,
         fake_with_ophyd_sim,
     )
+
+
+@skip_device(lambda: BL == "s03")
+def flux(wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False) -> Flux:
+    """Get the i03 synchrotron device, instantiate it if it hasn't already been.
```
DominicOram, line 260: Should: This isn't the synchrotron

src/dodal/beamlines/i03.py
```diff
@@ -252,3 +253,17 @@ def attenuator(
...
+    If this is called when already instantiated in i03, it will return the existing object.
+    """
+    return device_instantiation(
+        Flux,
+        "flux",
+        "",
```
DominicOram, line 266: Should: General preference is to put the prefix here rather than in the device, see https://github.com/DiamondLightSource/dodal/wiki/Device-Standards

DominicOram (changes requested):
Thanks, couple of comments

src/dodal/beamlines/i03.py
noemifrisina: Facepalm ...

DominicOram (approved):
Great, thanks!

PR #115, opened 2023-07-12
Add p38 dodal module

DiamondJoseph:


DiamondJoseph:
Currently just a copy of the p45 module: which devices on p38 do we have existing support for?

DiamondJoseph:
deprecated by #127

PR #123, opened 2023-07-18
Fix use of class variables

DominicOram:
See https://github.com/DiamondLightSource/python-artemis/issues/810

To test:
* Confirm class variables removed and behaviour unchanged


dperl-dls (approved):
seems fine, tests pass, etc.

Issue #124, opened 2023-07-24
i23: implement gonio


Tom-Willemsen:
Add I23's goniometer into dodal.

PR #125, opened 2023-07-24
124 implement i23 gonio

Tom-Willemsen:
Fixes https://github.com/DiamondLightSource/dodal/issues/124

Note: builds on https://github.com/DiamondLightSource/dodal/pull/59 - that should be reviewed first.

Add dodal device for I23 goniometer & corresponding system test.


DominicOram (approved):
System tests work. We're going to have to think about how we generalise the concept of goniometers at some point so that we can get plans to work across beamlines but that's a wider discussion

PR #127, opened 2023-07-25
Add p38 beamline definition

keithralphs:


DiamondJoseph (changes requested):
We know the devices work right now, but would be useful to have the same tests as the other beamlines to keep an eye on that and keep the test coverage from slipping

DiamondJoseph (approved):

PR #131, opened 2023-07-31
Return a status that include OAV wiring changes in zoom level change

DominicOram:
Changing zoom level now changes wiring so waiting on it to finish should also wait on these changes

Fixes issue seen in https://github.com/DiamondLightSource/python-artemis/issues/830


dperl-dls (approved):
looks good, tests seem to behave correctly, I like the addition of `__contains__` to `AndStatus`

PR #133, opened 2023-08-01
Added better docstrings for the attenuator

DominicOram:
See https://github.com/DiamondLightSource/python-artemis/issues/835

Hopefully makes it clearer what units to use


dperl-dls (approved):
very helpful

PR #134, opened 2023-08-01
Added limit checking helper function and tests

DominicOram:
See https://github.com/DiamondLightSource/python-artemis/pull/832#discussion_r1280251795


dperl-dls (approved):
great, thanks

PR #140, opened 2023-08-08
dodal#139 Adds i19-1 devices required for pin tip centring

CoePaul:
- Adds 
 --  i19-1 diffractometer with x,y,z and phi,omega,2theta and detector distance motion (as per EPICS controls) 
  -- on axis viewer OAV1 
  -- off axis viewer OAV2

- See also Jira ticket I19-484 ( source for dodal#139 )


DominicOram:
Closing as this has been implemented in GDA instead

PR #142, opened 2023-08-09
Add some basic devices for i04

DominicOram:
Fixes #141 also fixes #155

To test:
* Run `python tests/system_tests/test_i03_system.py` - confirm devices all connect
* Run `python tests/system_tests/test_i04_system.py` - confirm devices all connect

PR #160, opened 2023-09-03
Add XBPM feedback for i03

DominicOram:
Merge https://github.com/DiamondLightSource/dodal/pull/142 first.

Needed for https://github.com/DiamondLightSource/hyperion/issues/796


src/dodal/devices/i04/transfocator.py
```diff
@@ -0,0 +1,58 @@
...
+
+    start: EpicsSignal = Cpt(EpicsSignal, "START.PROC")
+    start_rbv: EpicsSignalRO = Cpt(EpicsSignalRO, "START_RBV")
+
+    vertical_lens_rbv: EpicsSignalRO = Cpt(EpicsSignalRO, "VER", kind=Kind.hinted)
+
+    TIMEOUT = 120
+    _POLLING_WAIT = 0.01
+
+    def polling_wait_on_stat_rbv(self, for_value):
```
Tom-Willemsen, line 24:
```suggestion
    def polling_wait_on_start_rbv(self, for_value):
```

Unless it's meant to be `stat_rbv`, in which case the field name (and PV name) are likely wrong

src/dodal/devices/i04/transfocator.py
```diff
@@ -0,0 +1,58 @@
...
+        # For some reason couldn't get monitors working on START_RBV
+        # (See https://github.com/DiamondLightSource/dodal/issues/152)
+        for _ in range(int(self.TIMEOUT / self._POLLING_WAIT)):
+            RBV_value = self.start_rbv.get()
+            sleep(self._POLLING_WAIT)
+            if RBV_value == for_value:
+                return
+        raise TimeoutError()
```
Tom-Willemsen, line 32: This might take much longer than `TIMEOUT` seconds, especially if `.get()` is slow for any reason.

How about something like:

```python
def polling_wait_on_stat_rbv(self, for_value):
    start_time = time.time()
    while time.time() < start_time + self.TIMEOUT:
        RBV_value = self.start_rbv.get()
        if RBV_value == for_value:
                return
        sleep(self._POLLING_WAIT)

    # last try
    if self.start_rbv.get() != for_value:
        raise TimeoutError()
```

src/dodal/devices/i04/transfocator.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def set(self, beamsize: float) -> DeviceStatus:
+        subscriber: int
+        status = DeviceStatus(self, timeout=self.TIMEOUT)
+
+        def set_based_on_predicition(old_value, value, *args, **kwargs):
+            if old_value != value:
```
Tom-Willemsen, line 39: Do we want to compare with a floating-point tolerance here?

src/dodal/devices/i04/transfocator.py
```diff
@@ -0,0 +1,58 @@
...
+                self.predict_vertical.unsubscribe(subscriber)
+
+                value = round(value)
```
Tom-Willemsen, line 42: As someone who isn't an expert on this device this seems a bit surprising. Can we justify it with a comment?

src/dodal/devices/i04/transfocator.py
```diff
@@ -0,0 +1,58 @@
...
+
+        LOGGER.info(f"Transfocator setting {beamsize} beamsize")
+        if self.beamsize_set.get() != beamsize:
+            subscriber = self.predict_vertical.subscribe(
+                set_based_on_predicition, run=False
+            )
+            self.beamsize_set.set(beamsize)
+        else:
+            status.set_finished()
+        return status
```
Tom-Willemsen, line 58: There's a load of fiddly stuff going on in this method which seems non-standard to me. Can we add a docstring on the class explaining what the device physically does and how this code achieves that goal?

src/dodal/devices/xbpm_feedback.py
```diff
@@ -0,0 +1,27 @@
+from ophyd import Component, Device, EpicsSignal, EpicsSignalRO
+from ophyd.status import StableSubscriptionStatus, StatusBase
+
+
+class XBPMFeedback(Device):
+    """The XBPM feedback device is an IOC that moves the DCM, HFM and VFM to automatically
+    hold the beam into place, as measured by the XBPM sensor."""
+
+    # The time that the feedback needs to be high to be considered stable
```
Tom-Willemsen, line 9: Good that we have a comment but can we expand it a little for people who aren't so familiar with the device?

Something along the lines of:

> We need to wait for `<something>` to go high, wait for `<some>` seconds, and then wait for it to go low again to ensure that `<some process>` has completed succesfully.

tests/devices/i04/test_transfocator.py
```diff
@@ -0,0 +1,52 @@
...
+    assert fake_transfocator.number_filters_sp.get() == 157
+    assert fake_transfocator.start.get() == 1
+
+    assert not set_status.done
+
+    fake_transfocator.start_rbv.sim_put(1)
+    sleep(0.02)
+    fake_transfocator.start_rbv.sim_put(0)
+
+    sleep(0.02)
```
Tom-Willemsen, line 51: This looks like it is a race condition against the `poll_period`? Is there not a way we can write this test with mocking etc to avoid these sleep calls?

I'm thinking you should be able to set up the mock to return `1` on the first call to `get()` and `0` subsequently.

pyproject.toml
```diff
@@ -14,17 +14,17 @@ classifiers = [
 ]
 description = "Ophyd devices and other utils that could be used across DLS beamlines"
 dependencies = [
-    "ophyd@git+https://github.com/Tom-Willemsen/ophyd@40e4b72c582e65e63d13a1650f76de709e5c70bb",  # Switch back to just "ophyd" once <relevant PR> merged.
+    "ophyd@git+https://github.com/DominicOram/ophyd@31a1762435bcb275d2555068233549885eab02e7", # Switch back to just "ophyd" once ophyd#1148, ophyd#1155, ophyd#1140 merged.
```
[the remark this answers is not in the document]
DominicOram: I discussed this with the core `ophyd` developers yesterday. I think @callumforrester was proposing something similar to your suggestion but I would rather not fork if we don't have to. The general consensus in the meeting was that individual facilities should be freer to merge things into `ophyd` without other facilities and we came up with a list of named maintainers on ophyd for bigger changes. I think we try with that for now and see how far we get.

src/dodal/devices/i04/transfocator.py
DominicOram: Scratch that, it stupidly is a float

tests/devices/i04/test_transfocator.py
DominicOram: Thanks, I've removed a couple of the sleeps, I think one might have to remain though.

pyproject.toml
```diff
@@ -14,17 +14,17 @@ classifiers = [
...
     "pillow",
     "requests",
     "graypy",
     "pydantic<2.0",
-    "opencv-python-headless",  # For pin-tip detection.
-    "aioca",  # Required for CA support with Ophyd V2.
-    "p4p",  # Required for PVA support with Ophyd V2.
+    "opencv-python-headless",                                                                  # For pin-tip detection.
+    "aioca",                                                                                   # Required for CA support with Ophyd V2.
+    "p4p",                                                                                     # Required for PVA support with Ophyd V2.
```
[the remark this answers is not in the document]
Tom-Willemsen: Ah ok didn't realise our auto-formatter touched `.toml`... In that case I agree with you.

src/dodal/devices/i04/transfocator.py
```diff
@@ -21,38 +31,51 @@ class Transfocator(Device):
...
+                # calculation in the IOC returns the theoretical float number of lenses
                 value = round(value)
                 LOGGER.info(f"Transfocator setting {value} filters")
                 self.number_filters_sp.set(value).wait()
                 self.start.set(1).wait()
-                self.polling_wait_on_stat_rbv(1)
-                self.polling_wait_on_stat_rbv(0)
+                self.polling_wait_on_start_rbv(1)
+                self.polling_wait_on_start_rbv(0)
                 status.set_finished()
```
Tom-Willemsen, line 71:
```suggestion
            status.set_finished()
```

Otherwise I think this could deadlock if it gets into this subscriber but doesn't match the `not isclose` condition, since the status would never be set to finished?

src/dodal/devices/i04/transfocator.py
```diff
@@ -21,38 +31,51 @@ class Transfocator(Device):
...
-        raise TimeoutError()
+            sleep(self._POLLING_WAIT)
+
+        # last try
+        if self.start_rbv.get() != for_value:
+            raise TimeoutError()
 
-    def set(self, beamsize: float) -> DeviceStatus:
+    def set(self, beamsize_microns: float) -> DeviceStatus:
+        """To set the beamsiz on the transfocator we must:
```
Tom-Willemsen, line 49: nit:

```suggestion
        """To set the beamsize on the transfocator we must:
```

Tom-Willemsen (commented):
Sorry, just one more substantive comment I noticed...

DominicOram:
From https://jira.diamond.ac.uk/browse/I03-870 it looks like the stability check is no longer needed in `ophyd`.

src/dodal/devices/i04/transfocator.py
DominicOram: I don't think we'll end up with a deadlock. The `status` will eventually just timeout in error. It's reasonable to ask if we want that to happen though. I think you're right that actually if the calculation gives the same number of lenses as before we should just immediately set finished. My worry here is that we're assuming the IOC behaves well and doesn't send us a  repeat first, I hope that's a reasonable assumption.

Tom-Willemsen: Ah ok and the timeout doesn't cause an error? Fair enough, I assumed that if it timed out that would cause the scan to fail (or if there wasn't a timeout that it would deadlock).

DominicOram: Sorry, you're right it will cause an error. In which case this is required

src/dodal/devices/xbpm_feedback.py
```diff
@@ -11,18 +11,14 @@ class XBPMFeedback(Device):
...
     pause_feedback: EpicsSignal = Component(EpicsSignal, "-EA-FDBK-01:FB_PAUSE")
     x: EpicsSignalRO = Component(EpicsSignalRO, "-EA-XBPM-02:PosX:MeanValue_RBV")
     y: EpicsSignalRO = Component(EpicsSignalRO, "-EA-XBPM-02:PosY:MeanValue_RBV")
 
     def trigger(self) -> StatusBase:
-        status = StableSubscriptionStatus(
-            self.pos_ok,
+        return SubscriptionStatus(
+            self.pos_stable,
             lambda value, *args, **kwargs: value == 1,
```
Tom-Willemsen, line 22: [nit, ignorable]

> [Signature should be `f(*, old_value, value, **kwargs)`](https://blueskyproject.io/ophyd/user_v1/generated/ophyd.status.SubscriptionStatus.html#ophyd.status.SubscriptionStatus)

I know this is technically correct but maybe it would be clearer to use the exact signature advised in the ophyd docs?

Tom-Willemsen (approved):
one minor suggestion, feel free to ignore and merge this if you like. Otherwise LGTM.

Tom-Willemsen (approved):

PR #161, opened 2023-09-04
Add a backlight device for I24

### LINKED ISSUE #156: Make a backlight device specific for I24
The backlight on I24 is different from i03/i04, in that it has two lights that can be used at the same time. 

Brief description:
The I24 "dual" backlight has 2 LEDs (PVs: BL24I-DI-LED-0{1,2}:TOGGLE):
LED1 has an on/off toggle and power setting. It also has a positioner (BL24I-MO-BL-01:P:) that moves it to set values - the only position not currently in use is the Diode because it needs to be fixed. 
LED2 is a "front light" which has an on/off toggle and power setting. This one doesn't move.

They might use both at the same time or just one.

noemifrisina:
Closes #156 

To test:
- check new tests make sense and pass

### DIFF
--- a/src/dodal/devices/i24/dual_backlight.py
+++ b/src/dodal/devices/i24/dual_backlight.py
@@ -1,0 +1,17 @@
+from ophyd import Component, Device, EpicsSignal, StatusBase
+
+
+class BacklightPositioner(Device):
+    """Device to control the backlight position."""
+
+    position: EpicsSignal = Component(EpicsSignal, "MP:SELECT")
+
+
+class DualBlacklight(Device):
+    led1: EpicsSignal = Component(EpicsSignal, "-DI-LED-01:TOGGLE")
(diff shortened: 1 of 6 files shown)


src/dodal/devices/i24/dual_backlight.py
```diff
@@ -0,0 +1,67 @@
...
+        if position not in self.pos1.allowed_backlight_positions:
+            raise ValueError(
+                f"""
+                Unknown requested position for LED1.
+                Allowed values: {self.pos1.allowed_backlight_positions}.
+            """
+            )
+
+        status = self.pos1.pos_level.set(position)
+        if position == self.pos1.zrst.get():
```
Tom-Willemsen, line 63: This implicitly hard-codes the mapping of `ZRST` -> `OUT`, I think. Which is probably unavoidable, but in that case, it might be clearer just to make it *explicit* instead and not try to read zrst/onst etc.

Tom-Willemsen (approved):

PR #164, opened 2023-09-09
Deploy as dls dodal

### LINKED ISSUE #145: Rename to allow pypi hosting
Currently we can't host `dodal` on pypi as the name clashes. Instead we should use `dls-dodal`. This was started at https://github.com/DiamondLightSource/dodal/pull/72 but we need to get the action to work, which will require renaming this repo to `dls-dodal`

DominicOram:
Fixes #145

This will mean that dodal is installed by `pip install dls-dodal` but used with `import dodal`. I think this is fine, various other python packages have a different distribution name .vs import name but interested if others disagree.

To test this pip install https://test.pypi.org/project/dls-dodal/ (this is from https://github.com/DominicOram/dodal/commits/0.0.8, which is this branch but pointed at test pypi and a fixed ophyd version)

To actually release we will need to:
* Get https://github.com/bluesky/ophyd/pull/1148 merged
* Convince ophyd to make a release
* Move our ophyd off the git version


graeme-winter (approved):
Seems like a sensible action - FWIW this is a pretty common pattern as far as I can tell

Actual change set IRL is tiny - do any docs need updating to go alongside this?

Tom-Willemsen (approved):
> This will mean that dodal is installed by pip install dls-dodal but used with import dodal. I think this is fine

I agree, this seems fine, lots of packages do this and there is minimal risk of confusion/conflicts here.

> Convince ophyd to make a release

I guess this might be complicated by the recent split to ophyd / ophyd-async - I think we probably need _both_ to release as it would be confusing to have two copies of ophyd V2 in our dependencies (and we'll likely need to adjust a bunch of our imports etc anyway)?

DominicOram:
> Seems like a sensible action - FWIW this is a pretty common pattern as far as I can tell
> 
> Actual change set IRL is tiny - do any docs need updating to go alongside this?

Docs were changed in https://github.com/DiamondLightSource/dodal/pull/72. So we've actually been in a state where the docs are wrong for the last couple of months.

> I guess this might be complicated by the recent split to ophyd / ophyd-async - I think we probably need both to release as it would be confusing to have two copies of ophyd V2 in our dependencies (and we'll likely need to adjust a bunch of our imports etc anyway)?

Yes, you're right that now we're using v2 we'll need them to release both. I think the removal of v2 is going to happen at the end of this week but I can discuss with them next week

DominicOram:
Given this affects `blueapi` a bit too I will wait on @callumforrester before merging

callumforrester (approved):
Happy to go ahead

PR #175, opened 2023-09-15
Increase OAV snapshot timeout

### LINKED ISSUE #174: Increase timeout on OAV snapshots
Due to file system issues the OAV snapshots may take some time, we should increase the timeout and confirm where the time is being taken. Ultimately this is better fixed by https://github.com/DiamondLightSource/hyperion/issues/885 but this is a good hotfix

DominicOram:
Fixes #174


dperl-dls (approved):
lgtm

PR #177, opened 2023-09-15
Increase timeout for all frames on eiger

### LINKED ISSUE #176: Increase timeout on Eiger saving all frames
For very large grids it may take longer than 30s for all frames to be collected.

Note that we might be able to reduce this back again when we've done https://github.com/DiamondLightSource/dodal/issues/172

DominicOram:
Fixes #176


dperl-dls (approved):
lgtm

PR #185, opened 2023-10-02
Improve logging

dperl-dls:
for artemis #911

### DIFF
--- a/tests/devices/unit_tests/i24/test_dual_backlight.py
+++ b/tests/devices/unit_tests/i24/test_dual_backlight.py
@@ -38,7 +38,7 @@ def test_when_led1_not_out_it_switches_on(fake_backlight: DualBacklight):
 def test_led2_independent_from_led1_position(fake_backlight: DualBacklight):
     fake_backlight.led2.sim_put("OFF")
     RE = RunEngine()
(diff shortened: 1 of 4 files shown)


tests/devices/unit_tests/i24/test_dual_backlight.py
```diff
@@ -38,6 +38,7 @@ def test_when_led1_not_out_it_switches_on(fake_backlight: DualBacklight):
 def test_led2_independent_from_led1_position(fake_backlight: DualBacklight):
     fake_backlight.led2.sim_put("OFF")
     RE = RunEngine()
-    RE(bps.abs_set(fake_backlight, fake_backlight.IN))
+    RE(bps.abs_set(fake_backlight, fake_backlight.IN, group="backlight in"))
+    RE(bps.wait("backlight in"))
```
DominicOram, line 42: Should: I'm surprised this works and the group is persistent between RE calls, can't you just do `RE(bps.abs_set(fake_backlight, fake_backlight.IN, wait=True))`

DominicOram (changes requested):
Extra logging is nice. Minor comment but think we should fix it.

DominicOram (approved):
This looks good, thank you!

PR #186, opened 2023-10-03
Switch to ophyd async

rosesyrett:
Simple switch to replace ophyd.v2 references with ophyd_async

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ classifiers = [
 description = "Ophyd devices and other utils that could be used across DLS beamlines"
 dependencies = [
     "ophyd-async@git+https://github.com/bluesky/ophyd-async.git",
-    "ophyd@git+https://github.com/DominicOram/ophyd@31a1762435bcb275d2555068233549885eab02e7", # Switch back to just "ophyd" once ophyd#1148, ophyd#1155, ophyd#1140 merged.
+    "ophyd",                                                      # Switch back to just "ophyd" once ophyd#1148, ophyd#1155, ophyd#1140 merged.
     "bluesky",
     "pyepics",
     "dataclasses-json",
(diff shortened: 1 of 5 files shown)


DominicOram (approved):
Looks good to me, @Tom-Willemsen I think this is mergable, right?

Tom-Willemsen:
So I think now that we have https://github.com/bluesky/ophyd-async/pull/19 in ophyd-async, we need to remove the array-reshaping bits from within the pin image recognition device - we were previously using a fork of ophyd that supported `ndarray` but *didn't* do the reshaping at the ophyd level.

I started doing this in a draft PR at https://github.com/DiamondLightSource/dodal/pull/162 but still need to look at why one of the unit tests is failing there (I haven't had time for this recently). I don't think it will work if the reshaping is effectively being done both here and in ophyd.

pyproject.toml
```diff
@@ -14,17 +14,18 @@ classifiers = [
...
     "pillow",
     "requests",
     "graypy",
     "pydantic<2.0",
-    "opencv-python-headless",                                                                  # For pin-tip detection.
-    "aioca",                                                                                   # Required for CA support with Ophyd V2.
-    "p4p",                                                                                     # Required for PVA support with Ophyd V2.
+    "opencv-python-headless", # For pin-tip detection.
+    "aioca",                  # Required for CA support with Ophyd V2.
+    "p4p",                    # Required for PVA support with Ophyd V2.
```
Tom-Willemsen, line 28: nit-pick: maybe we could replace the reference in these comments from "Ophyd V2" to "Ophyd async" now that the project has been renamed?

```suggestion
    "aioca",                  # Required for CA support with ophyd-async.
    "p4p",                    # Required for PVA support with ophyd-async.
```

rosesyrett:
Ok; no problem. This PR can wait until that is done, or we can close it in favour of #162 since that seems to also remove references to `ophyd.v2`.

Tom-Willemsen:
I'll try to get the other one in a state ready to merge as soon as I have a spare 5 mins and then close off this one once that is merged. Keith/Calum were also asking about this indirectly this morning...

PR #189, opened 2023-10-09
Add Slits Ophyd Classes

dan-fernandes:
Adds Ophyd classes to represent slits, as well as specific classes for the I24 Slits 04 Virtual Motors, and slit simulator.


callumforrester (changes requested):
Looks promising, just a few comments and CI needs fixing

DominicOram:
As discussed, @dan-fernandes, if you could drop me a message or add me as a reviewer when you're ready for another opinion I'll check it lines up with what we're doing in MX, thanks

PR #193, opened 2023-10-10
add directory provider and i22 example

rosesyrett:

PR #198, opened 2023-10-12
Add initial linkam3 implementation

tpoliaw:


src/dodal/devices/linkam3.py
```diff
@@ -0,0 +1,100 @@
...
+                break
+
+    def set(self, new_position: float, timeout: Optional[float] = None) -> AsyncStatus:
+        watchers: List[Callable] = []
+        coro = asyncio.wait_for(self._move(new_position, watchers), timeout=timeout)
+        return AsyncStatus(coro, watchers)
+
+    # TODO: Check bitshift order
+    async def in_error(self) -> bool:
+        return int(await self.status.get_value()) & 1
```
callumforrester, line 82: This and a number of other methods cause mypy to complain because they return an `int` and say they return a `bool`, who's right?

callumforrester: Do you think it's more readable to put `> 0` or just wrap the whole thing in `bool(...)`?

tpoliaw:
I think this has been made obsolete by #193

Issue #195, opened 2023-10-11
Flaky test in test_beamline_utils.py for Python 3.11.x (at least)


keithralphs:
The `test_wait_for_v2_device_connection_passes_through_timeout` test in `test_beamline_utils.py` fails about 40% of the time when run by the GitHub Action or running the command the action uses (`pytest --random-order -m "not s03"`) locally in a terminal. I've only seen this on Python 3.11.x so far, but it's possible that it might occur on other versions given the nature of the test (waiting for a call to the event loop during a specified timeout). When it fails it does so for both the default 5.0 timeout setting and the 15.0 one, so it seems like it doesn't really matter what the value is (I've tried others). I've seen the error on both the main branch and my p38-h5 branch.

Steps to reproduce;   In a workspace terminal on the main branch, run `pytest --random-order -m "not s03"` several times until it fails (usually within 3 or 4 attempts).

Platform: Ubuntu 23.04, Python 3.11.6

Actual error:
`FAILED tests/beamlines/unit_tests/test_beamline_utils.py::test_wait_for_v2_device_connection_passes_through_timeout[kwargs1-15.0] - AssertionError: Expected 'call_in_bluesky_event_loop' to be called once. Called 0 times.
FAILED tests/beamlines/unit_tests/test_beamline_utils.py::test_wait_for_v2_device_connection_passes_through_timeout[kwargs0-5.0] - AssertionError: Expected 'call_in_bluesky_event_loop' to be called once. Called 0 times.`

PR #201, opened 2023-10-12
Use patch.object() rather than patch()

Tom-Willemsen:
Hopefully fixes https://github.com/DiamondLightSource/dodal/issues/195

To test:
- run unit tests against python 3.11 a bunch of times, this test shouldn't fail.


DominicOram (approved):
I can't reproduce the issue locally on main anyway. This looks fine, I'm happy to just merge and keep an eye on it

PR #205, opened 2023-10-13
Fix logging to real graylog

Tom-Willemsen:
We were adding a *real* handler to the *real* logger, but patching a single function within that handler within this test only. Now adding a mocked handler instead.

To test: ensure `python -m pytest -m "not s03" --random-order` does not put anything in real graylog


DominicOram (approved):
Great, thank you!

PR #206, opened 2023-10-13
Add pilatus detector

mpdickie:


graeme-winter:
Comment / request: if adding Pilatus detector probably worth making sure that MX use case is supported e.g. CBF writing - there is also the `mxsettings` camserver command which I think is aliased in the Pilatus area detector interface, which would be good to have in there as part of the PR.

Justification:
- use on i19-1
- use on i24
- use on i15

rosesyrett:
Just as some context, this PR comes out of https://github.com/bluesky/ophyd-async/pull/24 in ophyd-async (used to be called ophyd.v2). So this is an ophyd async device, which means it works slightly differently to an ophyd device. But bluesky can definitely mix and match between the two.

I agree with your comment, although I think it's best leaving the pilatus as generic as possible; I don't yet know enough about `mxsettings` but presumably this is a settings file that stores PV values for the pilatus? We are currently developing this for ophyd-async devices, and we can talk about exactly where the MX case fits in.

I think I'll make this a draft PR as realistically it's probably not polished to go in yet (and at the moment, the pyproject.toml looks at the branch of ophyd-async that allows it work which is not ideal...)

rosesyrett:
Note: this will be merged into directory_provider, along with a few other things, and the tests will be cleaned up before directory_provider is merged into main.

rosesyrett:
Abandoning this, as it's genuinely easier to add it in to directory_provider instead of rebasing this...

PR #210, opened 2023-10-17
Allow using ophyd pin tip centring

Tom-Willemsen:
Adds reference to pin tip centring ophyd-async device.

Supporting PR for https://github.com/DiamondLightSource/hyperion/pull/938


DominicOram (approved):
Looks good thanks. We should have a think about codecov on these files as a whole and how useful it is

Issue #207, opened 2023-10-17
VMXm: Add devices for fast grid scan


DominicOram:
VMXm would like to be able to run the fast grid scan program, similar to the one we're running in Hyperion. To do this they will need the following devices:
* Eiger
* Fast grid scan
* Zebra

PR #211, opened 2023-10-17
[#207] Add VMXm fast grid scan devices

Tom-Willemsen:
Fixes https://github.com/DiamondLightSource/dodal/issues/207

- Adds VMXm devices
- Makes some of our infrastructure more generic for dealing with beamline names that don't look like `ixx`.


src/dodal/beamlines/vmxm.py
```diff
@@ -0,0 +1,57 @@
...
+def fast_grid_scan(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> FastGridScan:
+    """Get the vmxm fast_grid_scan device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in vmxm, it will return the existing object.
+    """
+    return device_instantiation(
+        device_factory=FastGridScan,
+        name="fast_grid_scan",
+        prefix="-MO-SAMP-11:FGS:",  # TODO: currently needs Pxxxx prefixes in EPICS on VMXm, ask controls if we can alias to versions without these prefixes?
```
Tom-Willemsen, line 40: Have asked controls (Andy) about this, await response before deciding what to do with this

DominicOram: We also have an issue where the i03 FGS uses `DWELL_TIME` and VMXm uses `EXPOSURE_TIME`. This could warrant two different ophyd devices, but they should both have a standard `exposure_time` signal, as that's more user friendly. i.e. if you do make two devices rename the i03 one

DominicOram:
Does a system test run against the new devices? I think from a discussion we had in the office a while ago we decided on using the most user friendly name for the dodal module so this should be `vmxm` (I appreciate this should be documented)

Tom-Willemsen:
> Does a system test run against the new devices

Not yet but in process of writing some

> I think from a discussion we had in the office a while ago we decided on using the most user friendly name for the dodal module so this should be vmxm

That's what I started with, but we've made assumptions in a few places in dodal that we can translate `BEAMLINE` env var directly into prefixes (e.g. `BeamlinePrefix` class) and module names (this one is my fault; `get_beamline_based_on_environment_variable`).

src/dodal/beamlines/vmxm.py
Tom-Willemsen: I'll ask Andy whether this interface can be made the same as i03. It would be _nice_ to avoid needing two separate ophyd devices for what is conceptually very similar/the same I think?

Or do you think long term we'll go towards separate ophyd devices _anyway_?

Tom-Willemsen: Maybe we can create an "alias" in ophyd so that both work?

DominicOram: I agree having one device is better. If we have to have two then they should be interchangeable at the plan level. Long term we'll end up getting rid of the FGS on i03 anyway

src/dodal/beamlines/vmxm.py
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+
+
+@skip_device(lambda: BL == SIM_BEAMLINE_NAME)
+def backlight(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> Backlight:
+    """Get the i03 backlight device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i03, it will return the existing object.
```
Tom-Willemsen, line 89:
```suggestion
    """Get the vmxm backlight device, instantiate it if it hasn't already been.
    If this is called when already instantiated on vmxm, it will return the existing object.
```

src/dodal/beamlines/vmxm.py
```diff
@@ -0,0 +1,114 @@
...
+        fake=fake_with_ophyd_sim,
+    )
+
+
+@skip_device(lambda: BL == SIM_BEAMLINE_NAME)
+def synchrotron(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> Synchrotron:
+    """Get the i03 synchrotron device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i03, it will return the existing object.
```
Tom-Willemsen, line 105:
```suggestion
    """Get the vmxm synchrotron device, instantiate it if it hasn't already been.
    If this is called when already instantiated on vmxm, it will return the existing object.
```

src/dodal/devices/backlight.py
```diff
@@ -7,9 +7,9 @@ class Backlight(Device):
     OUT = 0
     IN = 1
 
-    pos: EpicsSignal = Component(EpicsSignal, "-EA-BL-01:CTRL")
+    pos: EpicsSignal = Component(EpicsSignal, "-DI-IOC-02:LED:INOUT")
     # Toggle to switch it On or Off
-    toggle: EpicsSignal = Component(EpicsSignal, "-EA-BLIT-01:TOGGLE")
+    toggle: EpicsSignal = Component(EpicsSignal, "-EA-OAV-01:FZOOM:TOGGLE")
```
Tom-Willemsen, line 12: Urgh, need to find a way to make this generic.

src/dodal/devices/detector.py
```diff
@@ -40,7 +40,7 @@ class DetectorParams(BaseModel):
     use_roi_mode: bool
     det_dist_to_beam_converter_path: str
     trigger_mode: TriggerMode = TriggerMode.SET_FRAMES
-    detector_size_constants: DetectorSizeConstants = EIGER2_X_16M_SIZE
+    detector_size_constants: DetectorSizeConstants = EIGER2_X_9M_SIZE # This looks like it's always using the default and not taken from the json
```
Tom-Willemsen, line 43: Fix

src/dodal/devices/eiger.py
```diff
@@ -58,6 +58,8 @@ def with_params(
         return det
 
     def set_detector_parameters(self, detector_params: DetectorParams):
+        LOGGER.info(f"Exposure time {detector_params.exposure_time}")
```
Tom-Willemsen, line 61: Remove

src/dodal/devices/eiger.py
```diff
@@ -103,7 +105,7 @@ def stage(self):
             self.async_stage().wait(timeout=self.ARMING_TIMEOUT)
 
     def stop_odin_when_all_frames_collected(self):
-        LOGGER.info("Waiting on all frames")
+        LOGGER.info(f"Waiting on all frames, expected {self.detector_params.full_number_of_images}")
```
Tom-Willemsen, line 108: Probably fine to keep this.

src/dodal/devices/eiger.py
```diff
@@ -151,6 +153,8 @@ def change_roi_mode(self, enable: bool) -> Status:
             else self.detector_params.detector_size_constants.det_size_pixels
         )
 
+        LOGGER.info(f"Setting height and width on odin to {detector_dimensions.height}, {detector_dimensions.width}")
```
Tom-Willemsen, line 156: This is likely useful, keep.

src/dodal/devices/eiger.py
```diff
@@ -313,8 +322,9 @@ def disarm_detector(self):
     def do_arming_chain(self) -> Status:
         functions_to_do_arm = []
         detector_params: DetectorParams = self.detector_params
-        if detector_params.use_roi_mode:
-            functions_to_do_arm.append(lambda: self.change_roi_mode(enable=True))
+        # Default is no ROI on i03 so we only switych one way
+        #if detector_params.use_roi_mode:
+        functions_to_do_arm.append(lambda: self.change_roi_mode(detector_params.use_roi_mode))
```
Tom-Willemsen, line 327: Think carefully/make generic.

src/dodal/devices/fast_grid_scan.py
```diff
@@ -208,35 +209,38 @@ def clean_up(self):
 class FastGridScan(Device):
     x_steps: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "X_NUM_STEPS")
     y_steps: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Y_NUM_STEPS")
-    z_steps: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Z_NUM_STEPS")
+    #z_steps: EpicsSignalWithRBV = Component(EpicsSignalWithRBV, "Z_NUM_STEPS")
+    z_steps: Signal = Component(Signal, kind=Kind.hinted, value=1)
```
Tom-Willemsen, line 213: Think carefully about how to make this generic.

Or duplicate the whole thing as `SingleFastGridScan` and make the old one `DualFastGridScan` to reflect the fact it does two independent grids.

src/dodal/devices/fast_grid_scan.py
```diff
@@ -208,35 +209,38 @@ def clean_up(self):
...
     run_cmd: EpicsSignal = Component(EpicsSignal, "RUN.PROC")
     stop_cmd: EpicsSignal = Component(EpicsSignal, "STOP.PROC")
-    status: EpicsSignalRO = Component(EpicsSignalRO, "SCAN_STATUS")
+    status: EpicsSignalRO = Component(EpicsSignalRO, "SCAN_STATUS_RBV")
 
     expected_images: Signal = Component(Signal)
 
     # Kickoff timeout in seconds
-    KICKOFF_TIMEOUT: float = 5.0
+    KICKOFF_TIMEOUT: float = 20.0
```
Tom-Willemsen, line 243: Is it possible to make this generic/configurable?

Also we might be able to do this better, the main delay is waiting for completion callbacks on each parameter we set, if we set them all and then wait for all callbacks in parallel it may be significantly faster. Testing with Andy showed a 0.5-1s delay per parameter.

src/dodal/devices/utils.py
```diff
@@ -2,7 +2,7 @@
 from typing import Callable, Optional
 
 from ophyd import Component, Device, EpicsSignal
-from ophyd.status import Status, StatusBase
+from ophyd.status import AndStatus, Status, StatusBase
```
Tom-Willemsen, line 5: Shouldn't be needed...?

src/dodal/devices/vmxm/vmxm_attenuator.py
```diff
@@ -0,0 +1,45 @@
...
+
+from ophyd import Component, Device, EpicsSignal, Kind
+from ophyd.status import Status, SubscriptionStatus
+
+from dodal.devices.detector import DetectorParams
+from dodal.devices.status import await_value
+from dodal.log import LOGGER
+
+
+class VmxmAttenuator(Device):
```
Tom-Willemsen, line 11: It would be nice to have an `Attenuator` _interface_ - the details of this class are quite different to i03 but the concept is the same. Might be out of scope for this issue though - write up separate issue.

src/dodal/utils.py
```diff
@@ -210,12 +220,20 @@ def is_v1_device_type(obj: Type[Any]) -> bool:
     return is_class and follows_protocols and not is_v2_device_type(obj)
 
 
+# Special case for i02-1 -> VMXm
+# Scientists refer to it as VMXm but $BEAMLINE env var is i02-1
+BEAMLINE_NAME_TO_MODULE_NAME_OVERRIDES = {
+    "i02-1": "vmxm",
```
Tom-Willemsen, line 226: Pre-populate this with other known cases e.g. VMXi

src/dodal/devices/vmxm/vmxm_attenuator.py
Tom-Willemsen: See https://github.com/DiamondLightSource/hyperion/issues/956

src/dodal/devices/eiger.py
```diff
@@ -35,7 +36,9 @@ def set(self, value, *, timeout=None, settle_time=None, **kwargs):
 
     STALE_PARAMS_TIMEOUT = 60
     GENERAL_STATUS_TIMEOUT = 10
-    ALL_FRAMES_TIMEOUT = 120
+    ALL_FRAMES_TIMEOUT = (
+        600  # FIXME: VMXm hack - we need to get this passed in a a param probably?
```
Tom-Willemsen, line 40: Per code comment.

src/dodal/devices/eiger.py
```diff
@@ -286,6 +293,9 @@ def set_num_triggers_and_captures(self) -> Status:
 
     def _wait_for_odin_status(self) -> Status:
         self.forward_bit_depth_to_filewriter()
+        sleep(
+            2
+        )  # FIXME - VMXm hack - on VMXm ODIN will *occasionally* fail to start without this sleep. Need a better solution.
```
Tom-Willemsen, line 298: We might be able to do some kind of cleverer wait for ODIN to be ready to start capturing. Needs careful testing on VMXm if we adjust this logic as it was only failing to start ~10% of gridscans before this change was added.

src/dodal/devices/vmxm/vmxm_sample_motors.py
```diff
@@ -0,0 +1,7 @@
+from ophyd import Component as Cpt
+from ophyd import EpicsMotor
+from ophyd.epics_motor import MotorBundle
+
+
+class VmxmSampleMotors(MotorBundle):
```
Tom-Willemsen, line 6: We needed this to work around a bug/feature(?) of the VMXm motion system, we needed to manually resend omega in order to get the expected motion.

Add a comment along these lines?

src/dodal/devices/backlight.py
```diff
@@ -18,3 +17,21 @@ def set(self, position: int) -> StatusBase:
...
+    pos: EpicsSignal = Component(EpicsSignal, "-DI-IOC-02:LED:INOUT")
+    toggle: EpicsSignal = Component(EpicsSignal, "-EA-OAV-01:FZOOM:TOGGLE")
+
+    def set(self, position: int) -> StatusBase:
+        status = self.pos.set(position)
+        if position == self.OUT:
+            status &= self.toggle.set("Off")
+        else:
+            status &= self.toggle.set("On")
+        return status
```
olliesilvester, line 37: This device is the same as I03's backlight other than the PV names. I think we should make a base Backlight device where the PV names are parameters

src/dodal/devices/eiger.py
olliesilvester: Alternatively/in addition, we could make this variable default to a certain value based on the current beamline

src/dodal/devices/eiger.py
olliesilvester: Any idea what was causing it to fail when the sleep wasn't there?

src/dodal/devices/eiger.py
```diff
@@ -306,16 +316,20 @@ def _finish_arm(self) -> Status:
 
     def forward_bit_depth_to_filewriter(self):
         bit_depth = self.bit_depth.get()
-        self.odin.file_writer.data_type.put(f"UInt{bit_depth}")
+        self.odin.file_writer.data_type.set(f"UInt{bit_depth}").wait(
```
olliesilvester, line 319: Was this changed in attempt to fix the _wait_for_odin_status error?

src/dodal/devices/eiger.py
DominicOram: It's really based on the experiment. It's how long do we expect the detector to be exposing for in total

src/dodal/devices/eiger.py
Tom-Willemsen: Not really. Some kind of race condition at the ODIN level, I think. 2 seconds is a number plucked from thin air and we might be able to lower it or put in some better logic, but it'll need testing on VMXm if we do that. Should be possible on a Tuesday or in a shutdown if you ask the scientists - we shouldn't need beam to test this.

For this issue, I'd maybe suggest doing some variant of:
```
if beamline == "vmxm":
    sleep(2)
```

and write an issue to investigate and fix this properly.

src/dodal/devices/backlight.py
Tom-Willemsen: I agree, but in Ophyd V1 it's actually surprisingly tricky to do that (due to some of the `Component` magic).

Suggest spinning that out into an issue if there isn't already one?

src/dodal/devices/eiger.py
Tom-Willemsen: Yes, but I believe from a discussion with Dom on the beamline this is actually the "more correct" option anyway as it waits for a completion callback.

It wasn't sufficient to fix the issue on VMXm but I still think this is a good change to make.

DominicOram: Yes, I agree this is more correct

olliesilvester:
These changes look good to me, it looks like there is only one thing which will break how things currently work if we merged. We need to:
- Move `ALL_FRAMES_TIMEOUT` in `eiger.py` into some parameter. If this is based off the total exposure time, this value could perhaps be calculated from other grid scan parameters?


And there are a few things I can make separate issues for which aren't as essential:

- Making a BaseBackLight rather than two very similar versions for i03 and vmxm - maybe this is easier with opyhd-async devices? 
- Moving all the common bits of code to move into `fast_grid_scan_common.py`
- Find out why we need to manually resend omega in order to get expected motion (in `vmxm_sample_motors.py`)

There's also the issue we already made about having variations of a device within a plan
- https://github.com/DiamondLightSource/hyperion/issues/956

stan-dot:
@DominicOram  found this while scouring the issues for i22-related stuff

DominicOram:
> @DominicOram found this while scouring the issues for i22-related stuff

Yes, I'm aware this is still open.

The issue is on hold whilst we're working on the best solution that will work for VMXm and i03. Did you have a specific question about it?

stan-dot:
nothing specific, I just thought that Tom-Willemsen 's PR might have been lost

dperl-dls:
Closing as stale, please feel free to reopen if this is still needed

PR #216, opened 2023-10-25
correct return type for XBPMFeedback factory

dperl-dls:
No relevant issue, shouldn't really affect anything except static checking


DominicOram (approved):
Looks good, thanks. Docs build failures are unrelated

PR #225, opened 2023-10-30
Remove python 3.8 from pyproject

noemifrisina:


DominicOram:
Can you add a note in the wiki about what our policy is on supported versions? I think we decided to just match https://numpy.org/neps/nep-0029-deprecation_policy.html

DominicOram (approved):

noemifrisina:
Done

PR #228, opened 2023-11-03
Hyperion 949 - Allow multiple loggers to be set up

dperl-dls:
Needed for https://github.com/DiamondLightSource/hyperion/issues/949
Basically just adds the logger to be setup as an argument to the setup functions, so that applications can also use these for loggers which need to log to different files in a different process.

1. Run tests

### Checks for reviewer

### DIFF
--- a/src/dodal/log.py
+++ b/src/dodal/log.py
@@ -138,8 +138,6 @@ def set_up_logging_handlers(
         logging_path: The location to store log files, if left as None then puts them in the default location.
     """
     logging_level = logging_level or "INFO"
-    if dev_mode is None:
-        dev_mode = True
     stream_handler = logging.StreamHandler()
     _add_handler(logger, stream_handler, logging_level)
     graylog_handler = set_up_graylog_handler(logging_level, dev_mode, logger)
--- a/src/dodal/log.py
+++ b/src/dodal/log.py
@@ -16,7 +16,6 @@ LOGGER = logging.getLogger("Dodal")
 LOGGER.setLevel(logging.DEBUG)
 ophyd_logger.parent = LOGGER
 bluesky_logger.parent = LOGGER
-
 DEFAULT_FORMATTER = logging.Formatter(
     "[%(asctime)s] %(name)s %(module)s %(levelname)s: %(message)s"
 )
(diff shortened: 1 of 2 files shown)


tests/unit_tests/test_log.py
```diff
@@ -1,11 +1,19 @@
...
 from pathlib import Path
 from unittest.mock import MagicMock, patch
 
 import pytest
 
 from dodal import log
+from dodal.log import EnhancedRollingFileHandler, GELFTCPHandler
+
+
+def get_mock_streamhandler(level=20):
```
DominicOram, line 13: Nit: Feels cleaner to use the enums rather than the numbers for the levels in this file. They're only tests though so meh

DominicOram (approved):
Looks good, thanks. One minor comment, take it or leave it

src/dodal/log.py
```diff
@@ -124,28 +126,31 @@ def set_up_file_handler(
 
 def set_up_logging_handlers(
     logging_level: Optional[str] = "INFO",
-    dev_mode: bool = False,
+    dev_mode: Optional[bool] = False,
```
DominicOram, line 129: Should: I think this typing is wrong. Following it back through the code it looks like it's only ever specified in prod through the command line with a `store_true`, according to https://docs.python.org/3/library/argparse.html#action this is either True or False, I don't see how it could be `None`. Feel free to spin into another issue, in which case remove this optional and the check below.

DominicOram (changes requested):
Thanks. One minor thing.

src/dodal/log.py
```diff
@@ -127,25 +129,28 @@ def set_up_logging_handlers(
...
     """Set up the default logging environment.
     Args:
         logging_level: The level of logs that should be saved to file/graylog. Defaults to INFO.
         dev_mode: True if in dev mode, will not log to graylog in dev. Defaults to False.
         logging_path: The location to store log files, if left as None then puts them in the default location.
     """
-    logging_level = logging_level if logging_level else "INFO"
+    logging_level = logging_level or "INFO"
+    if dev_mode is None:
+        dev_mode = True
```
DominicOram, line 142: Should: Now that the `dev_mode` is not optional I think we can leave it up to the type checker to check this

DominicOram (changes requested):
One comment still, sorry

noemifrisina (approved):
Looks okay, thanks

PR #236, opened 2023-11-15
Function to verify undulator gap added to dodal

olliesilvester:
Partially fixes [#795](https://github.com/DiamondLightSource/hyperion/issues/795). Once this is merged, we still need to add it to the hyperion plans and change GDA to no longer do this step

1. Confirm that the energy set in `/dodal/devices/undulator_dcm.py` provides the same functionality as `verifyUndulatorGap` in GDA's `i03-config/scripts/beamLineSpecificEnergy.py`
2. Are the undulator gap PV's correct and used in the right way?
4. Confirm unit tests are sufficient and sensible.

### Checks for reviewer


tests/plans/unit_tests/test_verify_undulator_gap.py
```diff
@@ -0,0 +1,106 @@
...
+    fake_undulator.gap_motor.set = MagicMock()
+    fake_dcm.energy_in_kev.user_readback.sim_put(14000)
+
+    assert verify_undulator_gap(fake_undulator, fake_dcm) is True
+    mock_logger.warning.assert_called_once()
+    fake_undulator.gap_motor.set.assert_called_once_with(5.674)
+
+
+@patch("dodal.plans.verify_undulator_gap.LOGGER")
+def test_if_gap_is_correct_then_return_true_with_no_warnings(
```
olliesilvester, line 95: I'm not sure if this is technically a unit test but it seems useful to have anyway

src/dodal/plans/verify_undulator_gap.py
```diff
@@ -0,0 +1,70 @@
...
+        table_start = False
+        for line in table:
+            line = line.strip()
+            if line.startswith("Units"):
+                table_start = True
+                continue
+            if table_start:
+                value = line.split()
+                energy_to_distance_table[float(value[0])] = float(value[1])
+    return energy_to_distance_table
```
DominicOram, line 18: Should: I think we can just do the same that we do in `parse_table` in `det_dist_to_beam_converter.py` here, which is:
```python
from numpy import loadtxt
rows = loadtxt(self.lookup_file, delimiter=" ", comments=["#", "Units"])
```
In fact, given this is a common GDA file format can we put it in some util?

src/dodal/plans/verify_undulator_gap.py
```diff
@@ -0,0 +1,70 @@
...
+) -> float:
+    table_energies = energy_to_distance_table.keys()
+    table_energy_value_to_use: float = 0
+    closest_distance_to_energy_from_table = dcm_energy
+    for energy in table_energies:
+        distance = abs(dcm_energy - energy)
+        if distance < closest_distance_to_energy_from_table:
+            closest_distance_to_energy_from_table = distance
+            table_energy_value_to_use = energy
+    return energy_to_distance_table[table_energy_value_to_use]
```
DominicOram, line 32: Should: If you load into a `numpy` array as suggest above you will be able to do this in one or two lines using `numpy`

src/dodal/plans/verify_undulator_gap.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def verify_undulator_gap(undulator: Undulator, DCM: DCM) -> bool:
```
DominicOram, line 35: Should: If this is a plan then we need to use `yield from bps.set` etc. throughout rather than plain old `.set()`. I'm not sure if it should be though, I wonder if it makes more sense to just have an `energy` signal on the undulator that when set does this, it will make the waiting on it to change nicer too. Alternatively we could have a `UndulatorDCM` composite device that we change energy on? @dperl-dls, @olliesilvester and @rtuck99 shall we have a chat about this?

DominicOram: Should: I also don't really like the name `verify` as it implies that we're just checking it. We're not because we're also moving it. I would just call it `set_undulator_gap_for_dcm_energy`

src/dodal/plans/verify_undulator_gap.py
```diff
@@ -0,0 +1,70 @@
...
+    If undulator access level is enabled, check that the undulator gap matches the
+    energy from the DCM. If it doesn't match, set the undulator gap.
+    """
+
+    access_level = undulator.gap_access.get()
+    if access_level == UndulatorGapAccess.DISABLED.value:
+        LOGGER.warning(
+            "Undulator gap access is disabled. Unable to verify undulator gap"
+        )
+        return False
```
DominicOram, line 46: Should: I think it would be better to just throw an exception here and not return True/False

src/dodal/plans/verify_undulator_gap.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # Check if undulator gap is close enough to the value from the DCM
+    current_gap = undulator.current_gap.get()
+
+    if abs(gap_to_match_dcm_energy - current_gap) > undulator.gap_discrepancy_tolerance:
+        LOGGER.warning(
+            f"Undulator gap mismatch. {abs(gap_to_match_dcm_energy-current_gap):.3f} is outside tolerance.\
+            Restoring gap to nominal value, {gap_to_match_dcm_energy}"
+        )
+        undulator.gap_motor.set(gap_to_match_dcm_energy).wait(10)
```
DominicOram, line 68: Must: We won't always want to wait, and the timeout may change. If we keep this as a plan (see above) we should pass in an option for if we want to wait or not

tests/plans/unit_tests/test_verify_undulator_gap.py
```diff
@@ -0,0 +1,106 @@
...
+def create_energy_distance_table(path):
+    # Create lookup text file similar to the lookup table we see on I03
+    with open(path, mode="a") as file:
+        file.write(
+            "#test file\n\
+            Units    eV    mm\n\
+            5700    5.4606\n\
+            14000   5.674\n\
+            6000    5.681"
+        )
```
DominicOram, line 46: Should: better to mock out the `file.open` than to create an actual file on disk

tests/plans/unit_tests/test_verify_undulator_gap.py
DominicOram: I think it's a good test, would basically be a unit test if you mock out the file open

src/dodal/plans/verify_undulator_gap.py
```diff
@@ -0,0 +1,70 @@
...
+    # Use the lookup table to get the undulator gap associated with this dcm energy
+    gap_to_match_dcm_energy = _get_closest_gap_to_dcm_energy(
+        dcm_energy, energy_to_distance_table
+    )
+
+    # Check if undulator gap is close enough to the value from the DCM
+    current_gap = undulator.current_gap.get()
+
+    if abs(gap_to_match_dcm_energy - current_gap) > undulator.gap_discrepancy_tolerance:
+        LOGGER.warning(
```
DominicOram, line 64: Should: Warning seems strong here, I don't think it's a problem, info seems better

DominicOram (changes requested):
Good start, think we should discuss a bit more though

src/dodal/devices/undulator.py
```diff
@@ -1,5 +1,20 @@
-from ophyd import Component, Device, EpicsMotor
+from enum import Enum
+
+from ophyd import Component, Device, EpicsMotor, EpicsSignalRO
+
+UNDULATOR_DISCREPANCY_THRESHOLD = 1e-3  # This property was read from GDA although it didn't appear to exist anywhere, it defaults to this value
```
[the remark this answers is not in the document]
npat113: Never. It's really just a tolerance for position.

src/dodal/plans/verify_undulator_gap.py
npat113: Det/distance is a simple straight line interpolation. Gap/Energy is more complex and we may need to just sit together and work out how it is going to operate as we switch gap across the energy range.

src/dodal/plans/verify_undulator_gap.py
dperl-dls: I like the idea of an UndulatorDCM composite device with an energy signal, that sounds clean to me, it would be great to be able to `yield from bps.abs_set(energy, 12700)`

olliesilvester: Yes, I think this is what we will go with

src/dodal/devices/undulator.py
```diff
@@ -1,5 +1,20 @@
...
 
 class Undulator(Device):
-    gap: EpicsMotor = Component(EpicsMotor, "BLGAPMTR")
+    gap_motor: EpicsMotor = Component(EpicsMotor, "BLGAPMTR")
+    current_gap: EpicsSignalRO = Component(EpicsSignalRO, "CURRGAPD")
+    gap_access: EpicsSignalRO = Component(EpicsSignalRO, "IDBLENA")
+    lookup_table_path: str = (
+        "/dls_sw/i03/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt"
+    )
+    gap_discrepancy_tolerance: float = UNDULATOR_DISCREPANCY_THRESHOLD
```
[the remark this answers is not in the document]
olliesilvester: If the set energy will now be in a composite device with the DCM, should this constant remain within the undulator class?

src/dodal/plans/verify_undulator_gap.py
olliesilvester: > Should: I think we can just do the same that we do in `parse_table` in `det_dist_to_beam_converter.py` here, which is:
> 
> ```python
> from numpy import loadtxt
> rows = loadtxt(self.lookup_file, delimiter=" ", comments=["#", "Units"])
> ```
> 
> In fact, given this is a common GDA file format can we put it in some util?

I don't think we even need the util function, we can just use `loadtxt` as is

src/dodal/devices/undulator.py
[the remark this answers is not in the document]
DominicOram: Yes, I think keep in the undulator actually because I think it is specific to the undulator not the DCM/Undulator system as a whole

src/dodal/devices/undulator_dcm.py
```diff
@@ -0,0 +1,72 @@
...
+# Composite device to handle changing beamline energies
+class UndulatorDCM(Device):
+    class EnergySignal(Signal):
+        def set(self, value, *, timeout=None, settle_time=None, **kwargs) -> Status:
+            access_level = self.parent.undulator.gap_access.get()
+            if access_level == UndulatorGapAccess.DISABLED.value:
+                LOGGER.error("Undulator gap access is disabled. Contact Control Room")
+                raise AccessError(
+                    "Undulator gap access is disabled. Contact Control Room"
+                )
```
DominicOram, line 36: Should: The raised exception will get logged by the RE anyway, no need to log twice

src/dodal/devices/undulator_dcm.py
```diff
@@ -0,0 +1,72 @@
...
+
+            # Get 2d np.array converting energies to undulator gap distance, from lookup table
+            energy_to_distance_table = _get_energy_distance_table(
+                self.parent.undulator.lookup_table_path
+            )
+            LOGGER.info(f"Setting DCM energy to {value:.2f} kev")
+            self.parent.dcm.energy_in_kev.move(value)
```
DominicOram, line 43: Should: `move` returns a status, we should `and` all the statuses and return that on this `set`

src/dodal/devices/undulator_dcm.py
```diff
@@ -0,0 +1,72 @@
...
+
+            if (
+                abs(gap_to_match_dcm_energy - current_gap)
+                > self.parent.undulator.gap_discrepancy_tolerance_mm
+            ):
+                LOGGER.info(
+                    f"Undulator gap mismatch. {abs(gap_to_match_dcm_energy-current_gap):.3f}mm is outside tolerance.\
+                    Moving gap to nominal value, {gap_to_match_dcm_energy:.3f}mm"
+                )
+                return self.parent.undulator.gap_motor.move(gap_to_match_dcm_energy)
```
DominicOram, line 61: Should: Rather than return this we should and it with the status as above and return that

src/dodal/devices/undulator_dcm.py
```diff
@@ -0,0 +1,72 @@
...
+
+            complete_status = Status()
+            complete_status.set_finished()
+            return complete_status
+
```
DominicOram, line 66: Should: As above, deal with status' differently

src/dodal/devices/undulator_dcm.py
```diff
@@ -0,0 +1,72 @@
...
+    pass
+
+
+def _get_energy_distance_table(lookup_table_path: str) -> ndarray:
+    return loadtxt(lookup_table_path, comments=["#", "Units"])
+
+
+def _get_closest_gap_to_dcm_energy(
+    dcm_energy: float, energy_to_distance_table: ndarray
+) -> float:
```
DominicOram, line 21: Nit: I would remove reference to the `dcm` here. Just `get_closest_gap_for_energy`

src/dodal/devices/undulator_dcm.py
```diff
@@ -0,0 +1,72 @@
...
+                    f"Undulator gap mismatch. {abs(gap_to_match_dcm_energy-current_gap):.3f}mm is outside tolerance.\
+                    Moving gap to nominal value, {gap_to_match_dcm_energy:.3f}mm"
+                )
+                return self.parent.undulator.gap_motor.move(gap_to_match_dcm_energy)
+
+            complete_status = Status()
+            complete_status.set_finished()
+            return complete_status
+
+    energy: EnergySignal = Component(EnergySignal)
```
DominicOram, line 67: Should: Can you add units on this please

DominicOram (changes requested):
Some comments in code. I think the structure of this is nicer, thanks!

src/dodal/devices/undulator_dcm.py
DominicOram: We should probably then have some tests that assert the statuses are an amalgamation of all the underlying ones

PR #242, opened 2023-11-24
Pin tip centring now takes a median of values

DominicOram:
Hot fixes part 2 of https://github.com/DiamondLightSource/hyperion/issues/996

1. Confirm tests pass and are sensible

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/areadetector/plugins/MXSC.py
+++ b/src/dodal/devices/areadetector/plugins/MXSC.py
@@ -1,9 +1,8 @@
-from datetime import datetime, timedelta
 from typing import List, Tuple
 
 import numpy as np
 from ophyd import Component, Device, EpicsSignal, EpicsSignalRO, Kind, Signal
-from ophyd.status import Status, SubscriptionStatus
+from ophyd.status import StableSubscriptionStatus, Status
 
 from dodal.log import LOGGER
 
@@ -11,7 +10,7 @@ from dodal.log import LOGGER
 def statistics_of_positions(positions: List[Tuple]):
     x_coords, y_coords = np.array(positions).T
 
-    median = (np.median(x_coords), np.median(y_coords))
+    median = (int(np.median(x_coords)), int(np.median(y_coords)))
     std = (np.std(x_coords), np.std(y_coords))
 
     return median, std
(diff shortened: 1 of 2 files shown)


src/dodal/devices/areadetector/plugins/MXSC.py
```diff
@@ -1,16 +1,31 @@
...
-from ophyd.status import Status, SubscriptionStatus
+from ophyd.status import StableSubscriptionStatus, Status
+
+from dodal.log import LOGGER
+
+
+def statistics_of_positions(positions: List[Tuple]):
+    x_coords, y_coords = np.array(positions).T
+
+    median = (int(np.median(x_coords)), int(np.median(y_coords)))
```
olliesilvester, line 13: Why do we want these to be integers?

DominicOram: Good question. They're pixel values so we assumed they were integers later on for doing things like drawing grids and putting values into ispyb. Maybe I should do this conversion at the point we definitely need integers though

DominicOram: Actually, there's so many places this could potentially break that I think conversion here makes sense, I will add a comment though.

DominicOram: So I added type hinting, which makes it clearer that these things are valid `Pixel` locations and so must be ints. Doesn't really answer your question 100% but hopefully makes it clearer?

src/dodal/devices/areadetector/plugins/MXSC.py
```diff
@@ -19,16 +34,37 @@ class PinTipDetect(Device):
 
     triggered_tip: Signal = Component(Signal, kind=Kind.hinted, value=INVALID_POSITION)
     validity_timeout: Signal = Component(Signal, value=5)
+    settle_time_s: Signal = Component(Signal, value=0.5)
+
+    tip_positions: List[Tuple] = []
```
olliesilvester: It looks like we don't need to define `tip_positions` here if it's also defined in `trigger()`

olliesilvester (approved):
This looks good, I think the tests are good too. I was going to comment on doing some tests regarding the logic of `StableSubscriptionStatus`, eg `test_trigger_status_timeout_if_pin_doesnt_stay_valid`, but I guess since this is a tested function in Ophyd then there's no need

PR #243, opened 2023-11-27
Create devices needed for aligning the DCM and mirrors when changing energy

rtuck99:
… for DCM/HFM/VFM auto alignment

Required for fix DiamondLightSource/hyperion#986

Can you please validate the dodal device changes against the actual EPICS endpoints, I've matched the endpoints that I believe are present in GDA for I03, with the exception of I0 which I am not really sure how to represent and is completely speculative.

All of the new contents of i03.py is magic that I have copied and pasted from elsewhere but am not sure if it is correct.

Also, whether it's worth having VFM and HFM as separate device classes or whether we should just have a generic FocusingMirror and instantiate horiz/vert instances.

### Checks for reviewer


src/dodal/devices/qbpm1.py
```diff
@@ -3,4 +3,9 @@
 
 
 class QBPM1(Device):
+    """Quadrant Beam Position Monitor"""
+
     intensity: EpicsSignalRO = Cpt(EpicsSignalRO, "-DI-QBPM-01:INTEN", kind=Kind.normal)
+    intensityC: EpicsSignalRO = Cpt(
```
DominicOram, line 9: Should: Could we have a more descriptive name than `IntensityC`?

src/dodal/devices/focusing_mirror.py
```diff
@@ -0,0 +1,145 @@
...
+    translation1_mm: EpicsMotor = Component(EpicsMotor, "X1")
+    translation2_mm: EpicsMotor = Component(EpicsMotor, "X2")
+
+    stripe: EpicsSignal = Component(EpicsSignal, "STRP:DVAL", string=True)
+    # apply the current set stripe setting
+    apply_stripe: EpicsSignal = Component(EpicsSignal, "CHANGE.PROC")
+
+    bragg_to_lat_lookup_table_path: str = (
+        "CONFIGURE_ME"  # configured in per-mirror beamline-specific setup
+    )
```
DominicOram, line 138: Should: If this is a required part of the device can you pass it in to the constructor?

src/dodal/devices/focusing_mirror.py
```diff
@@ -0,0 +1,145 @@
...
+    BARE = "Bare"
+    PLATINUM = "Platinum"
+
+
+class MirrorVoltageDevice(Device):
+    _actual_v: EpicsSignal = Component(EpicsSignal, "R")
+    _setpoint_v: EpicsSignal = Component(EpicsSignal, "D")
+    _demand_accepted: EpicsSignal = Component(EpicsSignal, "DSEV")
+
+    def set(self, value, *, timeout=None, settle_time=0, **kwargs):
```
DominicOram, line 28: Should: My understanding on what this function is doing is:
* Set the setpoint
* Periodically poll the `demand_accepted` signal until it is 1

One issue here may be that is it possible the demand accepted is still a stale 1 from the previous value?

In any case, I think a cleaner solution is to use [subscribe](https://nsls-ii.github.io/ophyd/_as_gen/ophyd.ophydobj.OphydObject.subscribe.html) on the `demand_accepted` and wait for the event where the value changes from 0 -> 1. One potential gotcha on this is you want to set the subscription up before you do the set, to avoid race conditions. There's some precedent on doing this [here](https://github.com/DiamondLightSource/dodal/blob/f5fba125239e8dab128f13cbb7fe5aa080570c01/src/dodal/devices/fast_grid_scan.py#L179)

rtuck99:
> Additionally we're starting to paint ourselves into a corner with all the references to `/dls_sw/i03/software/daq_configuration/...`. I think at the least we should pull this location into a constant. Ideally the locations of lookup tables etc. should be passed into devices when they are made in `i03.py` so that we can change it for each beamline.

I will address some of it here but have raised a separate issue for this as it turned out to be too involved for this PR to do for everything

https://github.com/DiamondLightSource/dodal/issues/264

src/dodal/devices/focusing_mirror.py
```diff
@@ -0,0 +1,145 @@
...
+                "_channel14_voltage_device",
+                "_channel15_voltage_device",
+                "_channel16_voltage_device",
+                "_channel17_voltage_device",
+                "_channel18_voltage_device",
+                "_channel19_voltage_device",
+                "_channel20_voltage_device",
+                "_channel21_voltage_device",
+            ]
+        ]
```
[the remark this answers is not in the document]
rtuck99: This wasn't the original code - I ran into issues with ophyd magic at one point. However I suspect subsequent refactorings elsewhere may have fixed this as the more obvious solution now seems to work.

src/dodal/devices/focusing_mirror.py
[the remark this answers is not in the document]
rtuck99: Will update this with comments. I do have one question, in this code we create a Status `check_timer` purely in order to create an async timer (the status always times out and we don't need the result). I feel that it is possible to consider this somewhat of an abuse of the Status object, however I wasn't sure whether it's better to have my waits all managed by ophyd and on a threadpool/event loop managed by it or whether it was better to use async-await and then have to manage that myself.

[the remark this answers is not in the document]
DominicOram: I think `ophyd-async` will give us better management of that as we can use the inbuilt async framework better. I think for this particular case refactoring as in https://github.com/DiamondLightSource/dodal/pull/243#discussion_r1422984037 will solve it.

tests/devices/unit_tests/test_focusing_mirror.py
```diff
@@ -0,0 +1,89 @@
...
+    vfm_mirror_voltages: VFMMirrorVoltages,
+):
+    vfm_mirror_voltages._channel14_voltage_device._setpoint_v.set.return_value = (
+        NullStatus()
+    )
+    vfm_mirror_voltages._channel14_voltage_device._demand_accepted.sim_put(
+        DEMAND_ACCEPTED_OK
+    )
+
+    status: Status = vfm_mirror_voltages.voltage_channels[0].set(100)
```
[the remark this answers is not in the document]
rtuck99: I see the issue, but it doesn't trigger any warnings when I run ruff or in pycharm. Is there some other linter output I don't know about.

src/dodal/devices/qbpm1.py
rtuck99: No! I would probably have changed but honestly I don't know what this PV actually should be called. I believe it's some kind of aggregate measurement, but then if that's the case I don't know what `intensity` is measuring. In any case, it's no longer needed so I will remove it.

DominicOram: Remove it is best. If something like this comes up in the future we should make the effort to understand it (by discussion with Neil or controls) then put in a more sensible name. Part of the aim of this project is to get a better understanding of what we're actually doing

tests/devices/unit_tests/test_focusing_mirror.py
```diff
@@ -0,0 +1,119 @@
...
+):
+    vfm_mirror_voltages_with_set._channel14_voltage_device._setpoint_v.set.return_value = Status(
+        success=False, done=True
+    )
+
+    status: StatusBase = vfm_mirror_voltages_with_set.voltage_channels[0].set(100)
+    try:
+        status.wait()
+    except Exception:
+        pass
```
DominicOram, line 74: Should: 
```suggestion
    with pytest.raises(Exception):
        status.wait()
```

tests/devices/unit_tests/test_focusing_mirror.py
```diff
@@ -0,0 +1,119 @@
...
+    ].set(100)
+
+    actual_exception = None
+    try:
+        status.wait()
+    except Exception as e:
+        actual_exception = e
+
+    # Cannot assert because ophyd discards the original exception
+    # assert isinstance(actual_exception, WaitTimeoutError)
```
DominicOram, line 98: Should: is it not accessible through `__cause__`?

tests/devices/unit_tests/test_focusing_mirror.py
```diff
@@ -0,0 +1,119 @@
...
+
+    status: StatusBase = vfm_mirror_voltages_with_set_timing_out.voltage_channels[
+        0
+    ].set(100)
+
+    actual_exception = None
+    try:
+        status.wait()
+    except Exception as e:
+        actual_exception = e
```
DominicOram, line 95: Should: I think this removes the need for the assert too
```suggestion
    with pytest.raises(Exception) as e:
        status.wait()
```

tests/devices/unit_tests/util/test_lookup_tables.py
```diff
@@ -0,0 +1,22 @@
+from pytest import mark
+
+from dodal.devices.util.lookup_tables import (
+    linear_interpolation_lut,
+)
+
+
```
DominicOram, line 7: Should: Would be good to have a test in here for a non-monatomic file too

src/dodal/devices/util/lookup_tables.py
```diff
@@ -0,0 +1,42 @@
...
+        if not np.all(np.diff(s_values) > 0):
+            LOGGER.error(
+                f"Configuration file {filename} lookup table does not monotonically increase or decrease."
+            )
+            raise AssertionError(
+                f"Configuration file {filename} lookup table does not monotonically increase or decrease."
+            )
+
+    def s_to_t2(s: float) -> float:
+        # XXX numpy.interp doesn't do extrapolation, whereas GDA does, do we need this?
```
DominicOram, line 39: Should: As I said in the other PR, I still think this comment doesn't add anything. I don't think we need extrapolation so happy to leave it to raise an exception if it's ever tried

DominicOram (changes requested):
Apologies, a few comments but mostly just on tidying things up. Otherwise good work, thanks.

src/dodal/devices/focusing_mirror.py
```diff
@@ -0,0 +1,152 @@
...
+            return Status(success=True, done=True)
+
+        LOGGER.debug(f"setting {setpoint_v.name} to {value}")
+        demand_accepted_status = Status(self, DEFAULT_SETTLE_TIME_S)
+
+        subscription: dict[str, Any] = {"handle": None, "old_value": None}
+
+        def demand_check_callback(old_value, value, **kwargs):
+            # old_value parameter is unreliable when first called on subscribe, therefore we must
+            # save the current value and compare later to ensure that we definitely trigger on a rising edge.
```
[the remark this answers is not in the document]
DominicOram: It could possibly be that in your test you missed doing a `sim_put` to put `demand_accepted` low before doing the `set` on the voltages?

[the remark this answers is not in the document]
DominicOram: As discussed, issue here might be that we want to be sure the demand accepted is high before subscribe and that it goes 1->0->1. If it is at 0 to start then it's probably undefined behaviour and we want to throw an exception

pyproject.toml
```diff
@@ -76,9 +76,9 @@ ignore_missing_imports = true # Ignore missing stubs in imported modules
 markers = [
     "s03: marks tests as requiring the s03 simulator running (deselect with '-m \"not s03\"')",
 ]
+# --cov=dodal --cov-report term --cov-report xml:cov.xml
```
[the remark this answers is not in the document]
rtuck99: I discovered buried in the 7-year old bug report in pycharm that apparently --no-cov will override --cov so I have added it to the pycharm launcher config and no longer have to change `pyproject.toml`

tests/devices/unit_tests/util/test_lookup_tables.py
rtuck99: test added

DominicOram (approved):
Great, thank you!

PR #250, opened 2023-11-30
Add a device to get results from Zocalo

dperl-dls:
Needed for https://github.com/DiamondLightSource/hyperion/issues/997

1. Run tests, including system tests with `fake_zocalo`
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,281 @@
...
+            self.channel,
+            _receive_result,
+            acknowledgement=True,
+            allow_non_recipe_messages=False,
+        )
+
+
+def trigger_zocalo(zocalo: ZocaloResults):
+    """A minimal utility plan which will wait for analysis results to be returned from
+    Zocalo, and bundle them in a reading."""
```
DominicOram, line 245: Should: I'm not sure if trigger is the right word here, it kind of implies it's just starting it and not waiting for the end. I think ultimately in Hyperion we might want to separate start/end more to do parallel processing but we can think about that when we get to it.

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,281 @@
...
+    yield from bps.read(zocalo)
+    yield from bps.save()
+
+
+def get_processing_results(
+    zocalo: ZocaloResults,
+) -> Generator[Any, Any, Union[Tuple[np.ndarray, np.ndarray], Tuple[None, None]]]:
+    """A minimal plan which will extract an xray centring result and crystal bounding
+    box size from the zocalo results."""
+    results_recieved = yield from bps.rd(zocalo.results_valid)
```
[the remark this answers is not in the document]
dperl-dls: this was because we can't have optional signal values to deal with no results being returned - is resolved by making the results arrays

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+        finally:
+            self._kickoff_run = False
+
+    async def describe(self) -> dict[str, Descriptor]:
+        return OrderedDict(
+            [
+                (
+                    self._name + "-results",
+                    {
+                        "source": f"sim://{self._prefix}",
```
DominicOram, line 134: Should: This is confusing as `_prefix` is never set so it gives all these with a source of `sim://`. Do we need to give something that matches a backend? If not I think a better source description would just be something like `zocalo_service`

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+                        "shape": [-1, 3],
+                    },
+                ),
+                (
+                    self._name + "-bbox_sizes",
+                    {
+                        "source": f"sim://{self._prefix}",
+                        "dtype": "number",
+                        "shape": [-1, 3],
+                    },
```
DominicOram, line 155: Nit: I would pull this out as:
```python
zocalo_array_of_vectors =                     {
                        "source": f"sim://{self._prefix}",
                        "dtype": "number",
                        "shape": [-1, 3],
                    },
```
And reuse it in this and the above

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+        transport = lookup("PikaTransport")()
+        transport.connect()
+        return transport
+
+    def _subscribe_to_results(self):
+        transport = self._get_zocalo_connection()
+
+        def _receive_result(
+            rw: workflows.recipe.RecipeWrapper, header: dict, message: dict
+        ) -> None:
```
DominicOram, line 173: Should: It would be good to cover this in a unit test, even if it's a bit of a smoke test to confirm ack and the results are put in the right place

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+    yield from bps.trigger(zocalo, wait=True)
+    yield from bps.read(zocalo)
+    yield from bps.save()
+
+
+def get_processing_results(
+    zocalo: ZocaloResults,
+) -> Generator[Any, Any, Union[Tuple[np.ndarray, np.ndarray], Tuple[None, None]]]:
+    """A minimal plan which will extract the top ranked xray centre and crystal bounding
+    box size from the zocalo results."""
```
DominicOram, line 210: Could: The name of this function is deceptive as it only returns a single result. Would also be good if the doc string said that None meant no results were found

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+            ],
+        )
+
+    def _get_zocalo_connection(self):
+        zc = zocalo.configuration.from_file()
+        zc.activate_environment(self.zocalo_environment)
+
+        transport = lookup("PikaTransport")()
+        transport.connect()
+        return transport
```
DominicOram, line 166: Should: Can you pull this into a function in `zocalo_interaction` that can be used both here and in `ZocaloTrigger`

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+                        raw_results, key=lambda d: d[self.sort_key.value], reverse=True
+                    )
+                )
+
+            task = asyncio.create_task(_get_results())
+            try:
+                await asyncio.wait_for(asyncio.shield(task), self.timeout_s / 2)
+            except TimeoutError:
+                LOGGER.warning("Waited half of timeout for zocalo results - retrying")
+                await asyncio.wait_for(task, self.timeout_s / 2)
```
DominicOram, line 120: Should: Do we need this? It seems only really useful for testing?

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+            acknowledgement=True,
+            allow_non_recipe_messages=False,
+        )
+        LOGGER.info(f"Made zocalo queue subscription: {bool(subscription)}.")
+
+
+ZOCALO_READING_PLAN_NAME = "zocalo reading"
+
+
+def trigger_wait_and_read_zocalo(zocalo: ZocaloResults):
```
DominicOram, line 195: Nit: Can we just use https://nsls-ii.github.io/bluesky/generated/bluesky.plan_stubs.trigger_and_read.html?

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+    """A minimal utility plan which will wait for analysis results to be returned from
+    Zocalo, and bundle them in a reading."""
+
+    yield from bps.create(ZOCALO_READING_PLAN_NAME)
+    LOGGER.info("Running zocalo device .trigger()")
```
DominicOram, line 200: Nit: This log message isn't very friendly to the user, maybe "started waiting on zocalo results"?

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -0,0 +1,224 @@
...
+        None
+        if len(centres_of_mass) == 0  # type: ignore
+        else centres_of_mass[0] - np.array([0.5, 0.5, 0.5])  # type: ignore
+    )
+    LOGGER.info(f"Adjusted top centring result: {centre_of_mass}")
+    bbox_sizes = yield from bps.rd(zocalo.bbox_sizes, default_value=[])  # type: ignore
+    LOGGER.info(f"Bounding box sizes: {centres_of_mass}")
+    bbox_size = None if len(bbox_sizes) == 0 else bbox_sizes[0]  # type: ignore
+    LOGGER.info(f"Top bbox size: {bbox_size}")
+    return centre_of_mass, bbox_size
```
DominicOram, line 224: Nit: I think this is overly logged for production code, can we reduce them down a bit?

DominicOram (changes requested):
Thanks, some comments in code. Additionally I think that the way this is intended to work is now a bit confusing with the split of `ZocaloTrigger` (which isn't a device) and `ZocaloResults` (which is a device and must have `trigger` called on it). I think this is necessary confusion but can you write some docs (maybe just in this wiki for now) on how they should be used in general.

DominicOram (commented):
I'm struggling to run the system tests as the credentials aren't being created. @dperl-dls did you see a similar issue?

src/dodal/devices/zocalo/zocalo_results.py
dperl-dls: bbox and centre are separate results, aren't they? But sure, we can make it singular

src/dodal/devices/zocalo/zocalo_results.py
dperl-dls: I think this is important to avoid unnecessary crashes if we have a repeat of the issue we saw while testing. sure, maybe in  the long term this shouldn't be the solution, but it's at least necessary for soak testing imo.

src/dodal/devices/zocalo/zocalo_results.py
dperl-dls: got rid of the whole function

src/dodal/devices/zocalo/zocalo_results.py
DominicOram: That's true, it still caused me a double take though

src/dodal/devices/zocalo/zocalo_results.py
DominicOram: Ok, can we have some comments here about the issue we saw and so why we're doing a retry. Also, if we think we want a different long term solution can you make an issue for it?

DominicOram (approved):
Great, still haven't managed to run the system tests but trust that you have and I will sort it out at some point. Really love the docs, thanks!

PR #251, opened 2023-11-30
Add option to not set stub offsets

DominicOram:
See https://github.com/DiamondLightSource/hyperion/pull/1014

1. Confirm tests pass

### Checks for reviewer


dperl-dls (approved):
Agree that it's not very nice, but this is fine to me as a temporary fix

PR #258, opened 2023-12-07
Add logic to the OAV for re-scaling beam position and microns per pixel

### LINKED ISSUE #249: Add logic to OAV parameters to re-scale beam position and micronsperpixels
The beam position values found in `display.configuration` need to be re-scaled to an 1024x768 image e.g.
```
beam_center_i *= x_size / 1024
beam_center_j *= y_size / 768
```

This conversion should be done in the OAV paramaters right after the crosshair values are read.
The OAV image width/height are stored in PVs, for this we could use x_size and y_size from the `MJPG` plugin.

Similarly, micronsperpixels values need to be adjusted by
```
micronsperpixelsX *= 1024 / x_size
```

## Acceptance Criteria
* The beam position is updated to the correct value.
* Tests pass - including in hyperion.

noemifrisina:
Fixes #249 
The beam position and microns per pixels need to be rescaled to a "default" oav window size of (1024, 768)

1. Check that calculations are correct and run tests.
2. Run hyperion tests against this branch and check they still pass (scaling factor on i03 should amount to 1)

### Checks for reviewer


olliesilvester (changes requested):
Looks good - could we have a test for rescaled pixels per micron? Similar to the `test_extract_rescaled_beam_position`

olliesilvester (approved):

PR #273, opened 2024-01-03
Automatically find next run_number by default

olliesilvester:
Partially fixes https://github.com/DiamondLightSource/hyperion/issues/1008

I have added a function in dodal's utilities to look at all nexus files in a specified directory to find the next run number. `DetectorParams` is changed so that it uses this function for `run_number` by default, but so that it can still be overridden if explicitly provided.

1. Check that the above solution is sensible and implemented correctly
2. Check new tests

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/detector.py
+++ b/src/dodal/devices/detector.py
@@ -1,6 +1,7 @@
 from enum import Enum, auto
 from typing import Any, Optional, Tuple
 
+from hyperion.utils.get_run_number import get_run_number
 from pydantic import BaseModel, validator
 
 from dodal.devices.det_dim_constants import (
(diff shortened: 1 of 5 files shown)


dperl-dls (approved):
Looks good, your tests convince me.

It would be nice if we could make it not `optional` (eg using `default_factory`) but since we need the directory I don't really see a way around it.

olliesilvester:
What's the problem with using Optional?

dperl-dls:
Well, it's not _really_ optional, we have a default value we always use, so it's guaranteed to always be something, but the type checker will make us check it's `not None` anytime we want to use it now. There's a few other places where we have the same thing, but I haven't found a good way around it yet for things like this where you need to access another value to initialise it.

PR #279, opened 2024-01-08
Add PandA grid scans

olliesilvester:
Fixes #[870](https://github.com/DiamondLightSource/hyperion/issues/870)
For now we essentially copy the zebra fast grid scan devices and add extra panda functionality on top of it.

1. Confirm changes are sensible
2. Check new tests

### Checks for reviewer


src/dodal/beamlines/i03.py
```diff
@@ -204,6 +207,21 @@ def fast_grid_scan(
     )
 
 
+def panda_fast_grid_scan(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> PandAFastGridScan:
+    """Get the i03 fast_grid_scan device, instantiate it if it hasn't already been.
```
DominicOram, line 213: Should: The description here doesn't mention `panda`, I think some mention of how this is the one that should be used when triggering with the panda (rather than the FGS) would be good too

src/dodal/devices/panda_fast_grid_scan.py
```diff
@@ -0,0 +1,288 @@
...
+
+    @property
+    def end(self):
+        """Gives the point where the final frame is taken"""
+        # Note that full_steps is one indexed e.g. if there is one step then the end is
+        # refering to the first position
+        return self.steps_to_motor_position(self.full_steps - 1)
+
+    def is_within(self, steps):
+        return 0 <= steps <= self.full_steps
```
DominicOram, line 43: Could: This is exactly the same as the old one, can we just put it in a common place?

src/dodal/devices/panda_fast_grid_scan.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+class PandaGridScanParams(BaseModel, AbstractExperimentParameterBase):
```
DominicOram, line 46: Could: A lot of these params are the same FGS one, could we get them both to inherit from something common?

src/dodal/devices/panda_fast_grid_scan.py
```diff
@@ -0,0 +1,288 @@
...
+
+        self._target_count = self.device.y_steps.get()
+
+    def _running_changed(self, value=None, old_value=None, **kwargs):
+        if (old_value == 1) and (value == 0):
+            self.set_finished()
+            self.clean_up()
+
+    def clean_up(self):
+        self.device.status.clear_sub(self._running_changed)
```
DominicOram, line 177: Could: This is exactly the same as the old one, can we just put it in a common place?

tests/devices/unit_tests/test_panda_gridscan.py
```diff
@@ -0,0 +1,302 @@
...
+    PandaGridScanParams,
+    set_fast_grid_scan_params,
+)
+from dodal.devices.smargon import Smargon
+
+
+@pytest.fixture
+def fast_grid_scan():
+    FakeFastGridScan = make_fake_device(PandAFastGridScan)
+    fast_grid_scan: PandAFastGridScan = FakeFastGridScan(name="test fake FGS")
```
DominicOram, line 21: Nit: You can use `ophyd.sim.instantiate_fake_device` to do this in one line

DominicOram (changes requested):
Thanks, I know we probably don't want to go too far down the road of refactoring to make things common but there's some low hanging fruit that we could do. Also, as I said, if we pin to the `dodal` commits we can get the CI testing working too?

src/dodal/devices/panda_fast_grid_scan.py
```diff
@@ -0,0 +1,288 @@
...
+    KICKOFF_TIMEOUT: float = 5.0
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        def set_expected_images(*_, **__):
+            x, y, z = self.x_steps.get(), self.y_steps.get(), self.z_steps.get()
+            first_grid = x * y
+            second_grid = x * z
+            self.expected_images.put(first_grid + second_grid)
```
rtuck99, line 224: What's the significance of this offset x*y + x*z?

src/dodal/devices/panda_fast_grid_scan.py
```diff
@@ -0,0 +1,288 @@
...
+    z_step_size: float = 0.1
+    x_start: float = 0.1
+    y1_start: float = 0.1
+    y2_start: float = 0.1
+    z1_start: float = 0.1
+    z2_start: float = 0.1
+    x_axis: GridAxis = GridAxis(0, 0, 0)
+    y_axis: GridAxis = GridAxis(0, 0, 0)
+    z_axis: GridAxis = GridAxis(0, 0, 0)
+    runnup_distance_mm: float = 0.1
```
DominicOram, line 72: Nit: `runnup` looks odd to me, maybe it should be `run_up`?

src/dodal/devices/panda_fast_grid_scan.py
```diff
@@ -0,0 +1,145 @@
...
+                await_value(self.status, 1).wait()
+                st.set_finished()
+            except Exception as e:
+                st.set_exception(e)
+
+        threading.Thread(target=scan, daemon=True).start()
+        return st
+
+    def complete(self) -> DeviceStatus:
+        return GridScanCompleteStatus(self)
```
DominicOram, line 110: Must: This status looks at `position_counter` and waits for it to be the same as `set_expected_images` so I don't think it will ever complete for the panda.

tests/devices/unit_tests/test_panda_gridscan.py
```diff
@@ -0,0 +1,303 @@
...
+    grid_scan_params: PandAGridScanParams,
+    grid_position,
+    expected_x,
+    expected_y,
+    expected_z,
+):
+    motor_positions = grid_scan_params.grid_position_to_motor_position(grid_position)
+    np.testing.assert_allclose(
+        motor_positions, np.array([expected_x, expected_y, expected_z])
+    )
```
DominicOram, line 286: Should: These are actually testing things in the common grid scan code, which are presumably tested elsewhere, we don't need to test them again. May be true for other parts of this file too

DominicOram (changes requested):
Thanks, I appreciate this is taking ages to get in but almost there.

src/dodal/devices/panda_fast_grid_scan.py
olliesilvester: Thanks, I'd removed this after https://github.com/DiamondLightSource/dodal/pull/279#discussion_r1452384685, but the status did something slightly different

src/dodal/devices/panda_fast_grid_scan.py
olliesilvester: Re-added as the progress bar functionality is removed in the Panda version

DominicOram (approved):
Great, thank you!

PR #280, opened 2024-01-08
Hyperion 1036 Set energy: integrate set energy functions into hyperion plans

rtuck99:
Fixes DiamondLightSource/hyperion#1036

Refactors the DCM fixed offset so that now dodal populates an attribute on the device instead of making hyperion parse the beamline parameters.

Adds undulatorDCM as part of beamline I03 devices as it was missing.

- [x] Tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -95,21 +95,6 @@ def vfm_mirror_voltages(
     )
 
 
-@skip_device(lambda: BL == "s03")
-def hfm(
-    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
-) -> FocusingMirror:
-    return device_instantiation(
-        device_factory=FocusingMirror,
-        name="hfm",
-        prefix="-OP-HFM-01:",
-        wait=wait_for_connection,
-        fake=fake_with_ophyd_sim,
-        bragg_to_lat_lut_path=DAQ_CONFIGURATION_PATH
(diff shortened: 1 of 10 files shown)


tests/beamlines/unit_tests/test_i03.py
```diff
@@ -19,6 +19,7 @@ def test_device_creation():
     beamline_utils.clear_devices()
     devices = make_all_devices(i03, fake_with_ophyd_sim=True)
     for device_name in devices.keys():
+        print(device_name)
```
olliesilvester, line 22: We can remove this now

src/dodal/beamlines/i03.py
```diff
@@ -94,19 +95,6 @@ def vfm_mirror_voltages(
     )
 
 
-@skip_device(lambda: BL == "s03")
-def hfm(
```
olliesilvester, line 98: How come this device is removed now?

rtuck99: So initially when I first started working on the energy changes, I based it off of what GDA currently does, however after Neil put me right it turns out that we don't actually do any changes with it, everything is handled with feedback so no need to move it. Only the vertical mirror is needed.

olliesilvester (approved):
Looks good other than that print statement in the test

olliesilvester (approved):
Looks sensible, thanks!

PR #284, opened 2024-01-09
(#282) Extend DCM energy timeout from 10s to 30s

### LINKED ISSUE #282: DCM Energy timeout should be increased
While testing the changes for #280, Neil observed that the timeout value of 10s in undulator_dcm is too short as for long travels between extremes of the energy range, the travel time may approach or exceed this.

Therefore the timeout should be increased to a value that won't be reached during normal operation.

## Acceptance Criteria
* Energy can be changed over the full range without timing out

rtuck99:
Fixes #282 

1. Tests pass and changes are sensible.
2. Energy can now vary through the whole range without timeout

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/undulator_dcm.py
+++ b/src/dodal/devices/undulator_dcm.py
@@ -7,7 +7,8 @@ from dodal.devices.DCM import DCM
 from dodal.devices.undulator import Undulator, UndulatorGapAccess
 from dodal.log import LOGGER
 
-STATUS_TIMEOUT = 10
+ENERGY_TIMEOUT_S = 30
+STATUS_TIMEOUT_S = 10
 
 
 class AccessError(Exception):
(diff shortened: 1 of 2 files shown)


olliesilvester (approved):
Looks good. Are you sure we don't also need to increase the timeout of  `status &= self.parent.undulator.gap_motor.move` on line 69 of `undulator_dcm.py`?

rtuck99:
> Looks good. Are you sure we don't also need to increase the timeout of `status &= self.parent.undulator.gap_motor.move` on line 69 of `undulator_dcm.py`?



> Looks good. Are you sure we don't also need to increase the timeout of `status &= self.parent.undulator.gap_motor.move` on line 69 of `undulator_dcm.py`?

as per Neil's slack message:
Robert Tuck
  [10:19 AM](https://diamondlightsource.slack.com/archives/D063ZDQGK3R/p1704795553835599)
I notice in the code the undulator gap has the same timeout as the energy change - is 10s long enough for that or does that need to be increased as well?


Neil Paterson
:3robots:  [10:24 AM](https://diamondlightsource.slack.com/archives/D063ZDQGK3R/p1704795891194609)
undulator should be fast, but we can check once we have control of it again

PR #285, opened 2024-01-09
Fix for #283 undulator gap access check still passes even when undulator gap is disabled

### LINKED ISSUE #283: Undulator gap access check still passes even when undulator gap is disabled
When testing the energy changes, the undulator gap was disabled on i03 but AccessError exception wasn't raised.

## Acceptance Criteria
* The undulator gap access check works correctly
* There is a mechanism to allow the check to pass for the purpose of testing when the beamline is shut down and undulator gap access is disabled.

rtuck99:
Fixes #283 

1. Energy change should be correctly rejected when undulator gap access is disabled
2. This can be overridden by setting TEST_MODE=True in the undulator_dcm module

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/undulator_dcm.py
+++ b/src/dodal/devices/undulator_dcm.py
@@ -10,6 +10,9 @@ from dodal.log import LOGGER
 ENERGY_TIMEOUT_S = 30
 STATUS_TIMEOUT_S = 10
 
+# Enable to allow testing when the beamline is down, do not change in production!
+TEST_MODE = False
+
 
 class AccessError(Exception):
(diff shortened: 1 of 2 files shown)


olliesilvester (commented):
Looks okay to me, but it might be nicer to have the `TEST_MODE` as some sort of parameter rather than in this part of the code. What do you think? @dperl-dls @DominicOram

DominicOram:
`TEST_MODE` is fine for now. https://github.com/DiamondLightSource/hyperion/issues/1062 solves it properly

PR #289, opened 2024-01-14
Make the ophyd pin tip centre device more like the AD one

DominicOram:
See https://github.com/DiamondLightSource/hyperion/issues/863

1. Confirm tests pass and that the new pin tip detection device is more similar to the old one

### Checks for reviewer


dperl-dls (approved):
LGTM

PR #292, opened 2024-01-15
Fix type of detector size constants

DominicOram:
Detector dimensions are measured in floats and pixel size in integers, this actually enforces this rather than saying both could be floats or ints

1. Confirm this fixes the typing error on `get_data_shape` in `GridscanInternalParameters` in Hyperion

### Checks for reviewer


dperl-dls (approved):
nice!

PR #295, opened 2024-01-16
Fix multiple beam centre logging messages

olliesilvester:
Fixes Hyperion https://github.com/DiamondLightSource/hyperion/issues/1075

1. Check that new test fails when commenting out lines 216 and 217 on `oav_detector.py`
2. Check that this test passes when uncommenting
3. Confirm fix addresses issue 

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/oav/oav_detector.py
+++ b/src/dodal/devices/oav/oav_detector.py
@@ -204,6 +204,7 @@ class OAV(AreaDetector):
     def __init__(self, *args, params: OAVConfigParams, **kwargs):
         super().__init__(*args, **kwargs)
         self.parameters = params
+        self.subscription_id = None
 
     def wait_for_connection(self, all_signals=False, timeout=2):
         connected = super().wait_for_connection(all_signals, timeout)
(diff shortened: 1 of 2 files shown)


src/dodal/devices/oav/oav_detector.py
```diff
@@ -204,12 +204,17 @@ class OAV(AreaDetector):
...
     def wait_for_connection(self, all_signals=False, timeout=2):
         connected = super().wait_for_connection(all_signals, timeout)
         x = self.snapshot.x_size.get()
         y = self.snapshot.y_size.get()
 
         cb = partial(self.parameters.update_on_zoom, xsize=x, ysize=y)
-        self.zoom_controller.level.subscribe(cb)
+
+        if self.subscription_id:
+            self.zoom_controller.level.unsubscribe_all()
```
DominicOram, line 217: Should: As discussed, might be better just to unsubscribe the one we want otherwise we're causing side effects?

DominicOram (changes requested):
Thanks! One comment in code, otherwise looks good

DominicOram (approved):
Great, thank you

PR #296, opened 2024-01-17
Hyperion 1077 hyperion to populate ispyb params resolution instead of gda

rtuck99:
Fixes DiamondLightSource/hyperion#1077

1. Verify that the resolution calculation is appropriate. I ported the function that was being called in GDA's NewDetDistResolutionConverter although there are several versions in GDA and it seems that other experiment types / beamlines might require a different calculation.
2. The unit test cases I carried over from GDA, however the GDA tests are pretty bare-bones and artificial -hence odd values being injected (there are no tests in GDA for the ROI = true case and most of them are not for the mx case) and would be nice to have more and more realistic.

### Checks for reviewer


dperl-dls (changes requested):
This basically looks good.

Just for tidiness, I'd like to put all the detector stuff in a directory together (and export the stuff from current `detector.py` as well as this from its `__init__`)

For testing it could be good to compare:
- what GDA currently does, by just chainging settings in the gui and seeing the resulting "resolution"
- and what some analysis program (e.g. `dials.image_viewer`) thinks the res is for real datasets collected at some given settings (ask me about this)

dperl-dls (approved):
This looks good now, just make the tolerance on those tests like 5% or 10% or something (or change the values you're testing against, having manually confirmed they are close to GDA's behaviour), I think that's basically okay, and fix the linting please

PR #300, opened 2024-01-18
#298 Disconnect from zocalo on unstaging results device, to clear subscriptions

### LINKED ISSUE #298: ZocaloResults should subscribe on stage and unsubscribe on unstage
In https://github.com/DiamondLightSource/hyperion/issues/1080 we had some issues where Hyperion was ingesting the results of a GDA xray-centre. This ideally shouldn't be possible and so it would be best if `ZocaloResults` only subscribed whilst the device was staged.

## Acceptance Criteria
* On stage `ZocaloResults` subscribes to get zocalo messages
* If there are left over messages in the rabbitMQ queue at this point `ZocaloResults` should not get them
* On unstage `ZocaloResults` unsubscribes from messages

dperl-dls:
Fixes #298 
Should also merge https://github.com/DiamondLightSource/hyperion/pull/1084 for this to take effect

1. Run new system test

### Checks for reviewer


src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -106,17 +112,40 @@ def _clear_old_results(self):
         LOGGER.info("Clearing queue")
         self._raw_results_received = Queue()
 
-    def stage(self):
+    @AsyncStatus.wrap
+    async def stage(self, group=ZOCALO_STAGE_GROUP):
```
DominicOram, line 116: Should: It feels a bit weird to specify the group here rather than when doing the `bps.stage` in the plan. I don't think we do it anywhere else and I feel like it makes the plan a bit less readable 'cos you're waiting on something you haven't explicitly set up. Is there a reason to do it this way that I'm missing?

DominicOram (changes requested):
Sorry, one more comment. I also haven't tested ran the system test yet, will do so first thing Monday

src/dodal/devices/zocalo/zocalo_results.py
dperl-dls: nope, and it wouldn't even work, I just got it completely wrong. thanks for catching that

DominicOram (approved):
Great, thanks @dperl-dls!

PR #301, opened 2024-01-19
Allow reading aperture position name

stan-dot:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1032

### Checks for reviewer

### DIFF
--- a/tests/devices/unit_tests/test_aperture_scatterguard.py
+++ b/tests/devices/unit_tests/test_aperture_scatterguard.py
@@ -200,8 +200,8 @@ def test_aperture_positions_get_new_position_falsy(aperture_positions):
     large_missed_by_2_at_y = ApertureFiveDimensionalLocation(
         2.389, 42, 15.8, 5.25, 4.43
     )
-    new_position = aperture_positions.get_new_position(large_missed_by_2_at_y)
-    assert new_position is None
+    with pytest.raises(InvalidApertureMove):
+        aperture_positions.get_new_position(large_missed_by_2_at_y)
 
 
 def test_aperture_positions_get_new_position_robot_load_exact(aperture_positions):


stan-dot:
working through the failing tests.setters at lines 112 and 133 are troublesome
image

DominicOram (changes requested):
Sorry, I started reviewing this the other day but I think my signal was patchy so it never went through.

Some comments in code. Additionally, I think we need to be able to read the location of the device even if it is not Hyperion that has set it. This could be because Hyperion has just started or the device has been moved by some external program e.g. EPICS, GDA or physically moved.

src/dodal/devices/aperturescatterguard.py
```diff
@@ -9,21 +11,58 @@
...
 class InvalidApertureMove(Exception):
     pass
 
 
+class PositionName(str, Enum):
+    LARGE = "large"
+    MEDIUM = "medium"
+    SMALL = "small"
+    INVALID = "invalid"
+    ROBOT_LOAD = "robot_load"
```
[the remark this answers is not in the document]
DominicOram: This is a bit of a silly comment, sorry. We already have `AperturePositions`, name should just be another field on that

src/dodal/devices/aperturescatterguard.py
```diff
@@ -1,29 +1,67 @@
...
 from dodal.log import LOGGER
 
+Aperture5d = Tuple[float, float, float, float, float]
+
 
 class InvalidApertureMove(Exception):
     pass
 
 
+class ConfigurationRadiusMicrons(Optional[float], Enum):
```
DominicOram, line 22: Must: This fails on the tests with:
```python
ImportError while loading conftest '/scratch/ffv81422/hyperion/hyperion/tests/conftest.py'.
tests/conftest.py:10: in <module>
    from dodal.beamlines import beamline_utils, i03
../dodal/src/dodal/beamlines/i03.py:7: in <module>
    from dodal.devices.aperturescatterguard import AperturePositions, ApertureScatterguard
../dodal/src/dodal/devices/aperturescatterguard.py:22: in <module>
    class ConfigurationRadiusMicrons(Optional[float], Enum):
/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.10/lib/python3.10/typing.py:1109: in __mro_entries__
    raise TypeError(f"Cannot subclass {self!r}")
E   TypeError: Cannot subclass typing.Optional[float]
```

DominicOram (changes requested):
It's a bit hard to review this because it contains all the other code changes since it was branched so the view in https://github.com/DiamondLightSource/dodal/pull/301/files is a bit muddled. Could you either fix this by rebasing or doing a merge commit? We can talk through doing it together if you like. Additionally, there is one issue that stops me from running the unit tests. Once these are fixed I can review more thoroughly, thanks!

src/dodal/devices/aperturescatterguard.py
```diff
@@ -9,21 +11,58 @@
...
-    MEDIUM: Tuple[float, float, float, float, float]
-    SMALL: Tuple[float, float, float, float, float]
-    ROBOT_LOAD: Tuple[float, float, float, float, float]
+    LARGE: Aperture5d
+    MEDIUM: Aperture5d
+    SMALL: Aperture5d
+    ROBOT_LOAD: Aperture5d
+
+    # one micrometre tolerance
+    TOLERANCE_MM: float = 0.001
```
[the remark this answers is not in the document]
stan-dot: ok

src/dodal/devices/aperturescatterguard.py
```diff
@@ -58,31 +97,38 @@ def from_gda_beamline_params(cls, params):
...
+    def set(self, pos: Aperture5d) -> AndStatus:
         try:
             assert isinstance(self.aperture_positions, AperturePositions)
             assert self.aperture_positions.position_valid(pos)
         except AssertionError as e:
             raise InvalidApertureMove(repr(e))
+        self._update_name(pos)
         return self._safe_move_within_datacollection_range(*pos)
 
     def _safe_move_within_datacollection_range(
```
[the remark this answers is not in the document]
stan-dot: I'm happy to delete those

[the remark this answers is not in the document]
stan-dot: regarding the explanation of cases for invalid target position, or when it moves along Z - that was my attempt to reason about the code and then write down what each section does. it was not obvious to me from the code. I agree that they might be describing what happens incorrectly

src/dodal/devices/aperturescatterguard.py
```diff
@@ -14,85 +17,134 @@ class InvalidApertureMove(Exception):
...
     aperture_positions: Optional[AperturePositions] = None
+    selected_aperture = Cpt(Signal)
     APERTURE_Z_TOLERANCE = 3  # Number of MRES steps
 
     def load_aperture_positions(self, positions: AperturePositions):
         LOGGER.info(f"{self.name} loaded in {positions}")
         self.aperture_positions = positions
 
-    def set(self, pos: Tuple[float, float, float, float, float]) -> AndStatus:
+    def set(self, pos: ApertureFiveDimensionalLocation) -> AndStatus:
```
DominicOram, line 131: Should: Why not just set a `SingleAperturePosition` here? It simplifies things as we don't have to work out which position it is based on the values sent and we can just check it is equal to one of the objects defined in `AperturePositions`

DominicOram (changes requested):
Cleaner, thanks! One new comment as well as fixing the tolerance and removing the extra comments. Additionally, as I said at https://github.com/DiamondLightSource/dodal/pull/301#pullrequestreview-1834661761 it would be good to be able to read the position even if it was set outside Hyperion. There are two ways to do this:
1. Add callbacks on all the underlying motors so that we update `selected_aperture` when they change. Much like how `update_on_zoom` works now.
1. Override the read on the `selected_aperture` such that when you read from it it reads all the motor positions and works out which is selected from that

I think 2 will be cleaner as it doesn't rely on having loads of callbacks, which I think could get confusing.

src/dodal/devices/aperturescatterguard.py
stan-dot: hmm that sounds good. I was more hesitant to change this external API of the class as opposed to refactoring it internally

stan-dot: > Override the read on the selected_aperture such that when you read from it it reads all the motor positions and works out which is selected from that

I am not sure how to do that

stan-dot: will be extracted into a new ticket

DominicOram (approved):
Looks good, thanks. In the interests of getting this merged I have spun issues out to https://github.com/DiamondLightSource/hyperion/issues/1228

PR #304, opened 2024-01-24
Move plans, stubs and utils from dls-bluesky-core deprecated package to dodal

DiamondJoseph:
As discussed at i22 technical debt meeting

Includes pytest cases for added functionality:
`common` includes functions that are used for i22-linkam plans but may be useful to a wider audience
`plans`/`stubs` includes wrapping for standard bluesky plans to behave with BlueAPI, and generally useful ScanSpec plan

### Checks for reviewer

### DIFF
--- a/src/dodal/common/coordination.py
+++ b/src/dodal/common/coordination.py
@@ -1,0 +1,38 @@
+import uuid
+
+from dodal.common.types import Group
+
+
+def group_uuid(name: str) -> Group:
+    """
+    Returns a unique but human-readable string, to assist debugging orchestrated groups.
+
+    Args:
+        name (str): A human readable name
+
+    Returns:
+        readable_uid (Group): name appended with a unique string
+    """
+    return f"{name}-{str(uuid.uuid4())[:6]}"
+
+
+def inject(name: str):  # type: ignore
+    """
+    Function to mark a default argument of a plan method as a reference to a device
+    that is stored in the Blueapi context, as devices are constructed on startup of the
+    service, and are not available to be used when writing plans.
+    Bypasses mypy linting, returning x as Any and therefore valid as a default
+    argument.
+    e.g. For a 1-dimensional scan, that is usually performed on a consistent Movable
+    axis with name "stage_x"
+    def scan(x: Movable = inject("stage_x"), start: float = 0.0 ...)
+
+    Args:
+        name (str): Name of a device to be fetched from the Blueapi context
+
+    Returns:
+        Any: name but without typing checking, valid as any default type
+
+    """
+
+    return name
--- a/src/dodal/common/maths.py
+++ b/src/dodal/common/maths.py
@@ -1,0 +1,52 @@
+from typing import Tuple
+
+import numpy as np
+
+
+def step_to_num(start: float, stop: float, step: float) -> Tuple[float, float, int]:
+    """
+    Standard handling for converting from start, stop, step to start, stop, num
+    Forces step to be same direction as length
+    Includes a final point if it is within 1% of the final step, prevents floating
+    point arithmatic errors from giving inconsistent shaped scans between steps of an
+    outer axis.
+
+    Args:
+        start (float):
+            Start of length, will be returned unchanged
+        stop (float):
+            End of length, if length/step does not divide cleanly will be returned
+            extended up to 1% of step, or else truncated.
+        step (float):
+            Length of a step along the line formed from start to stop.
+            If stop < start, will be coerced to be backwards.
+
+    Returns:
+        start, adjusted_stop, num = Tuple[float, float, int]
+        start will be returned unchanged
+        adjusted_stop = start + (num - 1) * step
+        num is the maximal number of steps that could fit into the length.
+
+    """
+    # Make step be the right direction
+    step = abs(step) if stop >= start else -abs(step)
+    # If stop is within 1% of a step then include it
+    steps = int((stop - start) / step + 0.01)
+    return start, start + steps * step, steps + 1  # include 1st point
+
+
+def in_micros(t: float) -> int:
+    """
+    Converts between a positive number of seconds and an equivalent
+    number of microseconds.
+
+    Args:
+        t (float): A time in seconds
+    Raises:
+        ValueError: if t < 0
+    Returns:
+        t (int): A time in microseconds, rounded up to the nearest whole microsecond,
+    """
+    if t < 0:
+        raise ValueError(f"Expected a positive time in seconds, got {t!r}")
+    return int(np.ceil(t * 1e6))
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "opencv-python-headless",                                                                          # For pin-tip detection.
     "aioca",                                                                                           # Required for CA support with ophyd-async.
     "p4p",                                                                                             # Required for PVA support with ophyd-async.
+    "numpy",
 ]
 
 dynamic = ["version"]
@@ -87,6 +88,11 @@ addopts = """
     """
 # Doctest python code in docs, python code in src docstrings, test functions in tests
 testpaths = "docs src tests"
+
+[tool.coverage.report]
+exclude_also = [
+    '^"""', # Ignore the start/end of a file-level triple quoted docstring
+]
 
 [tool.coverage.run]
 data_file = "/tmp/dodal.coverage"
--- a/src/dodal/common/__init__.py
+++ b/src/dodal/common/__init__.py
@@ -1,0 +1,12 @@
+from .coordination import group_uuid, inject
+from .maths import in_micros, step_to_num
+from .types import MsgGenerator, PlanGenerator
+
+__all__ = [
+    "group_uuid",
+    "inject",
+    "in_micros",
+    "MsgGenerator",
+    "PlanGenerator",
+    "step_to_num",
+]
--- a/src/dodal/common/types.py
+++ b/src/dodal/common/types.py
@@ -1,0 +1,17 @@
+from typing import (
+    Annotated,
+    Any,
+    Callable,
+    Generator,
+)
+
+from bluesky.utils import Msg
+
+Group = Annotated[str, "String identifier used by 'wait' or stubs that await"]
+MsgGenerator = Annotated[
+    Generator[Msg, Any, None],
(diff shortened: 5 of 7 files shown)


DiamondJoseph:
I don't think the coverage is being calculated including tests marked async: tests/stubs/test_flyables.py::test_fly_and_collect is still marking flyables.py as 0 lines covered.

Also seems to be marking the beginning of triple quoted blocks of text as a testable line.

src/dodal/common/coordination.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    return f"{name}-{str(uuid.uuid4())[:6]}"
+
+
+def inject(name: str):  # type: ignore
+    """
+    Function to mark a default argument of a plan method as a reference to a device
+    that is stored in the Blueapi context.
+    Bypasses mypy linting, returning x as Any and therefore valid as a default
+    argument.
```
DominicOram, line 22: Should: This looks very weird as a function that's like a noop. Could you give an example in the docstrings in how to use it?

src/dodal/common/maths.py
```diff
@@ -0,0 +1,50 @@
+from typing import Tuple
+
+import numpy as np
+
+
+def step_to_num(start: float, stop: float, step: float) -> Tuple[float, float, int]:
+    """
+    Standard handling for converting from start, stop, step to start, stop, num
+    Forces step to be same direction as length
+    Includes a final point if it is within 1% of the end point (consistent with GDA)
```
DominicOram, line 10: Nit: In general I've tried to avoid comments mentioning GDA as it seems cargocult-y the decision should stand on it's own, not reference GDA

src/dodal/common/scanspecs.py
```diff
@@ -0,0 +1,33 @@
+from typing import List, Optional
+
+import numpy as np
+from scanspec.core import Frames
+from scanspec.specs import DURATION
```
DominicOram, line 5: Should: Having this file called `scanspecs` and referencing `scanspec` is going to get confusing. Can we either move this function into scanspec if it makes sense to do so or call this something like `scanspec_utils`?

src/dodal/plans/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+and any common functionality which may be useful to multiple plans extracted to stubs/.
+
+Plans:
+- Must have type hinted arguments, Should use the loosest sensible bounds
+- Must have docstrings describing behaviour and arguments of the function
+- Must not have variadic args or kwargs, Should pass collections instead
+- Must have optional argument named 'metadata' to add metadata to run(s)
+- Must add 'plan_args' to metadata with complete representation including defaults, None
+- Must add 'detectors', 'motors' metadata with list of names of relevant devices
+- Should pass 'shape' to metadata if the run's shape is knowable
```
DominicOram, line 17: Should: Can we enforce this in CI?

src/dodal/plans/scanspec.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Args:
+        detectors: List of readable devices, will take a reading at
+                                    each point
+        axes_to_move: All axes involved in this scan, names and
+            objects
+        spec: ScanSpec modelling the path of the scan
+        metadata: Key-value metadata to include
+                                                          in exported data, defaults to
+                                                          None.
```
DominicOram, line 36: Nit: Not sure what's happening with whitespace here?

src/dodal/plans/scanspec.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Returns:
+        Cycler: A new cycler
+    """
+
+    midpoints = spec.frames().midpoints
+    midpoints = {axes[name]: points for name, points in midpoints.items()}
+
+    # Need to "add" the cyclers for all the axes together. The code below is
+    # effectively: cycler(motor1, [...]) + cycler(motor2, [...]) + ...
```
DominicOram, line 78: :heart:

src/dodal/plans/wrapped.py
```diff
@@ -0,0 +1,50 @@
...
+        metadata (Optional[Mapping[str, Any]], optional): Key-value metadata to include
+                                                          in exported data.
+                                                          Defaults to None.
+
+    Returns:
+        MsgGenerator: _description_
+
+    Yields:
+        Iterator[MsgGenerator]: _description_
+    """
```
DominicOram, line 36: Should: I would rather see these removed than left at `_description_`

src/dodal/plans/wrapped.py
```diff
@@ -0,0 +1,50 @@
...
+    plan_args = (
+        {  # If bp.count added delay to plan_args, we could remove all md handling
```
DominicOram, line 38: Should: Personally, I don't like comments like this. What information is it actually giving a future developer? Is this body of work going to be done? Should it be done? If so make an issue for it and link the issue here saying we can remove this code when the issue is complete, at least then the developer has an action they are able to do.

src/dodal/stubs/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+Functions that yield from multiple stubs and offer a complete workflow
+should be moved to plans/.
+
+This package should never have a dependency on plans/.
+
+Stubs:
+- Must have type hinted arguments, Should use the loosest sensible bounds
+- Must have docstrings describing behaviour and arguments of the function
+- Must not have variadic args or kwargs, Should pass collections instead
+- Allow metadata to be propagated through if calling other stubs that take metadata
```
DominicOram, line 22: Should: As above, is this CI enforceable?

src/dodal/stubs/flyables.py
```diff
@@ -0,0 +1,60 @@
+import bluesky.plan_stubs as bps
+from bluesky.protocols import Flyable
+
+from dodal.common import MsgGenerator, group_uuid
+
+
+def fly_and_collect(
+    flyer: Flyable,
```
DominicOram, line 8: Should: The docstring says this should implement `Flyable` and `Collectable`, can the type hint reflect that?

src/dodal/stubs/flyables.py
```diff
@@ -0,0 +1,60 @@
...
+        MsgGenerator: Plan
+
+    Yields:
+        Iterator[MsgGenerator]: Bluesky messages
+    """
+    yield from bps.kickoff(flyer)
+    complete_group = group_uuid("complete")
+    yield from bps.complete(flyer, group=complete_group)
+    done = False
+    while not done:
```
DominicOram, line 49: Must: If the flyer never completes do we just end up with an infinite loop here? It feels like we need some way to set an actual timeout

src/dodal/stubs/wrapped.py
```diff
@@ -0,0 +1,141 @@
...
+
+"""
+Wrappers for Bluesky built-in plan stubs with type hinting
+"""
+
+Group = Annotated[str, "String identifier used by 'wait' or stubs that await"]
+T = TypeVar("T")
+
+
+def set_absolute(
```
DominicOram, line 17: Must: This is really horrible that we have to have this here. Is there a long term plan to add typing to `bluesky`?

DominicOram (changes requested):
Thanks, some comments in code. Some of it may be just differences in style but would be good to chat about it.

src/dodal/plans/__init__.py
DiamondJoseph: `test_plans_comply` in tests/plans/test_compliance checks


- [ ] Type hinted args, [loosest sensible bounds is judgement call for review]
- [ ] Has doc string [whether they adequately describe behaviour/args is left to reviewer, could parse the docstring str to ensure all args are mentioned if we wanted to]
- [ ] No VarArgs/VarKwargs
- [ ] Must have optional argument named 'metadata' with default None not {}
- [ ] Adds 'plan_args' to metadata with complete representation including defaults, None
- [ ] Must add 'detectors', 'motors' metadata with list of names of relevant devices
- [ ] Should pass 'shape' to metadata if the run's shape is knowable

DiamondJoseph: So everything that is knowable just from the signature/method object without trying to pass any arguments and step through it. 
Same for the the Stubs requirements, which is just the first 3 and the 4th as optional behaviour (e.g. not for move_relative, which doesn't call a stub that takes md).

src/dodal/common/coordination.py
DiamondJoseph: Added more detail

src/dodal/common/maths.py
DiamondJoseph: Adjusted to spell out the reason, may the planes return soon

src/dodal/common/scanspecs.py
DiamondJoseph: Will make a PR into Scanspec

src/dodal/plans/scanspec.py
DiamondJoseph: Fixed

src/dodal/plans/wrapped.py
DiamondJoseph: Adjusted to reference a PR I've opened on Bluesky with the required change.

src/dodal/stubs/__init__.py
DiamondJoseph: as above

src/dodal/stubs/flyables.py
DiamondJoseph: Adjusted, FlyableCollectable added to core.types, it's a useful construction for how we're using hardware scans.

src/dodal/stubs/flyables.py
DiamondJoseph: This plan isn't in use yet, will chat with TC/CF

src/dodal/common/scanspecs.py
DiamondJoseph: Merged https://github.com/dls-controls/scanspec/pull/93

DiamondJoseph:
I've pared this change down to just the functionality that may be useful for writing plans, with the wrapped plans and stubs moving back into BlueAPI, as the requirements only make sense there. With the scratch editable area now including the plans and stubs module going forward.

abbiemery (commented):
From what you explained this seams fine. Are there other services which depend on these methods that will have to be altered to reflect their move?

DominicOram (approved):
Great, ta

Issue #261, opened 2023-12-11
 Skip pin tip device on i03 system test


DominicOram:
Currently running `python tests/system_tests/test_i03_system.py` does not work as the i03 pin tip detect device cannot connect. This is because:
* It is an ophyd device and we're not set up to connect to them on system tests
* It is a PVA device and on the beamline without a PVA gateway

## Acceptance Criteria
* There is a generic way to skip some devices in the system tests

DominicOram:
@coretl - am I right in saying there's no plan for PVA gateway being rolled out?

coretl:
> @coretl - am I right in saying there's no plan for PVA gateway being rolled out?

That is correct, it is on the list of things to do for D-II, but there is no date on this yet. Definitely not this year.

PR #316, opened 2024-02-01
Add PV to read the barcode from the robot

DominicOram:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1109

1. Confirm unit tests pass
2. Confirm system tests pass (you will have to comment out the pin tip detection because of https://github.com/DiamondLightSource/dodal/issues/261)

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -18,6 +18,7 @@ from dodal.devices.oav.oav_detector import OAV, OAVConfigParams
 from dodal.devices.oav.pin_image_recognition import PinTipDetection
 from dodal.devices.panda_fast_grid_scan import PandAFastGridScan
 from dodal.devices.qbpm1 import QBPM1
+from dodal.devices.robot import BartRobot
 from dodal.devices.s4_slit_gaps import S4SlitGaps
 from dodal.devices.sample_shutter import SampleShutter
 from dodal.devices.smargon import Smargon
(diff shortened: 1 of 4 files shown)


rtuck99 (approved):

olliesilvester (approved):
Nice - we should say somewhere that any new dodal devices should be created in ophyd-async. Since it's specifically for waveform PVs, something like 'SingleIndexWaveformReadable' might be a more useful name. It might be better to put this class in a utils file

olliesilvester:
The system test fails when making the robot but the robot PV isn't in S03

DominicOram:
> Nice - we should say somewhere that any new dodal devices should be created in ophyd-async. 

Agreed, done so [here](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards#ophyd-vs-ophyd-async).

> Since it's specifically for waveform PVs, something like 'SingleIndexWaveformReadable' might be a more useful name. 
:+1: 

> It might be better to put this class in a utils file
I thought about this but from discussion [here](https://diamondlightsource.slack.com/archives/C04RRAM68LD/p1706792314614049) it sounds unlikely we have many of these PVs and this one will likely be fixed to be a proper string soon anyway, at which point I'll remove this.

> The system test fails when connecting to this and the panda...
Not sure about the panda one but I think the robot was a genuine issue with the device that I've now fixed. (note that if you're testing it today i03's OAV is off)

PR #318, opened 2024-02-02
Add some tests for status chaining util

dperl-dls:
Doesn't fix ~~Fixes~~ [Hyperion #1096](https://github.com/DiamondLightSource/hyperion/issues/1096)

doesn't replace ~~Replaces~~ `lambda`s with `partial` or other functions to ensure they are not called in the staging process.

adds some tests

1. Confirm old tests aren't broken
2. Confirm new tests make sense

### Checks for reviewer


DominicOram (changes requested):
I love the new tests but are you sure this fixes it? Running the same tests against the eiger in main passes fine

dperl-dls:
Oh :( Should have thought to check whether lambdas of functions actually execute them first

DominicOram (changes requested):
As discussed, given this didn't fix the issue can we change it back to lambdas but keep the tests. Then we'll merge this as it improves the codebase. But keep the original issue open as it doesn't fix it.

DominicOram (approved):
Looks good, even if didn't fix the issue

PR #326, opened 2024-02-12
325 close zocalo connections

### LINKED ISSUE #325: Don't leave open Zocalo connections
https://github.com/DiamondLightSource/hyperion/pull/1084 was incorrect, the `unstage()` method created and then closed a new connection, leaving the old one dangling. Instead, we should keep a reference to the connection made on subscription, and close that.

dperl-dls:
Fixes #325
Holds on to the Zocalo transport object created during stage and disconnects from it during unstage

1. See changes make sense and tests aren't broken
2. Confirm with analysis that issues seen the week of 06/02 did not repeat the weekend of 10-11/02
3. Confirm that now we are really not receiving zocalo results with hyperion turned off

### Checks for reviewer

### DIFF
--- a/tests/beamlines/unit_tests/test_i24.py
+++ b/tests/beamlines/unit_tests/test_i24.py
@@ -9,6 +9,8 @@ with patch.dict("os.environ", {"BEAMLINE": "i24"}, clear=True):
 
 
 def test_device_creation():
+    beamline_utils.BL = "i24"
+    beamline_utils.clear_devices()
     devices = make_all_devices(i24, fake_with_ophyd_sim=True)
     assert len(devices) > 0
     for device_name in devices.keys():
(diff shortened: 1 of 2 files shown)


tests/beamlines/unit_tests/test_i24.py
```diff
@@ -9,6 +9,8 @@
 
 
 def test_device_creation():
+    beamline_utils.BL = "i24"
+    beamline_utils.clear_devices()
```
DominicOram, line 13: Should: Can you fix this test by merging main instead of this please?

DominicOram (changes requested):
Looks good, thanks, only one minor comment. I added a test that I think tests this, it breaks on main and works here at least, could you check it makes sense?

olliesilvester (approved):
Looks good, I'm convinced the connection can't be left open now, since it unstages in the tidyup plan too.

Issue #305, opened 2024-01-25
Add ophyd-async Directory Provider


callumforrester:
ophyd-async detectors hold a reference to a [`DirectoryProvider`](https://blueskyproject.io/ophyd-async/main/user/generated/ophyd_async.core.DirectoryProvider.html#ophyd_async.core.DirectoryProvider), an object that can tell them where to write their data. It would be useful to create and access a directory provider object (a singleton) in dodal that can be applied to all ophyd-async detectors, so that their data-writing locations can be customised.

The use case is replicating the current data writing patterns of GDA on beamlines such as I22, for each new scan, we can tell the detectors to write to `<data>/<scan_number>-<detector_name>` or similar.

## Acceptance Criteria
* There is a means of injecting a common object into multiple ophyd-async detectors
* There is an implementation that can communicate with GDA, to ensure there are no clashes
* The I22 detectors (aravis and pilatus) are merged and using it

PR #327, opened 2024-02-12
Use ophyd-async DirectoryProvider

DiamondJoseph:
Exposes a global DirectoryProvider for ophyd-async devices to make use of, with a setter for services making use of dodal (e.g. BlueAPI) to choose their own implementation. Solves #305 
Extracted from https://github.com/DiamondLightSource/dodal/tree/directory_provider to allow i22 changes to be isolated and 

1. Instantiate an Ophyd-async device which requires a DirectoryProvider in its init
2. Confirm that the device can write to the directory provided.

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/beamline_utils.py
+++ b/src/dodal/beamlines/beamline_utils.py
@@ -1,10 +1,12 @@
 import inspect
+import tempfile
 from typing import Callable, Dict, Final, List, Optional, TypeVar, cast
 
 from bluesky.run_engine import call_in_bluesky_event_loop
 from ophyd import Device as OphydV1Device
 from ophyd.sim import make_fake_device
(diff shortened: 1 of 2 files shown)


DominicOram (approved):
This looks good to me, thanks. Do we have some docs on the `DirectoryProvider` and how it should be used?

DiamondJoseph:
> Do we have some docs on the `DirectoryProvider` and how it should be used?

[There's some incidental in this PR of documentation to ophyd-async](https://github.com/bluesky/ophyd-async/pull/107), which I will try and get over the line along with all the outstanding changes in dodal.

TLDR: call it, get a directory and an optional prefix for the files that an ophyd-async device is going to write. BlueAPI will have an implementation which calls off to GDA to get the current visit directory and current scan number and use those.

PR #336, opened 2024-02-14
Make the ophyd pin tip detection triggerable like the AD plugin

DominicOram:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1069

1. Confirm new tests pass
2. Confirm behavior matches that of `PinTipDetection` (without the averaging that does) 

### Checks for reviewer


src/dodal/devices/oav/pin_image_recognition/__init__.py
```diff
@@ -151,27 +134,32 @@ async def connect(self, sim: bool = False):
...
+                try:
+                    set_sim_value(
+                        self.triggered_tip, await self._get_tip_position(value)
+                    )
+                except Exception as e:
+                    LOGGER.warn(
+                        f"Failed to detect pin-tip location, will retry with next image: {e}"
+                    )
+                else:
+                    return
```
olliesilvester, line 155: When would this else statement be entered?

src/dodal/devices/oav/pin_image_recognition/__init__.py
```diff
@@ -151,27 +134,32 @@ async def connect(self, sim: bool = False):
...
-                (self._name, {"value": tip_pos, "timestamp": timestamp}),
-            ]
-        )
+    @AsyncStatus.wrap
+    async def trigger(self):
+        async def _set_triggered_tip():
+            """Monitors the camera data and updates the triggered_tip signal.
+
+            If a tip is found it will update the signal and stop monitoring
+            If no tip is found it will retry with the next monitored value
```
olliesilvester, line 143: As discussed I'll add a comment suggesting to add this to ophyd_async docs as a good example of `observe_value`

src/dodal/devices/oav/pin_image_recognition/__init__.py
```diff
@@ -151,27 +134,32 @@ async def connect(self, sim: bool = False):
...
+            """
+            async for value in observe_value(self.array_data):
+                try:
+                    set_sim_value(
+                        self.triggered_tip, await self._get_tip_position(value)
```
olliesilvester, line 148: Does there need to be something in `_get_tip_position` which raises an `Exception` if it finds invalid positions?

olliesilvester (approved):
In the trigger loop, if `self._get_tip_position(value)` gives an invalid position, this will exit the loop straight away rather than retry? Do we need to make sure an exception is raised so that it gets caught and the loop retries?

I might be wrong about the above, and looks good otherwise, so I'll approve

olliesilvester:
@dperl-dls if you try my new test before that 'raise' statement, it should fail, and pass with it

src/dodal/devices/oav/pin_image_recognition/__init__.py
```diff
@@ -15,20 +13,32 @@
...
-from dodal.devices.ophyd_async_utils import create_soft_signal_rw
+from dodal.devices.ophyd_async_utils import create_soft_signal_r, create_soft_signal_rw
 from dodal.log import LOGGER
 
+Tip = tuple[Optional[int], Optional[int]]
+
+
+def set_pin_value(*args, **kwargs):
+    if args[1] == PinTipDetection.INVALID_POSITION:
+        raise
```
dperl-dls, line 24: Don't you need to create some Exception here to raise? Even if this works somehow, I think you should raise a custom exception here so that the warning message later is more informative whether it was an invalid position or some other problem

dperl-dls (changes requested):
Seems okay to me, just one comment in code

src/dodal/devices/oav/pin_image_recognition/__init__.py
olliesilvester: `raise` just raises a generic `Exception`, and the try block catches a generic exception. But yes, I agree, might as well make a custom exception for this

dperl-dls (approved):
Looks good!

PR #343, opened 2024-02-16
Do string conversion on barcode waveform read

### LINKED ISSUE #342: Convert robot's barcode to a string type
Waveform PV reads output lists of type `str_`. This data type seems to cause issues when getting sent to ispyb, so we need to convert it to a regular string. We should try and remember this for future waveforms

## Acceptance Criteria
* `str_` converting to `str` in the robot device

olliesilvester:
Fixes #342

1. Confirm correct conversion

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/robot.py
+++ b/src/dodal/devices/robot.py
@@ -29,7 +29,7 @@ class SingleIndexWaveformReadable(StandardReadable):
     async def read(self) -> Dict[str, Reading]:
         underlying_read = await self.bare_signal.read()
         pv_reading = underlying_read[self.bare_signal.name]
-        pv_reading["value"] = pv_reading["value"][self.index]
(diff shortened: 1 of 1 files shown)


DominicOram (changes requested):
This maybe needs a little more investigation. Where we have `Sequence[str]` inside `SingleIndexWaveformReadable` should it actually be a `Sequence[_str]`?

olliesilvester:
According to https://pyepics.github.io/pyepics/overview.html#epics.caget:
`For integer or double array data from Epics waveform records, the regular value will be a numpy array (or a python list if numpy is not installed). The string representation will be something like ‘<array size=128, type=int>’`
I think a numpy.str_ is what you get when you convert a list of strings into a numpy aray, so this would make sense

DominicOram (approved):
This looks good, tests are failing on main too so happy to get this merged anyway

PR #347, opened 2024-02-21
Fix robot barcode type hint and setting pin value

olliesilvester:
Fixes first two bullet points of https://github.com/DiamondLightSource/hyperion/issues/1176

1. Confirm changes are sensible

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/robot.py
+++ b/src/dodal/devices/robot.py
@@ -2,7 +2,6 @@ from collections import OrderedDict
 from typing import Dict, Sequence
 
 from bluesky.protocols import Descriptor, Reading
-from numpy import str_
 from ophyd_async.core import StandardReadable
 from ophyd_async.epics.signal import epics_signal_r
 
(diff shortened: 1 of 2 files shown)


src/dodal/devices/robot.py
```diff
@@ -23,7 +22,7 @@ def __init__(
             pv (str): The waveform PV that contains a list of strings
             index (int, optional): The index to read. Defaults to 0.
         """
-        self.bare_signal = epics_signal_r(Sequence[str_], pv)
+        self.bare_signal = epics_signal_r(Sequence[str], pv)
```
DominicOram, line 25: Could: Does this mean we no longer need the cast to `str` below?

DominicOram (approved):
Thank you! One minor comment, take it or leave it. Otherwise looks good

src/dodal/devices/robot.py
olliesilvester: We still need the cast. Not sure exactly where it happens, but reading the PV  converts this from a `str` to a `str_`

PR #349, opened 2024-02-29
hyperion #1168 set stdout log to info

dperl-dls:
needed for https://github.com/DiamondLightSource/hyperion/issues/1186


DominicOram (approved):
I don't mind doing this but why is it required to fix the issue? Can you elaborate a bit on the reasoning.

dperl-dls:
Well, it's probably not *only* needed for this issue but in general too - but basically because there are otherwise too many ophyd logs potentially in `start_log` as well as in the test logs, etc.

PR #351, opened 2024-02-29
Fix pin tip detection to not use set_sim_value in production

DominicOram:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1181

1. Add:
```python 
@patch(
    "dodal.devices.oav.pin_image_recognition.set_sim_value",
    new=MagicMock(side_effect=Exception()),
)
```
onto `test_given_find_tip_fails_twice_when_triggered_then_tip_invalid_and_tried_twice` to simulate the issue mentioned in https://github.com/DiamondLightSource/hyperion/issues/1176. Confirm it fails on main and passes here

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/oav/pin_image_recognition/__init__.py
+++ b/src/dodal/devices/oav/pin_image_recognition/__init__.py
@@ -4,7 +4,7 @@ from typing import Optional
 
 import numpy as np
 from numpy.typing import NDArray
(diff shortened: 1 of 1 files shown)


dperl-dls (approved):
Seems good to do this in a consistent way everywhere

PR #361, opened 2024-03-06
Intial implementation of Synchrotron device

keithralphs:
Fixes [DAQ-4934](https://jira.diamond.ac.uk/browse/DAQ-4934)

1. Open up an iPython terminal on machine of the beamline network (e.g. a beamline workstation)
2. Along with the necessary imports for a Run engine and the Count plan, so the following import: from bluesky.run_engine import call_in_bluesky_event_loop, and run %autoawait call_in_bluesky_event_loop to cope with the current multi event loop problem described in https://github.com/bluesky/ophyd-async/issues/135.
3. Import the synchrotron .py file and creae an OASynchrotron device using the with DeviceCollector form
4. Run the Count plan on the device, apssing in lambda x,y: print(y) as the second parameter to the run engine to print out the emitted documents
5. Confirm that the full set of documents describing the readable signals of the device and their metadat are emitted and that no errors are reported

### Checks for reviewer


DiamondJoseph:
@stan-dot here's the initial implementation of the Synchrotron device, we're working on attaching units etc. more appropriately, but Synchrotron is Device, which is a tree of other Devices (ring_current, etc.), with leaves of signals (ring_current.current), and reading the Synchrotron reads all of the internal devices.

src/dodal/devices/synchrotron.py
```diff
@@ -14,19 +32,115 @@ class SynchrotronMode(Enum):
...
+        super().__init__(name=name)
+
+
+class OASynchrotron(OADevice):
+    def __init__(self, prefix: str = _SIGNAL_PREFIX, name: str = "synchrotron"):
+        self.machine_status = OASynchrotronMachineStatus()
+        self.topup = OASynchrotronTopUp()
+        self.ring_current = OASynchrotronRingCurrent()
+
+        self.components = [self.ring_current, self.topup, self.machine_status]
```
DiamondJoseph, line 116: I think we can replace 

self.components = [self.ring_current, self.topup, self.machine_status] with self.components = [child[1] for child in self.children()], which future proofs it a little. children() is defined on OADevice

or 
```
for _, device in self.children():
   device.thing()

```

src/dodal/devices/synchrotron.py
```diff
@@ -14,19 +32,115 @@ class SynchrotronMode(Enum):
...
+
+        super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def stage(self) -> None:
+        for device in self.components:
+            await device.stage().task
```
DiamondJoseph, line 123: Probably want to use an asyncio.gather here

src/dodal/devices/synchrotron.py
keithralphs: On reflection, this actually made the code more complex

src/dodal/devices/synchrotron.py
keithralphs: :+1:

DiamondJoseph (changes requested):
Once the ophyd_async signal changes go in we may need to refactor the tests for the new metadata

src/dodal/devices/synchrotron.py
```diff
@@ -1,9 +1,23 @@
...
+class PV(str, Enum):
+    STATUS_PREFIX = "CS-CS-MSTAT-01:"
+    TOP_UP_PREFIX = "SR-CS-FILL-01:"
+    SIGNAL_PREFIX = "SR-DI-DCCT-01:"
+    SIGNAL = "SIGNAL"
+    MODE = "MODE"
+    USRCNTDN = "USERCOUNTDN"
+    BEAM_ENERGY = "BEAMENERGY"
+    CNTDN = "COUNTDOWN"
+    ENDCNTDN = "ENDCOUNTDN"
```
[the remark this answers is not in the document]
keithralphs: Any particular reason? - I did it at the time to make the tes structure more iterative, but that has since become less of an issue

tests/devices/unit_tests/test_synchrotron.py
```diff
@@ -0,0 +1,223 @@
...
+    set_sim_value(sim_synchrotron.ring_current, RING_CURRENT)
+    set_sim_value(sim_synchrotron.machine_user_countdown, USER_COUNTDOWN)
+    set_sim_value(sim_synchrotron.topup_start_countdown, START_COUNTDOWN)
+    set_sim_value(sim_synchrotron.top_up_end_countdown, END_COUNTDOWN)
+    set_sim_value(sim_synchrotron.beam_energy, BEAM_ENERGY)
+    set_sim_value(sim_synchrotron.synchrotron_mode, MODE)
+    return sim_synchrotron
+
+
+async def test_oasynchrotron_read(sim_synchrotron: OASynchrotron):
```
DominicOram, line 130: Must: All these tests need marking with `@pytest.mark.asyncio` otherwise pytest won't pick them up. I suspect this is why codecov is upset

src/dodal/devices/synchrotron.py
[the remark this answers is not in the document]
DominicOram: Nit: My 2 cents is I feel like pulling them into constants in general is overkill. We haven't in any other device, do we think it should be the standard? (I particularly dislike `CNTDN = "COUNTDOWN"` where it feels like we've named the constant to be less clear than just the string). But take or leave it, minor nit pick really

DominicOram (changes requested):
Sorry to jump in, mainly just commenting about the testing issue. I also really think we should just delete the old one as part of this if we're replicating all it's functionality anyway. Then we can remove all the `OA` bits in the names

tests/devices/unit_tests/test_synchrotron.py
keithralphs: Callum put ayncio=auto in the pyproject.toml file (which I realise i haven't included in the push yet) so that the annotation wasn't necessary, is thatttt sufficient (assuming I include it)?

src/dodal/devices/synchrotron.py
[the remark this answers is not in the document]
DominicOram: > I guess it depends how long the non-OASynchrotron will persist. If we're all agreed that it can be done away with, then each value is only being used once and don't pollute the namespace.

Yh, I'll make a Hyperion PR to use this instead. We also have some generic "wait for topup" plans that I'll move into dodal.

DiamondJoseph:
Will benefit from https://github.com/bluesky/ophyd-async/pull/137 being merged prior to moving to depend on ophyd_async main (or next release), as the Descriptor is about to get a lot richer.

tests/devices/unit_tests/test_synchrotron.py
DominicOram: I don't know, I guess so. In which case we should remove all the asyncio marks from elsewhere. Sounds like a new PR though. Has Callum put that change in a separate one?

DiamondJoseph: Keith is away afternoons this week and all next, I'm happy to isolate and push that change (I'm blocked waiting on p38 hardware)

src/dodal/devices/synchrotron.py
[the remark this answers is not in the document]
keithralphs: Decided to go for the two enums approach to simplify importing in the test and prevent name clashes with the test data, as discussed.

src/dodal/devices/synchrotron.py
```diff
@@ -14,19 +28,48 @@ class SynchrotronMode(Enum):
...
+            SynchrotronMode, PV.STATUS_PREFIX + PV.MODE
+        )
+        self.machine_user_countdown = epics_signal_r(
+            float, PV.STATUS_PREFIX + PV.USRCNTDN
+        )
+        self.beam_energy = epics_signal_r(float, PV.STATUS_PREFIX + PV.BEAM_ENERGY)
+        self.topup_start_countdown = epics_signal_r(float, PV.TOP_UP_PREFIX + PV.CNTDN)
+        self.top_up_end_countdown = epics_signal_r(
+            float, PV.TOP_UP_PREFIX + PV.ENDCNTDN
+        )
```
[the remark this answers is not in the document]
keithralphs: Done, with new Enums

tests/devices/unit_tests/test_synchrotron.py
keithralphs: Joseph say the above is merged so should work now

DominicOram (approved):
I've done some work at https://github.com/DiamondLightSource/dodal/pull/379 and https://github.com/DiamondLightSource/hyperion/pull/1253 to remove the old Synchroton code but to avoid merge issues I think we just merge this with the both of them and I'll rebase my changes.

DiamondJoseph (approved):

pyproject.toml
```diff
@@ -14,7 +14,7 @@ classifiers = [
 description = "Ophyd devices and other utils that could be used across DLS beamlines"
 dependencies = [
     "ophyd",
-    "ophyd_async@git+https://github.com/bluesky/ophyd-async@ec5729640041ee5b77b4614158793af3a34cf9d8", #Use a specific branch from ophyd async until https://github.com/bluesky/ophyd-async/pull/101 is merged
+    "ophyd_async@git+https://github.com/bluesky/ophyd-async", #Use a specific branch from ophyd async until https://github.com/bluesky/ophyd-async/pull/101 is merged
```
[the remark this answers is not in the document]
DiamondJoseph: FWIW, it is still pinned to main, but the linked commit has made it into a release: do we still require main?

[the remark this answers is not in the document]
DominicOram: We don't need `main` but we do need `>0.2.0`, which isn't in pypi. So we could change this to `ophyd_async@git+https://github.com/bluesky/ophyd-async@0.2.1` if we want. Or we just leave it on `main` until `ophyd-async` gets a proper pypi release (https://diamondlightsource.slack.com/archives/C06JZPSU7LJ/p1709286528347809)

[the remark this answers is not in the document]
DominicOram: Sounds good to me, thanks Keith

DominicOram:
Code looks good to me. I'm not 100% sure why the container build is upset, clearly it's due to unpinning the dependency but that should be fine. I think we merge anyway.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ classifiers = [
 description = "Ophyd devices and other utils that could be used across DLS beamlines"
 dependencies = [
     "ophyd",
-    "ophyd_async@git+https://github.com/bluesky/ophyd-async@ec5729640041ee5b77b4614158793af3a34cf9d8", #Use a specific branch from ophyd async until https://github.com/bluesky/ophyd-async/pull/101 is merged
+    "ophyd_async@git+https://github.com/bluesky/ophyd-async",
```
callumforrester, line 17: Try this, think it might fix the container build, `ophyd-async` is the correct package name. Not sure why `ophyd_async` works at all to be honest.

```suggestion
    "ophyd-async@git+https://github.com/bluesky/ophyd-async",
```

DiamondJoseph: Looks like it was ophyd-async prior to being pinned, and the underscore change was hidden by the direct git dependency

callumforrester: Does anyone know if it _needs_ to be pinned to main anyway? @DiamondJoseph @DominicOram

callumforrester:
@DominicOram From discussion this morning I think we'd like to either fix or get rid of the container build, whichever is easier. Any objection to the latter? I don't think dodal in a container is ever used since it's a library

callumforrester:
PR if we decide to remove the build: https://github.com/DiamondLightSource/dodal/pull/388

DominicOram:
Yes, happy to remove it

pyproject.toml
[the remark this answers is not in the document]
coretl: Why can't we just tag a new release of ophyd-async right now?

pyproject.toml
coretl: container build doesn't work with main, needs to be pointing directly at a SHA. This is fixed in the copier template

pyproject.toml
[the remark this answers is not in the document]
DominicOram: > Why can't we just tag a new release of ophyd-async right now?

@abbiemery was looking at it. I believe that `ophyd-async` is still pinned to an unreleased `bluesky`. So yh, we could pin to GH version of `ophyd-async`. Either way, with removing the container build I think the dependency issue is outside the scope of this change

PR #363, opened 2024-03-06
Tidy up aperture reading

DominicOram:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1228

1. Confirm tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/util/motor_utils.py
+++ b/src/dodal/devices/util/motor_utils.py
@@ -1,0 +1,5 @@
+from ophyd import Component, EpicsMotor, EpicsSignalRO
+
+
+class EpicsMotorWithMRES(EpicsMotor):
+    motor_resolution: Component[EpicsSignalRO] = Component(EpicsSignalRO, ".MRES")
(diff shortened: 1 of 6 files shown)


src/dodal/devices/util/motor_utils.py
```diff
@@ -0,0 +1,5 @@
+from ophyd import Component, EpicsMotor, EpicsSignalRO
+
+
+class EpicsMotorWithMRES(EpicsMotor):
```
dperl-dls, line 4: Could: make this `ExtendedEpicsMotor` or something like that, so that we can add `.ACCL` as well

dperl-dls (approved):
LGTM, one optional comment

src/dodal/devices/util/motor_utils.py
DominicOram: I think if we want to add that we can do so when we have a usecase. I'm also keen to move to `ophyd-async` before then, where we hopefully get these into the motor proper

PR #371, opened 2024-03-08
Add raw set method to aperturescatterguard

dperl-dls:
Used in hyperion #818

### Checks for reviewer


DominicOram (approved):
I originally thought it was bad to provide this backdoor but after discussion with @dperl-dls I am convinced that the backdoor already exists and this just makes it more explicit that it's unsafe

Issue #378, opened 2024-03-13
Move to copier


DominicOram:
We should move to the new `copier` template rather than the old skeleton. https://github.com/DiamondLightSource/python-copier-template is the template, https://diamondlightsource.github.io/python-copier-template/main/tutorials/adopt-existing.html is some instructions

## Acceptance Criteria
* We have moved to the copier template
* All CI behaviour is maintained

joeshannon:
I might have some time to take a look at this - I have some experience as I previously converted blueapi.

Probably the most fiddly part was probably moving from the separate "developer" and "user" guides to a single manual. Is that the plan for dodal too?

DominicOram:
Yep, we can combine the docs too

PR #388, opened 2024-03-18
Remove Container Build from CI

callumforrester:
Fixes ongoing issues in CI caused by dependency pinning
Dodal in a container is not used and building conflicts with ophyd-async dependency pinning often, slowing down the merging of some pull requests. This PR removes it, which will hopefully speed things up.

1. Ensure CI still passes

### Checks for reviewer


DiamondJoseph:
Should remove the "Run in a Container" page of the docs too if we remove the container build.

Is there a ticket for switching to the Copier template?

DiamondJoseph (changes requested):
Remove the "Run in a Container" user docs too

callumforrester:
Done, can make a ticket for copier

DiamondJoseph:
> Done, can make a ticket for copier

If you could, then we can discuss with MX about whether we want to adopt or not. It should be comparatively simple as already using skeleton

DominicOram:
>Is there a ticket for switching to the Copier template?
https://github.com/DiamondLightSource/dodal/issues/378

DominicOram (approved):
Looks good to me, thanks!

DiamondJoseph (approved):

PR #390, opened 2024-03-18
Remove MXSC (edge detection) AD plugin device

dperl-dls:
Removes the OAV MXSC device now that we are only using the new PinTipDetection device

1. Check nothing is broken

### Checks for reviewer


olliesilvester:
It looks like some of the functions in `oav/oav_calculations.py`are no longer being used, eg `find_midpoint`

olliesilvester (approved):
Looks good, can you just double check the files I mentioned to see if they have things which can be removed?

PR #394, opened 2024-03-18
Move transmission to the experiment param base class

dperl-dls:
Needed for hyperion https://github.com/DiamondLightSource/hyperion/issues/1033

1. Run tests

### Checks for reviewer


src/dodal/parameters/experiment_parameter_base.py
```diff
@@ -1,7 +1,15 @@
...
 
+from pydantic import BaseModel
+
+
+class AbstractExperimentParameterBase(BaseModel, ABC):
+    pass
+
+
+class AbstractExperimentWithBeamParams(AbstractExperimentParameterBase):
+    transmission_fraction: float
```
olliesilvester, line 11: Wouldn't we always need a transmission fraction? Could this just be in the original `AbstractExperimentParameterBase`?

olliesilvester (approved):
Nice change, I had one comment on the Hyperion PR, https://github.com/DiamondLightSource/hyperion/pull/1271#pullrequestreview-1965716929

src/dodal/parameters/experiment_parameter_base.py
dperl-dls: we might want to expose endpoints which only do things without beam, e.g. optical centring, or load a sample only, and those don't need transmission

PR #395, opened 2024-03-19
Add aperture position readbacks #385

### LINKED ISSUE #385: Use existing PVs for Aperture Scatterguard location
We are too strict on the current location of the aperture scatterguard, leaving us unable to move them. We can instead use the existing PVs to know where we are. The PVs are, each of these will go high if in the selected position:

* `BL03I-MO-MAPT-01:Y:SMALL_CALC`
* `BL03I-MO-MAPT-01:Y:MEDIUM_CALC`
* `BL03I-MO-MAPT-01:Y:LARGE_CALC`

Unfortunately there are is no equivalent for robot load, we can either do that manually based on the Y or add that in to controls too

## Acceptance Criteria
* Reading the aperture scatterguard position is taken from the state of these PVs

katesmith280:
Fixes #385 

1. Run tests

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/aperturescatterguard.py
+++ b/src/dodal/devices/aperturescatterguard.py
@@ -88,8 +88,6 @@ class ApertureScatterguard(InfoLoggingDevice):
     scatterguard = Cpt(Scatterguard, "-MO-SCAT-01:")
     aperture_positions: Optional[AperturePositions] = None
     TOLERANCE_STEPS = 3  # Number of MRES steps
-    ROBOT_LOAD_TOLERANCE = 5  # Number of mm from ROBOT_LOAD_Y
-    ROBOT_LOAD_Y = 31.4  # Robot load position in mm
 
     class SelectedAperture(SignalRO):
         def get(self):
@@ -128,19 +126,20 @@ class ApertureScatterguard(InfoLoggingDevice):
         """
         Returns the current aperture position using readback values
         for SMALL, MEDIUM, LARGE. ROBOT_LOAD position defined when
-        mini aperture y <= ROBOT_LOAD_Y + ROBOT_LOAD_TOLERANCE.
+        mini aperture y <= ROBOT_LOAD.location.aperture_y + tolerance.
         If no position is found then raises InvalidApertureMove.
         """
         assert isinstance(self.aperture_positions, AperturePositions)
         current_ap_y = float(self.aperture.y.user_readback.get())
-
+        robot_load_ap_y = self.aperture_positions.ROBOT_LOAD.location.aperture_y
+        tolerance = self.TOLERANCE_STEPS * self.aperture.y.motor_resolution.get()
         if int(self.aperture.large.get()) == 1:
             return self.aperture_positions.LARGE
         elif int(self.aperture.medium.get()) == 1:
             return self.aperture_positions.MEDIUM
         elif int(self.aperture.small.get()) == 1:
             return self.aperture_positions.SMALL
-        elif current_ap_y <= self.ROBOT_LOAD_Y + self.ROBOT_LOAD_TOLERANCE:
+        elif current_ap_y <= +robot_load_ap_y + tolerance:
             return self.aperture_positions.ROBOT_LOAD
 
         raise InvalidApertureMove("Current aperture/scatterguard state unrecognised")
--- a/src/dodal/devices/aperturescatterguard.py
+++ b/src/dodal/devices/aperturescatterguard.py
@@ -139,7 +139,7 @@ class ApertureScatterguard(InfoLoggingDevice):
             return self.aperture_positions.MEDIUM
         elif int(self.aperture.small.get()) == 1:
             return self.aperture_positions.SMALL
-        elif current_ap_y <= +robot_load_ap_y + tolerance:
+        elif current_ap_y <= robot_load_ap_y + tolerance:
             return self.aperture_positions.ROBOT_LOAD
 
         raise InvalidApertureMove("Current aperture/scatterguard state unrecognised")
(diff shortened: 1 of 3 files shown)


DominicOram:
Adding @dperl-dls as a reviewer too as I don't have lots of time at the moment. We'll see who gets to it first though

src/dodal/devices/aperturescatterguard.py
```diff
@@ -123,22 +123,23 @@ def _set_raw_unsafe(self, positions: ApertureFiveDimensionalLocation) -> AndStat
...
+        current_ap_y = float(self.aperture.y.user_readback.get())
+
+        if int(self.aperture.large.get()) == 1:
+            return self.aperture_positions.LARGE
+        elif int(self.aperture.medium.get()) == 1:
+            return self.aperture_positions.MEDIUM
+        elif int(self.aperture.small.get()) == 1:
+            return self.aperture_positions.SMALL
+        elif current_ap_y <= self.ROBOT_LOAD_Y:
+            return self.aperture_positions.ROBOT_LOAD
```
dperl-dls, line 142: Should: I feel that the robot load position should still have some tolerance on it - probably the same `MRES` * some n steps as before - otherwise we'll still end up with similar issues where the state is unrecognised when the `current_ap_y` is just a hair short of the canonical `ROBOT_LOAD` position. Since this position is well out of the way of collisions we can probably increase the default number of steps to 5 or something.

tests/devices/unit_tests/test_aperture_scatterguard.py
```diff
@@ -175,32 +176,46 @@ def set_underlying_motors(
...
-    )
-    set_underlying_motors(ap_sg, large_missed_by_2_at_y)
+    ap_sg.aperture.large.sim_put(0)  # type: ignore
+    ap_sg.aperture.medium.sim_put(0)  # type: ignore
+    ap_sg.aperture.small.sim_put(0)  # type: ignore
+    ap_sg.aperture.y.set(34.0)  # type: ignore
+    assert ap_sg._get_current_aperture_position() == aperture_positions.ROBOT_LOAD
+
+
+def test_aperture_positions_robot_load_unsafe(
```
dperl-dls, line 210: Should: a test similar to these two for the above tolerance would be good

dperl-dls (changes requested):
Looks great, just one minor comment

src/dodal/devices/aperturescatterguard.py
DominicOram: Sorry, I suggested this logic to @katesmith280. I'm happy with a tolerance or without given that it's just a less than at the moment (so has some tolerance built in). If you think tolerance improves it then that's fine by me

katesmith280: I pushed a new change that decoupled the the robot load position (31.4mm) and the permitted tolerance (5 mm). So the aperture position value ROBOT_LOAD now has a bit more meaning, i.e. position plus tolerance.

dperl-dls: Oh, sorry. Yeah it's not critical, we could always address it later if it causes problems - I did see the `<=` but was just worried that that tolerance was only one-sided and won't work if the move to robot load stopped slightly before the position, so I do think it's more robust with that in.

src/dodal/devices/aperturescatterguard.py
```diff
@@ -89,11 +88,13 @@ class ApertureScatterguard(InfoLoggingDevice):
     scatterguard = Cpt(Scatterguard, "-MO-SCAT-01:")
     aperture_positions: Optional[AperturePositions] = None
     TOLERANCE_STEPS = 3  # Number of MRES steps
+    ROBOT_LOAD_TOLERANCE = 5  # Number of mm from ROBOT_LOAD_Y
```
dperl-dls, line 91: Oh, sorry, I notice this before - we can't use a hardcoded position here because they need to be able to be adjusted in a configuration file (used by both GDA and Hyperion), so we load the positions in to `self.aperture_positions` on creating this device. 5 mm is also too big of a tolerance, I was suggesting 5 times the spatial resolution of the motor.

So then in `_get_current_aperture_position` you can compare `current_ap_y` to that dynamically loaded position value with the calculated tolerance loaded from the motor, something like:

```Python
def _get_current_aperture_position(self) -> SingleAperturePosition:
    ...
    robot_load_ap_y = self.aperture_positions.ROBOT_LOAD.location.aperture_y
    tolerance = self.TOLERANCE_STEPS * self.aperture.y.motor_resolution.get()
    ...
    elif current_ap_y <=  + robot_load_ap_y + tolerance:
    ...
```

dperl-dls (changes requested):
Sorry I missed this detail in my last review, that part is quite important - Hopefully the new comment is clearer but feel free to ask me if you have any questions about it

src/dodal/devices/aperturescatterguard.py
katesmith280: Thanks @dperl-dls - I've added in these abstractions now and just updated the tests to also call them (instead of the hard coded within ant outside tolerance values)

src/dodal/devices/aperturescatterguard.py
```diff
@@ -123,22 +122,25 @@ def _set_raw_unsafe(self, positions: ApertureFiveDimensionalLocation) -> AndStat
...
+        current_ap_y = float(self.aperture.y.user_readback.get())
+        robot_load_ap_y = self.aperture_positions.ROBOT_LOAD.location.aperture_y
+        tolerance = self.TOLERANCE_STEPS * self.aperture.y.motor_resolution.get()
+        if int(self.aperture.large.get()) == 1:
+            return self.aperture_positions.LARGE
+        elif int(self.aperture.medium.get()) == 1:
+            return self.aperture_positions.MEDIUM
+        elif int(self.aperture.small.get()) == 1:
+            return self.aperture_positions.SMALL
+        elif current_ap_y <= +robot_load_ap_y + tolerance:
```
dperl-dls, line 142:
```suggestion
        elif current_ap_y <= robot_load_ap_y + tolerance:
```

dperl-dls (approved):
Great, thanks! I added a suggestion to fix a typo, otherwise looks good to me, you can merge this whenever you like now

PR #402, opened 2024-03-27
Add Beam Current Monitors for I22 and P38

### LINKED ISSUE #401: Tetramm Device
As an I22 scientist I would like to capture beam current data via the beamline's current monitors (connected to TETRAMM ADCs) so it can be used in my processing pipeline as calibration data.
Required for I22 pressure jump and stop flow experiments. 

## Acceptance Criteria
* Tetramm device available in dodal
* Configuration for i0 and it monitors included in i22 beamline module
* Configuration for mockup i0 monitor included on p38 beamline module

callumforrester:
Fixes #401 

Add a Tetramm Device based on the I22 user experiment in November, including tests. Add beam current monitors for i22 and mockup beam current monitor for p38.

1. Run unit tests
2. Confirm they pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/tetramm.py
+++ b/src/dodal/devices/tetramm.py
@@ -1,0 +1,273 @@
+import asyncio
+from enum import Enum
+from typing import Generator, Sequence
+
+import bluesky.plan_stubs as bps
+from bluesky.protocols import Hints
+from ophyd_async.core import (
+    AsyncStatus,
+    DetectorControl,
+    DetectorTrigger,
+    Device,
+    DirectoryProvider,
+    ShapeProvider,
+    StandardDetector,
+    set_and_wait_for_value,
+    set_signal_values,
+    walk_rw_signals,
+)
+from ophyd_async.core.device_save_loader import Msg
+from ophyd_async.epics.areadetector.utils import ad_r, ad_rw, stop_busy_record
+from ophyd_async.epics.areadetector.writers import HDFWriter, NDFileHDF
+from ophyd_async.epics.signal import epics_signal_r, epics_signal_rw
+
+
+class TetrammRange(str, Enum):
+    uA = "+- 120 uA"
+    nA = "+- 120 nA"
+
+
+class TetrammTrigger(str, Enum):
+    FreeRun = "Free run"
+    ExtTrigger = "Ext. trig."
+    ExtBulb = "Ext. bulb"
+    ExtGate = "Ext. gate"
+
+
+class TetrammChannels(str, Enum):
+    One = "1"
+    Two = "2"
+    Four = "4"
+
+
+class TetrammResolution(str, Enum):
+    SixteenBits = "16 bits"
+    TwentyFourBits = "24 bits"
+
+
+class TetrammGeometry(str, Enum):
+    Diamond = "Diamond"
+    Square = "Square"
+
+
+class TetrammDriver(Device):
+    def __init__(
+        self,
+        prefix: str,
+        name: str = "",
+    ):
+        self._prefix = prefix
+        self.range = ad_rw(TetrammRange, prefix + "Range")
+        self.sample_time = ad_r(float, prefix + "SampleTime")
+
+        self.values_per_reading = ad_rw(int, prefix + "ValuesPerRead")
+        self.averaging_time = ad_rw(float, prefix + "AveragingTime")
+        self.to_average = ad_r(int, prefix + "NumAverage")
+        self.averaged = ad_r(int, prefix + "NumAveraged")
+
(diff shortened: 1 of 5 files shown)


src/dodal/devices/tetramm.py
```diff
@@ -0,0 +1,273 @@
...
+        self.minimum_values_per_reading = minimum_values_per_reading
+        self.readings_per_frame = readings_per_frame
+
+    def get_deadtime(self, exposure: float) -> float:
+        # 2 internal clock cycles. Best effort approximation
+        return 2 / self.base_sample_rate
+
+    async def arm(
+        self,
+        num: int,
```
callumforrester, line 119: This parameter is ignore but it _has_ to be there to satisfy the interface, meaning the user can pass whatever they want. Possible an issue we should raise with ophyd-async.

src/dodal/devices/tetramm.py
```diff
@@ -0,0 +1,273 @@
...
+        trigger: DetectorTrigger,
+        exposure: float,
+    ) -> AsyncStatus:
+        if trigger not in {DetectorTrigger.edge_trigger, DetectorTrigger.constant_gate}:
+            raise ValueError("Only edge triggers are supported")
```
DiamondJoseph, line 124: The exception text doesn't seem to match the allowed cases: a constant_gate may be actually triggering off a rising or falling edge, but the message implies it can only support edge_trigger exactly?

src/dodal/devices/tetramm.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+class TetrammShapeProvider(ShapeProvider):
+    max_channels = 11
+
+    def __init__(self, controller: TetrammController) -> None:
+        self.controller = controller
+
+    async def __call__(self) -> Sequence[int]:
+        return [self.max_channels, self.controller.readings_per_frame]
```
DiamondJoseph, line 239: Can we use the value of `drv.num_channels`? That seems to be the actual shape when deployed on p38

DiamondJoseph (changes requested):

src/dodal/devices/tetramm.py
callumforrester: So the actual shape should be `[drv.num_channels, controller.readings_per_frame]`?

DiamondJoseph: That is my understanding. On p38 I have /entry/data/data with actual shape [1, 1000], if it is related to the number of channels, it shouldn't be the max_channels (wherever that comes from) but the configured number of channels

callumforrester: ...In conclusion, we're not sure, @Tom-Willemsen or @tpoliaw, do you have any input?

Tom-Willemsen: From a purely theoretical perspective the currently-configured number of channels makes more sense to me than the maximum number of channels... but I don't really know much about this specifically.

What is currently done in GDA?

DiamondJoseph: In GDA NcdTetrammDetector ~ `controller.dimensions = [*scan_shape, frameCount]`, but `getChannels()` is used as the value of `getDataDimensions()`

DiamondJoseph: (i.e., shapeProvider should only [frameCount] for the former, but [channels, frameCount] if the latter

tpoliaw: I wouldn't use p38 as an example of data shape, I don't think it is correctly set up to record all the channels whatever the value of `num_channels`. From memory, I think the data written out on I22 is always the same size as well, just some of the channels are 0 if the num channels is set below the max. **Please double check this**, it's been a while since I've worked with it.

callumforrester: Thanks, I'm going to suggest that we keep this as it was in November (i.e. as it is in this PR now) and raise additional tickets based on Tuesday testing if need be. What does everyone think?

DiamondJoseph: I'd agree, presumably it was not-wrong for November and we can investigate further when we need another TetrAMM: if the p38 one is configured differently we should try and get it configured the same, it has been updated to use the same IOC version as the one on i22 IIRC.

src/dodal/devices/tetramm.py
callumforrester: Continued in https://github.com/bluesky/ophyd-async/issues/162

DiamondJoseph (approved):

callumforrester:
Fix for CI errors: https://github.com/DiamondLightSource/dodal/pull/408

PR #403, opened 2024-03-27
Eiger stop now waits on all parts

olliesilvester:
Might fix https://github.com/DiamondLightSource/hyperion/issues/1167 , but can't be sure until some beamline testing has been done

1. Think about if this could be causing the issue and check fix

### Checks for reviewer


DominicOram (approved):
I'm not sure it will fix it but I also don't think this will cause harm. I suggest we merge and give it a go

PR #405, opened 2024-04-02
Update AdSim devices and configuration to use ophyd-async

### LINKED ISSUE #404: Update adsim to ophyd_async
The AreaDetector simulator backed by the Epics AreaDetectorSimulator is split into three parts: `dodal/src/dodal/devices/adsim.py` contains a 5-axis MotorBundle; `dodal/src/dodal/devices/areadetector/adsim.py` contains a simulated detector; `dodal/src/dodal/adsim.py` is equivalent to a beamline. Having these maintained and using ophyd_async may enable fast testing of plan behaviour and would give an easily run up simulated beamline for local development.

## Acceptance Criteria
* Blueapi can be run up with the areadetector simulator devices exposed and connected with ophyd_async signals and an example grid_scan may be run with data being emitted

DiamondJoseph:
Fixes #404 

1. Follow instructions for starting the AreaDetector simulator [included in the beamline module](https://github.com/DiamondLightSource/dodal/blob/adsim_async/src/dodal/beamlines/t01.py#L30) or from [the epics-containers docs](https://epics-containers.github.io/main/tutorials/launch_example.html)
2. Run the system tests

### Checks for reviewer


stan-dot:
@Relm-Arrowny  you've done some simulation work for p99  afaik, how useful is this PR?

Relm-Arrowny:
> @Relm-Arrowny you've done some simulation work for p99 afaik, how useful is this PR?

It nice to have, at the moment I either use fake or something from bluesky.sim so being able to just add area detector without thinking about it would save time.

DiamondJoseph:
How far does `fake_with_ophyd_sim` get you? Do you need something that is capable of writing files?

e.g.

```python
def det(
    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = True
) -> AravisDetector:
    return device_instantiation(
        AravisDetector,
        "det",
        "-EA-MAP-01:",
        wait_for_connection,
        fake_with_ophyd_sim,
        directory_provider=get_directory_provider(),
    )
```

Relm-Arrowny:
> How far does `fake_with_ophyd_sim` get you? Do you need something that is capable of writing files?

It got me as far as getting arming/trigger to work so I can use it to test plans and emitted the right docs.  [This ](https://github.com/DiamondLightSource/p99-bluesky/blob/834a91ea54da4df07c504bd3cdd9b5b27a99ec95/tests/test_stxm.py#L45)is an example of what I do atm.
As for writing file, I am not sure, atm I am not concern about writing files but I can see I may in the future.

src/dodal/devices/areadetector/adsim.py
```diff
@@ -1,47 +1,45 @@
...
+    DirectoryProvider,
+    StandardDetector,
+)
+from ophyd_async.epics.areadetector.controllers import ADSimController
+from ophyd_async.epics.areadetector.drivers import ADBase, ADBaseShapeProvider
+from ophyd_async.epics.areadetector.writers import HDFWriter, NDFileHDF, NDPluginStats
 
-from .adutils import Hdf5Writer, SingleTriggerV33, SynchronisedAdDriverBase
 
+class AdSimDetector(StandardDetector, HasHints):
```
DiamondJoseph, line 11: Already defined in ophyd-async

https://github.com/bluesky/ophyd-async/blob/e4fb8492bfad29dff9f539f6adc63c8152f68493/src/ophyd_async/epics/adsimdetector/_sim.py#L9

src/dodal/devices/adsim.py
```diff
@@ -1,13 +1,14 @@
...
-    z = Component(EpicsMotor, "M3")
-    theta = Component(EpicsMotor, "M4")
-    load = Component(EpicsMotor, "M5")
+    def __init__(self, prefix: str, name: str = "sim"):
+        self.x = Motor(prefix + "M1")
+        self.y = Motor(prefix + "M2")
+        self.z = Motor(prefix + "M3")
+        self.theta = Motor(prefix + "M4")
+        self.load = Motor(prefix + "M5")
+        super().__init__(name=name)
```
DiamondJoseph, line 14: I'd extend StandardReadable and have these motors as readable children

i.e.
```
def __init__(self, prefix, name):
        self.x = Motor(prefix+"M1")
        ...
```

tests/common/beamlines/test_device_instantiation.py
```diff
@@ -6,6 +6,8 @@
 from dodal.common.beamlines import beamline_utils
 from dodal.utils import BLUESKY_PROTOCOLS, make_all_devices
 
+ALL_BEAMLINES = {"adsim", "i03", "i04", "i04_1", "i23", "i24", "p38", "p45"}
```
DiamondJoseph, line 9: Looks like these tests now automatically populate the list of all beamlines, so this isn't required

src/dodal/beamlines/adsim.py
```diff
@@ -0,0 +1,38 @@
...
+set_log_beamline(BL)
+set_utils_beamline(BL)
+
+
+def stage(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> SimStage:
+    return device_instantiation(
+        SimStage,
+        "sim_motors",
```
DiamondJoseph, line 21: I'd have device.name and the function share the same name, so either make device.name "stage" or the function renamed to "sim_motors"

tests/common/beamlines/test_device_instantiation.py
ZohebShaikh: I thought it will be added every time the tests runs and git will show it as file changed ..so was thinking it will be better if it is in github..Can remove it if required

DiamondJoseph: It's been replaced with a call to this function

https://github.com/DiamondLightSource/dodal/blob/32ff2397a4e6c9d4f420c02182ffd0bbd02f8ad4/src/dodal/beamlines/__init__.py#L18

src/dodal/devices/areadetector/adsim.py
ZohebShaikh: https://github.com/bluesky/ophyd-async/blob/ce3b85ded7811843d7c984a4911c5799ff05eb55/src/ophyd_async/epics/demo/demo_ad_sim_detector.py#L10

This is the only place sim_detector is present and it is not exported 

Should I increase the version of ophyd_async but it will be a big change

DiamondJoseph: Ah, I didn't realise that we hadn't updated to ophyd-async yet. Let's find out who's doing that.

DominicOram (changes requested):
As someone who has never used the `ADSim` before I think this needs some more info. Ideally I would like a few system tests that spawn the `ADSim` and run some stuff tests against it using these files. If this is out of scope then I would at least like some documentation at the top of `adsim.py` pointing to how I would get started using the `ADSim` (and an issue to do the proper tests).

src/dodal/beamlines/adsim.py
```diff
@@ -0,0 +1,93 @@
...
+        BL,
+        Path("/scratch/adsim/bluesky"),
+        client=LocalDirectoryServiceClient(),
+    )
+)
+
+"""
+Usage Example
+-------------
+
```
stan-dot, line 31: needs explanation what adsim is, link do docs maybe

src/dodal/devices/areadetector/adaravis.py
```diff
@@ -1,101 +0,0 @@
...
-        Hdf5Writer,
-        suffix="HDF5:",
-        root="",
-        write_path_template="",
-    )
-    _priming_settings: Mapping[Signal, Any]
-
-    def __init__(self, *args, **kwargs) -> None:
-        super().__init__(*args, **kwargs)
-        self.hdf.kind = "normal"
```
stan-dot, line 25: why is `AdAravis` related?

stan-dot (approved):

DiamondJoseph:
@ZohebShaikh re: discussion on docs and use: since the AD sim that we use on workstations isn't going to be containerised or available in exactly the same form outside of Diamond anytime soon, CF and I discussed and we can instead back it with the example Simulation Beamline that is spun up by EC:

https://epics-containers.github.io/main/tutorials/launch_example.html

It has fewer axes, but still has a camera and should be possible to run up from anywhere with the tutorial.

[Please walk through the tutorial for this and alter the beamline to connect to these devices!](https://epics-containers.github.io/main/tutorials/launch_example.html)

src/dodal/devices/areadetector/adaravis.py
DiamondJoseph: The AdSim was a final remaining "used" ophyd device in this module, so with it being deleted the whole module may as well be removed.

stan-dot:
found a bug, possibly with #841 that when in ipython context a device does not connect while in the `dodal connect beamlinename` it does

src/dodal/beamlines/t01.py
callumforrester, line 1: Should: I prefer `adsim.py` to `t01.py`, we already have `training_rig.py`, so named because it doesn't cover a single beamline. It is a generic configuration that is reusable across test rigs and thus has a descriptive name so it is easily findable. I suggest the same here.

pyproject.toml
```diff
@@ -150,7 +151,7 @@ allowlist_externals =
     sphinx-build
     sphinx-autobuild
 commands =
-    tests: pytest -m 'not s03' --cov=dodal --cov-report term --cov-report xml:cov.xml {posargs}
+    tests: pytest -m 'not (s03 or adsim)' --cov=dodal --cov-report term --cov-report xml:cov.xml {posargs}
```
DiamondJoseph, line 154: `-m 'not s03' -m 'not adsim'` only respects the last one passed in.

src/dodal/beamlines/__init__.py
```diff
@@ -16,6 +16,7 @@
     "p47": "training_rig",
     "p48": "training_rig",
     "p49": "training_rig",
+    "t01": "adsim",
```
callumforrester, line 19: Could: Do we still need this? It will make `dodal connect t01` redirect to `adsim` but does anyone know what `t01` is?

DiamondJoseph: I don't think it hurts to have it. t01 is used in the docs for the [example-services](https://github.com/epics-containers/example-services) that deploy the containerised IOCs.

PR #410, opened 2024-04-03
Hyperion 1169 Attempted fix for failure re-arming the detector after beam dump

rtuck99:
Fixes #1169

1. To reproduce the issue without the fix, set dodal.devices.eiger.TEST_1169_INJECT = True, and dodal.devices.eiger.TEST_1169_FIX = False, this will cause the stage to be re-attempted, it should fail before it reaches the  cam acquire and fan wait
2. To test whether the fix works, leave the values at their defaults - it should now successfully re-arm the detector

If the tests pass we can tidy the code.

### Checks for reviewer


olliesilvester (approved):
Looks sensible, let's go down to test together at some point. Could you fix the mypy error?

DominicOram:
Merge this. @rtuck99 and @DominicOram to test tomorrow.

PR #421, opened 2024-04-05
Add generic ADAravis ophyd-async device for OAV/GigE camera

DiamondJoseph:
ADAravis are commonly used as OAV camera on multiple beamlines, including i22, p38. AKA GigE cameras.

Known Driver/Model combinations in use*:
Manta_G-125B, Manta_G-125C, Manta G-145B, Manta_G-145C, Manta G-235B, Manta G-235C, Manta G-609B, Manta G-895C
Mako G-040C, Mako G-125B, Mako G-125C, Mako G-234B, Mako G-234C, Mako G-507C
GC1020C

*retrieved from the Model_RBV of every OAV/Camera that I could find on every beamline's synoptic screen, with the proviso that many devices are currently unplugged or disabled and it's a lot of screens to look at, and I am very unobservant.

These devices share many of the same behaviours and fields, and if they can be treated generically would save a lot of duplicated code. However, it seems the installed firmware on each camera is not necessarily consistent. I propose getting this change to a mergable state, merging it, then opening new issues when inconsistency between the cameras cannot be resolved by updating their firmware: some fields may come from an internally stored read-only XML file, which is vulnerable to radiation damage. 

* inconsistent model naming for some Manta instances. 
* TriggerSource enum [as defined for IOC DB](https://github.com/areaDetector/ADGenICam/tree/master/GenICamApp/Db) records for many cameras inconsistent with actually exposed fields on cameras (often only line_1 or line_1, line_2 actually exposed, but handling required for external triggering).
* Some detector deadtimes not available from cited documentation due to discontinuation

1. Run unit tests
2. Check any known (and enabled) ADAravis OAVs for compatibility

### Checks for reviewer


src/dodal/devices/areadetector/adaravis_deadtime.csv
```diff
@@ -0,0 +1,36 @@
...
+Mako G-040,101e-6,Mono8,Mono12Packed,BayerRG8,BayerRG12Packed,YUV411Packed
+Mako G-040,140-e6,Mono12,BayerRG12,YUV422Packed
+Mako G-040,217-6,RGB8Packed,BGR8Packed,YUV444Packed
+
+Mako G-125,70e-6
+
+# Assume 12 bits: 10 bits = 275e-6
+Mako G-234,356-6,Mono8,BayerRG8,BayerRG12,BayerRG12Packed,YUV411Packed,YUV422Packed
+# Assume 12 bits: 10 bits = 563e-6
+Mako G-234,726-e6,RGB8Packed,BRG8Packed,YUV444Packed
```
Tom-Willemsen, line 32: I'm surprised this file parsed properly...

```suggestion
Manta G-2460,1304e-6,Mono12,BayerRG12,YUV422Packed
Manta G-2460,1961e-6,RGB8Packed,BGR8Packed,YUV444Packed

# cite: https://cdn.alliedvision.com/fileadmin/content/documents/products/cameras/various/appnote/GigE/GigE-Cameras_AppNote_PIV-Min-Time-Between-Exposures.pdf retrieved 2024-04-05
Manta G-609,47e-6

#cite: https://cdn.alliedvision.com/fileadmin/content/documents/products/cameras/Mako/techman/Mako_TechMan_en.pdf retrieved 2024-04-05
Mako G-040,101e-6,Mono8,Mono12Packed,BayerRG8,BayerRG12Packed,YUV411Packed
Mako G-040,140e-6,Mono12,BayerRG12,YUV422Packed
Mako G-040,217e-6,RGB8Packed,BGR8Packed,YUV444Packed

Mako G-125,70e-6

# Assume 12 bits: 10 bits = 275e-6
Mako G-234,356e-6,Mono8,BayerRG8,BayerRG12,BayerRG12Packed,YUV411Packed,YUV422Packed
# Assume 12 bits: 10 bits = 563e-6
Mako G-234,726e-6,RGB8Packed,BRG8Packed,YUV444Packed
```

src/dodal/devices/areadetector/adaravis.py
```diff
@@ -1,99 +1,186 @@
...
+            ),
+            config_sigs=(drv.acquire_time, drv.acquire),
+            name=name,
+        )
+
+    def get_external_trigger_gpio(self):
+        return self.controller.gpio_number
+
+    def set_external_trigger_gpio(self, gpio_number: AdAravisController.GPIO_NUMBER):
+        assert gpio_number in get_args(AdAravisController.GPIO_NUMBER)
```
Tom-Willemsen, line 180: This is a bit fiddly - asserting on introspection on an aliased type-hint... I'm not saying it's wrong but maybe could use a brief comment?

callumforrester (changes requested):
In order to fix #409, there should also be config for the OAV in `i22.py`

src/dodal/devices/areadetector/adaravis.py
```diff
@@ -1,99 +1,186 @@
...
+        self.trigger_source = ad_rw(ADAravisTriggerSource, prefix + "TriggerSource")
+        self.model = ad_r(str, prefix + "Model")
+        self.pixel_format = ad_r(str, prefix + "PixelFormat")
+        self.deadtime: Optional[float] = None
+        super().__init__(prefix, name=name)
+
+    async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
+        await super().connect(sim=sim, timeout=timeout)
+        model: str = await self.model.get_value()
+        pixel_format: str = await self.pixel_format.get_value()
```
Tom-Willemsen, line 96: There's an implicit assumption here that the pixel format won't change at runtime (i.e. between `connect()` and the plan running). I guess that assumption _might_ be correct, but it would be worth making this more explicit. Or maybe recalculating the deadtime in `arm` or somewhere a bit closer to where it's actually used.

src/dodal/devices/areadetector/adaravis_deadtime.csv
DiamondJoseph: It's just a csv until a value is actually selected, good catches, there's a reason I'm not in data entry

src/dodal/devices/areadetector/adaravis.py
```diff
@@ -1,99 +1,186 @@
...
+    line_2 = "Line2"
+    line_3 = "Line3"
+    line_4 = "Line4"
+
+
+_supported_external_triggers = (
+    DetectorTrigger.constant_gate,
+    DetectorTrigger.edge_trigger,
+)
+_csv_path = Path(__file__).parent / "adaravis_deadtime.csv"
```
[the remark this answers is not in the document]
DiamondJoseph: I was specifically trying to avoid that (mostly when thinking about every single supported camera, now that I've cut down to just the ones in use at Diamond I'm not as opposed). It's only ~30 entries, but there's got to be a good way of handling these configuration/lookup tables/magic numbers (delete as cynicism demands) generally...

src/dodal/devices/areadetector/adaravis.py
```diff
@@ -1,99 +1,186 @@
...
+    def get_external_trigger_gpio(self):
+        return self.controller.gpio_number
+
+    def set_external_trigger_gpio(self, gpio_number: AdAravisController.GPIO_NUMBER):
+        assert gpio_number in get_args(AdAravisController.GPIO_NUMBER)
+        self.controller.gpio_number = gpio_number
+
+    @property
+    def hints(self) -> Hints:
+        assert isinstance(self.writer, HasHints)
```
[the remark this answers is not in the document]
DiamondJoseph: Found a slightly neater way of resolving the issue (it was never going to fail the assertion) from the Pilatus patchset, have used that here now

src/dodal/devices/areadetector/adaravis.py
DiamondJoseph: Ideally it'd be in AdAravisDetector.prepare, as that's where it's used, arm is definitely closer. I'll see what I can do

src/dodal/devices/areadetector/adaravis.py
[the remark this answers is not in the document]
callumforrester: I would say in general there are three options for configuration/lookup tables/magic numbers:
1. Package with dodal, as you've done here
2. Package with IOC and make available via a PV
3. Put in some other database that can be queried at runtime

My personal favourite is 2. because it's the most standard thing you can do. Possibly worth asking @coretl if such a thing can be done long-term for aravis cameras. For now I think 1. is a good solution, certainly better than 3. which adds an unnecessary dependency. 

So we're doing roughly the right thing here, but I don't know if it will work with packaging, depending on how `pyproject.toml` is set up. I haven't checked, this is a smell thing, I just haven't always found packaging non-`.py` files in `src` to work reliably so I prefer to embed stuff in `.py` files because, again, it's the most standard thing you can do. 

I'm willing to be overruled on files, however I'd like a different format than csv. We seem to be using yaml everywhere else so if we want to agree on a consistent approach to lookup tables, shall we stick with that?

src/dodal/devices/areadetector/adaravis.py
DiamondJoseph: Moved to prepare

src/dodal/beamlines/i22.py
```diff
@@ -30,8 +34,50 @@ def it(
...
+        directory_provider=get_directory_provider(),
+    )
+
+
+def oav_driver(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> ADAravisDriver:
+    return device_instantiation(
+        ADAravisDriver,
+        "oav_driver",
```
DiamondJoseph, line 49: I believe these names are required due to device_instantiation device tree walking?

DiamondJoseph:
duplicate branch

PR #422, opened 2024-04-05
Refactor and create ophyd-async FGS devices

### LINKED ISSUE #413: Make a common ophyd-async FastGridScan device
The `FastGridScan` and `PandaFastGridScan` devices are very similar. We should create a common `FastGridScan` device which contains its core functionality. We should also use this opportunity to convert the FGS devices to `ophyd-async`

## Acceptance Criteria
* CommonFastGridScan device created which contains only the common behaviour between the two current types of scan.
* Children `ZebraFastGridScan` and `PandaFastGridScan` devices created from this
* We make it clear that the 'fast' in `FastGridScan` means that it's a hardware triggered scan
* Hyperion plans still work
* The `CommonFastGridScan`, `FastGridScan`, and `PandAFastGrid` scan are now in `ophyd-async`

olliesilvester:
Fixes #413 

See associated Hyperion change at: https://github.com/DiamondLightSource/hyperion/pull/1294

Combines the FGS parameters and devices for panda and zebra scans, and converts them to ophyd-async. 

1. Check overall logic is unchanged
2. Run tests

### Checks for reviewer


src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +193,153 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
         return dwell_time_ms
 
 
-class GridScanCompleteStatus(DeviceStatus):
+class PandAGridScanParams(GridScanParamsCommon):
```
olliesilvester, line 196: It might be nicer if I separate this file into two. One for the params and one for the devices

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +193,153 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+        self.y_steps = epics_signal_rw_rbv(int, "X_NUM_STEPS")
+        self.z_steps = epics_signal_rw_rbv(int, "X_NUM_STEPS")
+        self.x_step_size = epics_signal_rw_rbv(float, "X_STEP_SIZE")
+        self.y_step_size = epics_signal_rw_rbv(float, "Y_STEP_SIZE")
+        self.z_step_size = epics_signal_rw_rbv(float, "Z_STEP_SIZE")
+        self.x_start = epics_signal_rw_rbv(float, "X_START")
+        self.y1_start = epics_signal_rw_rbv(float, "Y_START")
+        self.y2_start = epics_signal_rw_rbv(float, "Y2_START")
+        self.z1_start = epics_signal_rw_rbv(float, "Z_START")
+        self.z2_start = epics_signal_rw_rbv(float, "Z2_START")
```
olliesilvester, line 244: Should add comments for any unclear parameters, and add units to everything

src/dodal/devices/fast_grid_scan.py
```diff
@@ -155,19 +170,16 @@ def grid_position_to_motor_position(self, grid_position: ndarray) -> ndarray:
...
-    a grid in x-z.
-
-    The grid specified is where data is taken e.g. it can be assumed the first frame is
-    at x_start, y1_start, z1_start and subsequent frames are N*step_size away.
+    Params for standard Zebra FGS, which includes the dwell time
     """
 
     dwell_time_ms: float = 10
 
+    def get_param_positions(self):
```
olliesilvester, line 178: Not sure if this way of adding to the dictionary is any cleaner than the previous way of handling this

src/dodal/devices/fast_grid_scan.py
olliesilvester: I'll do this in a separate PR, since that will be annoying to review combined with this

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +190,176 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+        self.run_cmd = epics_signal_rw(int, "RUN.PROC")
+        self.stop_cmd = epics_signal_rw(int, "STOP.PROC")
+        self.status = epics_signal_r(float, "SCAN_STATUS")
+
+        self.expected_images = ExpectedImages(parent=self)
+
+        self.motion_program = MotionProgram(prefix)
+
+        # Kickoff timeout in seconds
+        self.KICKOFF_TIMEOUT: float = 5.0
```
rtuck99, line 272: better to be KICKOFF_TIMEOUT_S

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +190,176 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+
+        self.COMPLETE_STATUS: float = 60.0
```
rtuck99, line 274: feels like a suspiciously time-in-secondsy status...

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +190,176 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+            "z1_start": self.z1_start,
+            "z2_start": self.z2_start,
+        }
 
-class FastGridScan(Device):
-    x_steps = Component(EpicsSignalWithRBV, "FGS:X_NUM_STEPS")
-    y_steps = Component(EpicsSignalWithRBV, "FGS:Y_NUM_STEPS")
-    z_steps = Component(EpicsSignalWithRBV, "FGS:Z_NUM_STEPS")
+    def is_invalid(self) -> bool:
+        if "GONP" in self.scan_invalid.source:
```
rtuck99, line 291: I feel like this could do with some explanation

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +190,176 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+        LOGGER.debug("Running scan")
+        await self.run_cmd.set(1)
+        LOGGER.info("Waiting for FGS to start")
+        await wait_for_value(self.status, 1, self.KICKOFF_TIMEOUT)
+        LOGGER.debug("FGS kicked off, exiting FGS kickoff thread")
 
-    x_step_size = Component(EpicsSignalWithRBV, "FGS:X_STEP_SIZE")
-    y_step_size = Component(EpicsSignalWithRBV, "FGS:Y_STEP_SIZE")
-    z_step_size = Component(EpicsSignalWithRBV, "FGS:Z_STEP_SIZE")
+    def collect(self):
```
rtuck99, line 309: Why do we have these empty methods?

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +190,176 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
-    )
-    x_counter = Component(EpicsSignalRO, "FGS:X_COUNTER")
-    y_counter = Component(EpicsSignalRO, "FGS:Y_COUNTER")
-    scan_invalid = Component(EpicsSignalRO, "FGS:SCAN_INVALID")
 
-    run_cmd = Component(EpicsSignal, "FGS:RUN.PROC")
-    stop_cmd = Component(EpicsSignal, "FGS:STOP.PROC")
-    status = Component(EpicsSignalRO, "FGS:SCAN_STATUS")
+class FastGridScan(FastGridScanCommon):
+    """Device for standard Zebra FGS. In this scan, the goniometer's velocity profile follows a parabloic shape between X steps,
```
rtuck99, line 321:
```suggestion
    """Device for standard Zebra FGS. In this scan, the goniometer's velocity profile follows a parabolic shape between X steps,
```

src/dodal/devices/fast_grid_scan.py
DominicOram: This is cargo-culted in from GDA, I actually think we can probably remove it

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +190,176 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+        running = await self.motion_program.running.get_value()
+        if running:
+            LOGGER.info(f"Motion program {curr_prog} still running, waiting...")
+            await wait_for_value(self.motion_program.running, 0, self.KICKOFF_TIMEOUT)
+
+        LOGGER.debug("Running scan")
+        await self.run_cmd.set(1)
+        LOGGER.info("Waiting for FGS to start")
+        await wait_for_value(self.status, 1, self.KICKOFF_TIMEOUT)
+        LOGGER.debug("FGS kicked off, exiting FGS kickoff thread")
```
DominicOram, line 307: Nit: This isn't really a thread, it's `asyncio` so I guess just `FGS kicked off`?

src/dodal/devices/fast_grid_scan.py
olliesilvester: I have no idea actually, they were left over from the old device. I'll remove them, unless @DominicOram objects

src/dodal/devices/fast_grid_scan.py
olliesilvester: Hmmm, we have loads of timeout variables without mentioning the units and they are _always_ in seconds, I'm inclined to leave it like this unless you care loads

src/dodal/devices/fast_grid_scan.py
DominicOram: I don't think they're needed, feel free to remove

src/dodal/devices/fast_grid_scan.py
olliesilvester: Same response as below!

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +193,165 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+
+
+class MotionProgram(Device):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        super().__init__(name)
+        self.running = epics_signal_r(float, prefix + "PROGBITS")
+        self.program_number = epics_signal_r(float, prefix + "CS1:PROG_NUM")
+
+
+class ExpectedImages(SignalRW):
```
DominicOram, line 216: Should: I don't think this should be writable and so just needs to inherit from `SignalR`. It's also going to be an int so can inherit from `SignalR[int]`

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +193,165 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+    def __init__(self, parent) -> None:
+        super().__init__(SimSignalBackend(int, "sim://expected_images"))
+        self.parent = parent
```
DominicOram, line 219: Should: 
```suggestion
    def __init__(self, parent: "FastGridScanCommon") -> None:
        super().__init__(SimSignalBackend(int, "sim://expected_images"))
        self.parent: "FastGridScanCommon" = parent
```

Means you can remove the type ignore later on

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +193,165 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
-        params.z1_start,
-        scan.z2_start,
-        params.z2_start,
-        scan.position_counter,
-        0,
-    )
+def set_fast_grid_scan_params(scan: FastGridScanCommon, params: GridScanParamsCommon):
+    assert set(params.get_param_positions().keys()) == set(
+        scan.movable_params.keys()
+    ), "Scan parameters don't match the scan device"
```
DominicOram, line 346: Could: We can get the type checker to check this.
```python

ParamType = TypeVar("ParamType", bound=GridScanParamsCommon)

class FastGridScanCommon(Generic[ParamType], StandardReadable, ABC):
    ...
    
class ZebraFastGridScan(FastGridScanCommon[ZebraGridScanParams]):
    ....
   
class PandAFastGridScan(FastGridScanCommon[PandAGridScanParams]):
    ....
    
def set_fast_grid_scan_params(scan: FastGridScanCommon[ParamType], params: ParamType):
   ....
```

Which will make the type checker highlight code like
```python
    assert isinstance(parameters.experiment_params, PandAGridScanParams)
    yield from set_flyscan_params(
        fgs_composite.zebra_fast_grid_scan, parameters.experiment_params
    )
```

I would be interested to see if @callumforrester or @dperl-dls have thoughts on using the type system like this.

DominicOram (changes requested):
Couple of comments but ran out of time for a full review, sorry

src/dodal/devices/fast_grid_scan.py
dperl-dls: I like it

rtuck99 (approved):
Approved as my comments addressed

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +196,165 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+        # grid
+        self.run_up_distance_mm = epics_signal_rw_rbv(float, "RUNUP_DISTANCE")
+        self.movable_params["run_up_distance_mm"] = self.run_up_distance_mm
 
-    def clean_up(self):
-        self.device.position_counter.clear_sub(self._notify_watchers)
-        self.device.status.clear_sub(self._running_changed)
 
+def set_fast_grid_scan_params(
+    scan: FastGridScanCommon[ParamType], params: GridScanParamsCommon
```
DominicOram, line 345: Must: `params` should be of type `ParamType`. This means we can then remove the assert and just trust the type system to catch when the params don't match

src/dodal/devices/fast_grid_scan.py
```diff
@@ -181,181 +196,165 @@ def non_integer_dwell_time(cls, dwell_time_ms: float) -> float:
...
+        super().__init__(SoftSignalBackend(int))
+        self.parent = parent
+
+    async def get_value(self):
+        x = int(await self.parent.x_steps.get_value())  # type: ignore
+        y = int(await self.parent.y_steps.get_value())  # type: ignore
+        z = int(await self.parent.z_steps.get_value())  # type: ignore
+        first_grid = x * y
+        second_grid = x * z
+        return first_grid + second_grid
```
DominicOram, line 230: Should: I think if you specify the type of the parent you can fix these type issues:
```suggestion
        self.parent: "FastGridScanCommon" = parent

    async def get_value(self):
        x = await self.parent.x_steps.get_value()
        y = await self.parent.y_steps.get_value()
        z = await self.parent.z_steps.get_value()
        first_grid = x * y
        second_grid = x * z
        return first_grid + second_grid
```

tests/devices/unit_tests/test_gridscan.py
```diff
@@ -25,97 +27,49 @@ def discard_status(st: Status | DeviceStatus):
...
-    mock_run_set_status = mock()
-    when(fast_grid_scan.run_cmd).put(ANY).thenReturn(mock_run_set_status)
-    fast_grid_scan.status.subscribe = lambda func, **_: func(1)  # type: ignore
 
-    status = fast_grid_scan.kickoff()
-
-    status.wait()
-    assert status.exception() is None
+@pytest.fixture
+async def panda_fast_grid_scan(request):
```
DominicOram, line 38: Should: Why do we need `request` here and in the other fixture? They don't seem to be used

DominicOram (changes requested):
Great, thanks, almost there. Couple of comments. Also, most of the tests only run against the zebra version, maybe just parameterize a couple to run against the panda version too?

olliesilvester:
> Great, thanks, almost there. Couple of comments. Also, most of the tests only run against the zebra version, maybe just parameterize a couple to run against the panda version too?

I'll add a few now - the tests will change once https://github.com/DiamondLightSource/hyperion/issues/1432 is done anyway

tests/devices/unit_tests/test_gridscan.py
```diff
@@ -27,49 +27,67 @@ def discard_status(st: Status | DeviceStatus):
...
+    use_pgs,
     zebra_fast_grid_scan: ZebraFastGridScan,
+    panda_fast_grid_scan: PandAFastGridScan,
 ):
-    set_mock_value(zebra_fast_grid_scan.scan_invalid, False)
-    set_mock_value(zebra_fast_grid_scan.position_counter, 0)
-    set_mock_value(zebra_fast_grid_scan.status, 1)
+    grid_scan: ZebraFastGridScan | PandAFastGridScan = (
+        panda_fast_grid_scan if use_pgs else zebra_fast_grid_scan
+    )
```
DominicOram, line 56: Nit: So I think you could be [clever](https://x.com/TitusWinters/status/1143595692113481728) here with fixtures (https://engineeringfordatascience.com/posts/pytest_fixtures_with_parameterize/#using-requestgetfixturevalue-) and do something like:
```python
@pytest.mark.parametrize(
    "grid_scan",
    ["zebra_fast_grid_scan", "panda_fast_grid_scan"],
)
async def test_given_settings_valid_when_kickoff_then_run_started(
    grid_scan_type: str, request
):
    grid_scan: ZebraFastGridScan | PandAFastGridScan = request.getfixturevalue(grid_scan_type)
```

But I appreciate that this may cause more confusion/wackiness down the line so leave it up to you if you think it's nicer or not.

DominicOram (approved):
Great, thanks! Minor nit comment if you want to take it, up to you

tests/devices/unit_tests/test_gridscan.py
olliesilvester: I just had a play with this, I think there are some issues with using `request.getfixturevalue` with `asyncio` fixtures. There's probably ways around this, but I'll leave it for now

PR #426, opened 2024-04-09
Return empty string if barcode is an empty list

### LINKED ISSUE #425: Gracefully handle no barcode on robot
When the BART BARCODE is empty reading the device will fail with:

```
  File "/dls_sw/i03/software/bluesky/hyperion_v9.0.0/dodal/src/dodal/devices/robot.py", line 38, in read
    pv_reading["value"] = str(pv_reading["value"][self.index])
IndexError: index 0 is out of bounds for axis 0 with size 0
```

## Acceptance Criteria
* When the barcode is empty reading from the robot does not throw an exception

DominicOram:
Fixes #425

1. Run the following on `main` and this branch:
```python
import os

os.environ["BEAMLINE"] = "i03"

import bluesky.plan_stubs as bps
from dodal.beamlines.i03 import robot

RE = RunEngine()

my_bart = robot()


def my_plan():
    barcode = yield from bps.rd(my_bart.barcode)
    print(barcode)


RE(my_plan())
```
2. Confirm fails on main, passes here (this is based on the state of i03 when writing this so if there is a sample loaded when you test it will pass on both)
3. Confirm new unit test also passes

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/robot.py
+++ b/src/dodal/devices/robot.py
@@ -35,10 +35,8 @@ class SingleIndexWaveformReadable(StandardReadable):
     async def read(self) -> Dict[str, Reading]:
         underlying_read = await self.bare_signal.read()
         pv_reading = underlying_read[self.bare_signal.name]
-        try:
-            pv_reading["value"] = str(pv_reading["value"][self.index])
-        except Exception:
-            pv_reading["value"] = ""
+        old_reading = pv_reading["value"]
+        pv_reading["value"] = str(old_reading[self.index]) if old_reading else ""
         return OrderedDict([(self._name, pv_reading)])
 
     async def describe(self) -> dict[str, Descriptor]:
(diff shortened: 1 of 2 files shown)


dperl-dls (approved):

src/dodal/devices/robot.py
```diff
@@ -35,7 +35,10 @@ def __init__(
     async def read(self) -> Dict[str, Reading]:
         underlying_read = await self.bare_signal.read()
         pv_reading = underlying_read[self.bare_signal.name]
-        pv_reading["value"] = str(pv_reading["value"][self.index])
+        try:
+            pv_reading["value"] = str(pv_reading["value"][self.index])
+        except Exception:
```
dperl-dls, line 40: Could: make this more specific

dperl-dls: or just do
```python
pv_reading["value"] = str(pv_reading["value"][self.index]) if pv_reading["value"] else ""
```

PR #436, opened 2024-04-11
Create webcam device for robot load snapshots on i03

### LINKED ISSUE #435: Device to get an image from a URL
Many webcams will provide an image via a URL e.g. `http://i03-webcam1/axis-cgi/jpg/image.cgi`. For diagnostic purposes we want to take an image from a webcam, save it to a specified filepath and read the saved path back. Much like we do with the `MJPG` stream from the OAV.

## Acceptance Criteria
* There is a device that we can use to save a webcam image and retrieve it from disk

DominicOram:
Fixes #435

Run:
```python
import bluesky.plan_stubs as bps
from dodal.beamlines.i03 import webcam

RE = RunEngine()

my_webcam = webcam()


def my_plan():
    yield from bps.mv(my_webcam.filename, "test_file")
    yield from bps.mv(my_webcam.directory, "/tmp/")
    yield from bps.trigger(my_webcam, wait=True)
    print((yield from bps.rd(my_webcam)))

RE(my_plan())
```
and confirm image gets stored in `tmp`.

### Checks for reviewer


DominicOram:
@callumforrester or @DiamondJoseph before I finish writing the tests for this is there a more idiomatic way of doing it? `last_saved_path` is effectively a `datum` right? So would it make sense to put it in an event like that? If so how do I do it?

DiamondJoseph:
> @callumforrester or @DiamondJoseph before I finish writing the tests for this is there a more idiomatic way of doing it? `last_saved_path` is effectively a `datum` right? So would it make sense to put it in an event like that? If so how do I do it?

Are you only going to be saving images during a data collection, or at any time?

callumforrester:
@DominicOram you need to implement `read` and `collect_stream_docs`, see this PR: https://github.com/bluesky/ophyd-async/pull/144/files#diff-2cd995da8cacab74d7a14c898acc1270646df44d1aaabba2d5c2a35a86271911 

Which is a fake ophyd-async detector that writes real HDF5 files

DominicOram:
> Are you only going to be saving images during a data collection, or at any time?

For the first usecase inside a collection but I don't want to rule it out

> @DominicOram you need to implement read and collect_stream_docs, see this PR...

Thank you, do you think it's appropriate here? The use of `stream` is a bit odd as we only really want the one image.

Basically, this is how I would implement the webcam but very keen to hear from you guys on if you think there is a better way

callumforrester:
I think `stream` is still appropriate for a single image, but could ask @coretl for a second opinion

DiamondJoseph:
As an external file, I think if you're going to emit anything it'd be a StreamResource to describe the file itself then 0-N StreamDatum for correlating data within the file. 

What is consuming the last_saved_path?

DominicOram:
> What is consuming the last_saved_path?

We're just dumping the path into ispyb

DiamondJoseph:
Warning that StreamResource/StreamDatum are due to undergo a change in a bluesky version to be released after June, so anything that parses the SR will need to be modified after that point.

AFAIK those documents can't be emitted unless the RunEngine thinks it's in a run. You can make a run just to emit the documents (a Run is any correlated data collection, collecting data from just the Webcam is still a valid Run) if you need to start sending the path outside of a data collection.

https://github.com/bluesky/event-model/pull/301

DominicOram:
Ok, looking through the example @callumforrester gave I think the added complexity isn't worth it in this case. What do you guys think?

DiamondJoseph:
```python
    async def collect_stream_docs(
        self, indices_written: int
    ) -> AsyncIterator[StreamAsset]:
        filename = await self.filename.get_value()
        directory = await self.directory.get_value()
        yield from compose_stream_resource(
                spec="image/png",  # prep for switch to mimetype
                root=directory,
                data_key="last_saved_path",
                resource_path=f"{filename}.png",
                resource_kwargs={},
            ).stream_resource_doc

    def stream_data(self, indices_written: int) -> Iterator[StreamDatum]:
        # don't emit Datum
        return None
```

It should be relatively simple, you know what you're writing (a png), and where (the root/resource_dir distinction will matter when using softiocs in the cluster)

DominicOram:
Spent a couple of hours on trying to do it with streams, ended up more messy without much gain so using existing solution

tests/plans/test_topup_plan.py
```diff
@@ -15,6 +15,7 @@
 
 @pytest.fixture
 def synchrotron() -> Synchrotron:
+    RunEngine()
```
olliesilvester, line 18: Should we move hyperion's RE fixture to dodal? The

```
@pytest.fixture
def RE():
    RE = RunEngine({}, call_returns_result=True)
    RE.subscribe(
        VerbosePlanExecutionLoggingCallback()
    )  # log all events at INFO for easier debugging
    yield RE
    try:
        RE.halt()
    except Exception as e:
        print(f"Got exception while halting RunEngine {e}")
    finally:
        stopped_event = threading.Event()

        def stop_event_loop():
            RE.loop.stop()  # noqa: F821
            stopped_event.set()

        RE.loop.call_soon_threadsafe(stop_event_loop)
        stopped_event.wait(10)
    del RE
```

DominicOram: Yes, probably, made https://github.com/DiamondLightSource/dodal/issues/449

DominicOram:
> the png file saved by the manual test can't be opened properly - were you expecting it to be able to?

This works for me @olliesilvester. Potentially the webcam was down when you tried? Can you try again and check what `http://i03-webcam1/axis-cgi/jpg/image.cgi` looks like if it doesn't work?

olliesilvester:
> > the png file saved by the manual test can't be opened properly - were you expecting it to be able to?
> 
> This works for me @olliesilvester. Potentially the webcam was down when you tried? Can you try again and check what `http://i03-webcam1/axis-cgi/jpg/image.cgi` looks like if it doesn't work?

The issue was the PNG library on my workstation, the files work OK on another computer

olliesilvester (approved):

PR #445, opened 2024-04-19
Move Diamond implementation of DirectoryProvider into dodal

### LINKED ISSUE #305: Add ophyd-async Directory Provider

DiamondJoseph:
Refactor to remove use of globals, make per-beamline, rename to make explicit that current impl is partial one-visit-static implementation. 

As the naming scheme is Diamond specific, and is used in instantiating devices, belongs in dodal
Closes #305 

1. Run up a (sim) ophyd_async StandardDetector
2. Ensure tries to write data to correct location

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i22.py
+++ b/src/dodal/beamlines/i22.py
@@ -1,8 +1,8 @@
-from functools import lru_cache
-
-from ophyd_async.core import DirectoryProvider
-
-from dodal.beamlines.beamline_utils import device_instantiation
+from dodal.beamlines.beamline_utils import (
+    device_instantiation,
+    get_directory_provider,
+    set_directory_provider,
+)
 from dodal.beamlines.beamline_utils import set_beamline as set_utils_beamline
 from dodal.common.visit import StaticVisitDirectoryProvider
 from dodal.devices.tetramm import TetrammDetector
@@ -12,17 +12,15 @@ from dodal.utils import get_beamline_name
 BL = get_beamline_name("i22")
 set_log_beamline(BL)
 set_utils_beamline(BL)
-
-
-@lru_cache
-def get_directory_provider() -> DirectoryProvider:
-    """VISIT is a placeholder value for reference.
+"""VISIT is a placeholder value for reference.
     Requires that all StandardDetector IOCs are running with /data mapping to /dls/i22/data
     """
-    return StaticVisitDirectoryProvider(
+set_directory_provider(
+    StaticVisitDirectoryProvider(
         BL,
         "/data/i22/2024/VISIT",
     )
+)
 
 
 def i0(
--- a/src/dodal/beamlines/i22.py
+++ b/src/dodal/beamlines/i22.py
@@ -1,5 +1,10 @@
-from dodal.beamlines.beamline_utils import device_instantiation, get_directory_provider
+from functools import lru_cache
+
+from ophyd_async.core import DirectoryProvider
+
+from dodal.beamlines.beamline_utils import device_instantiation
 from dodal.beamlines.beamline_utils import set_beamline as set_utils_beamline
+from dodal.common.visit import StaticVisitDirectoryProvider
 from dodal.devices.tetramm import TetrammDetector
 from dodal.log import set_beamline as set_log_beamline
 from dodal.utils import get_beamline_name
@@ -7,6 +12,17 @@ from dodal.utils import get_beamline_name
 BL = get_beamline_name("i22")
 set_log_beamline(BL)
 set_utils_beamline(BL)
+
+
+@lru_cache
+def get_directory_provider() -> DirectoryProvider:
+    """VISIT is a placeholder value for reference.
+    Requires that all StandardDetector IOCs are running with /data mapping to /dls/i22/data
+    """
+    return StaticVisitDirectoryProvider(
+        BL,
+        "/data/i22/2024/VISIT",
+    )
 
 
 def i0(
--- a/src/dodal/beamlines/beamline_utils.py
+++ b/src/dodal/beamlines/beamline_utils.py
@@ -1,12 +1,10 @@
 import inspect
-import tempfile
 from typing import Callable, Dict, Final, List, Optional, TypeVar, cast
 
 from bluesky.run_engine import call_in_bluesky_event_loop
 from ophyd import Device as OphydV1Device
 from ophyd.sim import make_fake_device
 from ophyd_async.core import Device as OphydV2Device
-from ophyd_async.core import DirectoryProvider, StaticDirectoryProvider
 from ophyd_async.core import wait_for_connection as v2_device_wait_for_connection
 
 from dodal.utils import AnyDevice, BeamlinePrefix, skip_device
@@ -15,7 +13,6 @@ DEFAULT_CONNECTION_TIMEOUT: Final[float] = 5.0
 
 ACTIVE_DEVICES: Dict[str, AnyDevice] = {}
 BL = ""
-DIRECTORY_PROVIDER: DirectoryProvider | None = None
 
 
 def set_beamline(beamline: str):
@@ -124,17 +121,3 @@ def device_instantiation(
     if post_create:
         post_create(device_instance)
     return device_instance
-
-
-def set_directory_provider(provider: DirectoryProvider):
-    global DIRECTORY_PROVIDER
-
-    DIRECTORY_PROVIDER = provider
-
-
-def get_directory_provider() -> DirectoryProvider:
-    if DIRECTORY_PROVIDER is None:
-        set_directory_provider(
-            StaticDirectoryProvider(tempfile.NamedTemporaryFile().name, "")
-        )
-    return DIRECTORY_PROVIDER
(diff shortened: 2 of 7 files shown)


src/dodal/common/data.py
```diff
@@ -0,0 +1,145 @@
...
+        return DataCollectionIdentifier(collectionNumber=self._count)
+
+
+class StaticVisitDirectoryProvider(DirectoryProvider):
+    """
+    Gets information from a remote service to determine how their files should be named.
+    Supports a single Visit which should be passed as a Path relative to the root of the Detector IOC mounting.
+    i.e. for a Kubernetes IOC that mounts `/dls/ixx/data` as `/data`: root="/data/YYYY/visit/"
+    but for a Detector running with a view to the whole DLS filesystem, root="/dls/ixx/data/YYYY/visit"
+    Requires that all detectors are running with a mutual view on the filesystem.
```
Tom-Willemsen, line 89: Is this actually true?

src/dodal/beamlines/i22.py
```diff
@@ -9,6 +14,14 @@
 set_utils_beamline(BL)
 
 
+@lru_cache
+def get_directory_provider() -> DirectoryProvider:
+    return StaticVisitDirectoryProvider(
+        BL,
+        "/data/i22/2024/VISIT",
```
Tom-Willemsen, line 21: I think I must not be understanding this PR at all, so bear with me, but:
- Why are we hard-coding `2024`?
- What is `VISIT`? Is that a placeholder that's means to get magically replaced elsewhere?
- If we're `lru_caching` this doesn't that mean `blueapi` needs a full restart just to change visits? That seems suboptimal.
- In p38 we're *also* literally hard-coding `cm37282-2`, that really doesn't seem right??

src/dodal/common/data.py
```diff
@@ -0,0 +1,145 @@
...
+from pathlib import Path
+from typing import Optional
+
+from aiohttp import ClientSession
+from ophyd_async.core import DirectoryInfo, DirectoryProvider
+from pydantic import BaseModel
+
+
+class DataCollectionIdentifier(BaseModel):
+    # VisitService API
```
Tom-Willemsen, line 11: ?

src/dodal/common/data.py
```diff
@@ -0,0 +1,145 @@
...
+        self._url = url
+
+    async def create_new_collection(self) -> DataCollectionIdentifier:
+        async with ClientSession() as session:
+            async with session.post(f"{self._url}/numtracker") as response:
+                if response.status == 200:
+                    json = await response.json()
+                    return DataCollectionIdentifier.parse_obj(json)
+                else:
+                    raise Exception(response.status)
```
Tom-Willemsen, line 54: Can this be a more specific exception class.

Also please can we make the error message actually descriptive about what it was trying to do.

src/dodal/common/data.py
```diff
@@ -0,0 +1,145 @@
...
+        self._session = None
+
+    async def update(self) -> None:
+        """
+        Calls the visit service to create a new data collection in the current visit.
+        """
+        # TODO: After visit service is more feature complete:
+        # TODO: Allow selecting visit as part of a request, pass down to this class
+        # TODO: Query visit service to get information about visit and data collection
+        # TODO: Use AuthN information as part of verification with visit service
```
Tom-Willemsen, line 117: Are there github or JIRA issues associated with these things? If not can you create and link to them from here...

src/dodal/common/data.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def _generate_directory_info(
+        self,
+        collection_id_info: DataCollectionIdentifier,
+    ) -> DirectoryInfo:
+        # Diamond standard file naming
+        collection_id = collection_id_info.collectionNumber
+        return DirectoryInfo(
+            root=self._root,
+            # All Path into Visit is root, no resource_dir
```
Tom-Willemsen, line 134: I don't understand this comment...

src/dodal/common/data.py
```diff
@@ -0,0 +1,145 @@
...
+        # TODO: Allow selecting visit as part of a request, pass down to this class
+        # TODO: Query visit service to get information about visit and data collection
+        # TODO: Use AuthN information as part of verification with visit service
+
+        try:
+            collection_id_info = await self._client.create_new_collection()
+            self._current_collection = self._generate_directory_info(collection_id_info)
+        except Exception as e:
+            self._current_collection = None
+            raise e
```
Tom-Willemsen, line 124:
```suggestion
        except Exception:
            self._current_collection = None
            raise
```

I think it would also be useful to debug log here, something like "got error so clearing current collection".

src/dodal/common/data.py
```diff
@@ -0,0 +1,145 @@
...
+            prefix=f"{self._beamline}-{collection_id}",
+        )
+
+    def __call__(self) -> DirectoryInfo:
+        if self._current_collection is not None:
+            return self._current_collection
+        else:
+            raise ValueError(
+                "No current collection, update() needs to be called at least once"
+            )
```
Tom-Willemsen, line 145: Q: could we just make it call update itself if `_current_collection` is initially None?

DiamondJoseph:
Requires https://github.com/bluesky/ophyd-async/pull/245

src/dodal/common/data.py
Tom-Willemsen: The comment seems to suggest that there are both "detectors mounting the whole DLS filesystem" and "detectors mounting only the data directory". So my question was, is it the case that these detectors cannot and will not ever coexist on a beamline?

Ideally I'd like to find a nice way to _assert_ that's true at runtime, rather than waiting for an obscure error, but happy to spin that out into a separate issue...

src/dodal/beamlines/i22.py
DiamondJoseph: You're understanding perfectly. 

Unfortunately right now we can't authenticate the blueapi user, so we cannot (according to the security policy) expose the list of visits that the user could write to (because the visits that user X is on is privileged information that user Y is not allowed to access), so we set blueapi up for a single visit. This is why the year, visit are hardcoded right now. VISIT is just a placeholder for comment and review at the moment. The p38/p45 commissioning visits are the correct ones for writing data into currently.

This is how the hardcoding was done in November (in one part, beamline and visit). Eventually the beamline will be hardcoded and the visit will be updating with (potentially) every request. This is a step to get there. 

The lru_caching will still be on the DirectoryProvider, but the DirectoryProvider will return a new DirectoryInfo that has relevant information for the current visit for each Run.

src/dodal/common/data.py
DiamondJoseph: Just marking that these classes are part of the partial VisitService implemenation in the GDA server. I can remove.

Tom-Willemsen: Happy for you to keep it if you link to the overarching JIRA/GH issue so that people like me can figure out what "VisitService API" means ;) It's just not clear to an outsider currently what is meant to be permanent and what is a stop-gap solution.

src/dodal/common/data.py
DiamondJoseph: Currently we're writing all of our h5 files to /data/year/visit. If we want to write them to visit/experiment/, then we need to distinguish between root (the part of the path that is semantic and not shared between different views of the file system) and resource_path, which is required for all services (we have aims of mounting only ever /data -> /dls, /data -> /dls/ixx/data or /data -> /dls/ixx/data/year/visit, of which everything inside visit/ is resource_path and everything up to it is root)

Tom-Willemsen: Same thoughts as a few of the other threads then, if you expand the comment a bit with either part of ^that explanation or a link to it/relevant issue/etc then that's fine.

src/dodal/beamlines/i22.py
callumforrester: An option if we don't want to hard-code it here is to use an environment variable or similar if anyone has thoughts on that

Tom-Willemsen: I was originally imagining you'd have a plan exposed by blueAPI that could set the year/visit - because then it can be changed at runtime and avoids needing to edit `dodal` code just to change experiments.

But tbh I'm happy with just documentation if this is an "intermediate" solution.

src/dodal/common/data.py
DiamondJoseph: I think it needs to be a separate issue. I'm not sure how we can check other than trying to set a filepath to them and it failing.

DiamondJoseph: Created new issue to track this https://github.com/DiamondLightSource/dodal/issues/455

Tom-Willemsen (approved):
Let's just get this in and we can always refine later as needed.

PR #458, opened 2024-04-23
432 create sample stage device for p99

### LINKED ISSUE #432: Create Sample stage device for p99
p99 is being used as the testbed for the flagship CSXID beamline(i17).
A device on p99 that is not currently captured by other dodal devices
This tickets is for p99 sample stage: it is physically a combination of a standard xyz stage and a rotational stage with 2 axises, but it also include a virtual xyz stage (in different reference frame),  a  "fliter wheel" that select a predetermine position of the sample stage and a theta angle of the incident beam. 

## Acceptance Criteria
- [ ] An ophyd-async sampleStage device has been created that can be instantiated against the signals of i20-1.

### Tasks
- [ ] create devices for p99 sample stage.
- [ ] The sample stage has 10 motors 
- [ ] sample x, y ,z,  (sample frame) virtual
- [ ] sample (Lab) x, y, z (most likely real motor position)
- [ ] Theta, pitch and roll, pitch and roll are most likely mirror and theta are most likely incident angle (virtual)
- [ ] There is also a Filter Wheel

###Ref
image

Relm-Arrowny:
Fixes #432 see [p99#7](https://github.com/Relm-Arrowny/p99Bluesky/issues/7) and [p99#13](https://github.com/Relm-Arrowny/p99Bluesky/issues/13) for full detail

Added :
- Lab xyz-stage and labxyz-stage for p99. 
- The angles pitch roll and theta are not standard motors, they only have set and feedback and it is not stopable. so a SetReadOnlyMotor device is added to capture this.
- p99 Filter wheel is added as a FilterMotor device with an enum class for the selections.

1. Unit test should cover all of the devices.
2. IRL test can be done on p99 using bluesky and this [Jupiter doc](https://github.com/Relm-Arrowny/p99Bluesky/blob/main/tests/jupyter_tests/stagetest.ipynb).

### Checks for reviewer


src/dodal/beamlines/p99.py
```diff
@@ -0,0 +1,62 @@
...
+        fake=fake_with_ophyd_sim,
+    )
+
+
+def sample_lab_xyz_stage(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> ThreeAxisStage:
+    return device_instantiation(
+        FilterMotor,
+        prefix="-MO-STAGE-02:LAB:",
```
DominicOram, line 58: Should: This has the same prefix as `sample_xyz_stage` so presumably is writing to the same PVs. How is it a different device?

src/dodal/devices/stages.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, prefix: str, name: str, infix: list[str] | None = None):
+        if infix is None:
+            infix = ["X", "Y", "Z"]
+        self.x = Motor(prefix + infix[0])
+        self.y = Motor(prefix + infix[1])
+        self.z = Motor(prefix + infix[2])
+        super().__init__(name=name)
+
+
+class SingleBasicStage(Device):
```
DominicOram, line 41: Should: This isn't used anywhere other than tests. Why do we need it?

src/dodal/devices/epics/setReadOnlyMotor.py
```diff
@@ -0,0 +1,86 @@
...
+                    time_elapsed=time.monotonic() - start,
+                )
+
+        self.user_readback.subscribe_value(update_watchers)
+        try:
+            await self.user_setpoint.set(new_position)
+        finally:
+            self.user_readback.clear_sub(update_watchers)
+        if not self._set_success:
+            raise RuntimeError("Motor was stopped")
```
DominicOram, line 73: Should: When is this ever going to run, it looks like `_set_success` is always `True`? In which case can we remove this and `_set_success`?

src/dodal/devices/epics/setReadOnlyMotor.py
```diff
@@ -0,0 +1,86 @@
...
+class SetReadOnlyMotor(StandardReadable, Movable):
+    """Device that moves a motor with only setpoint and readback.
+    Parameters
+    ----------
+    prefix:
+        EPICS PV (None common part up to and including :).
+    name:
+        Name for the motor.
+    suffix:
+        Th last part of any EPICS PV, default is the [".VAL", ".RBV"].
```
DominicOram, line 19: Typo:
```suggestion
        The last part of any EPICS PV, default is the [".VAL", ".RBV"].
```

src/dodal/devices/p99/sample_stage.py
```diff
@@ -0,0 +1,45 @@
...
+    Zn5um = ("Zn 5um",)
+    Zr = ("Zr (empty)",)
+    Mo = ("Mo (empty)",)
+    Rh = ("Rh (empty)",)
+    Pd = ("Pd (empty)",)
+    Ag = ("Ag (empty)",)
+    Cd25um = ("Cd 25um",)
+    W = ("W (empty)",)
+    Pt = ("Pt (empty)",)
+    User = ("User",)
```
DominicOram, line 39: Why are these tuples? I think if the EPICS signal is just a usual `MBBO` then these can just be strings. If there is a specific reason a comment would be good.

DominicOram (changes requested):
Sorry to jump in here without knowing much background. I've put some minor comments in the code. In general I'm not sure I like that we have to have a `SetReadOnlyMotor` as I'm not sure this is the right solution in general. Has the possibility of wrapping a `SoftMotor` around  it in IOC been discussed with controls? Then the implementation details get hidden from the ophyd layer and up.

src/dodal/devices/p99/sample_stage.py
Relm-Arrowny: No reason at all, I read the enum from epics and left the comma in by mistake so the tuples was vscode auto correct!!!

src/dodal/devices/p99/sample_stage.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, prefix: str, name: str):
+        self.theta = SetReadOnlyMotor(
+            prefix, name, suffix=["WRITETHETA", "WRITETHETA:RBV", "WRITETHETA.EGU"]
+        )
+        self.roll = SetReadOnlyMotor(
+            prefix, name, suffix=["WRITETHETA", "WRITETHETA:RBV", "WRITETHETA.EGU"]
+        )
+        self.pitch = SetReadOnlyMotor(
+            prefix, name, suffix=["WRITETHETA", "WRITETHETA:RBV", "WRITETHETA.EGU"]
+        )
```
DominicOram, line 19: Should: I assume you want these to actually be different PVs and this is a typo?

src/dodal/devices/stages.py
```diff
@@ -0,0 +1,36 @@
+from ophyd_async.core import Device
+from ophyd_async.epics.motion.motor import Motor
+
+
+class ThreeAxisStage(Device):
```
DominicOram, line 5: Should: Can we combine this with [XYZPositioner](https://github.com/DiamondLightSource/dodal/blob/00c5c72031e1e8948b59c61ec54b95919fbfae7a/src/dodal/devices/motors.py#L10)

DominicOram (changes requested):
Sorry, I completely forgot that I started reviewing this. One comment, it also looks like it won't work with the latest `main` as `beamline_utils` has been renamed. Otherwise happy to merge.

DominicOram:
Do you still need this @Relm-Arrowny? Would be good to get it merged if useful or dropped otherwise, just so we don't have too many stale PRs

Relm-Arrowny:
Sorry p99 is being move and this slipped my mine, it should be good to go.

src/dodal/devices/motors.py
```diff
@@ -3,8 +3,35 @@
...
+        EPICS PV, default is the ["X", "Y", "Z"].
+    Notes
+    -----
+    Example usage::
+        async with DeviceCollector():
+            xyz_stage = ThreeAxisStage("BLXX-MO-STAGE-XX:")
+    Or::
+        with DeviceCollector():
+            xyz_stage = ThreeAxisStage("BLXX-MO-STAGE-XX:", suffix = [".any",
+              ".there", ".motorPv"])
```
DominicOram, line 27: Should: Docs now reference the old device name, I will fix

DominicOram (approved):
Great, thank you! I couldn't test that the PVs are right as I couldn't `ssh` into P99, maybe it's down? So going to trust you on that. Otherwise just one suggestion, which I will fix myself quickly

PR #460, opened 2024-04-23
Use set and wait rather than put to be more explicit

### LINKED ISSUE #448: Change `put(...)` to `set(...).wait`
We had previously assumed that for an ophyd device a `put(...)` was equivalent to `set(...).wait()`. From a re-analysis of the code I'm not sure this is actually true. This may explain some of the race condition errors we've been seeing in the `eiger`. I think given there is no risk in moving to using `set(...).wait()` everywhere anyway and it's arguably more explicit we should just do that.

## Acceptance Criteria
* We move to using `set(...).wait()` rather than `put(...)`
* All tests still pass

DominicOram:
Fixes #448

1. Confirm tests still pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -120,7 +120,7 @@ class EigerDetector(Device):
                 # In free run mode we have to manually stop odin
                 self.stop_odin_when_all_frames_collected()
 
-            self.odin.file_writer.start_timeout.put(1)
+            self.odin.file_writer.start_timeout.set(1).wait()
             LOGGER.info("Waiting on filewriter to finish")
             self.filewriters_finished.wait(30)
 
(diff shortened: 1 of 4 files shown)


rtuck99:
I have a general comment re. the originating bug - if put doesn't wait, what are the semantics of put? And why do we have to look through the code, it feels like it should be documented somewhere as it's fairly fundamental...

rtuck99 (approved):

DominicOram:
> I have a general comment re. the originating bug - if put doesn't wait, what are the semantics of put? 

Put will wait on a callback from EPICS if `put_complete` has been set. Otherwise it will not. This is in contrast to `set`, which I believe gives a returning status that is only complete when the readback of the PV changes. 

> And why do we have to look through the code, it feels like it should be documented somewhere as it's fairly fundamental...

The docstring on `put` does specifically say `Low-level method for writing to a Signal.`, which I guess implies we shouldn't really be using it. If we think it's worth adding more documentation into `ophyd` around it then we can prioritise that but with moving to `ophyd_sync` anyway soon I'm not too fussed about digging into this too far.

olliesilvester:
> The docstring on `put` does specifically say `Low-level method for writing to a Signal.`, which I guess implies we shouldn't really be using it. If we think it's worth adding more documentation into `ophyd` around it then we can prioritise that but with moving to `ophyd_sync` anyway soon I'm not too fussed about digging into this too far.

I agree with the documentation point. We should remove `put` from all our dodal devices given that we were using it with the assumption that waited. We can just replace with `set(value).wait(timeout)`

DominicOram:
@olliesilvester made a good point in comment here (which is now deleted?) I should have put timeouts on these waits, right now they're infinite. Will make a new PR

PR #463, opened 2024-04-24
Add timeouts to waits on sets

### LINKED ISSUE #448: Change `put(...)` to `set(...).wait`

DominicOram:
Fixes #448

1. Confirm this now has sensible timeouts

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -120,7 +120,7 @@ class EigerDetector(Device):
                 # In free run mode we have to manually stop odin
                 self.stop_odin_when_all_frames_collected()
 
-            self.odin.file_writer.start_timeout.set(1).wait()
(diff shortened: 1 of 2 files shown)


olliesilvester (approved):
Looks good. There are some flaky tests in hyperion's `device_setup_plans.test_utils.py`, but I'm not sure if they're caused by this change (it failed 1/10 times)
Also, except for a few exceptions, I think we should almost never be doing waits with no timeout. This could be a separate discussion/PR though

PR #461, opened 2024-04-24
Port `UndulatorDCM` to ophyd-async

callumforrester:
Works towards #389 and #397 

Changes:
- Rename `DCM.py` to `dcm.py` to comply with pep8: https://peps.python.org/pep-0008/#package-and-module-names
- Port `UndulatorDCM` and its two components `Undulator` and `DCM` to ophyd-async as `StandardReadable`s. 
- Update tests 

### Instructions to reviewer on how to test:
1. Check unit tests pass

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards)

PR #467, opened 2024-04-24
Convert FocusingMirror to ophyd_async

### LINKED ISSUE #398: Ophyd-async devices for HFM and VFM
Required for #377

We need to implement an HFM and VFM device exposed the fields required for I22 nexus files (see linked issue).

## Acceptance Criteria
* HFM device available in dodal
* VFM device available in dodal
* HFM and VFM config included in i22 beamline module
* Fake HFM and VFM config included in p38 beamline module

DominicOram:
Fixes #398

1. Confirm unit tests still pass
2. Confirm i03 system tests still pass (after commenting out the panda/pin tip detection due to https://github.com/DiamondLightSource/dodal/issues/261)
3. Confirm new i22 system tests pass

### Checks for reviewer


src/dodal/devices/focusing_mirror.py
```diff
@@ -107,30 +114,49 @@ def voltage_channels(self) -> list[MirrorVoltageDevice]:
...
 
 
-class FocusingMirror(Device):
+class FocusingMirror(StandardReadable):
     """Focusing Mirror"""
 
-    def __init__(self, bragg_to_lat_lut_path, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(
+        self, name, prefix, bragg_to_lat_lut_path=None, x_y_sufffixes=("X", "Y")
```
Tom-Willemsen, line 121: I wonder if a signature like 

```
self, name, prefix, *, bragg_to_lat_lut_path=None, x_suffix="X", y_suffix="Y"
```

would be a bit clearer? I don't love that we have to do `x_y_sufffixes[1]` later.

src/dodal/devices/focusing_mirror.py
```diff
@@ -107,30 +114,49 @@ def voltage_channels(self) -> list[MirrorVoltageDevice]:
...
+    ):
         self.bragg_to_lat_lookup_table_path = bragg_to_lat_lut_path
+        self.yaw_mrad = Motor(prefix + "YAW")
+        self.pitch_mrad = Motor(prefix + "PITCH")
+        self.roll_mrad = Motor(prefix + "ROLL")
+        self.x_mm = Motor(prefix + x_y_sufffixes[0])
```
Tom-Willemsen, line 127: Maybe one too many f's

```suggestion
        self.x_mm = Motor(prefix + x_y_suffixes[0])
```

src/dodal/devices/focusing_mirror.py
```diff
@@ -107,30 +114,49 @@ def voltage_channels(self) -> list[MirrorVoltageDevice]:
...
+        self.y_mm = Motor(prefix + x_y_sufffixes[1])
+        self.jack1_mm = Motor(prefix + "Y1")
+        self.jack2_mm = Motor(prefix + "Y2")
+        self.jack3_mm = Motor(prefix + "Y3")
+        self.translation1_mm = Motor(prefix + "X1")
+        self.translation2_mm = Motor(prefix + "X2")
+
+        # See https://manual.nexusformat.org/classes/base_classes/NXmirror.html
+        self.type, _ = soft_signal_r_and_backend(str, "single")
```
Tom-Willemsen, line 136: Since the only two allowable values are defined in the nexus standard ("single" and "multi"), could this be an enum? Or does that interact badly with the soft backend in some way?

src/dodal/devices/focusing_mirror.py
```diff
@@ -107,30 +114,49 @@ def voltage_channels(self) -> list[MirrorVoltageDevice]:
...
+        # The device is in the beamline co-ordinate system so pitch is the incident angle
+        # regardless of orientation of the mirror
+        self.incident_angle = Motor(prefix + "PITCH")
```
Tom-Willemsen, line 139: Maybe you've thought about this already but is there a reason for this over
```suggestion
        self.incident_angle = self.pitch_mrad
```

?

Tom-Willemsen (approved):
Substantially looks fine from my perspective, feel free to take or leave my nit-picky comments...

src/dodal/devices/focusing_mirror.py
DominicOram: I thought about that but if you have one signal like that I'm not sure how the name gets set on it? Probably would get the last one set so might be alright but then would you get `incident_angle` when you read `pitch`?

To be honest, I'm not sure of the best way to structure this, I assume you want the name to be `incident_angle` for the nexus writer?

Tom-Willemsen: Yep fair enough sounds like I just missed some complexity (naming the signals) - feel free to leave as it was.

src/dodal/devices/focusing_mirror.py
```diff
@@ -12,7 +19,7 @@
 DEFAULT_SETTLE_TIME_S = 60
 
 
-class MirrorStripe(Enum):
+class MirrorStripe(str, Enum):
```
rtuck99, line 22: I feel conflicted about this. It's just far too pythony for me. But each to their own...

DominicOram: It's an `ophyd_async` requirement that if you want to specify a signal as an `Enum` type it also has to inherit from `str`.

tests/system_tests/test_i22_system.py
```diff
@@ -0,0 +1,23 @@
...
+
+from dodal.utils import make_all_devices
+
+if __name__ == "__main__":
+    """This test runs against the real beamline and confirms that all the devices connect
+    i.e. that we match the beamline PVs. Obviously this must be run on the DLS network
+    and could be flaky if parts of the beamline are down for maintenance.
+
+    This is not implemented as a normal pytest test as those tests run using the S03
+    EPICS ports and switching ports at runtime is non-trivial
```
callumforrester, line 13: Nit: This isn't true when running against the real beamline

src/dodal/devices/focusing_mirror.py
```diff
@@ -107,30 +121,48 @@ def voltage_channels(self) -> list[MirrorVoltageDevice]:
...
+        self.roll_mrad = Motor(prefix + "ROLL")
+        self.x_mm = Motor(prefix + x_suffix)
+        self.y_mm = Motor(prefix + y_suffix)
+        self.jack1_mm = Motor(prefix + "Y1")
+        self.jack2_mm = Motor(prefix + "Y2")
+        self.jack3_mm = Motor(prefix + "Y3")
+        self.translation1_mm = Motor(prefix + "X1")
+        self.translation2_mm = Motor(prefix + "X2")
+
+        self.type, _ = soft_signal_r_and_backend(MirrorType, MirrorType.SINGLE)
```
callumforrester, line 142: Happy with this as a preferred way to include config for now, will do the same with Undulator and DCM (although in a separate PR to #461)

src/dodal/beamlines/i22.py
```diff
@@ -57,6 +58,32 @@ def it(
...
+    wait_for_connection: bool = True,
+    fake_with_ophyd_sim: bool = False,
+) -> FocusingMirror:
+    return device_instantiation(
+        FocusingMirror,
+        "hfm",
+        "-OP-KBM-01:HFM:",
+        wait_for_connection,
+        fake_with_ophyd_sim,
+    )
```
callumforrester, line 84: Should: Please could you also make versions on these in the P38 config with `fake_with_ophyd_sim: bool = True`? This allows us to produce documents from P38 with enough info to make a nexus file.

callumforrester (changes requested):
Looks good, thanks, only the P38 point to urgently address.

tests/system_tests/test_i22_system.py
DominicOram: What do you mean? It is true that it's not a normal pytest because switching at runtime is non-trivial?

src/dodal/beamlines/p38.py
```diff
@@ -162,6 +163,32 @@ def slits_6(
     )
 
 
+def vfm(
+    wait_for_connection: bool = True,
+    fake_with_ophyd_sim: bool = False,
```
DiamondJoseph, line 168:
```suggestion
    fake_with_ophyd_sim: bool = True,
```

src/dodal/beamlines/p38.py
```diff
@@ -162,6 +163,32 @@ def slits_6(
...
+        "vfm",
+        "-OP-KBM-01:VFM:",
+        wait_for_connection,
+        fake_with_ophyd_sim,
+    )
+
+
+def hfm(
+    wait_for_connection: bool = True,
+    fake_with_ophyd_sim: bool = False,
```
DiamondJoseph, line 181:
```suggestion
    fake_with_ophyd_sim: bool = True,
```

DiamondJoseph (changes requested):
nit: fake p38 mirrors

DiamondJoseph (approved):

tests/system_tests/test_i22_system.py
callumforrester: Sorry, should have specified, the ports point is not true. S03 runs on non-standard EPCIS port but a real beamline runs on the standard ones.

callumforrester (approved):
Happy for merge, thanks for this!

Issue #415, opened 2024-04-04
 Support parallel and lazy connection of Devices


coretl:
At the moment we have something like:
https://github.com/DiamondLightSource/dodal/blob/4f646f86a94f20d08997786afd478eb48de602ad/src/dodal/beamlines/i03.py#L46-L57

This allows optional serial connection and creation of fake devices instead of real devices via `device_instantiation`. The ophyd-async mechanism for this is using the `connect` method:

```async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):```

(you could argue that `fake` would be a better name than `sim`, so that could also be changed)

I would like to be able to:
- Create all the Devices for a beamline
- Have blueapi know what those Devices are
- Connect those Devices in the fastest possible way so we can get to running a plan quickly

I will make some suggestions and I'd like to encourage a vigorous discussion of the alternatives here!

## Suggestion 1: Background idempotent connect

Rather than an explicit connect when creating the Device, add it to a background queue so it will be connected eventually. I would also suggest we make `connect()` idempotent so that it will complete immediately if it has already been run successfully. If we also use Devices as arguments to plans then blueapi could call `connect()` on all the Devices that are plan ags (or some other decorator solution a-la hyperion composites).

```python
# ophyd-async
class Device:
    # None if connect hasn't started, an Event if it has, a set Event if it's done
    _connect_task: Optional[asyncio.Task] = None
    async def connect(self, sim: bool = False, timeout: float = DEFAULT_TIMEOUT):
        if not self._connect_task or (self._connect_task.done() and self._connect_task.exception()):
            # connect never run, or it errored, so try to connect now
            coros = {
                name: child_device.connect(sim, timeout=timeout)
                for name, child_device in self.children()
            }
            self._connect_task = asyncio.create_task(wait_for_connection(**coros))
        # Wait for it to complete
        await self._connect_task
```

```python
# blueapi
async def run_plan_sketch(plan, kwargs):
    coros = {name: device.connect() for name, device in get_devices(plan, kwargs)}
    await wait_for_connection(**coros)
    RE(plan(**kwargs))
```
  
## Suggestion 2: Remove device factories

If connect is no longer happening at Device instantiation then it should be safe to always have them created at startup, then we can just import them with the minimum boilerplate. Connection can be done outside this function.

```python
# dodal/beamlines/i22.py
BL = "BL22I"
panda1 = PandAHDF(f"{BL}-EA-PANDA-01")
[...]

PR #469, opened 2024-04-25
Add fswitch device for i22

### LINKED ISSUE #399: Ophyd-async device for F-Switch
Required for #377

We need to implement an F-Switch device exposing the fields required for I22 nexus files (see linked issue). The F-Switch is a group of 4 motors.

## Acceptance Criteria
* F-Switch class available in dodal
* F-Switch device config included in i22 beamline module
* F-Switch fake device config included in p38 beamline module

joeshannon:
Fixes #399 

1. Create device pointing at a beamline fswitch
2. Perform a read plan and verify that the num_filters result matches the number of filters on the edm screen (manually adding up the lights)

### Checks for reviewer


joeshannon:
This is an initial implementation thought to achieve the desired result however it needs to be checked with beamline staff.

There is some scope in future for combining with the i04 fswitch (currently transfocator.py) but the epics interfaces are currently incompatible.

src/dodal/devices/i22/fswitch.py
```diff
@@ -0,0 +1,43 @@
...
+    Device for i22's fswitch. This currently only implements the minimum
+    functionality for retrieving the number of lenses inserted.
+
+    Eventually this should be combined with the transfocator device in the i04
+    module but is currently incompatible as the Epics interfaces are different.
+    """
+
+    NUM_FILTERS = 128
+
+    def __init__(self, prefix: str, name: str = "") -> None:
```
DiamondJoseph, line 28: Nit: I'd prefer for top-level devices like this to require a name, but it looks like the Slits made it in with a default so I can't point to precedence =(

tests/devices/i22/test_fswitch.py
```diff
@@ -0,0 +1,42 @@
...
+    expected_reading: Mapping[str, Any],
+) -> None:
+    await device.stage()
+    reading = await device.read()
+    await device.unstage()
+    assert reading == expected_reading
+
+
+@mock.patch("time.time", mock.MagicMock(return_value=12345))
+async def test_reading_slits_reads_gaps_and_centres(fswitch: FSwitch):
```
joeshannon, line 29: This test needs renaming

src/dodal/devices/i22/fswitch.py
```diff
@@ -0,0 +1,43 @@
...
+            }
+        )
+        super().__init__(name)
+
+    async def read(self) -> Dict[str, Reading]:
+        await super().read()
+        result = await asyncio.gather(
+            *(filter.get_value() for filter in self.filters.values())
+        )
+        num_in = sum(r.value == FilterState.IN for r in result)
```
DiamondJoseph, line 42: Be worth a comment explaining that IN seems to mean "In the box", as opposed to "In the beam"

src/dodal/devices/i22/fswitch.py
```diff
@@ -0,0 +1,43 @@
...
+        self.filters = DeviceVector(
+            {
+                i: epics_signal_r(FilterState, f"{prefix}FILTER-{i:03}:STATUS_RBV")
+                for i in range(FSwitch.NUM_FILTERS)
+            }
+        )
+        super().__init__(name)
+
+    async def read(self) -> Dict[str, Reading]:
+        await super().read()
```
DiamondJoseph, line 38: Should do something with the result of this, for when we make a standard way of reading children devices

src/dodal/devices/i22/fswitch.py
DiamondJoseph: result.count(FilterState.IN)

?

tests/devices/i22/test_fswitch.py
```diff
@@ -0,0 +1,42 @@
...
+    set_sim_value(fswitch.filters.get(0), FilterState.OUT)
+    set_sim_value(fswitch.filters.get(1), FilterState.OUT)
+    set_sim_value(fswitch.filters.get(2), FilterState.OUT)
+
+    await assert_reading(
+        fswitch,
+        {
+            "num_filters": {
+                "timestamp": 12345,
```
DiamondJoseph, line 38: Can also do Any then don't have to patch time.time

src/dodal/devices/i22/fswitch.py
```diff
@@ -0,0 +1,43 @@
...
+from ophyd_async.core.device import DeviceVector
+from ophyd_async.epics.signal import epics_signal_r
+
+
+class FilterState(str, Enum):
+    IN = "IN"
+    OUT = "OUT"
+
+
+class FSwitch(StandardReadable):
```
DiamondJoseph, line 17: Are there signals for the desired config fields: lens geometry, cylindrical, lens_material?

Tom-Willemsen: This is going to be one for the scientists. I imagine it'll be constant (for i22), but I can't see it mentioned in the beamline paper or https://www.diamond.ac.uk/default/Instruments/Soft-Condensed-Matter/small-angle/I22/Ops-modes/I22-TR-SAXS-WAXS.html. I can tag it onto some other questions I've fired at Andy if that's helpful...

tests/devices/i22/test_fswitch.py
```diff
@@ -0,0 +1,42 @@
...
+    return fswitch
+
+
+async def assert_reading(
+    device: StandardReadable,
+    expected_reading: Mapping[str, Any],
+) -> None:
+    await device.stage()
+    reading = await device.read()
+    await device.unstage()
```
DominicOram, line 24: Should: Does this have to be staged/unstaged for the test to work?

src/dodal/devices/i22/fswitch.py
callumforrester: @coretl thinks the opposite so we should probably resolve this at some point. Large point of diverging effort otherwise.

src/dodal/devices/i22/fswitch.py
callumforrester: Could also just rename the enum values to be more explicit?

src/dodal/devices/i22/fswitch.py
callumforrester:
```python
default_reading = await super().read()
return {"num_filters": Reading(value=num_in, timestamp=time.time()), **default_reading}
```

tests/devices/i22/test_fswitch.py
callumforrester: No, that's a copy of my mistake from #431, fix in #474

tests/devices/i22/test_fswitch.py
callumforrester: *`unittest.mock.ANY`, not `typing.Any`

src/dodal/devices/i22/fswitch.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Device for i22's fswitch. This currently only implements the minimum
+    functionality for retrieving the number of lenses inserted.
+
+    Eventually this should be combined with the transfocator device in the i04
+    module but is currently incompatible as the Epics interfaces are different.
```
callumforrester, line 23: Should: I would prefer this as a Github issue rather than a docstring, could just link to the issue here

callumforrester (approved):
Along with I22 config, could also do with P38 config for a fake fswitch, see other optics components already there

callumforrester (changes requested):
Sorry, wrong button!

src/dodal/devices/i22/fswitch.py
DominicOram: Agreed, as discussed at https://github.com/DiamondLightSource/dodal/issues/399#issuecomment-2049824696 I think the first thing to do to combine them would be to reach out to i22 and see if they want the logic for beamsize calculation. @aragaod (i04 beamline scientist) has said he'd be happy to chat to them about how i04 use it if we arrange the discussion.

src/dodal/devices/i22/fswitch.py
joeshannon: Does this need to be recorded in a ticket or is it already noted somewhere?

fajinyuan:
This is my first comment in dodal or Bluesky, so I don't know the class naming convention please clarify for me. So my question is why this class us FSwitch not FiltersSwitch which will be more readable and easier for later joined developer to find what is already available so it can reduce potential duplication and make resue easier.

src/dodal/devices/i22/fswitch.py
Tom-Willemsen: For i22:

> F-switch only contains one type of lens which are paraboloid, cylindrical, beryllium lenses.

joeshannon:
> This is my first comment in dodal or Bluesky, so I don't know the class naming convention please clarify for me. So my question is why this class use "FSwitch", not "FiltersSwitch" which will be more readable and easier for later joined developer to find what is already available so it can reduce potential duplication and make resue easier.

The device is called an "F-Switch" see:
https://accelconf.web.cern.ch/medsi2016/papers/wepe22.pdf

Do people think a better name should be used? I'm not really familiar with the device so open to suggestions.

src/dodal/devices/i22/fswitch.py
joeshannon: After discussion with controls, I have switched the values over.

callumforrester (approved):

src/dodal/devices/i22/fswitch.py
Tom-Willemsen: > new version contains a PV that provides this value already

I presume controls will check the maths with i22 scientists directly? Very much wouldn't assume that i04's maths necessarily just works for i22...

joeshannon: I'm assuming that for the number of lenses in the beam it is just counting up the filter PVs that are "IN_BEAM" as this device is doing? Or is there some complexity I'm missing?

fajinyuan:
Thank @joeshannon. I thought using abbreviation is 'sole right' of software engineers, did not realise hardware enginners also prefer to use it. now I understand it is the name of the Device!

DiamondJoseph (approved):

src/dodal/devices/i22/fswitch.py
coretl: In https://github.com/DiamondLightSource/dodal/issues/415#issuecomment-2043044664 I suggest using the name of the device factory function to name the device. This would mean you could do:

```python
# i22.py
@device_factory(lazy=True)
def fswitch():
    return FSwitch(f"{BL}-MO-FSWT-01:")

assert fswitch().name == "fswitch"
```

If @DominicOram is happy with this approach I will write this up as an ADR for Dodal along with the connection strategy

DominicOram: Yes, I believe that's the default we've tried to go with. Thank you @coretl

src/dodal/devices/i22/fswitch.py
Tom-Willemsen: Suggest just making these soft signals as per https://github.com/DiamondLightSource/dodal/blob/a3e68efaeae6ff16d74ffd7470f056c978feabd9/src/dodal/devices/focusing_mirror.py#L136, and either pass the initial values into the device constructor or default them to something like `""`.

DominicOram: Suggest just making these....

Note: I think we will need to be on the latest `ophyd-async` first. I'm working on it, should be done this morning (https://github.com/DiamondLightSource/dodal/issues/475)

joeshannon: Thanks for the suggestion @Tom-Willemsen I think we would like to have some further thoughts on the soft signals/metadata but may add it in a follow up PR.

PR #477, opened 2024-04-29
(#466) Add udc_directory_provider.py to configure pandas with HDF file directory

### LINKED ISSUE #466: PandAs in ophyd-async 0.3 require DirectoryProvider information
In ophyd-async 0.3, PandAs extend to be StandardDetectors, able to write HDF files. This requires that they are provided with a DirectoryProvider during initialisation.

non-mx beamlines are anticipated to use an implementation of UpdatingDirectoryProvider and the `attach_metadata` decorator to ensure that for each Run, a new collection_id is fetched from the visit service (currently: GDA). Due to requiring Auth information to get the list of visits for a user, the current DirectoryProvider uses a single static visit.

However, currently, the i03 PandA does not write hdf files, or emit StreamResources.

## Acceptance Criteria
* i03's beamline module can be instantiated after upgrading to ophyd-async 0.3

rtuck99:
* Update minimum ophyd-async version to 0.3a1
* Change references to ophyd_async.panda.PandA to HDFPanda

Fixes #466 

See also accompanying hyperion change: DiamondLightSource/hyperion#1337

1. All tests pass
2. hyperion saves panda HDF files in the folder denoted by the `storage_directory` request parameter

### Checks for reviewer

### DIFF
--- a/src/dodal/common/udc_directory_provider.py
+++ b/src/dodal/common/udc_directory_provider.py
@@ -1,0 +1,25 @@
+from pathlib import Path
+
+from ophyd_async.core import DirectoryInfo
+
+_directory_info: DirectoryInfo
+
+
+def set_directory(directory: Path):
+    """
+    Set the current base directory of the UDC DirectoryProvider.
+    Files will be stored in the 'panda' subdirectory.
+    """
+    global _directory_info
+    _directory_info = DirectoryInfo(
+        root=directory, resource_dir=Path("panda"), prefix="", suffix=""
+    )
+
+
+def get_udc_directory_provider():
+    """Get the singleton instance of the UDC DirectoryProvider"""
+    return _get_directory
+
+
+def _get_directory() -> DirectoryInfo:
+    return _directory_info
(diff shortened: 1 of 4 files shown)


src/dodal/common/udc_directory_provider.py
```diff
@@ -0,0 +1,25 @@
...
+    )
+
+
+def get_udc_directory_provider():
+    """Get the singleton instance of the UDC DirectoryProvider"""
+    return _get_directory
+
+
+def _get_directory() -> DirectoryInfo:
+    return _directory_info
```
DiamondJoseph, line 25: Is there a reason to have a seperate get_udc_directory_provider rather than setting the get_directory_provider in beamline_utils to the UDC implementation?

rtuck99: the one in `beamline_utils` requires `UpdatingDirectoryProvider` however the update() method on it doesn't accept a parameter. That  means that we would then have to register some kind of hook to fetch the directory, or just ignore the method entirely.
It seems more straightforward not to implement this method rather than introduce another layer of callbacks.

DiamondJoseph: UpdatingDirectoryProvider.update() isn't a fixed signature right now, I'd be happy for it to become **kwargs: it was just intended as a generic pattern that can be updated in some form

PR #470, opened 2024-04-25
Replace black with ruff

callumforrester:
Closes #311 

CI was starting to fail due to minor disagreements between black and ruff. This PR switches to the ruff formatter as standard like the copier template: https://github.com/DiamondLightSource/python-copier-template/blob/main/.pre-commit-config.yaml

### Instructions to reviewer on how to test:
1. Ensure CI passes
2. Optional: rebase into #461 and see if it fixes CI

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards)

PR #480, opened 2024-04-29
416 port shutter to ophyd async

### LINKED ISSUE #416: Port `shutter` to `ophyd-async`
## Acceptance Criteria
Shutter device is available with ophyd async on a beamline like i20-1, i03 or i04.
Proposed device name: `Shutter`.

Might need a 'reset' method. SampleShutter might be different from HutchShutter - the latter needs a way to reset the interlock.

## current SampleShutter
in `/dodal/devices/sample_shutter.py`
https://github.com/DiamondLightSource/dodal/blob/871e0526fa496e9b068eedd7d19c23027449cc26/src/dodal/devices/sample_shutter.py#L14

stan-dot:
Fixes #416

### Checks for reviewer

### based on these other tests

https://github.com/DiamondLightSource/dodal/pull/469/files#diff-9f79aeed2a351d7575245f7a92dcb60548f1181e2b2ba61d96aa1b1f42dba338

https://github.com/DiamondLightSource/dodal/blob/main/tests/devices/unit_tests/test_tetramm.py


stan-dot:
the errors are quite odd

DominicOram:
> the errors are quite odd
Some of the test errors are due to failures on `main`, if you merge `main` in they should be fixed now

DominicOram (changes requested):
Great, thank you! Can you replace the existing `sample_shutter` with this one please?

tests/devices/unit_tests/test_shutter.py
```diff
@@ -0,0 +1,29 @@
...
+        (OpenState.OPEN),
+        (OpenState.CLOSE),
+    ],
+)
+async def test_set_opens_and_closes_shutter(state: OpenState, sim_shutter: Shutter):
+    status = sim_shutter.set(state)
+    assert not status.done
+    reading = await sim_shutter.read()
+    assert reading == {}
+    # todo check reading makes sense
```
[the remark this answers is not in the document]
stan-dot: yes, it's not complete as I am not sure what output to expect but byt dev setup is a hindrance to run this locally atm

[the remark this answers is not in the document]
stan-dot: will replace the sample_shutter once I get the testing right

stan-dot:
what is this?  @DominicOram I'm stuck
image

DominicOram:
> what is this?

I don't know, could be wierdness in `black` versions. Can you try re-installing `dodal`?

stan-dot:
reloaded dodal, still getting the issue: ` black --version
black, 24.4.2 (compiled: yes)
Python (CPython) 3.11.9`

stan-dot:
could it be that ruff and black disagree here?

DominicOram:
> could it be that ruff and black disagree here?

Possibly, in which case we should increase the priority of getting https://github.com/DiamondLightSource/dodal/pull/470 merged. For now you could just do `--no-verify`

stan-dot:
@DominicOram  it's fixed this error now. working on that test

stan-dot:
> ERROR tests/beamlines/unit_tests/test_i24.py::test_device_creation[i24] - RuntimeError: Bluesky event loop not running
looks like this test should have been put into system tests?

DominicOram:
> looks like this test should have been put into system tests?

This error happens when you try and create a `ophyd-async` device when you don't have a run engine yet. Which is a limitation `ophyd` didn't have. You can fix it by just getting the test to depend on the `RE` fixture.

stan-dot:
both the devcontainer (black, 24.4.2 (compiled: yes)) and the CI (,black==24.4.2) align in the version of black.

the error is very odd, I have no idea https://github.com/DiamondLightSource/dodal/actions/runs/8907643447/job/24461870330?pr=480

stan-dot:
https://github.com/python/mypy/issues/8138

similar issue with mypy

.vscode/settings.json
```diff
@@ -1,5 +1,7 @@
 {
-    "python.testing.pytestArgs": [],
+    "python.testing.pytestArgs": [
+        "tests"
+    ],
```
callumforrester, line 4: Was this causing a problem?

.devcontainer/devcontainer.json
```diff
@@ -20,11 +20,9 @@
         }
     },
     // Set *default* container specific settings.json values on container create.
-    "settings": {
-        "python.defaultInterpreterPath": "/venv/bin/python"
-    },
     "customizations": {
         "vscode": {
+            "python.defaultInterpreterPath": "/venv/bin/python",
```
callumforrester, line 25: Could you make this look like the copier template: https://github.com/DiamondLightSource/python-copier-template/blob/main/.devcontainer/devcontainer.json

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,27 @@
...
+        super().__init__(
+            name=name,
+        )
+        self.set_readable_signals([self.position])
+
+    def set(self, open_val: int | OpenState):
+        if isinstance(open_val, OpenState):
+            open_val = open_val.value
+        shutter_status: AsyncStatus = self.position.set(open_val)
+        return shutter_status
```
callumforrester, line 27: Can't `self.position` simply be a signal of type `OpenState`?

callumforrester (changes requested):

.vscode/settings.json
stan-dot: yes, every time on opening this needed to be configured

.devcontainer/devcontainer.json
stan-dot: ah sure, bad nesting on my end

src/dodal/devices/shutter.py
stan-dot: could be

DiamondJoseph: It could even be 

```python
self.open = epics_signal_rw(
            write_pv=prefix + "CTRL2", read_pv=prefix + "STA", datatype=bool, name=name
        )
```

stan-dot: I'm not sure I understand the case for simplicity here.

In my model of this, I require the 'set' method to override the device 'set' method. 

I agree that OpenState is a bit of a redundant thing.


But I'd leave it at int as bool is less flexible and IDK how does it go throguh epics channels

DiamondJoseph: Less flexible is good, the shutter has literally 2 states. A bool is the perfect representation of that.

here https://github.com/bluesky/ophyd-async/blob/f6f9cc0762f8c4a9c6555a3734799b0269ae266c/src/ophyd_async/epics/_backend/_p4p.py#L203
and here https://github.com/bluesky/ophyd-async/blob/f6f9cc0762f8c4a9c6555a3734799b0269ae266c/src/ophyd_async/epics/_backend/_aioca.py#L128
are where bool is handled.

The signal should either be an Enum or a bool, I am insistent on that. int is just incorrect.

DominicOram (commented):
I agree with what others have said. Additionally, can you make sure https://github.com/DiamondLightSource/hyperion/pull/1351 gets merged at the same time as this?

src/dodal/devices/shutter.py
stan-dot: dropped

stan-dot:
@DominicOram I am running into a new error to do with slits, what is going on there?

DominicOram:
> @DominicOram I am running into a new error to do with slits, what is going on there?

Try merging in main?

DiamondJoseph:
I have merged main in to get the CI happier, be sure to pull before doing any work @stan-dot

stan-dot:
I was assuming that the main is free from bugs detectable in CI. And that pulling mid-PR is chaotic.

callumforrester:
@stan-dot no, PR should be kept up-to-date with main to avoid hidden bugs. Possibly we should make the repo require the branch to be up-to-date before merging

stan-dot:
>  PR should be kept up-to-date with main to avoid hidden bugs
PRs should be short and quickly merged back into main.  a branch does not go into deployment so hidden bugs don't matter.

you're not very specific what 'hidden' means here. "bugs detectable in CI " was a wording I precisely used

.devcontainer/devcontainer.json
```diff
@@ -1,54 +1,51 @@
...
+    "source=${localEnv:HOME}/.inputrc,target=/root/.inputrc,type=bind",
+    // map in home directory - not strictly necessary but useful
+    "source=${localEnv:HOME},target=${localEnv:HOME},type=bind,consistency=cached"
+  ],
+  // make the workspace folder the same inside and outside of the container
+  "workspaceMount": "source=${localWorkspaceFolder},target=${localWorkspaceFolder},type=bind",
+  "workspaceFolder": "${localWorkspaceFolder}",
+  // After the container is created, install the python project in editable form
+  "postCreateCommand": "pip install -e .[dev]"
+}
```
DiamondJoseph, line 51: These changes are not related and should be their own PR if you think they are required, possibly into the DiamondLightSource/python-copier-template.

.vscode/settings.json
DiamondJoseph, line 4: As above, not related to this change. I agree that it's generally useful, but it should go into python-copier-template.

tests/devices/unit_tests/test_shutter.py
```diff
@@ -0,0 +1,47 @@
...
+@pytest.mark.parametrize(
+    "state",
+    [
+        (OpenState.OPEN),
+        (OpenState.CLOSE),
+    ],
+)
+async def test_set_opens_and_closes_shutter(state: OpenState, sim_shutter: Shutter):
+    status = sim_shutter.set(state)
+    assert not status.done
```
DiamondJoseph, line 26: Why assert not status.done?

tests/devices/unit_tests/test_shutter.py
```diff
@@ -0,0 +1,47 @@
...
+    reading = await sim_shutter.read()
+
+    shutter_position = reading.get("shutter-position", {})
+    assert (
+        shutter_position["value"] in OpenState
+    ), f"Unexpected value: {shutter_position['value']}"
```
DiamondJoseph, line 32: Should assert that the state is the expected state

tests/devices/unit_tests/test_shutter.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # Assert that 'alarm_severity' is 0
+    assert shutter_position.get("alarm_severity") is not None, "Alarm severity is None"
+    severity = shutter_position["alarm_severity"]  # type: ignore
+    assert severity == 0, f"Alarm severity is not 0: {severity}"
+
+    if state == OpenState.CLOSE:
+        assert shutter_position["value"] == OpenState.CLOSE
+    if state == OpenState.OPEN:
+        assert shutter_position["value"] == OpenState.OPEN
```
DiamondJoseph, line 47: This can be done above and simpler

tests/devices/unit_tests/test_shutter.py
DiamondJoseph, line 32:
```suggestion
    assert (
        shutter_position["value"] is state
    ), f"Unexpected value: {shutter_position['value']}"
```

DiamondJoseph (changes requested):

DiamondJoseph:
>  Possibly we should make the repo require the branch to be up-to-date before merging

I'll make an issue for this as I think it definitely should if not already

stan-dot:
I am expecting to rebaseon /merge in the main branch but only when the PR is all ready to merge. weird issues before that imply that something is wrong with the change commits. 


a project should assume that the main is bug free (definitely in CI) and if the deployment state is something very odd, discovering a bug that escaped the CI, then a 'hotfix' branch to the main is done. 

nowhere is a 'frequent merge' needed from main into branch. afaik

.vscode/settings.json
stan-dot: valid

.devcontainer/devcontainer.json
stan-dot: not valid point, as already this pattern is here. 

> each small change should be own PR

what difference does it make, it's just paperwork

https://github.com/DiamondLightSource/python-copier-template/blob/main/.devcontainer/devcontainer.json

tests/devices/unit_tests/test_shutter.py
stan-dot: as that is what was in the `sample_shutter` test:
`import pytest
from ophyd.sim import make_fake_device

from dodal.devices.sample_shutter import OpenState, SampleShutter


@pytest.fixture
def fake_sample_shutter():
    FakeSampleShutter = make_fake_device(SampleShutter)
    fake_sample_shutter: SampleShutter = FakeSampleShutter(name="sample_shutter")
    return fake_sample_shutter


def set_pos_rbv(fake_sample_shutter: SampleShutter, value):
    fake_sample_shutter.pos_rbv.sim_put(value)  # type: ignore


@pytest.mark.parametrize(
    "state",
    [
        (OpenState.OPEN),
        (OpenState.CLOSE),
    ],
)
def test_set_opens_and_closes_shutter(state, fake_sample_shutter):
    status = fake_sample_shutter.set(state)
    assert not status.done
    set_pos_rbv(fake_sample_shutter, state.value)
    status.wait(1)
`

I don't know much about shutters, so kept that behavior

tests/devices/unit_tests/test_shutter.py
stan-dot: that is asserted some lines later, should we skip dealing with the timestamp?

stan-dot: changed as requested

tests/devices/unit_tests/test_shutter.py
stan-dot: done

.devcontainer/devcontainer.json
stan-dot: I see the point definitely for different part of the codebase that do different features.


but this small dev-env change has been requested by vscode. the change is to put:
`1  "vscode": {
   2  "settings": {
     3   "python.defaultInterpreterPath": "/venv/bin/python"
    4},`
    line 3 into vscode and not the global settings. 
    
    it's just git that's complaining about many lines

.devcontainer/devcontainer.json
callumforrester, line 1: Are these changes intentional?

src/dodal/devices/shutter.py
callumforrester: Going to unresolve this one as I still think it should be a signal. That would also make the tests easier as you can use `set_sim_value` to mock the state and `get_value` rather than `read`

callumforrester (changes requested):

.devcontainer/devcontainer.json
stan-dot: only 1 line change actually. main must have got out of sync with the copier template
it's about the location of the python interpreter - inside vscode/settings, not outside

stan-dot:
image
@callumforrester  it is a signal, right? unless I'm mistaken about how ophyd-async works (highly likely)

.devcontainer/devcontainer.json
callumforrester: What was the intentional change? Or just rebase and we can see

src/dodal/devices/shutter.py
callumforrester: @stan-dot https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/undulator.py

These are signals, you shouldn't have to override `set` on the class inheriting from `StandardReadable`, you should just be able to add signals to it.

.devcontainer/devcontainer.json
stan-dot: ? not sure what you mean ?

that's bluesky for instanc e- here the settings.python.defaultinterpreter path is inside vscode customizations.
that's what my intentional change was towards. copier template has the same thing
image

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,28 @@
+from enum import Enum
+
+from bluesky.protocols import Movable
+from ophyd_async.core import AsyncStatus, StandardReadable
+from ophyd_async.epics.signal import epics_signal_rw
+
+
+class OpenState(Enum):
+    CLOSE = 0
+    OPEN = 1
```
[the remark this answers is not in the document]
callumforrester: I think this will currently fail, have you tested it against the real shutter PVs?

tests/devices/unit_tests/test_shutter.py
callumforrester: `status` should be awaited at the end of the test then, to avoid leakage, which also preserves the original behavoir

.devcontainer/devcontainer.json
callumforrester: Happy for PRs that move it towards the copier template (or just wholesale adopt it) but could they please be _separate_ PRs?

src/dodal/devices/shutter.py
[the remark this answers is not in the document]
stan-dot: why would it fail? I haven't tested it on a beamline

tests/devices/unit_tests/test_shutter.py
stan-dot: that was just an example, not this PR

.devcontainer/devcontainer.json
stan-dot: ok will do

.vscode/settings.json
```diff
@@ -1,5 +1,4 @@
 {
-    "python.testing.pytestArgs": [],
```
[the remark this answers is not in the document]
stan-dot: yes

stan-dot:
@callumforrester the main thing is about the signal

.devcontainer/devcontainer.json
stan-dot: extracting the changes. `.devcontainer` was not changed when it was moving towards the new template...
image

tests/devices/unit_tests/test_shutter.py
callumforrester: It still needs to be awaited

.vscode/settings.json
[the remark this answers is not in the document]
callumforrester: I mean this: https://github.com/DiamondLightSource/dodal/pull/480#discussion_r1595282308

I'm afraid I am really not in favour of PRs that make ancillary changes to the vscode config on top of other things (that the PR is actually supposed to be about).

PRs that are kept small and self-contained are easier to review and show a clearer history. There is nothing stopping you from raising lots of small PRs btw.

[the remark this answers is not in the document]
stan-dot: but I didn't change this specific file. I do not make ancillary changes to vscode config, it does it automatically. I go along with it as I don't want to fight against my IDE. possibly git ignoring `.vscode ` would be a kind of solution to this.

stan-dot:
image

this works on my machine, is it using a different version of ophyd_async I presume

coretl:
> image
> 
> this works on my machine, is it using a different version of ophyd_async I presume

This was changed recently as part of the sim -> mock changes. However this import path was never meant to be part of the public interface. This is why we need to get https://github.com/bluesky/ophyd-async/pull/26 done so we tie down our public interface and make everything private.

stan-dot:
> @github-actions
Code CI / test (ubuntu-latest, 3.10, .[dev]) (push) In progress — This check has started...
[Details](https://github.com/DiamondLightSource/dodal/actions/runs/9093802386/job/24993518220?pr=480)




this job takes over an hour, odd

stan-dot:
how soon would that be fixed @coretl  ? or should I just pin to the previous version? if yes, which version number would be safe from that change?

dev-requirements.txt
callumforrester, line 1: Should this have been committed?

stan-dot: my bad

src/dodal/devices/shutter.py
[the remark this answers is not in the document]
coretl: `dlsPlc` provided shutter has 5 status:
- Open
- Closed
- Opening
- Closing
- Fault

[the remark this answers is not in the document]
stan-dot: fantastic, so we do enum not bool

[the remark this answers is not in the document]
DiamondJoseph: > fantastic, so we do enum not bool

But it **must** be OpenState(str, Enum) and **must** capture all 5 states, else the [signalbackend will prevent connection when using a non-sim backend](https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/epics/_backend/common.py).

tests/devices/unit_tests/test_shutter.py
stan-dot: ok, done

src/dodal/devices/shutter.py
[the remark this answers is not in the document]
stan-dot: changed to support all 5 types

stan-dot:
it doesn't like not implementing '`set`, @callumforrester , what to do?
image

src/dodal/common/types.py
```diff
@@ -18,4 +18,4 @@
 
 class UpdatingDirectoryProvider(DirectoryProvider, ABC):
     @abstractmethod
-    def update(self, **kwargs) -> None: ...
```
DiamondJoseph, line 21: This change is required for Hyperion, please revert (or rebase on top of master)

DiamondJoseph:
> it doesn't like not implementing '`set`

It's extending Movable but doesn't define a set. Either remove it extending Movable (in which case only shutter.position is Movable, or define a set method that defers to the underlying signal.

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,27 @@
...
+    FAULT = "fault"
+
+
+class Shutter(StandardReadable, Movable):
+    def __init__(self, prefix: str, name: str):
+        self.position = epics_signal_rw(
+            write_pv=prefix + "CTRL2",
+            read_pv=prefix + "STA",
+            datatype=OpenState,
+            name=name,
```
DiamondJoseph, line 22: shutter.position.name shouldn't be equal to shutter.name (as this change currently does). shutter.position.name should retain the default "" (i.e., don't pass it into this call) and it will be named by the DeviceCollector.

DiamondJoseph (changes requested):

src/dodal/devices/shutter.py
stan-dot: ok

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Shutter(StandardReadable):
+    def __init__(self, prefix: str, name: str):
+        self.position = epics_signal_rw(
+            write_pv=prefix + "CTRL2", read_pv=prefix + "STA", datatype=OpenState
+        )
+        super().__init__(
+            name=name,
+        )
+        self.set_readable_signals([self.position])
```
DiamondJoseph, line 23: nit: Slightly neater maybe, but I won't hold the change up anymore.

```suggestion
        with self.add_children_as_readables(ConfigSignal):
            self.position = epics_signal_rw(
                write_pv=prefix + "CTRL2", read_pv=prefix + "STA", datatype=OpenState
           )
        super().__init__(name=name)
```

stan-dot: I had no idea this method exists on the StandardReadable. what difference does it make?

if it's just another way to do the same thing ie syntactic sugar, it's not Pythonic
Zen of Python line 13
> There should be one-- and preferably only one --obvious way to do it.

DiamondJoseph: It was added during the bluesky hackathon (i.e. "very" recently) and shows up in the auto-generated docs for the StandardReadable, but I agree that it isn't obvious which is why it's a nit.

https://github.com/bluesky/ophyd-async/pull/240/files

tests/devices/unit_tests/test_shutter.py
DiamondJoseph: Unresolving as I think this will be non-deteministic in the tests. 

I think we should await the status here instead: it's the status for the move, if we're not mocking out the move, then we can't reliably say that it hasn't completed already; we're also immediately after reading the signal and asserting that its value is the value that we have told it to move to (i.e. that the move has completed). We should instead request the move, await the move, then assert that we have moved to the requested position.

tests/beamlines/unit_tests/test_i24.py
```diff
@@ -4,7 +4,10 @@
 
 
 @pytest.mark.parametrize("module_and_devices_for_beamline", ["i24"], indirect=True)
-def test_device_creation(RE, module_and_devices_for_beamline):
+def test_device_creation(
+    RE,
+    module_and_devices_for_beamline,
+):
```
DiamondJoseph, line 10: Can this unrelated formatting change (if it doesn't break the CI) be reverted?

stan-dot: done

src/dodal/devices/shutter.py
stan-dot: done.

tests/beamlines/unit_tests/test_i24.py
DiamondJoseph: Not quite, a space snuck in

src/dodal/devices/shutter.py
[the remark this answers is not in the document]
DominicOram: Must: This does not currently work against the real device. Running `python tests/system_tests/test_i03_system.py` gives:
```
ERROR:root:device `position` raised unexpected exception TypeError
  File "/scratch/ffv81422/hyperion/hyperion/.venv/lib/python3.11/site-packages/ophyd_async/core/signal.py", line 77, in connect
    await self._backend.connect(timeout=timeout)
  File "/scratch/ffv81422/hyperion/hyperion/.venv/lib/python3.11/site-packages/ophyd_async/epics/_backend/_aioca.py", line 202, in connect
    self.converter = make_converter(self.datatype, self.initial_values)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/hyperion/hyperion/.venv/lib/python3.11/site-packages/ophyd_async/epics/_backend/_aioca.py", line 145, in make_converter
    supported_values = get_supported_values(pv, datatype, pv_choices)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/hyperion/hyperion/.venv/lib/python3.11/site-packages/ophyd_async/epics/_backend/common.py", line 18, in get_supported_values
    raise TypeError(
TypeError: BL03I-EA-SHTR-01:STA has choices ('Close', 'Open'), which do not match <enum 'OpenState'>, which has ('closed', 'open', 'opening', 'closing', 'fault')
  File "/scratch/ffv81422/hyperion/dodal/tests/system_tests/test_i03_system.py", line 22, in <module>
    make_all_devices(i03)
  File "/scratch/ffv81422/hyperion/dodal/src/dodal/utils.py", line 128, in make_all_devices
    devices: dict[str, AnyDevice] = invoke_factories(factories, **kwargs)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/hyperion/dodal/src/dodal/utils.py", line 153, in invoke_factories
    devices[dependent_name] = factories[dependent_name](**params, **kwargs)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/hyperion/dodal/src/dodal/utils.py", line 101, in wrapper
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/hyperion/dodal/src/dodal/beamlines/i03.py", line 401, in sample_shutter
    return device_instantiation(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/hyperion/dodal/src/dodal/utils.py", line 101, in wrapper
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/scratch/ffv81422/hyperion/dodal/src/dodal/beamlines/beamline_utils.py", line 112, in device_instantiation
    _wait_for_connection(device_instance, mock=fake)
  File "/scratch/ffv81422/hyperion/dodal/src/dodal/beamlines/beamline_utils.py", line 55, in _wait_for_connection
    call_in_bluesky_event_loop(
  File "/scratch/ffv81422/hyperion/hyperion/.venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 2782, in call_in_bluesky_event_loop
    return fut.result(timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/scratch/ffv81422/hyperion/hyperion/.venv/lib/python3.11/site-packages/ophyd_async/core/utils.py", line 118, in wait_for_connection
    raise NotConnected(exceptions)
ophyd_async.core.utils.NotConnected: 
coros: NotConnected:
    position: TypeError: BL03I-EA-SHTR-01:STA has choices ('Close', 'Open'), which do not match <enum 'OpenState'>, which has ('closed', 'open', 'opening', 'closing', 'fault')
```

It also seems bad that a user can set the position to `Opening` or `Fault`... I think we need two separate signals, with two separate enums for setpoint and readback.

DominicOram (changes requested):
Thanks @stan-dot! Main issue is https://github.com/DiamondLightSource/dodal/pull/480#discussion_r1606783981, which we must resolve before merging. 

> It's extending Movable but doesn't define a set. Either remove it extending Movable (in which case only shutter.position is Movable, or define a set method that defers to the underlying signal.

My strong preference would be to make this `Movable` and define the `set`. I think it's more readable for a user to do `yield from bps.abs_set(shutter, OpenState.Open)` than `yield from bps.abs_set(shutter.position, OpenState.Open)`.

Must: We also need to make sure that we wait on the shutter to finish moving before returning from the `set`. This is what the old one did by comparing to the `RBV`. We might be able to get away with relying on the put complete but you will need to check with controls. I think just comparing to the RBV is probably cleaner

src/dodal/devices/shutter.py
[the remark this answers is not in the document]
DiamondJoseph: Which makes some sense with the different suffixes for read/set. CTRL with datatype `ShutterState(str, Enum): OPEN, CLOSED` and STA with datatype str?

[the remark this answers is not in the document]
DominicOram: `STA` can be an `enum` too, with the states that Tom mentions above

tests/beamlines/unit_tests/test_i24.py
stan-dot: done

stan-dot:
@DominicOram I don't understand the relation of your comment to this comment
https://github.com/DiamondLightSource/dodal/pull/480#discussion_r1596315940


I am clearly not understanding everything. and why is it only now that this requirement of 'set returns a verified value' shows up?

src/dodal/devices/shutter.py
```diff
@@ -1,21 +1,41 @@
...
-from ophyd_async.epics.signal import epics_signal_rw
+from bluesky.protocols import Movable
+from ophyd_async.core import AsyncStatus, StandardReadable, wait_for_value
+from ophyd_async.epics.signal import epics_signal_r, epics_signal_w
 
 
-class OpenState(str, Enum):
+class OpenStateSet(str, Enum):
+    CLOSED = "close"
+    OPEN = "open"
```
DiamondJoseph, line 11: This needs reset

src/dodal/devices/shutter.py
```diff
@@ -1,21 +1,41 @@
...
+
+
+class OpenStateReadback(str, Enum):
```
DiamondJoseph, line 14: I think the enums should be prefixed with the device type name, similarly to how the StandardDetectors are being written (e.g. AravisTriggerMode to distinguish from other TriggerModes).

ShutterSetState and ShutterState?

src/dodal/devices/shutter.py
```diff
@@ -1,21 +1,41 @@
...
         with self.add_children_as_readables():
-            self.position = epics_signal_rw(
-                write_pv=prefix + "CTRL2", read_pv=prefix + "STA", datatype=OpenState
+            self.position_readback = epics_signal_r(
+                read_pv=prefix + "STA",
+                datatype=OpenStateReadback,
+            )
+            self.position_set = epics_signal_w(
+                write_pv=prefix + "CTRL2",
+                datatype=OpenStateSet,
```
DiamondJoseph, line 31: I don't think self.position_set should be a readback value. It should be outside of the add_children_as_readables() context

src/dodal/devices/shutter.py
```diff
@@ -1,21 +1,41 @@
...
             )
         super().__init__(name=name)
+
+    def set(self, position: OpenStateSet):
+
+        return AsyncStatus(
+            asyncio.wait_for(
+                await wait_for_value(self.position_readback, position, None), timeout=5
+            )
+        )
```
DiamondJoseph, line 41: This is never sending the set value down to the signal, just waiting for it to change to the correct value?

DiamondJoseph (changes requested):

src/dodal/devices/shutter.py
stan-dot: done

src/dodal/devices/shutter.py
stan-dot: what would be the logic for this then?

src/dodal/devices/shutter.py
stan-dot: done

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,40 @@
...
+                write_pv=prefix + "CTRL2",
+                datatype=ShutterSetState,
+            )
+        super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def set(self, position: ShutterSetState):
+        new_position = ShutterState(position)
+        await self.position_set.set(position)
+        return await wait_for_value(
```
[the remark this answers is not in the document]
DominicOram: Should:
```python
if position == ShutterSetState.reset:
    return
```

src/dodal/devices/shutter.py
stan-dot: done

src/dodal/devices/shutter.py
stan-dot: added, with an early return logic for 'set' comparison for readback

src/dodal/devices/shutter.py
stan-dot: yep

src/dodal/devices/shutter.py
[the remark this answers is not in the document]
stan-dot: done, now only linting

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,44 @@
...
+    OPEN = "open"
+    RESET = "reset"
+
+
+class ShutterState(str, Enum):
+    CLOSED = ShutterSetState.CLOSED
+    OPEN = ShutterSetState.OPEN
+    OPENING = "opening"
+    CLOSING = "closing"
+    FAULT = "fault"
```
DiamondJoseph, line 19: Must:
RE: Tom's comment with the potential states, emphasis mine:

CTRL:

    **O**pen
    **Close**
    **R**eset

STA:

    **O**pen
    **Closed**
    **O**pening
    **C**losing
    **F**ault

I'd make an explicit new string for ShutterState.OPEN instead of reusing ShutterSetState.OPEN too.

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,44 @@
...
+            self.position_readback = epics_signal_r(
+                read_pv=prefix + "STA",
+                datatype=ShutterState,
+            )
+        super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def set(self, position: ShutterSetState):
+        if position == ShutterSetState.RESET:
+            return
```
DiamondJoseph, line 38: @DominicOram why should this shortcut? What does set(Reset) do on the IOC?

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,44 @@
...
+        assert position in ShutterState
+        new_position = ShutterState(position)
+        await self.position_set.set(position)
+        return await wait_for_value(
+            signal=self.position_readback, match=new_position, timeout=50
+        )
```
DiamondJoseph, line 44: Must: this all then doesn't work because Closed vs Close

tests/devices/unit_tests/test_shutter.py
```diff
@@ -0,0 +1,32 @@
...
+)
+async def test_set_opens_and_closes_shutter(state: ShutterState, sim_shutter: Shutter):
+    status = sim_shutter.set(state)
+    reading = await sim_shutter.read()
+
+    shutter_position = reading.get("shutter-position", {})
+    assert (
+        shutter_position["value"] is state
+    ), f"Unexpected value: {shutter_position['value']}"
+    await status
```
DiamondJoseph, line 32:
```suggestion
    await sim_shutter.set(state)
    reading = await sim_shutter.read()

    shutter_position = reading.get("shutter-position", {})
    assert (
        shutter_position["value"] is state
    ), f"Unexpected value: {shutter_position['value']}"
```

nit: Awaiting the set rather than holding onto a reference that you then need to await later?

DiamondJoseph (changes requested):

src/dodal/devices/shutter.py
DiamondJoseph: Open, Close, Reset are then commands not states.

ShutterCommand, ShutterState?

src/dodal/devices/shutter.py
DominicOram: I don't know. To be honest, I'm not sure we should even expose it as an option but the Enum has to be exhaustive.

src/dodal/devices/shutter.py
stan-dot: capitalization - yeah, I must have missed it

src/dodal/devices/shutter.py
stan-dot: yeah, good catch

tests/devices/unit_tests/test_shutter.py
stan-dot: sounds good

src/dodal/devices/shutter.py
stan-dot: these are commands, but also states of the channel. IMO many devices have this 'state'

stan-dot:
all callback put on mock https://blueskyproject.io/ophyd-async/main/how-to/write-tests-for-devices.html

stan-dot:
todo: add a sim version

stan-dot:
unexpected argument `wait`
image

tests/devices/unit_tests/test_shutter.py
```diff
@@ -0,0 +1,44 @@
...
+    assert (
+        shutter_position["value"] is ShutterState.OPEN
+    ), f"Unexpected value: {shutter_position['value']}"
+
+
+async def test_set_shutter_close(sim_shutter: Shutter):
+    desired_state: ShutterSetState = ShutterSetState.CLOSE
+    callback_on_mock_put(
+        sim_shutter.position_set,
+        lambda _: set_mock_value(sim_shutter.position_readback, ShutterState.CLOSED),
```
DiamondJoseph, line 36: nit: I'd prefer to see this on the fixture, with sensible open/open close/closed

stan-dot: ok but it's not passing the test. it might have to do with ophyd-async

stan-dot: not sure how this is supposed to go @DiamondJoseph

coretl:
Ok, ignore all feedback I have given on this PR...

```bash
tmc43@pc0105:~ $ caget -d 31 BL04I-EA-SHTR-01:STA
BL04I-EA-SHTR-01:STA
    Native data type: DBF_ENUM
    Element count:    1
    Value:            Close
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            ( 2)
                      [ 0] Close
                      [ 1] Open
tmc43@pc0105:~ $ caget -d 31 BL04I-EA-SHTR-01:CTRL2
BL04I-EA-SHTR-01:CTRL2
    Native data type: DBF_ENUM
    Element count:    1
    Value:            Close
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            ( 2)
                      [ 0] Close
                      [ 1] Open
tmc43@pc0105:~ $ caget BL04I-EA-SHTR-01:CTRL2.OUT
BL04I-EA-SHTR-01:CTRL2.OUT     BL04I-EA-ZEBRA-01:SOFT_IN:B1 PP NMS
```

This is a shutter controlled by a zebra output, I thought you meant a "port shutter" which is the PLC controlled shutter to the optics or experimental hutch, which has the Open/Close/Reset. I have no idea what this thing actually is!

src/dodal/devices/shutter.py
coretl, line 1: I would call this a `ZebraControlledShutter` or something similar. I have never seen this used outside MX.

src/dodal/devices/shutter.py
```diff
@@ -0,0 +1,47 @@
...
+    OPEN = "Open"
+    RESET = "Reset"
+
+
+class ShutterState(str, Enum):
+    CLOSED = "Closed"
+    OPEN = "Open"
+    OPENING = "Opening"
+    CLOSING = "Closing"
+    FAULT = "Fault"
```
coretl, line 19:
```suggestion
```suggestion
class ZebraShutterState(str, Enum):
    CLOSE = "Close"
    OPEN = "Open"
```
```

File: src/dodal/devices/shutter.py [hunk 26]
```diff
@@ -0,0 +1,47 @@
...
+            return
+        new_position = (
+            ShutterState.CLOSED
+            if desired_position == ShutterSetState.CLOSE
+            else ShutterState.OPEN
+        )
+        await self.position_set.set(desired_position)
+        return await wait_for_value(
+            signal=self.position_readback, match=new_position, timeout=50
+        )
```
coretl, line 47:
```suggestion
    async def set(self, state: ZebraShutterState):
        await self.position_set.set(state)
        return await wait_for_value(
            signal=self.position_readback, match=state, timeout=DEFAULT_TIMEOUT
        )
```

coretl (changes requested):

src/dodal/devices/shutter.py
stan-dot: ok so how many shutters there are? I didn't make it with a particular PV in mind, but as a generic shutter, so to speak. 

for i20-1, i03 and i04. https://github.com/DiamondLightSource/dodal/issues/416 

@DominicOram is the goal of having a generic shutter viable or is it necessary to split into 2 devices?

DominicOram: There will be a shutter on every beamline. Which beamline do you particularly want this for? i20_1? In which case it looks like the interface is how Tom described it:
```
BL20J-PS-SHTR-02:CON
    Native data type: DBF_ENUM
    Element count:    1
    Value:            Open
    Status:           UDF
    Severity:         INVALID
    Enums:            ( 3)
                      [ 0] Open
                      [ 1] Close
                      [ 2] Reset
$ caget -d31 BL20J-PS-SHTR-02:STA
BL20J-PS-SHTR-02:STA
    Native data type: DBF_ENUM
    Element count:    1
    Value:            Closed
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            ( 5)
                      [ 0] Fault
                      [ 1] Open
                      [ 2] Opening
                      [ 3] Closed
                      [ 4] Closing
```
Given that the names of the PVs are different and the enums are different I don't see how we could share the exact same ophyd device for both. However, I think they should share an interface. What is the purpose of this issue? Do you want a shutter for i20_1? Do you want to convert i03/i04s? I suggest you do one of those and close this PR out, if we want to make a grand unified shutter we can make a new PR.

stan-dot: `i20_1` is what I wanted a shutter for, but not it's no longer relevant. 

I agree now that with hindsight that this could be another PR on its own https://github.com/DiamondLightSource/dodal/pull/480#pullrequestreview-2029310222

DominicOram: Well if `i20_1` is no longer needed I suggest you either call this just a conversion of i03/i04's

stan-dot: @DominicOram should I just go with the suggestion to call the device `ZebraControlledShutter` and when enums to use then?

DominicOram: Sounds great, thanks!

src/dodal/devices/shutter.py
stan-dot: added

tests/devices/unit_tests/test_shutter.py
stan-dot: not sure how to apply which after the change

tests/devices/unit_tests/test_shutter.py
```diff
@@ -0,0 +1,49 @@
...
+        set_mock_value(sim_shutter.position_readback, ZebraShutterState.CLOSE)
+
+    callback_on_mock_put(sim_shutter.position_set, set_correct_mock_value)
+    await sim_shutter.set(desired_state)
+    reading = await sim_shutter.read()
+
+    shutter_position = reading.get("shutter-position_readback", {})
+    assert (
+        shutter_position["value"] is ZebraShutterState.CLOSE
+    ), f"Unexpected value: {shutter_position['value']}"
```
DiamondJoseph, line 49:
```suggestion
@pytest.fixture
async def sim_shutter():
    async with DeviceCollector(mock=True):
        sim_shutter: ZebraShutter = ZebraShutter(
            prefix="sim_shutter",
            name="shutter",
        )

    def propagate_status(value:  ZebraShutterState):
         set_mock_value(sim_shutter.position_readback, value)
    callback_on_mock_put(sim_shutter.position_set, propagate_status)
    return sim_shutter


@pytest.mark.parametrize("new_state", [ZebraShutter.OPEN, ZebraShutter.CLOSE])
async def test_set_shutter_sets_child(sim_shutter: ZebraShutter, new_state: ZebraShutterState):
    await sim_shutter.set(new_state)
    reading = await sim_shutter.read()

    shutter_position = reading.get("shutter-position_readback", {})
    assert (
        shutter_position["value"] is new_state
    ), f"Unexpected value: {shutter_position['value']}"
```

Does this work?

src/dodal/devices/zebra_controlled_shutter.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, prefix: str, name: str):
+        self.position_set = epics_signal_w(
+            write_pv=prefix + "CTRL2",
+            datatype=ZebraShutterState,
+        )
+        with self.add_children_as_readables():
+            self.position_readback = epics_signal_r(
+                read_pv=prefix + "STA",
+                datatype=ZebraShutterState,
+            )
```
DiamondJoseph, line 23: nit: these signals used to be called `pos` and `pos_rbv`

If Hyperion expects them to be called something in particular and ever reads rather than just setting this is a Must not a Nit.

src/dodal/devices/zebra_controlled_shutter.py
```diff
@@ -0,0 +1,31 @@
...
+        super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def set(self, desired_position: ZebraShutterState):
+        await self.position_set.set(desired_position)
+        return await wait_for_value(
+            signal=self.position_readback, match=desired_position, timeout=50
+        )
```
DiamondJoseph, line 31: nit: prefer ophyd-async's DEFAULT_TIMEOUT over this magic number (let's use one common magic number!)

DiamondJoseph (changes requested):

src/dodal/devices/zebra_controlled_shutter.py
stan-dot: @DominicOram how do you stand about this? I'm all for readable variable names

tests/devices/unit_tests/test_shutter.py
stan-dot: yep, that works, adding this now

stan-dot: done

src/dodal/devices/zebra_controlled_shutter.py
stan-dot: done

src/dodal/devices/zebra_controlled_shutter.py
dperl-dls: I'm also all for readable names, and Since the device name is changing it will have to be updated in Hyperion anyway. I would like it if you would update the Hyperion PR to match this one. Also may I suggest `position_setpoint` instead?

stan-dot: added `position_setpoint`

DiamondJoseph (approved):

DiamondJoseph:
I am dismissing reviews from @coretl because he's on leave for a week

DominicOram (approved):
Great, thank you. Can you make sure to approve + merge https://github.com/DiamondLightSource/hyperion/pull/1419 at the same time?

stan-dot:
Yes, I keep in memory to do that, will do around 3pm hopefully

PR #505, opened 2024-05-07
Convert UDCDirectoryProvider to standard pattern

DiamondJoseph:
1. Check Hyperion PR still behaves as expected
2. https://github.com/DiamondLightSource/hyperion/pull/1352

### Checks for reviewer


src/dodal/common/udc_directory_provider.py
```diff
@@ -2,24 +2,23 @@
...
-def set_directory(directory: Path):
-    """
-    Set the current base directory of the UDC DirectoryProvider.
-    Files will be stored in the 'panda' subdirectory.
-    """
-    global _directory_info
-    _directory_info = DirectoryInfo(
-        root=directory, resource_dir=Path("panda"), prefix="", suffix=""
-    )
+class UDCDirectoryProvider(UpdatingDirectoryProvider):
```
olliesilvester, line 8: I think change the name of this class since it should be used specifically for storing panda postition data, something like `PandaPcapDirectoryProvider`? 

Also,  this `DirectoryProvider` points to a `panda` subdirectory which doesn't exist by default. We should add a comment to warn that this directory must already exist when using with a PandA - I don't think there's logic in the HDF writer or the IOC to create the directory for us

src/dodal/common/udc_directory_provider.py
```diff
@@ -2,24 +2,23 @@
...
+    resource_dir = Path("panda")
 
+    def __init__(self, directory: Path | None = None):
+        self._directory_info = (
+            DirectoryInfo(root=directory, resource_dir=self.resource_dir)
+            if directory
+            else None
```
olliesilvester, line 15: Could: Log a warning if we try and initialise this without specifying a directory

olliesilvester (changes requested):
Couple of nitty comments, but the logic and test looks good!

src/dodal/common/udc_directory_provider.py
DiamondJoseph: Something coming down the pipe on that respects: https://github.com/bluesky/ophyd-async/blob/5c116e293afb54b437ffa2ab380635a3c96b19b2/src/ophyd_async/core/_providers.py#L116 Will be handled in the PandA hdf writer at some point. I'll revisit tomorrow

olliesilvester (approved):
Thanks! I'll add something to create the panda directory if it doesn't already exist, in a separate PR

PR #510, opened 2024-05-07
Pin sphinx-autobuild to play nice with blueapi's fastapi version

DominicOram:
See https://github.com/DiamondLightSource/hyperion/issues/1346 and https://github.com/DiamondLightSource/blueapi/pull/456

I feel like this shouldn't be necessary as the pin is already in `blueapi` itself and this should just use that version but `hyperion` doesn't and it doesn't seem worth going down the rabbithole.

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dev = [
     "pytest-cov",
     "pytest-random-order",
     "ruff",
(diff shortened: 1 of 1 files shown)


dperl-dls (approved):
:(

PR #511, opened 2024-05-07
Add NeXus-friendly soft signals to Undulator

### LINKED ISSUE #389: IVU Compatible Insertion Device
Follow-up to insertion device for I22 requirement in #377. I22 has an IVU insertion device, which is different to the CPMUs used by I03 and I04. Further consultation is needed to see if the PVs are sufficiently similar that the current [Undulator](https://github.com/DiamondLightSource/dodal/blob/c79e88295a066eb1630592b3fa031d6f033bafa8/src/dodal/devices/undulator.py#L16) device can be generalised (and ported to ophyd-async) or if it makes sense to create a new one.

@Tom-Willemsen I see you've done some work on I23, which has the same type as I22, did you ever look into making an ophyd device for the ID.

## Acceptance Criteria
* ophyd-async IVU device available
* Insertion device config included in i22 beamline module
* Fake insertion device config included in p38 beamline module

callumforrester:
Add soft signals to Undulator that provide static information needed for the NXInsertionDevice class in the NeXus standard: poles and length.

Fixes #389

1. Run unit tests
2. Run system tests against i22, i03, i04 and i04-1

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/undulator.py
+++ b/src/dodal/devices/undulator.py
@@ -1,6 +1,6 @@
 from enum import Enum
 
-from ophyd_async.core import StandardReadable
+from ophyd_async.core import ConfigSignal, StandardReadable, soft_signal_r_and_backend
 from ophyd_async.epics.motion import Motor
 from ophyd_async.epics.signal import epics_signal_r
 
@@ -23,12 +23,28 @@ class Undulator(StandardReadable):
     def __init__(
         self,
         prefix: str,
+        poles: int,
+        length: float,
         name: str = "",
     ) -> None:
+        """Constructor
+
+        Args:
+            prefix: PV prefix
+            poles (int): Number of magnetic poles built into the undulator
+            length (float): Length of the undulator in meters
+            name (str, optional): Name for device. Defaults to "".
+        """
+
         with self.add_children_as_readables():
             self.gap_motor = Motor(prefix + "BLGAPMTR")
             self.current_gap = epics_signal_r(float, prefix + "CURRGAPD")
             self.gap_access = epics_signal_r(UndulatorGapAccess, prefix + "IDBLENA")
-        self.gap_discrepancy_tolerance_mm: float = UNDULATOR_DISCREPANCY_THRESHOLD_MM
+
+        with self.add_children_as_readables(ConfigSignal):
+            self.poles, _ = soft_signal_r_and_backend(int, poles)
+            self.length, _ = soft_signal_r_and_backend(float, length)
+
+        self.gap_discrepancy_tolerance_mm = UNDULATOR_DISCREPANCY_THRESHOLD_MM
 
         super().__init__(name)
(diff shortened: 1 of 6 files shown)


src/dodal/devices/undulator.py
```diff
@@ -23,12 +23,28 @@ class Undulator(StandardReadable):
     def __init__(
         self,
         prefix: str,
+        poles: int,
+        length: float,
```
callumforrester, line 27: @DominicOram Interested in your thoughts about requiring this information in the constructor before I go ahead and find it out for all beamlines. It's a bit forceful but does guarantee that all `Undulator` instances carry enough information for making an `NXInsertionDevice`

src/dodal/devices/undulator.py
```diff
@@ -23,12 +23,28 @@ class Undulator(StandardReadable):
...
             self.gap_motor = Motor(prefix + "BLGAPMTR")
             self.current_gap = epics_signal_r(float, prefix + "CURRGAPD")
             self.gap_access = epics_signal_r(UndulatorGapAccess, prefix + "IDBLENA")
-        self.gap_discrepancy_tolerance_mm: float = UNDULATOR_DISCREPANCY_THRESHOLD_MM
+
+        with self.add_children_as_readables(ConfigSignal):
+            self.poles, _ = soft_signal_r_and_backend(int, poles)
+            self.length, _ = soft_signal_r_and_backend(float, length)
+
+        self.gap_discrepancy_tolerance_mm = UNDULATOR_DISCREPANCY_THRESHOLD_MM
```
callumforrester, line 48: @DominicOram and @rtuck99 could this also be a soft signal, for consistency?

src/dodal/devices/undulator.py
DominicOram: I'm happy for this to be required information but it's not required for a NXInsertionDevice to be written, both those fields are optional in https://manual.nexusformat.org/classes/base_classes/NXinsertion_device.html#nxinsertion-device

src/dodal/devices/undulator.py
DominicOram: I'm happy for it to be a signal, make sure to change `undulator_dcm` to read it as a signal though

src/dodal/devices/undulator.py
callumforrester: That's a good point! It would be ideal if it was optional then, and the value only made it into the `read` output if a user supplied it. Not sure how ophyd-async deals with optional signals/if it should. @coretl might have some comments.

callumforrester: @coretl says that similar things in panda are optional signals (i.e. the attribute is just `None` if no value supplied)

src/dodal/devices/undulator.py
```diff
@@ -24,11 +24,36 @@ def __init__(
...
+            )
+            if poles is not None:
+                self.poles, _ = soft_signal_r_and_backend(int, poles)
+            else:
+                self.poles = None
+
+            if length is not None:
+                self.length, _ = soft_signal_r_and_backend(float, length)
+            else:
+                self.length = None
```
callumforrester, line 57: For some reason, the initial values are not propagating to the signal backends, TODO: investigate

callumforrester: Resolved by #525

DiamondJoseph (changes requested):
Need a sim version for p38 too

DiamondJoseph (approved):

PR #516, opened 2024-05-09
Change logging level on PandASubdirectoryProvider

olliesilvester:
In a previous PR I suggested adding this logging message, but later realised it would appear every time Hyperion started, so debug is better

1. Confirm changes and tests sensible
2. Run unit tests

### Checks for reviewer


src/dodal/common/udc_directory_provider.py
```diff
@@ -7,11 +8,14 @@
...
 class PandASubdirectoryProvider(UpdatingDirectoryProvider):
+    """Directory provider for the HDFPanda. Points to, and optionally creates, a panda subdirectory relative to the
+    directory path provided"""
+
     resource_dir = Path("panda")
 
     def __init__(self, directory: Path | None = None):
         if directory is None:
-            LOGGER.warn(
+            LOGGER.debug(
```
olliesilvester, line 18: I changed this to a debug since i03 initialises with directory=None, so those messages would get very annoying

src/dodal/common/udc_directory_provider.py
```diff
@@ -20,7 +24,13 @@ def __init__(self, directory: Path | None = None):
             else None
         )
 
-    def update(self, directory: Path):
+    def update(self, directory: Path, create_directory=False):
+        if create_directory:
```
DiamondJoseph, line 28: As is, you're allowed to (but always not) pass in a Path on construction, but the path is only created on update. If passing a Path into the initialiser isn't desired, remove it, if it is desired sometimes, make sure that the directory may be created during the init

src/dodal/common/udc_directory_provider.py
```diff
@@ -20,7 +24,13 @@ def __init__(self, directory: Path | None = None):
...
+            panda_directory = f"{directory}/{self.resource_dir}"
+            # At some point in the future, ophyd_async will have the feature to create the requested directory
+            if not os.path.isdir(panda_directory):
```
DiamondJoseph, line 31: This is in the Hyperion/blueapi container: which user does Hyperion run as? blueapi runs as a kubernetes user without file system permissions as far as possible

DiamondJoseph (changes requested):

src/dodal/common/udc_directory_provider.py
olliesilvester: Ah I hadn't considered that, Hyperion is running on a control machine so should have permissions. it might be better to just stick this part within a Hyperion plan

DiamondJoseph: Running as gda2?

Either way, if it's in the plan or the DirProvider it'll need to be running as a user that can write to the visit directory

olliesilvester: Yeah, as gda2. We can either check, in dodal, if the user has permission to create a directory if `create_directory=True` before trying to create it, or have the logic in Hyperion where the user will always be able to. I think Hyperion is better for now, can always change later

olliesilvester: Moved the logic to Hyperion so this PR is pretty tiny now!

Issue #502, opened 2024-05-03
Automatically keep I22 and P38 in sync


callumforrester:
As a developer I would like to keep P38, at least from an ophyd point of view, as a close representation of I22. Currently they are treated as different beamlines with some similar components.

## Possible Solution

Could we somehow define different versions of particular devices?

## Acceptance Criteria
* TBD

DiamondJoseph:
A possible solution is to use the same beamline module for both, with careful @skip_device usage and care for visit directories- p38 is only ever going to be a commissioning visit(?)

callumforrester:
I would suggest more than just `d11` and `saxs` in the same beamline module. I would suggest that we day `d11` _is_ `saxs`, or at least a testing representation of it. So the ophyd devices of P38 are a 1:1 representation of I22, even if we have to mock some devices.

stan-dot:
are the PV names the same? What is stopping us from using the file as-is?

should we have BEAMLINE_NAME derived from venv?

DiamondJoseph:
`BL = get_beamline_name("p38")`

Checks Environment Variables to see if BEAMLINE is exported (it will be on a beamline control machine).

Comparing `p38.py` and `i22.py`:

(best guess):
d11 ~ saxs: "-DI-DCAM-03:" vs. "-EA-PILAT-01:",
d12 ~ waxs: "-DI-DCAM-04:" vs. "-EA-PILAT-03:",
d3 ~ oav: "-DI-DCAM-01:" vs. "-DI-OAV-01:"

i0 ~ i0: "-EA-XBPM-01:" vs. "-EA-XBPM-02:"
linkam ~ linkam: "-EA-LINKM-02:" vs. "-EA-TEMPC-05",

panda1/2/3: match
panda4, it: no equivalent device on p38
all others: sim devices on p38

d11, d12 shouldn't be exposed as -EA-PILAT-01, as they aren't Pilatuses (? should we allow this? Hopefully all plan stubs are written for StandardDetectors?)
it should be added as a sim device or else repeating the values from i0 (?)

Definitely the Linkam and TetrAMM could be exposed on the same prefix

stan-dot:
then we have the manual option of just applying a ternary expression for the PVs /  or an if block 

or we could make some decorator or such.

we'd also need to make it crash for panda4 I guess if there is an attempt to use it in a plan???

callumforrester:
[...]

PR #519, opened 2024-05-10
Create a CLI for connecting to a beamline

### LINKED ISSUE #506: Connection test utility for each beamline
As a developer I would like to easily run a utility that checks that a beamline module connects to the PVs it is supposed to connect to.

## Possible Solution

We already have the system tests, but they require significant boilerplate per beamline. We could have a CLI utility e.g.

```shell
dodal connect <beamline>
```

We want devices that do not connect to be reported as failures unless we specifically mark them as ignorable. See discussion in #261

`dodal connect all` may also be required.  

## Acceptance Criteria
* Easy utility for connecting to beamline
* System tests deleted

callumforrester:
Fixes #506 

1. Run unit tests
2. Run CLI system tests against all beamlines

### Checks for reviewer


src/dodal/beamlines/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+    "i03",
+    "i04",
+    "i04_1",
+    "i20_1",
+    "i22",
+    "i23",
+    "i24",
+    "p38",
+    "p45",
+}
```
callumforrester, line 11: This list of all beamline modules was previously maintained in the tests directory. It has been moved here so the CLI can reuse it. 

I'm not sure how much I like maintaining a list of all beamline modules as it seems prone to error. The advantage is that typing `dodal connect --help` will include a list of beamlines you can connect to.

The alternatives are some fancy auto-detection of beamline modules, perhaps by finding all modules in `dodal.beamlines` that match a pattern or simply parsing any string the user enters and printing an error message if it's not a beamline. 

I dislike the first because baking in beamline naming conventions has tended to lead to trouble in the past.

src/dodal/cli.py
```diff
@@ -0,0 +1,55 @@
...
+@click.group(invoke_without_command=True)
+@click.version_option(version=__version__, message="%(version)s")
+@click.pass_context
+def main(ctx: click.Context) -> None:
+    if ctx.invoked_subcommand is None:
+        print("Please invoke subcommand!")
+
+
+@main.command(name="connect")
+@click.argument("beamline", type=click.Choice(list(ALL_BEAMLINES)), required=True)
```
callumforrester, line 24: Passing the beamline module in as an argument may be confusing in cases where it conflicts with the `${BEAMLINE}` environment variable. However, in some cases the two are intentionally different. For example if running against s03 you load the `i03` beamline module but with `BEAMLINE=s03`, so `${BEAMLINE}` and the beamline module are two separate things with the same name and the same value in lots of cases, which seems a bit confusing. Not sure if it's worth tackling here or raising another issue.

src/dodal/beamlines/__init__.py
DominicOram: We could move `_device_helpers.py` etc. into some other location and just have a convention that only beamline modules live in `beamlines`. Then we can just auto-detect based on all the module names in that folder?

src/dodal/cli.py
DominicOram: Could we default to using `${BEAMLINE}` (which will be the case for 90%) of our users then bomb out if it's not set?

DominicOram (changes requested):
Thanks. Code looks good but it doesn't behave exactly how I would expect. If I do `dodal connect i03` it attempts to connect to `s03` rather than `i03`. I'm also not convinced that it tries to connect to subsequent devices after the first failure. Additionally:
* Could you add a line to the README that let's people know this exists?
* I guess we can now remove all the `test_XXX_system.py`s?

src/dodal/cli.py
DominicOram: Yh, you're right, let's just require it

callumforrester:
@DominicOram makes sense, the `dodal connect i03` behaviour mirrors the `i03` system tests, which try to connect to `s03` by default unless `${BEAMLINE}` is `i03`. I agree its confusing, which is why I was hoping for a better solution for https://github.com/DiamondLightSource/dodal/pull/519#discussion_r1598342114

src/dodal/cli.py
callumforrester: I think it's still confusing given https://github.com/DiamondLightSource/dodal/pull/519#issuecomment-2107575051

Some other options are:
- Completely remove the `${BEAMLINE}` variable from dodal, base everything on the beamline module
- Have s03 as some kind of profile for i03 rather than base on the environment variable so you could do something like `dodal connect --profile tickit i03`

callumforrester:
@DominicOram re: connecting to subsequent devices after the first failure: This is an issue in general with `make_all_devices`, hopefully one that  @ZohebShaikh will solve as part of https://github.com/DiamondLightSource/blueapi/issues/444

src/dodal/cli.py
callumforrester: Agreed, we then just move to solution 2 eventually as part of #502

src/dodal/beamlines/__init__.py
callumforrester: I felt that was worth its own PR: https://github.com/DiamondLightSource/dodal/pull/544

src/dodal/cli.py
callumforrester: @DominicOram I think I've probably misunderstood this because what I've done is the same thing again. I made it so you can type:

```shell
dodal connect i04-1
```

which will try to connect to `s04-1` or

```shell
export BEAMLINE=i04-1
dodal connect
```

which try to connect to i04. That feels about as confusing and probably not what you had in mind.

callumforrester: Have just thought of another requirement: the training rigs. They are all identical so should have one dodal module (I believe @paula-mg has prototyped one) and rely on `${BEAMLINE}` or similar to differentiate.

callumforrester:
CI failure should be fixed by https://github.com/DiamondLightSource/dodal/pull/544

README.rst
```diff
@@ -13,6 +13,26 @@ Documentation  https://DiamondLightSource.github.io/dodal
...
+
+You can test your connection to a beamline if it's PVs are visible to your machine with:
+
+.. code:: shell
+
+    # On any workstation:
+    dodal connect <BEAMLINE>
+
+    # On a beamline workstation, this should suffice:
+    dodal connect ${BEAMLINE}
```
DominicOram, line 27: Should: Sorry to scope creep a bit can we add this to the PR template while it's not in CI. As a reviewer it's basically the first thing I run and a lot of the time it fails.

src/dodal/cli.py
```diff
@@ -0,0 +1,55 @@
...
+    "attempt any I/O. Useful as a a dry-run.",
+    default=False,
+)
+def connect(beamline: str, all: bool, sim_backend: bool) -> None:
+    """Initialises a beamline module, connects to all devices, reports
+    any connection issues."""
+
+    module_name = module_name_for_beamline(beamline)
+
+    from bluesky import RunEngine
```
DominicOram, line 44: Nit: Does pylance not give you an error here about `RunEngine` not being exported from `bluesky`? I have to use `from bluesky.run_engine import RunEngine` to fix

src/dodal/cli.py
DominicOram: Sorry, I wasn't clear my suggestion is literally to do `os.environ["BEAMLINE"] = beamline` inside `connect`

DominicOram (changes requested):
Great, thank you. Some comments in code. I think we need to fix the  simulated beamlines being used so doing the `os.environ["BEAMLINE"] = beamline` is a Must unless you have a cleaner solution.

As I said before, can you remove the `test_iXX_system.py` files please?

src/dodal/cli.py
callumforrester: Interesting, yes it does, even though `RunEngine` is exported from the top level: https://github.com/bluesky/bluesky/blob/27ca36cc60b718ad3f2f2bb343b1d086aec77708/src/bluesky/__init__.py#L4

I suspect pylance is only looking for `__all__`, will change it to reduce the nasty red lines.

callumforrester: Also unsure why I'm doing the import inside the function, will change that too

callumforrester:
> As I said before, can you remove the test_iXX_system.py files please?

Whoops, sorry!

src/dodal/cli.py
callumforrester: Sounds good, also works for the training rig case

tests/unit_tests/test_cli.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_cli_sets_beamline_environment_variable(runner: CliRunner):
+    with patch.dict(os.environ, clear=True):
+        _mock_connect(EXAMPLE_BEAMLINE, runner=runner)
+        assert os.environ["BEAMLINE"] == EXAMPLE_BEAMLINE
+
+
+# We need to mock the environment because the CLI edits it and this could break other
+# tests
```
callumforrester, line 37: Setting `os.environ["BEAMLINE"]` caused unintended side effects, will make an issue to find a better way

callumforrester: Done in #569

src/dodal/cli.py
```diff
@@ -0,0 +1,58 @@
...
+    """Initialises a beamline module, connects to all devices, reports
+    any connection issues."""
+
+    os.environ["BEAMLINE"] = beamline
+
+    module_name = module_name_for_beamline(beamline)
+
+    RE = RunEngine()  # noqa: F841
+
+    devices = make_all_devices(
```
DominicOram, line 51: Nit: Would be nice if we could print something just before this saying that it's started and will take some time. The i03 ones take ~30s so not obvious it hasn't just hung.

DominicOram (approved):
Great, thank you! Minor nit-picks in code but otherwise good. I Tested a few of the beamlines. The `_1` beamlines are broken for a different reason I think (https://github.com/DiamondLightSource/dodal/issues/570) but others passed.

src/dodal/beamlines/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+def all_beamline_modules() -> Iterable[str]:
+    """
+    Get the names of all importable modules in beamlines
+
+    Returns:
+        Iterable[str]: Generator of beamline module names
+    """
+
+    # This is done bty inspecting file names rather than modules to avoid
```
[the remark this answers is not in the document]
callumforrester: I already use it, this is a case of pebcak...

PR #521, opened 2024-05-10
Fix i22 slits

### LINKED ISSUE #520: Fix i22 slits
Six sets of slits were added in #431.

There are a couple of issues:

* They were added with `fake_with_ophyd_sim = True` (which is correct for p38 but not i22)
* They were added with the assumption that all slits are the same however this doesn't appear to be the case - S4 is different to the others and doesn't have {X,Y} centre and size, instead 6 different motors

First part is to set `fake_with_ophyd_sim = False`.

Second part is to confirm if the hardware is fundamentally different or if the IOC can be changed to present itself as the other slits. If it is different then we probably need an additional slits implementation - in this case a new issue can be made.

## Acceptance Criteria
* All slit devices in the i22 connect or are skipped if a suitable device is not implemented yet.

joeshannon:
* These were added as fake by mistake
* Skip S4 for now as it has a different Epics interface

Fixes #520.

1. Check all i22 devices can be instantiated and connect with no error (easiest to do with blueapi)

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i22.py
+++ b/src/dodal/beamlines/i22.py
@@ -125,7 +125,7 @@ def hfm(
 
 def slits_1(
     wait_for_connection: bool = True,
-    fake_with_ophyd_sim: bool = True,
+    fake_with_ophyd_sim: bool = False,
 ) -> Slits:
     return numbered_slits(
         1,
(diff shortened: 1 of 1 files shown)


callumforrester (approved):
LGTM

PR #524, opened 2024-05-13
473 create a device for xpress3 areadetector

### LINKED ISSUE #473: Create a device for XPRESS3 AreaDetector
For SWIFT / i20-1 beamline for the initial tests with bluesky and blueapi.

## Acceptance Criteria
* Have the device runnable with plans from its directory

## Steps
-  [ ] check what are the CA channels used
-  [ ] compare Java implementation - there are many of them, also for fluoresencence. many for many different beamlines
-  [ ] compare Pilatus https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/epics/areadetector/pilatus.py
-  [ ] check main methods from ophyd-async

Relm-Arrowny:
Fixes #473 
- Replaced xspress3_mini(ophyd) with xspress3(ophyd_async) and added support for more than one channel. Extra channels are stored as a list. `xspress3mini.channels[<channel number>]`

- Removed function arm and unify it with stage.  Should be able to replace 
 `yield from bps.abs_set(xspress3mini.do_arm, 1, wait=True)` 
 with
`yield from bps.stage(xspress3mini, wait=True)`
in plans.

- Also removed signal with epics ip ending Erase as it is not connected or used for anything. 

1. Run unit test
2. Confirm the function stage is logically sound to replace express_mini.arm

### Checks for reviewer


src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_channel_oa.py
```diff
@@ -0,0 +1,28 @@
...
+
+class TimeSeriesValues(str, Enum):
+    START_VALUE = "Acquire"
+    STOP_VALUE = "Done"
+    UPDATE_VALUE = ""
+
+
+class Xspress3MiniChannel(Device):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        # Define some signals
```
DominicOram, line 15: Could: I'm not sure this comment provides much value. What would be better is a docstring on the device that gives context about what a channel is.

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+        self.dt_corrected_latest_mca = epics_signal_r(float, prefix + "ARR1:ArrayData")
+        self.set_num_images = epics_signal_rw(float, prefix + "NumImages")
+        self.detector_busy_states = [
+            DetectorState.ACQUIRE,
+            DetectorState.CORRECT,
+            DetectorState.ABORTING,
+        ]
+        super().__init__(name=name)
+
+    async def stage(self):
```
DominicOram, line 98: Should: I think you want to decorate this with `@AsyncStatus.wrap` to make sure it plays nicely with bluesky

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+        await self.arm()
+
+    async def arm(self) -> AsyncStatus:
```
DominicOram, line 101: Nit: You probably don't need separate `arm` and `stage` methods

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+        LOGGER.info("Arming Xspress3Mini detector...")
+        self.trigger_mode_mini.set(TriggerMode.BURST)
+        await wait_for_value(
+            self.detector_state, lambda v: v in self.detector_busy_states, timeout=1
+        )
+        self.erase.set(EraseState.ERASE)
```
DominicOram, line 107: Should: I think you want to `await` this set to finish

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+        await set_and_wait_for_value(self.acquire, AcquireState.ACQUIRE, timeout=1)
+        return AsyncStatus(wait_for_value(self.acquire, AcquireState.DONE, timeout=1))
```
DominicOram, line 109: Could: Can you pull all the timeouts into a constant?

tests/devices/unit_tests/test_xspress3mini_OA.py
```diff
@@ -0,0 +1,45 @@
...
+    AcquireState,
+    DetectorState,
+    Xspress3Mini,
+)
+
+
+def get_bad_status() -> Status:
+    status = Status("get_bad_status")
+    status.set_exception(Exception)
+    return status
```
DominicOram, line 19: Should: I don't think this function is used?

tests/devices/unit_tests/test_xspress3mini_OA.py
```diff
@@ -0,0 +1,45 @@
...
+    set_sim_value(fake_xspress3mini.detector_state, DetectorState.ACQUIRE)
+    status = await fake_xspress3mini.arm()
+    assert status.done is False
+    set_sim_value(fake_xspress3mini.acquire, AcquireState.DONE)
+    await asyncio.sleep(0.1)
+    assert status.done is True
+
+    set_sim_value(fake_xspress3mini.detector_state, DetectorState.IDLE)
+    with pytest.raises(TimeoutError):
+        await fake_xspress3mini.stage()
```
DominicOram, line 40: Should: It seems like the bits above are testing that the detector arms when all is good and this lines that the detector doesn't stage when all is not good, the stage timesout, could they be split into two different tests?

tests/devices/unit_tests/test_xspress3mini_OA.py
```diff
@@ -0,0 +1,45 @@
...
+async def fake_xspress3mini(prefix: str = "huh") -> Xspress3Mini:
+    async with DeviceCollector(sim=True):
+        fake_xspress3mini = Xspress3Mini(prefix, "Xspress3Mini")
+    return fake_xspress3mini
+
+
+async def test_arm_success_on_busy_state_OA(fake_xspress3mini: Xspress3Mini):
+    # set_sim_value(fake_xspress3mini.detector_state,DetectorState.IDLE)
+    set_sim_value(fake_xspress3mini.detector_state, DetectorState.ACQUIRE)
+    status = await fake_xspress3mini.arm()
```
DominicOram, line 32: Could: `stage` is the external facing API so better to test against that than `arm`

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+    SAVING = "Saving"
+    ERROR = "Error"
+    INTILTIALIZING = "Initializing"
+    DISCONNECTED = "Disconnected"
+    ABORTED = "Aborted"
+
+
+class Xspress3Mini(Device):
+    def __init__(self, prefix: str, name: str = "", num: int = 1) -> None:
+        # Define some signals
```
DominicOram, line 70: Should: As above, I don't think this comment is too helpful, a docstring about what the xspress3 mini is would be more helpful

DominicOram (changes requested):
This is a good start, some comments in code. Additionally can you:
* Remove the old `xspress3_mini`
* Remove `ophyd_async` and `oa` from the names, I think if there's only one device it's not needed
* Change `i03.py` to use this version and add it to `i20_1.py`
* Run the system tests for `i03`/`i20_1` and confirm the PVs all connect still

How would you like to use this device? All we do in `hyperion` is `arm` the device and directly read a couple of the signals into the plan. Presumably you want something more complicated with getting the data to disk etc? In which case we should think about what you want `read()` to do. We do have a requirement at some point to also read data out in a better way, supposedly the device can write HDF files? We'd be interested in doing that and embedding the data in our nexus files.

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_channel_oa.py
```diff
@@ -0,0 +1,28 @@
...
+        self.update_arrays = epics_signal_rw(float, prefix + "SCAS:TS:TSAcquire")
+
+        self.roi_high_limit = epics_signal_rw(float, prefix + "SCA5_HLM")
```
stan-dot, line 18: What is roi?

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_channel_oa.py
```diff
@@ -0,0 +1,28 @@
...
+        self.roi_llm = epics_signal_rw(float, prefix + "SCA5_LLM")
+
+        self.time = epics_signal_r(float, prefix + "SCA0:Value_RBV")
+        self.reset_ticks = epics_signal_r(float, prefix + "SCA1:Value_RBV")
+        self.reset_count = epics_signal_r(float, prefix + "SCA2:Value_RBV")
+        self.all_event = epics_signal_r(float, prefix + "SCA3:Value_RBV")
```
stan-dot, line 24: What does this mean?

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+    LVDS_both = "LVDS_Both"
+
+
+class UpdateRBV(str, Enum):
+    DISABLED = "Disabled"
+    ENABLED = "Enabled"
+
+
+class EraseState(str, Enum):
+    DONE = 0
```
stan-dot, line 45: Why str when it's an integer?

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+    ERASE = 1
+
+
+class AcquireState(str, Enum):
+    DONE = 0
+    ACQUIRE = 1
+
+
+class DetectorState(str, Enum):
+    ACQUIRE = "Acquire"
```
stan-dot, line 55: Why capitalize the first letter?

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+        # Define some signals
+        self.channels = DeviceVector(
+            {i: Xspress3MiniChannel(f"{prefix}:C{i}_") for i in range(1, num + 1)}
+        )
+        self.erase = epics_signal_rw(EraseState, prefix + "ERASE")
+        self.get_max_num_channels = epics_signal_r(
+            float, prefix + "MAX_NUM_CHANNELS_RBV"
+        )
+        self.acquire = epics_signal_rw_rbv(AcquireState, prefix + "Acquire")
+        self.get_roi_calc_mini = epics_signal_rw(float, prefix + "MCA1:Enable_RBV")
```
stan-dot, line 79: Roi again, mysterious

tests/devices/unit_tests/test_xspress3mini_OA.py
```diff
@@ -0,0 +1,45 @@
+import asyncio
+
+import pytest
+from bluesky import plan_stubs as bps
+from bluesky.run_engine import RunEngine
+from ophyd.status import Status
```
stan-dot, line 6: Why import from regular ophyd?

tests/devices/unit_tests/test_xspress3mini_OA.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_bad_status() -> Status:
+    status = Status("get_bad_status")
+    status.set_exception(Exception)
+    return status
+
+
+@pytest.fixture
+async def fake_xspress3mini(prefix: str = "huh") -> Xspress3Mini:
```
stan-dot, line 23: Fake or mock?
@callumforrester

stan-dot (changes requested):

tests/devices/unit_tests/test_xspress3mini_OA.py
callumforrester: With the latest `ophyd-asyn`, I think we should try to use `mock` universally

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
callumforrester: Should follow the format of `DONE = "DONE"`, ditto for the others

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class Xspress3Mini(Device):
+    def __init__(self, prefix: str, name: str = "", num: int = 1) -> None:
+        # Define some signals
+        self.channels = DeviceVector(
+            {i: Xspress3MiniChannel(f"{prefix}:C{i}_") for i in range(1, num + 1)}
+        )
+        self.erase = epics_signal_rw(EraseState, prefix + "ERASE")
+        self.get_max_num_channels = epics_signal_r(
```
callumforrester, line 75: Should: I don't think you should have `get` in the attribute names
```suggestion
        self.max_num_channels = epics_signal_r(
```

tests/devices/unit_tests/test_xspress3mini_OA.py
callumforrester: Can use `AsyncStatus` from ophyd-async

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_channel_oa.py
Relm-Arrowny: region of interest, you can think of xspress3 like a smart band pass filter so you have to tell it the lower and upper limit for it to work.

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_channel_oa.py
Relm-Arrowny: These are the xpress status so, time it been up, time it been reset etc, not sure we need to collect them for i20-1 but I kept them for historical and being save, I dont want to break something if someone is using them.

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
Relm-Arrowny: it is to match the c enum so I can either match "Acquire "or use the location in this case 0 to set them. Which is also the reason I have not change done = 0 above to string yet... I need to find out what it is.

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
Relm-Arrowny: The situation is actually worse than I though, what is the best way to deal with undefined Enum, atm I kind of side step it by giving it a enum but set it with int.

```
NotConnected: 
express3_i20: NotConnected:
    erase: TypeError: BL20I-EA-DET-06:ERASE has choices (' ', ' '), which do not match <enum 'EraseState'>, which has ('Done', 'Erasing')
express3_i3: NotConnected:
    erase: TypeError: BL03I-EA-XSP3-01:ERASE has choices (' ', ' '), which do not match <enum 'EraseState'>, which has ('Done', 'Erasing')
```
Also a related issue, what do we do if the rbv return a different enum, would it be reasonable to make 2 signals?
```
class AcquireRBVState(str, Enum):
    DONE = "Done"
    ACQUIRE = "Acquiring"

class AcquireState(str, Enum):
    DONE = "Done"
    ACQUIRE = "Acquire"

  self.acquire = epics_signal_rw(AcquireState, prefix + "Acquire")
  self.acquire_rbv = epics_signal_r(AcquireRBVState, prefix + "Acquire_RBV")
```

Relm-Arrowny:
>     * Remove the old `xspress3_mini`
I am a bit scare to do this as this will definitely break the plans that are being use on i03 so I added as single  channel xspress3 in i03.py so may be we can remove both once we know it works?
>     * Change `i03.py` to use this version and add it to `i20_1.py`
I have added xspress3 device for both.  
>     * Run the system tests for `i03`/`i20_1` and confirm the PVs all connect still
As i20-1 xpress3 is not functional atm, I have ran connection test on at my desk for i03 and i20 xpress3 it all connect but no put/set is tested. Question, Is there a standard way/automatic way of doing system tests?  

> How would you like to use this device? All we do in `hyperion` is `arm` the device and directly read a couple of the signals into the plan. Presumably you want something more complicated with getting the data to disk etc? In which case we should think about what you want `read()` to do. We do have a requirement at some point to also read data out in a better way, supposedly the device can write HDF files? We'd be interested in doing that and embedding the data in our nexus files.

@stan-dot  will know more than me on how they want to use xpress3 on i20-1.
However, as I am going through the code, it seem Hyperion and i20 use xpress3 in a very similar way but i20-1 appears to be much more complex and make extensive use of the MCA ROI/channels and arrays. So I am wondering would it make sense to impermanent it like area detector, where we will  have xpress diver, controller and detector.

tests/devices/unit_tests/test_xspress3mini_OA.py
callumforrester: @Relm-Arrowny @stan-dot Now that #525 is merged you can just use the mock backend, and you have to actually or the PR will break

src/dodal/devices/xspress3/xspress3.py
```diff
@@ -0,0 +1,146 @@
...
+            {i: Xspress3Channel(f"{prefix}C{i}_") for i in range(1, num_channels + 1)}
+        )
+        """MCA on/off switch readback"""
+        self.get_roi_calc_status = DeviceVector(
+            {
+                i: epics_signal_rw(float, f"{prefix}MCA{i}:Enable_RBV")
+                for i in range(1, num_channels + 1)
+            }
+        )
+        """start and size of the MCA array"""
```
stan-dot, line 94: what is MCA array specifically?

stan-dot (changes requested):
just expand this one abbreviation pls

src/dodal/devices/xspress3/xspress3.py
Relm-Arrowny: multi-channel analyser, unlike the single channel analyser(SCA) it return an array of SCA counts over a user defined time/counter/size.   So in term of xpress3, rather than arming/staging it every time you want 1 data point, you arm/stage it for x number of point and it return an array instead of a scaler.

stan-dot:
@Relm-Arrowny  honestly I don't know much more details on i20-1 uses that xpress3 detector.

DominicOram:
> I am a bit scare to do this as this will definitely break the plans that are being use on i03 so I added as single channel xspress3 in i03.py so may be we can remove both once we know it works?

Don't be, swap them and I'll take responsibility for any issues on i03.

> As i20-1 xpress3 is not functional atm, I have ran connection test on at my desk for i03 and i20 xpress3 it all connect but no put/set is tested. Question, Is there a standard way/automatic way of doing system tests?

Yes, those tests just ensure connection but they actually get you reasonably far. No, they're not automated yet, there are some issues in the backlog to move towards automating them but it's not trivial because they require a connection to the real PVs so need to be run on site. In general the idea for proper system tests is to create a device simulator in `tickit` that the real IOC talks to, practically this is quite hard though.

>So I am wondering would it make sense to impermanent it like area detector, where we will have xpress diver, controller and detector.

To answer this question properly we probably need to discuss with a wider group. @iain-hall might have a good idea of how they currently do things? I would suggest organising a meeting with some controls/DAQ/analysis and scientist to discuss how it is used and how we would like it to be used.

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
DominicOram: Yes, you would have to make two signals, see similar issue on the shutter https://github.com/DiamondLightSource/dodal/pull/480#discussion_r1606783981

src/dodal/devices/xspress3/xspress3.py
```diff
@@ -0,0 +1,147 @@
...
+    LVDS_both = "LVDS Both"
+
+
+class UpdateRBV(str, Enum):
+    DISABLED = "Disabled"
+    ENABLED = "Enabled"
+
+
+'''
+class EraseState(str, Enum):
```
stan-dot, line 44: pls delete empty

stan-dot (approved):

src/dodal/devices/ophyd_async_xspress3_mini/xspress3_mini_oa.py
Relm-Arrowny:
```
NotConnected: 
express3_i20: NotConnected:
    erase: TypeError: BL20I-EA-DET-06:ERASE has choices (' ', ' '), which do not match <enum 'EraseState'>, which has ('Done', 'Erasing')
express3_i3: NotConnected:
    erase: TypeError: BL03I-EA-XSP3-01:ERASE has choices (' ', ' '), which do not match <enum 'EraseState'>, which has ('Done', 'Erasing')

```
Not sure how to handle this,

Below is the only way I can get it to connect and passes all the check/test but I am not sure it will behave correctly with the real device.

```
class EraseState(str, Enum):
    DONE = ""
    ERASE = ""


await self.erase.set(EraseState.ERASE)  
```

`(int, enum)`  should work but backend would not allow it to connect same for just int type.

Anyone has an idea what's the best way to deal with this?

Relm-Arrowny: Just an up date below will work, dont know if it is intentional.
```
class EraseState(str, Enum):
    DONE = ""
    ERASE = ""


await self.erase.set(<int>)  int can be the position of the c enum  
```

DominicOram: Tagging @stan-dot as this is general good info. So to look up the type of a PV I do the following:
* `python2 /dls_sw/work/useful-scripts/find-pv.py BL20I-EA-DET-06:ERASE`
* This showed me the PV was defined in `/dls_sw/work/R3.14.12.7/support/BL20I-BUILDER/iocs/BL20I-EA-IOC-05/db/BL20I-EA-IOC-05_expanded.db` as a `bo` record.
* EPICS record types are documented at https://epics.anl.gov/base/R3-15/7-docs/RecordReference.html
* If you look at https://epics.anl.gov/base/R3-15/7-docs/boRecord.html you can see that `bo` records take values of 0 or 1, but they can define `ZNAM` and `ONAM` which are user friendly names for the 0/1 states
* So if we do:
```
$ caget BL03I-EA-XSP3-01:ERASE.ZNAM BL03I-EA-XSP3-01:ERASE.ONAM
BL03I-EA-XSP3-01:ERASE.ZNAM    
BL03I-EA-XSP3-01:ERASE.ONAM    
```
We can see that both are blank compared to a PV of the same type I found elsewhere:
```
$ caget BL03I-MO-SGON-01:CS1:DeferMoves.ZNAM BL03I-MO-SGON-01:CS1:DeferMoves.ONAM
BL03I-MO-SGON-01:CS1:DeferMoves.ZNAM Defer Off
BL03I-MO-SGON-01:CS1:DeferMoves.ONAM Defer On
```
* Given they're both `blank` this is why your string enum complains if they're anything else. Obviously this won't actually work in production as if you set `blank` on the PV how does it know which one you actually need?

So, we have some choices:
1. Talk to controls and ask them to provide named states for the PV, or change the PV to a type that is more appropriate
2. Use an `int` - This will work but is a bit of a shame as it's not super clear that writing 1 to the erase is what you want to do
3. Use an `IntEnum`  e.g.
```python
class EraseEnum(int, Enum):
    ERASE = 1
    DONE = 0
```
But are you saying that didn't work? In which case we should probably open a discussion up about if this is a bug in `ophyd_async`

I prefer option 1 but it will be the most work. I can live with option 2 for now in the interests of just getting things done, maybe we can then put in a ticket with controls to improve it at a later date

Relm-Arrowny: Thanks, knowing how to check pv type would have save me so much time :).

Both 2 and 3 will fail to connect when opyhd_asynco beckend type check the signal as it expect str type enum. As it is the only way for it to work is to trick opyhd_asynco  with two blanks for connection and set it with int later, which is not the best idea.
Who would be a good person to talk to about epics xspress3?  

> * Given they're both `blank` this is why your string enum complains if they're anything else. Obviously this won't actually work in production as if you set `blank` on the PV how does it know which one you actually need?

Something I notice playing around with the epic pv I have access to, if I set the Names  to "blank" they default into int, and option 2 works.  But it seem not all blanks are equal.

DominicOram: Ok, I think that is a bug in `ophyd_async` too then as we can't assume that all PVs will have valid string representations. Could you make an issue in `ophyd_async` about it?

> Who would be a good person to talk to about epics xspress3?
I will find out for you

Relm-Arrowny: Just for record keeping, issue raised in [ophyd_async] (https://github.com/bluesky/ophyd-async/issues/353).

After discussion with control, the problem signal is being use in any form so it is removed from this device and it is no longer an issue for this pr.

tests/devices/unit_tests/test_xspress3.py
```diff
@@ -0,0 +1,45 @@
...
+    return mock_xspress3mini
+
+
+async def test_stage_success_on_busy_state(mock_xspress3mini: Xspress3):
+    set_mock_value(mock_xspress3mini.detector_state, DetectorState.ACQUIRE)
+    status = await mock_xspress3mini.stage()
+    assert status.done is False
+    set_mock_value(mock_xspress3mini.acquire, AcquireState.DONE)
+    await asyncio.sleep(0.1)
+    assert status.done is True
```
DominicOram, line 39: Must: This test isn't doing what you think it is. if you remove `set_mock_value(mock_xspress3mini.acquire, AcquireState.DONE)` it still passes fine. I think you need to write some tests that confirm:
* That stage is setting everything you think it is
* That if acquire RBV does not go to DONE it fails
* That `RE(bps.stage(...))` is happy

DominicOram (changes requested):
This is really good work but a couple of comments in code

src/dodal/beamlines/i20_1.py
```diff
@@ -23,3 +24,20 @@ def turbo_slit(
         wait=wait_for_connection,
         fake=fake_with_ophyd_sim,
     )
+
+
+def express3(
```
DominicOram, line 29: Should: Sorry, missed this before. Can you make sure this matches the spelling of the device `xspress3`

src/dodal/devices/xspress3/xspress3.py
```diff
@@ -0,0 +1,144 @@
...
+    INTILTIALIZING = "Initializing"
+    DISCONNECTED = "Disconnected"
+    ABORTED = "Aborted"
+
+
+class Xspress3(Device):
+    """Xpress/XpressMini is a region of interest (ROI) picker that sums the detector
+    output into a scaler with user-defined regions. It is often used as a signal
+     discriminator to provide better energy resolution in X-ray detection experiments.
+     This currently only provide staging functionality.
```
DominicOram, line 65: Nit: Something seems off with the indenting here?

src/dodal/devices/xspress3/xspress3.py
```diff
@@ -0,0 +1,144 @@
...
+            lambda v: v in self.detector_busy_states,
+            timeout=self.timeout,
+        )
+        await self.acquire.set(AcquireState.ACQUIRE)
+        await wait_for_value(
+            self.acquire_rbv, AcquireRBVState.ACQUIRE, timeout=self.timeout
+        )
+        await wait_for_value(
+            self.acquire_rbv, AcquireRBVState.DONE, timeout=self.timeout
+        )
```
DominicOram, line 144: Should: I don't think it makes sense to wait on both of these `stage` basically means get the device ready to take data so I expect that we want to wait for `ACQUIRE` (it is now taking data) rather than `DONE` (it has finished taking data).

tests/devices/unit_tests/test_xspress3.py
```diff
@@ -0,0 +1,78 @@
...
+    mock_xspress3mini.acquire.set = AsyncMock()
+    status: AsyncStatus = mock_xspress3mini.stage()
+
+    assert status.done is False
+    while not status.done:
+        await asyncio.sleep(0.1)
+    assert status.done is True
+
+    mock_xspress3mini.trigger_mode.set.assert_called_once_with(TriggerMode.BURST)
+    mock_xspress3mini.acquire.set.assert_called_once_with(AcquireState.ACQUIRE)
```
DominicOram, line 58: Must: This test should actually be failing. Your `stage` waits for the RBV to go to `DONE` but you have no callback in this test that sets it to `DONE`. Additionally:
* You can use the in-built `get_mock_put` rather than setting your own mocks
* You can just `await ... .stage` to call it:
```suggestion
async def test_stage_success_on_busy_state(mock_xspress3mini: Xspress3):
    set_mock_value(mock_xspress3mini.detector_state, DetectorState.ACQUIRE)
    callback_on_mock_put(
        mock_xspress3mini.acquire,
        lambda *_, **__: set_mock_value(
            mock_xspress3mini.acquire_rbv, AcquireRBVState.ACQUIRE
        ),
    )
    await mock_xspress3mini.stage()

    get_mock_put(mock_xspress3mini.trigger_mode).assert_called_once_with(
        TriggerMode.BURST, wait=ANY, timeout=ANY
    )
    get_mock_put(mock_xspress3mini.acquire).assert_called_once_with(
        AcquireState.ACQUIRE, wait=ANY, timeout=ANY
    )
```

DominicOram (changes requested):
Couple more comments

tests/devices/unit_tests/test_xspress3.py
callumforrester: `Stageable` is a protocol, meaning you don't actually `need` to inherit from it, you can also type annotate uses of it and then it will be checked via duck typing. If you inherit from it you're just making it clearer in the code: this must be `Stageable`.

Both bullet points are good, I would quite like all devices to have RE tests, it would have caught several issues we've seen on the beamline. I'd also happily support an issue in bluesky to separate out the `Stageable` protocol. Something like:

```python
class Stageable(Protocol):
    def stage(self): ...

class StageableUnstageable(Protocol):
    def stage(self): ...
    def unstage(self): ...
```

.vscode/settings.json
callumforrester, line 1: Should: Please raise separate PRs for vscode config changes. Ideally upstream to the [copier template](https://github.com/DiamondLightSource/python-copier-template).

src/dodal/devices/xspress3/xspress3.py
Relm-Arrowny: I think I got confused with the return await_value(self.acquire, 0). 
```python 
acquire = Component(EpicsSignalWithRBV, "Acquire")

def arm(self) -> Status:
        LOGGER.info("Arming Xspress3Mini detector...")
        self.trigger_mode_mini.put(TriggerMode.BURST.value)
        arm_status = await_value_in_list(self.detector_state, self.detector_busy_states)
        self.erase.put(EraseState.ERASE.value)
        arm_status &= self.acquire.set(AcquireState.ACQUIRE.value)
        arm_status.wait(self.ARM_STATUS_WAIT)
        return await_value(self.acquire, 0)
 ```

would this be more correct?
 ```python
    @AsyncStatus.wrap
        LOGGER.info("Arming Xspress3 detector...")
        await self.trigger_mode.set(TriggerMode.BURST)
        await wait_for_value(
            self.detector_state,
            lambda v: v in self.detector_busy_states,
            timeout=self.timeout,
        )
        await self.acquire.set(AcquireState.ACQUIRE)
        await wait_for_value(
            self.acquire_rbv, AcquireRBVState.ACQUIRE, timeout=self.timeout
        )
    @AsyncStatus.wrap
        LOGGER.info("unstaging Xspress3 detector...")
        await wait_for_value(
            self.acquire_rbv, AcquireRBVState.DONE, timeout=self.timeout
        )
 ```

DominicOram: I think the waiting for the done in `unstage` makes more logically sense to me as something we would expect on the device. So let's leave it at that and make sure it works as expected as part of https://github.com/DiamondLightSource/hyperion/issues/1433

DominicOram (changes requested):
Great, thank you! I think I'm happy with this now. I put in some time for us to go and test on the beamline next month. Could you just fix the unit tests then we can merge

DominicOram (approved):
Great, thanks!

Issue #497, opened 2024-05-02
Implement the ophyd-async sim to mock changes in dodal


abbiemery:
The changes implemented in the [sim to mock change](https://github.com/bluesky/ophyd-async/issues/94) need to be reflected in dodal.

Issue #482, opened 2024-04-29
 Use new `ophyd_async` soft signal helpers 


DominicOram:
Following https://github.com/bluesky/ophyd-async/pull/217 we should be able to replace some of our helper functions. 

## Acceptance Criteria
* `dodal` helper functions are replaced with the equivalent `ophyd_async` funcs

DominicOram:
I think this is blocked on some discussion of:
* https://github.com/bluesky/ophyd-async/issues/267
* https://github.com/bluesky/ophyd-async/issues/266

PR #525, opened 2024-05-13
Compatibility with ophyd-async v0.3a4

callumforrester:
Update all compatibility issues with [ophyd-async v0.3a4](https://github.com/bluesky/ophyd-async/releases/tag/v0.3a4)

- Handle renaming of `sim` to `mock`
- Use ophyd-async's new soft signals, replacing dodal's home-grown ones

1. Confirm unit tests pass

### Checks for reviewer


src/dodal/devices/oav/pin_image_recognition/__init__.py
```diff
@@ -50,38 +51,30 @@ def __init__(self, prefix: str, name: str = ""):
...
-        self.triggered_tip = create_soft_signal_r(Tip, "triggered_tip", self.name)
-        self.triggered_top_edge = create_soft_signal_r(
+        self.triggered_tip, _ = soft_signal_r_and_setter(
+            Tip, "triggered_tip", self.name
+        )
+        self.triggered_top_edge, _ = soft_signal_r_and_setter(
             NDArray[np.uint32], "triggered_top_edge", self.name
         )
-        self.triggered_bottom_edge = create_soft_signal_r(
+        self.triggered_bottom_edge, _ = soft_signal_r_and_setter(
```
olliesilvester, line 60: As discussed, we can make a separate issue to make sure these signals use their setters rather than the backends in order to set values. Also, there may be scope to have a regular `soft_signal_r` function in ophyd-`async`

olliesilvester (approved):
Looks good, but I'm mainly trusting the unit tests! I'll make a separate issue regarding my comment

src/dodal/devices/webcam.py
```diff
@@ -3,18 +3,17 @@
...
 
 class Webcam(StandardReadable, Triggerable):
     def __init__(self, name, prefix, url):
         self.url = url
-        self.filename = create_soft_signal_rw(str, "filename", "webcam")
-        self.directory = create_soft_signal_rw(str, "directory", "webcam")
-        self.last_saved_path = create_soft_signal_rw(str, "last_saved_path", "webcam")
+        self.filename = soft_signal_rw(str, "filename", "webcam")
+        self.directory = soft_signal_rw(str, "directory", "webcam")
+        self.last_saved_path = soft_signal_rw(str, "last_saved_path", "webcam")
```
evalott100, line 16:
```suggestion
        self.filename = soft_signal_rw(str, name="filename")
        self.directory = soft_signal_rw(str, name="directory")
        self.last_saved_path = soft_signal_rw(str, name="last_saved_path")
```

`soft_signal_rw` has different arguments to `create_soft_signal_rw`.

We don't need to provide the source prefix to `soft_signal_rw`

evalott100 (changes requested):

DominicOram:
I think this Fixes https://github.com/DiamondLightSource/dodal/issues/482 and Fixes https://github.com/DiamondLightSource/dodal/issues/497

src/dodal/devices/tetramm.py
```diff
@@ -53,25 +53,25 @@ def __init__(
...
-        self.trigger_mode = ad_rw(TetrammTrigger, prefix + "TriggerMode")
-        self.bias = ad_rw(bool, prefix + "BiasState")
-        self.bias_volts = ad_rw(float, prefix + "BiasVoltage")
-        self.geometry = ad_rw(TetrammGeometry, prefix + "Geometry")
+        self.num_channels = epics_signal_rw(TetrammChannels, prefix + "NumChannels")
+        self.resolution = epics_signal_rw(TetrammResolution, prefix + "Resolution")
+        self.trigger_mode = epics_signal_rw(TetrammTrigger, prefix + "TriggerMode")
+        self.bias = epics_signal_rw(bool, prefix + "BiasState")
+        self.bias_volts = epics_signal_rw(float, prefix + "BiasVoltage")
+        self.geometry = epics_signal_rw(TetrammGeometry, prefix + "Geometry")
```
DominicOram, line 74: Must: I think the equivalent of `ad_rw` is `epics_signal_rw_rbv` see https://github.com/bluesky/ophyd-async/pull/274/files

src/dodal/devices/oav/pin_image_recognition/__init__.py
```diff
@@ -50,38 +51,30 @@ def __init__(self, prefix: str, name: str = ""):
...
-        self.validity_timeout = create_soft_signal_rw(
-            float, "validity_timeout", self.name
-        )
+        self.canny_upper_threshold = soft_signal_rw(int, "canny_upper", self.name)
+        self.canny_lower_threshold = soft_signal_rw(int, "canny_lower", self.name)
+        self.close_ksize = soft_signal_rw(int, "close_ksize", self.name)
+        self.close_iterations = soft_signal_rw(int, "close_iterations", self.name)
+        self.scan_direction = soft_signal_rw(int, "scan_direction", self.name)
+        self.min_tip_height = soft_signal_rw(int, "min_tip_height", self.name)
+        self.validity_timeout = soft_signal_rw(float, "validity_timeout", self.name)
```
DominicOram, line 77: Must: As @evalott100 has said below, the arguments here have changed and we're now actually making signals with an `initial_value` of e.g. "canny_upper".

DominicOram: Should: For bonus points could we use the initial values as in `connect`? I attempted to do this before but hit https://github.com/bluesky/ophyd-async/issues/266, which shouldn't be a problem now

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -80,15 +79,17 @@ def __init__(
...
             NDArray[np.uint64], "centres_of_mass", self.name
         )
-        self.bbox_sizes = create_soft_signal_r(
+        self.bbox_sizes, _ = soft_signal_r_and_setter(
             NDArray[np.uint64], "bbox_sizes", self.name
         )
-        self.ispyb_dcid = create_soft_signal_r(int, "ispyb_dcid", self.name)
-        self.ispyb_dcgid = create_soft_signal_r(int, "ispyb_dcgid", self.name)
+        self.ispyb_dcid, _ = soft_signal_r_and_setter(int, "ispyb_dcid", self.name)
+        self.ispyb_dcgid, _ = soft_signal_r_and_setter(int, "ispyb_dcgid", self.name)
```
DominicOram, line 92: Must: Again, the signature is different

DominicOram (changes requested):
Thanks. Mostly echoing @evalott100's comment but in some additional places too

src/dodal/devices/webcam.py
callumforrester: Concerning that that didn't make any tests fail...

DominicOram: Slightly, I'm not overly surprised that the initial_value isn't covered in tests. I don't think that matters for quite a few signals. We should probably cover the names better though.

It's also concerning that the type checker doesn't catch `soft_signal_rw(int, initial_value="string")` as a problem.

callumforrester: Hmm, I did this in part to use #497 because the initial values weren't working properly in my tests for the `Undulator`, and I decided to upgrade before tackling the problem. Is it possible that `initial_value` never worked?

src/dodal/devices/zocalo/zocalo_results.py
dperl-dls: This really ought to be caught by a test

callumforrester: https://github.com/DiamondLightSource/dodal/issues/527

DominicOram (changes requested):
Thanks, couple more commenst in the code. I think the `OAV` is broken for i22's system tests on main but otherwise `python tests/system_tests/test_i22_system.py` should pass

src/dodal/devices/tetramm.py
callumforrester: Hmm, have done this but the PVs seem to be wrong now, will investigate

callumforrester:
@DominicOram the OAV bug was fixed in TBD, have applied the tetramm fixes and now the system test is happy:

```shell
(venv) vid18871@pc0101:/scratch/vid18871/projects/dodal# python tests/system_tests/test_i22_system.py 
Making all i22 devices
Successfully connected
```

DominicOram (approved):
Great, thank you. Going to merge as this is breaking other things

DominicOram:
FYI, I believe this has slowed down are tests quite a bit https://github.com/DiamondLightSource/dodal/issues/533

Issue #528, opened 2024-05-14
 ApertureScatterguard has no prefix


DominicOram:
When running `python tests/system_tests/test_i03_system.py`(with the devices commented out as specified in https://github.com/DiamondLightSource/dodal/issues/261) I get the following error:

```
ophyd_async.core.utils.NotConnected: 
coros: NotConnected:
    aperture: NotConnected:
        x: NotConnected:
            motor_egu: NotConnected: ca://-MO-MAPT-01:X.EGU
            velocity: NotConnected: ca://-MO-MAPT-01:X.VELO
            user_readback: NotConnected: ca://-MO-MAPT-01:X.RBV
            user_setpoint: NotConnected: ca://-MO-MAPT-01:X.VAL
            max_velocity: NotConnected: ca://-MO-MAPT-01:X.VMAX
            acceleration_time: NotConnected: ca://-MO-MAPT-01:X.ACCL
            precision: NotConnected: ca://-MO-MAPT-01:X.PREC
            deadband: NotConnected: ca://-MO-MAPT-01:X.RDBD
            motor_done_move: NotConnected: ca://-MO-MAPT-01:X.DMOV
            low_limit_travel: NotConnected: ca://-MO-MAPT-01:X.LLM
            high_limit_travel: NotConnected: ca://-MO-MAPT-01:X.HLM
            motor_stop: NotConnected: ca://-MO-MAPT-01:X.STOP
        y: NotConnected:
            motor_egu: NotConnected: ca://-MO-MAPT-01:Y.EGU
            velocity: NotConnected: ca://-MO-MAPT-01:Y.VELO
            user_readback: NotConnected: ca://-MO-MAPT-01:Y.RBV
            user_setpoint: NotConnected: ca://-MO-MAPT-01:Y.VAL
            max_velocity: NotConnected: ca://-MO-MAPT-01:Y.VMAX
            acceleration_time: NotConnected: ca://-MO-MAPT-01:Y.ACCL
            precision: NotConnected: ca://-MO-MAPT-01:Y.PREC
            deadband: NotConnected: ca://-MO-MAPT-01:Y.RDBD
            motor_done_move: NotConnected: ca://-MO-MAPT-01:Y.DMOV
            low_limit_travel: NotConnected: ca://-MO-MAPT-01:Y.LLM
            high_limit_travel: NotConnected: ca://-MO-MAPT-01:Y.HLM
[...]

PR #532, opened 2024-05-14
Remove indexing logic on barcode and tests

### LINKED ISSUE #530: Bart Robot: Barcode can now be read as a normal string
On i03 controls have changed the barcode so that we can just read it as a string. We should change this on our side and remove the horrible string indexing logic.

## Acceptance Criteria
* String indexing logic is removed and bart can still read barcode

DominicOram:
Fixes #530 

1. Confirm unit tests pass
2. From i03.py commenting out aperture scatterguard (https://github.com/DiamondLightSource/dodal/issues/528), pin_detect and panda (https://github.com/DiamondLightSource/dodal/issues/261)
3. Confirm test_i03_system passes here and fails on main

### Checks for reviewer


olliesilvester (approved):
That's a nice change! Confirmed the system test passes here but not on main

PR #531, opened 2024-05-14
Add ophyd async logger to dodal logger

olliesilvester:
Fixes #434 

### Instructions to reviewer on how to test:
1. Confirm logs are correctly added and that tests pass

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly

PR #535, opened 2024-05-15
Remove dodgy logging test

olliesilvester:
In https://github.com/DiamondLightSource/dodal/pull/531 I added a test which uses global variables, so is very flakey. The other test added,`test_ophyd_async_logger_integrated`, tests basically the same thing anyway

### Checks for reviewer


DominicOram (approved):
Great, thank you

PR #544, opened 2024-05-15
Move non-beamline modules out of `beamlines` package

### LINKED ISSUE #543: Move non-beamline modules out of beamlines package
We could move `_device_helpers.py` etc. into some other location and just have a convention that only beamline modules live in `beamlines`. Then we can just auto-detect based on all the module names in that folder?

_Originally posted by @DominicOram in https://github.com/DiamondLightSource/dodal/pull/519#discussion_r1598357802_

## Acceptance Criteria
- Only beamline modules in `dodal.beamlines`
- Non-beamline modules still accessible, but via a different import path

callumforrester:
Fixes #543 

Suggested location, the alternative is turning `dodal.utils` into a directory.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


callumforrester:
Another alternative is to add to the `common` package that is already there and make `common.beamlines`

DominicOram (approved):
Thanks, I'm ambivalent on it's exact location. Here is fine. Could you add this convention in some docs somewhere probably both:
* In a `README` inside `beamlines`
* https://github.com/DiamondLightSource/dodal/wiki/Adding-a-Beamline

I'm happy for you to then merge without re-review

Issue #533, opened 2024-05-14
Improve test speed 


DominicOram:
Unit tests are now taking ~45 sec (https://github.com/DiamondLightSource/dodal/actions/runs/9084455713/job/24965427206) the last test only took ~20 sec (https://github.com/DiamondLightSource/dodal/actions/runs/9017299416/job/24775541622)

The main difference is https://github.com/DiamondLightSource/dodal/pull/525

## Acceptance Criteria
* Tests are reduced back down to < 30 s
* https://github.com/DiamondLightSource/dodal/pull/546 is reverted and still regularly passes

callumforrester:
From passing `--durations=10` to pytest:

The 10 slowest tests from before #525  (total local run time of 17s)
```shell
====================================================================================================================================== slowest 10 durations ======================================================================================================================================
3.00s call     tests/devices/unit_tests/test_focusing_mirror.py::test_mirror_set_voltage_sets_and_waits_settle_timeout_expires
2.01s call     tests/devices/unit_tests/test_zocalo_results.py::test_subscribe_only_on_called_stage
1.01s call     tests/devices/unit_tests/test_xspress3mini.py::test_stage_fails_in_failed_acquire_state
0.15s setup    tests/beamlines/unit_tests/test_device_instantiation.py::test_device_creation[i22]
0.14s setup    tests/beamlines/unit_tests/test_device_instantiation.py::test_devices_are_identical[i22]
0.14s call     tests/devices/unit_tests/test_zocalo_interaction.py::test_run_start[with_exception-expected_message1]
0.13s call     tests/beamlines/unit_tests/test_beamline_utils.py::test_device_is_new_after_clearing
0.13s call     tests/devices/unit_tests/test_gridscan.py::test_scan_within_limits_1d[-1-5-1-False]
0.11s setup    tests/devices/unit_tests/test_attenuator.py::test_set_transmission_in_run_engine
0.11s call     tests/devices/unit_tests/test_eiger.py::test_given_in_free_run_mode_and_not_all_frames_collected_in_time_when_unstaged_then_odin_stopped_and_exception_thrown
```

The 10 slowest tests after (total local run time of 35s)
```shell
====================================================================================================================================== slowest 10 durations ======================================================================================================================================
3.00s call     tests/devices/unit_tests/test_focusing_mirror.py::test_mirror_set_voltage_sets_and_waits_settle_timeout_expires
[...]

PR #546, opened 2024-05-16
Increase timeout to fix test

DominicOram:
Due to https://github.com/DiamondLightSource/dodal/issues/533 this test is now regularly timing out. This fixes that for now but I have also added a note to the speed up issue to revert this

1. Confirm this test is flaky in main
2. Confirm it is not so here

### Checks for reviewer


callumforrester (approved):
Hopefully not permanent and can be reversed if #533 is fixed

PR #547, opened 2024-05-16
Fix slit naming

### LINKED ISSUE #529: Misleading slits device name
https://github.com/DiamondLightSource/dodal/blob/050e9a4405e6a87e39843682df9963f6d8be3ced/src/dodal/beamlines/_device_helpers.py#L29

From @Tom-Willemsen: Should be called "slits" rather than "slit_gaps".

Tagging @callumforrester @joeshannon

keithralphs:
Fixes #529

Run unit tests which replicate the declaration as used in dodal/beamlines/i22.py but with a simulated ophyd device

### Checks for reviewer

### DIFF
--- a/tests/beamlines/unit_tests/test_device_helpers.py
+++ b/tests/beamlines/unit_tests/test_device_helpers.py
@@ -1,15 +1,15 @@
-from random import randint
+import pytest
 
 from dodal.beamlines._device_helpers import numbered_slits
 
 
-def test_slit_field_naming():
-    num = randint(1, 20)
+@pytest.mark.parametrize("num", [5, 17])
+def test_slit_field_naming(num):
     prefix = f"-AL-SLITS-{num:02}:"
     slits = numbered_slits(
-        num,
-        False,
-        True,
+        slit_number=num,
+        wait_for_connection=False,
+        fake_with_ophyd_sim=True,
     )
 
     assert slits.name == f"slits_{num}"
--- a/tests/beamlines/unit_tests/test_device_helpers.py
+++ b/tests/beamlines/unit_tests/test_device_helpers.py
@@ -1,0 +1,24 @@
+from random import randint
+from dodal.beamlines._device_helpers import numbered_slits
+
+
+def test_slit_field_naming():
+
+    num = randint(1, 20)
+    prefix = f"-AL-SLITS-{num:02}:"
+    slits = numbered_slits(
+        num,
+        False,
+        True,
+    )
+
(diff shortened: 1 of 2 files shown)


tests/beamlines/unit_tests/test_device_helpers.py
```diff
@@ -0,0 +1,23 @@
...
+from dodal.beamlines._device_helpers import numbered_slits
+
+
+def test_slit_field_naming():
+    num = randint(1, 20)
+    prefix = f"-AL-SLITS-{num:02}:"
+    slits = numbered_slits(
+        num,
+        False,
+        True,
```
Tom-Willemsen, line 12: I think it would be clearer to pass these as kwargs.

tests/beamlines/unit_tests/test_device_helpers.py
```diff
@@ -0,0 +1,23 @@
+from random import randint
+
+from dodal.beamlines._device_helpers import numbered_slits
+
+
+def test_slit_field_naming():
+    num = randint(1, 20)
```
Tom-Willemsen, line 7: In general I really don't like seeing `random` in tests - it's a recipe for flaky tests later.

If we want to test various cases, we can parameterize the test with e.g. `1, 2, 20`.

Tom-Willemsen (changes requested):
In reviewing this I also noticed that i22 is set up to have `fake_with_ophyd_sim=True` by default in the beamline module - that's likely not helpful behaviour (took me a while to understand why bluesky was saying all the slits were at 0 all the time!)

But the naming change looks fine.

tests/beamlines/unit_tests/test_device_helpers.py
keithralphs: Do you mean 

```
    slits = numbered_slits(
        num,
        wait_for_connection=False,
        fake_with_ophyd_sim=True,
    }
```

?

Tom-Willemsen: Yes exactly. Certainly for the two booleans, I'd probably do it for `num` as well personally but don't mind so much about that one

keithralphs: ok

tests/beamlines/unit_tests/test_device_helpers.py
keithralphs: Ok, parameterized it is

Tom-Willemsen (approved):

PR #548, opened 2024-05-16
(#494) Use ApertureScatterguard tolerances from parameter file

### LINKED ISSUE #494: Use tolerances in beamline parameters file for aperture scatterguard
For the aperture scatterguard to be in position we currently use the motor tolerances, we should instead use those in the beamline parameters file see https://diamondlightsource.slack.com/archives/C048RD0KFFF/p1714579073626219

## Acceptance Criteria
* Tolerances from beamline parameters file are used in the aperture/scatterguard

dperl-dls:
Fixes #494
Loads tolerances for `ApertureScatterguard` movements from the beamline parameters file

1. Run tests

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/aperturescatterguard.py
+++ b/src/dodal/devices/aperturescatterguard.py
@@ -28,6 +28,15 @@ ApertureFiveDimensionalLocation = namedtuple(
 
 
 @dataclass
+class ApertureScatterguardTolerances:
+    ap_x: float
+    ap_y: float
+    ap_z: float
+    sg_x: float
+    sg_y: float
+
+
+@dataclass
 class SingleAperturePosition:
     # Default values are needed as ophyd_async sim does not respect initial_values of
     # soft signal backends see https://github.com/bluesky/ophyd-async/issues/266
@@ -56,6 +65,16 @@ def position_from_params(
     )
 
 
+def tolerances_from_params(params: dict) -> ApertureScatterguardTolerances:
+    return ApertureScatterguardTolerances(
+        ap_x=params["miniap_x_tolerance"],
+        ap_y=params["miniap_y_tolerance"],
+        ap_z=params["miniap_z_tolerance"],
+        sg_x=params["sg_x_tolerance"],
+        sg_y=params["sg_y_tolerance"],
+    )
+
+
 @dataclass
 class AperturePositions:
     """Holds the motor positions needed to select a particular aperture size."""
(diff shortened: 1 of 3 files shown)


src/dodal/devices/aperturescatterguard.py
```diff
@@ -171,14 +192,13 @@ async def _get_current_aperture_position(self) -> SingleAperturePosition:
...
         robot_load_ap_y = self.aperture_positions.ROBOT_LOAD.location.aperture_y
-        tolerance = self.TOLERANCE_STEPS * await self.aperture.y.deadband.get_value()
         if await self.aperture.large.get_value(cached=False) == 1:
             return self.aperture_positions.LARGE
         elif await self.aperture.medium.get_value(cached=False) == 1:
             return self.aperture_positions.MEDIUM
         elif await self.aperture.small.get_value(cached=False) == 1:
             return self.aperture_positions.SMALL
-        elif current_ap_y <= robot_load_ap_y + tolerance:
+        elif current_ap_y <= robot_load_ap_y + self.aperture_positions.tolerances.ap_y:
```
olliesilvester, line 201: Is this definitely the correct variable? The issue says we should be using the scatterguard tolerance (`sg_y_tolerance`), but this is using the `miniap_y_tolerance`

src/dodal/devices/aperturescatterguard.py
```diff
@@ -203,13 +223,12 @@ async def _safe_move_within_datacollection_range(
             )
 
         current_ap_z = await self.aperture.z.user_readback.get_value()
-        tolerance = self.TOLERANCE_STEPS * await self.aperture.z.deadband.get_value()
         diff_on_z = abs(current_ap_z - aperture_z)
-        if diff_on_z > tolerance:
+        if diff_on_z > self.aperture_positions.tolerances.ap_z:
```
olliesilvester, line 227: Same comment as above - wrong tolerance maybe?

olliesilvester (changes requested):
Logically looks good, but I think you may be using the miniap tolerance instead of scatterguard? Feel free to dismiss this if I'm wrong though

src/dodal/devices/aperturescatterguard.py
dperl-dls: where do you see this in the issue? we're comparing positions for the aperture device so I don't see why we wouldn't use the tolerances for it too

olliesilvester (approved):

PR #549, opened 2024-05-16
Read aperture size in serialisable format

dperl-dls:
Needed for [Hyperion 1311](https://github.com/DiamondLightSource/hyperion/issues/1311)
- Add a Converter and method to convert SingleAperturePosition
- Add a util to get beam size x and y from the above

1. Run tests

### Checks for reviewer

### DIFF
--- a/src/dodal/beamline_specific_utils/i03.py
+++ b/src/dodal/beamline_specific_utils/i03.py
@@ -2,7 +2,7 @@ from dataclasses import dataclass
 
 from dodal.devices.aperturescatterguard import SingleAperturePosition
 
-I03_BEAM_HEIGHT = 20
+I03_BEAM_HEIGHT_UM = 20
 
 
 @dataclass
@@ -13,4 +13,4 @@ class BeamSize:
 
 def beam_size_from_aperture(position: SingleAperturePosition):
     aperture_size = position.radius_microns
-    return BeamSize(aperture_size, I03_BEAM_HEIGHT if aperture_size else None)
+    return BeamSize(aperture_size, I03_BEAM_HEIGHT_UM if aperture_size else None)
--- a/src/dodal/beamline_specific_utils/i03.py
+++ b/src/dodal/beamline_specific_utils/i03.py
@@ -1,0 +1,16 @@
+from dataclasses import dataclass
+
+from dodal.devices.aperturescatterguard import SingleAperturePosition
+
+I03_BEAM_HEIGHT = 20
+
+
+@dataclass
+class BeamSize:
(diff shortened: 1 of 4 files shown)


src/dodal/beamline_specific_utils/i03.py
```diff
@@ -0,0 +1,16 @@
...
+
+@dataclass
+class BeamSize:
+    x_um: float | None
+    y_um: float | None
+
+
+def beam_size_from_aperture(position: SingleAperturePosition):
+    aperture_size = position.radius_microns
+    return BeamSize(aperture_size, I03_BEAM_HEIGHT if aperture_size else None)
```
olliesilvester, line 16: I don't think this file has to be an i03-specific util. We could read the `I03_BEAM_HEIGHT` from some parameter and then have this as a generic util

src/dodal/devices/aperturescatterguard.py
```diff
@@ -93,18 +101,30 @@ def __init__(self, prefix: str = "", name: str = "") -> None:
...
-        self.selected_aperture = self.SelectedAperture(
-            backend=SoftSignalBackend(
-                SingleAperturePosition, AperturePositions.UNKNOWN
-            ),
+        aperture_backend = SoftSignalBackend(
+            SingleAperturePosition, AperturePositions.UNKNOWN
         )
+        aperture_backend.converter = self.ApertureConverter()
+        self.selected_aperture = self.SelectedAperture(backend=aperture_backend)
+        self.selected_aperture
```
olliesilvester, line 109: Remove typo

src/dodal/devices/aperturescatterguard.py
```diff
@@ -38,6 +38,14 @@ class SingleAperturePosition:
         0, 0, 0, 0, 0
     )
 
+    def dict(self):
```
olliesilvester, line 41: Can we use the `dataclass.asdict()` method for this instead?

olliesilvester (approved):
Looks good, thanks. There's one typo to remove and an optional comment about using `asdict`

src/dodal/beamline_specific_utils/i03.py
dperl-dls: yeah, I kind of agree, there will be more beamlines that could use this, but I think it's easy enough to move it when we need it

src/dodal/devices/aperturescatterguard.py
dperl-dls: yeah, I had actually already done that, just forgot to push :man_facepalming:

PR #550, opened 2024-05-16
added Exception handling for devices

ZohebShaikh:
closes https://github.com/DiamondLightSource/blueapi/issues/444

The main change in this PR is that `make_all_devices` returns tuple of `(all_devices, exception_devices)` 

1. Checkout branch `device-resilience` in dodal
2. In a common python virtual environment install both blueapi and dodal 
`pip install -e ../dodal[dev]`  `pip install -e .[dev]`
3. Test device does not fail when exception is occurred

### Checks for reviewer
- [ ] Achives requirement for  https://github.com/DiamondLightSource/blueapi/issues/444


tests/test_utils.py
```diff
@@ -66,7 +66,9 @@ def test_no_signature_builtins_not_devices() -> None:
     import tests.fake_beamline_misbehaving_builtins as fake_beamline
 
     devices = make_all_devices(fake_beamline)
-    assert not devices
+    assert not devices[0]
+    for exception in devices[1]:
```
ZohebShaikh, line 70: I just had a look at this, the device fake_beamlines don't create exceptions so I will need to add some devices in the fake_beamline that raises a exception when it does not connect

tests/beamlines/unit_tests/test_beamline_utils.py
```diff
@@ -69,7 +69,7 @@ def test_instantiate_v2_function_fake_makes_fake():
 
 def test_clear_devices(RE):
     devices = make_all_devices(i03, fake_with_ophyd_sim=True)
-    assert len(beamline_utils.ACTIVE_DEVICES) == len(devices.keys())
+    assert len(beamline_utils.ACTIVE_DEVICES) == len(devices[0].keys())
```
callumforrester, line 72: Nit: You could instead have

```python
devices, exceptions = make_all_devices(i03, fake_with_ophyd_sim=True)
assert len(beamline_utils.ACTIVE_DEVICES) == len(devices.keys()) and len(exceptions) == 0
```

...which is more readable, ditto for the other tests

src/dodal/utils.py
```diff
@@ -109,7 +110,7 @@ def wrapper(*args, **kwds) -> T:
 
 def make_all_devices(
```
callumforrester, line 111: Should: The docstring for this function needs updating

tests/test_utils.py
callumforrester: I think we need the following test cases:
- All devices pass
- All devices fail
- Mixture
If you have to make a few small fake beamline modules to cover those then go for it

src/dodal/utils.py
```diff
@@ -125,16 +126,19 @@ def make_all_devices(
...
     return devices
 
 
 def invoke_factories(
     factories: Mapping[str, AnyDeviceFactory],
     **kwargs,
-) -> Dict[str, AnyDevice]:
+) -> Tuple[Dict[str, AnyDevice], Dict[str, Exception]]:
     devices: dict[str, AnyDevice] = {}
+    exception_devices: dict[str, Exception] = {}
```
callumforrester, line 141: Nit: I think this makes more sense

```suggestion
    device_exceptions: dict[str, Exception] = {}
```

callumforrester (changes requested):
LGTM! Just the tests and a couple of nitpicks

src/dodal/utils.py
```diff
@@ -150,11 +158,18 @@ def invoke_factories(
...
-    all_devices = {device.name: device for device in devices.values()}
+        try:
+            devices[dependent_name] = factories[dependent_name](**params, **kwargs)
+        except Exception as e:
+            device_exceptions[dependent_name] = e
+
+    all_devices = {
+        device.name: device
+        for device in devices.values()
+        if not isinstance(device, Exception)
```
callumforrester, line 169: I don't think you need this line

callumforrester:
I'm taking over this PR in @ZohebShaikh's absence

tests/fake_beamline_all_devices_rasie_exception.py
```diff
@@ -0,0 +1,21 @@
...
+def device_a() -> Readable:
+    raise DestroyedError
+
+
+def device_b() -> EpicsMotor:
+    raise DisconnectedError
+
+
+def device_c() -> Cryo:
+    raise UnknownStatusFailure
```
DiamondJoseph, line 21: Currently one Protocol, 2 ophyd devices, add an ophyd-async device class?

DiamondJoseph (changes requested):
Overall looks good just some edge cases and minor optimisations

tests/fake_beamline_all_devices_rasie_exception.py
callumforrester: Given that we are moving in the direction of exclusive ophyd-async (and therefore hopefully dodal's handling of ophyd devices can go away one day) I might just make it ophyd-async exclusively

DiamondJoseph: I'd prefer to see both tested until there are no ophyd devices, given they are different execution paths

callumforrester: Have reverted cryostream

src/dodal/utils.py
```diff
@@ -120,41 +121,49 @@ def make_all_devices(
...
+    )
 
     return devices
 
 
 def invoke_factories(
     factories: Mapping[str, AnyDeviceFactory],
     **kwargs,
-) -> Dict[str, AnyDevice]:
+) -> Tuple[Dict[str, AnyDevice], Dict[str, Exception]]:
```
DominicOram, line 142: Could: I think it's worth a bit of a docstring here on how the two dicts relate to each other. e.g. that every device will be represented in one list, either the first if it succeeds or the second if it fails? I wonder if a type like `Dict[str, AnyDevice | Exception]` would make this more clear but I'm not too sure.

DominicOram: I also thing a general `docstring` on what's going on with the leaves etc. would be nice

DominicOram (changes requested):
Ta, I think mostly just need to fix the `cli`

src/dodal/utils.py
callumforrester: Have added some docstrings, the other option which I'm tempted by is to return a dataclass rather than a tuple. Will do that if no one objects.

DominicOram: Fine by me

DominicOram (approved):
Great, thank you, happy for this to be merged now but if you want to do the `dataclass` change then that's fine too

DiamondJoseph (approved):

PR #552, opened 2024-05-20
Disable visit service connection on I22 and P38

callumforrester:
Temporary measure to facilitate beamline testing: Don't require a connection with the GDA visit service to write data on I22 and P38.

1. Confirm unit tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i22.py
+++ b/src/dodal/beamlines/i22.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from ophyd_async.epics.areadetector import AravisDetector, PilatusDetector
 from ophyd_async.panda import HDFPanda
 
@@ -7,7 +9,7 @@ from dodal.beamlines.beamline_utils import (
     set_directory_provider,
 )
 from dodal.beamlines.beamline_utils import set_beamline as set_utils_beamline
-from dodal.common.visit import StaticVisitDirectoryProvider
+from dodal.common.visit import LocalDirectoryServiceClient, StaticVisitDirectoryProvider
 from dodal.devices.focusing_mirror import FocusingMirror
 from dodal.devices.i22.fswitch import FSwitch
 from dodal.devices.slits import Slits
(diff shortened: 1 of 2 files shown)


src/dodal/beamlines/i22.py
```diff
@@ -22,13 +24,17 @@
...
+# infrastructure that is still WIP.
+# Communication with GDA is also WIP so for now we determine an arbitrary scan number
+# locally and write the commissioning directory. The scan number is not guaranteed to
+# be unique and the data is at risk - this configuration is for testing only.
 set_directory_provider(
     StaticVisitDirectoryProvider(
         BL,
-        "/data/i22/2024/VISIT",
+        Path("/dls/i22/data/2024/cm37271-2/bluesky"),
+        client=LocalDirectoryServiceClient(),
```
DiamondJoseph, line 37: Can LocalDirectoryServiceClient take a init arg of initial_count: int = 0 which can be set arbitrarily high to prevent collision? I worry about deploying something that can overwrite data to the beamline.

Tom-Willemsen: Maybe I'm lacking context here but if blueapi is restarted mid-experiment I think this overwrites data - I think that's a pretty likely case for the June experiment...

As a simple interim solution before the proper GDA <-> bluesky communication is sorted, maybe we could mitigate this by the folder containing the timestamp at which `StaticVisitDirectoryProvider` was provided - then at least a subsequent restart wouldn't overwrite old data.

callumforrester: This absolutely can overwrite data, see the comment in the diff. It meant to just be a temporary change and only write to `<commissioning dir>/bluesky`, so as not to break anything else. I felt that was sufficient protection for the beamline but if you two don't, no problem, we can think of something else. 

@Tom-Willemsen this is not intended for the June experiment, just to unblock testing in the interim so blueapi can at least write _some_ data. Creating a unique directory is an interesting idea, but blueapi doesn't have write permissions to the visit directory and in my opinion it never should, then we don't have to deal with the gda2 problem.

DiamondJoseph (approved):
for now...

PR #553, opened 2024-05-20
Add devices for I24 serial

noemifrisina:
Needed for [MX#97](https://github.com/DiamondLightSource/mx_bluesky/issues/97)

Adds a couple of devices for I24

1. Tests pass

### Checks for reviewer


src/dodal/devices/i24/aperture.py
```diff
@@ -0,0 +1,35 @@
...
+    IN = "In"
+    OUT = "Out"
+    ROBOT = "Robot"
+    MANUAL = "Manual Mounting"
+
+
+class AperturePositioner(StandardReadable):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self.pos_select = epics_signal_rw(ApPosition, prefix + "MP:SELECT")
+        super().__init__(name)
```
DominicOram, line 18: Should: Why is this in a new class? Can we not just have the signal in the `Aperture` class?

src/dodal/devices/i24/aperture.py
```diff
@@ -0,0 +1,35 @@
...
+
+    When a position is selected, the x motor is moved.
+    The position of the y motor is calibrated at start up and is not changed.
+    """
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self.x = Motor(prefix + "X")
+        self.y = Motor(prefix + "Y")
+
+        self.pos = AperturePositioner(prefix, name)
```
DominicOram, line 34: Nit: I would call this `position` rather than `pos`

src/dodal/devices/i24/beamstop.py
```diff
@@ -0,0 +1,37 @@
+from enum import Enum
+
+from ophyd_async.core import StandardReadable
+from ophyd_async.epics.motion import Motor
+from ophyd_async.epics.signal import epics_signal_rw
+
+
+class BSPositions(str, Enum):
```
DominicOram, line 8: Nit: I would call this `BeamStopPositions`, prefer clarity over brevity

src/dodal/devices/i24/beamstop.py
```diff
@@ -0,0 +1,37 @@
...
+    CHECKBEAM = "CheckBeam"
+    DATACOLLECTION = "Data Collection"
+    DATACOLLECTIONFAR = "Data Collection Far"
+    TRAYMOUNT = "Tray Mount"
```
DominicOram, line 12: Nit: Would be good to have these hyphen separated e.g. `DATA_COLLECTION`, makes them a bit more readable

src/dodal/devices/i24/beamstop.py
```diff
@@ -0,0 +1,37 @@
...
+
+class Beamstop(StandardReadable):
+    """Device to move the beamstop.
+
+    The positioner moves the x,y,z motors when a position is selected.
+    The additional rot_y motor is independent of the positioner and can to be moved
+    on its own as needed.
+
+    WARNING. Before moving the rot_y motor away from 0, it is important to make sure
+    that the backlight is in the "OUT" position to avoid a collision.
```
DominicOram, line 25: Could: We should protect against this somehow. What we have done in other places where there are potential collisions is make a new composite device like `BeamStopAndBacklight` where moving `rot_y` would move the backlight first. Or we could not allow the rot_y to move at all in this case

src/dodal/devices/i24/beamstop.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self.x = Motor(prefix + "X")
+        self.y = Motor(prefix + "Y")
+        self.z = Motor(prefix + "Z")
+
+        self.roty = Motor(prefix + "ROTY")
+
+        self.pos_select = epics_signal_rw(BSPositions, prefix + "MP:SELECT")
```
DominicOram, line 35: Must: Does this move the `rot_y` too? In which case is there the same collision risk?

src/dodal/devices/i24/beamstop.py
```diff
@@ -0,0 +1,37 @@
...
+    WARNING. Before moving the rot_y motor away from 0, it is important to make sure
+    that the backlight is in the "OUT" position to avoid a collision.
+    """
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self.x = Motor(prefix + "X")
+        self.y = Motor(prefix + "Y")
+        self.z = Motor(prefix + "Z")
+
+        self.roty = Motor(prefix + "ROTY")
```
DominicOram, line 33: Nit: I think `y_rotation` is a bit more clear?

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -1,14 +1,36 @@
...
+
+class BLPositions(str, Enum):
+    OUT = "Out"
+    IN = "In"
+    LOADCHECK = "LoadCheck"
+    OAV2 = "OAV2"
+    DIODE = "Diode"
+
+
+class LedStatus(str, Enum):
```
DominicOram, line 15: Nit: LED is an acronym so I would suggest `LEDStatus`

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -1,14 +1,36 @@
...
+from ophyd_async.epics.signal import epics_signal_rw
 
-class BacklightPositioner(Device):
+
+class BLPositions(str, Enum):
+    OUT = "Out"
+    IN = "In"
+    LOADCHECK = "LoadCheck"
+    OAV2 = "OAV2"
+    DIODE = "Diode"
```
DominicOram, line 12: Nit: As above, I would prefer `BacklightPositions` and `LOAD_CHECK`

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -1,14 +1,36 @@
...
+    ON = "ON"
+
+
+class BacklightPositioner(StandardReadable):
     """Device to control the backlight position."""
 
-    # String description of the backlight position e.g. "In", "OAV2"
-    pos_level = Component(EpicsSignal, "MP:SELECT")
+    def __init__(self, prefix: str, name: str = "") -> None:
+        # String description of the backlight position e.g. "In", "OAV2"
```
DominicOram, line 24: Nit: This isn't a string, it's an enum?

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -26,18 +48,17 @@ class DualBacklight(Device):
...
-    IN = "In"
-
-    led1 = Component(EpicsSignal, "-DI-LED-01:TOGGLE")
-    pos1 = Component(BacklightPositioner, "-MO-BL-01:")
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self.led1 = epics_signal_rw(LedStatus, prefix + "-DI-LED-01:TOGGLE")
+        self.pos1 = BacklightPositioner(prefix + "-MO-BL-01", name)
 
-    led2 = Component(EpicsSignal, "-DI-LED-02:TOGGLE")
+        self.led2 = epics_signal_rw(LedStatus, prefix + "-DI-LED-02:TOGGLE")
```
DominicOram, line 55: Should: It would be good if these were more readable to the scientist e.g. `backlight_state` rather than `led1`?

DominicOram (changes requested):
Thanks! Some changes requested. I didn't get a chance to run `dodal connect i24` as I'm in a meeting at the moment, I assume this works?

noemifrisina:
> Thanks! Some changes requested. I didn't get a chance to run `dodal connect i24` as I'm in a meeting at the moment, I assume this works?

Looked fine bit I hadn't run it since before merging the pmac - and apparently forgot to for that one. A couple of pmac-related errors that should be fixed now

src/dodal/devices/i24/beamstop.py
noemifrisina: No, `pos_select` only moves x,y,z motors. The `rot_y` functionality was an add on on request of the beamline staff but is not moved by the positioner - and they're not really moving that motor it beyond setting it to 0 at the moment because of the collision issue

src/dodal/devices/i24/dual_backlight.py
noemifrisina: Renamed but I thought it was clearer to use just `backlight` and `frontlight` which is what the scientists used to refer to them. Left LED1 and LED2 in the docstring as that's what they're named in the backlight synoptic

src/dodal/devices/i24/beamstop.py
noemifrisina: For a device, I would pick the second option, so check if the backlight is in and if it is don't allow rot_y to move (the beamstop can still go out of the way by moving the other motors with the positioner). 
However, I'd make a separate issue for creating this device and discuss it better with the scientists - I got the impression when I asked that it's not really used - have yet to take a good look at what happens in GDA though. For the serial stuff, they only set this to 0 at the start of the collection to make sure it won't be in the backlight's way.

src/dodal/devices/i24/beamstop.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class Beamstop(StandardReadable):
+    """Device to move the beamstop.
+
+    The positioner moves the x,y,z motors when a position is selected.
+    The additional rot_y motor is independent of the positioner and can to be moved
+    on its own as needed.
+
+    WARNING. Before moving the rot_y motor away from 0, it is important to make sure
```
DominicOram, line 24: Should: `rot_y` in the docstring no longer matches the name of the signal.

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -39,26 +39,26 @@ class DualBacklight(StandardReadable):
...
 
     Note that the two LED are independently switched on and off. When LED1 is
     in "Out" position (switched off), LED2 might still be on.
     """
 
     def __init__(self, prefix: str, name: str = "") -> None:
-        self.led1 = epics_signal_rw(LedStatus, prefix + "-DI-LED-01:TOGGLE")
-        self.pos1 = BacklightPositioner(prefix + "-MO-BL-01", name)
+        self.backlight_state = epics_signal_rw(LEDStatus, prefix + "-DI-LED-01:TOGGLE")
+        self.bl_position = BacklightPositioner(prefix + "-MO-BL-01:", name)
```
DominicOram, line 53: Nit: I would go for `backlight_position` for clarity

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -39,26 +39,26 @@ class DualBacklight(StandardReadable):
...
+        self.frontlight_state = epics_signal_rw(LEDStatus, prefix + "-DI-LED-02:TOGGLE")
         super().__init__(name)
 
     @AsyncStatus.wrap
-    async def set(self, position: BLPositions):
-        await self.pos1.pos_level.set(position, wait=True)
-        if position == BLPositions.OUT:
-            await self.led1.set(LedStatus.OFF, wait=True)
+    async def set(self, position: BacklightPositions):
+        await self.bl_position.pos_level.set(position, wait=True)
```
DominicOram, line 60: Could: You've overridden the set so you should be able to just do `await self.bl_position.set(position, wait=True)`

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -17,27 +39,26 @@ class DualBacklight(Device):
 
     To set the position for LED1:
         b = DualBacklight(name="backlight)
-        b.pos1.pos_level.set("OAV2")
+        b.bl_position.pos_level.set("OAV2")
 
     To see get the available position values for LED1:
-        b.pos1.alowed_backlight_positions
+        b.bl_position.alowed_backlight_positions
```
DominicOram, line 45: Should: This isn't something you can do any more.

src/dodal/devices/i24/dual_backlight.py
```diff
@@ -17,27 +39,26 @@ class DualBacklight(Device):
 
     To set the position for LED1:
         b = DualBacklight(name="backlight)
-        b.pos1.pos_level.set("OAV2")
+        b.bl_position.pos_level.set("OAV2")
```
DominicOram, line 42: Could: You can just do `b.bl_position.set("OAV2")` because you have overridden the set

DominicOram (changes requested):
Great, thanks. Some final minor comments

src/dodal/devices/i24/beamstop.py
noemifrisina: Issue [here](https://github.com/DiamondLightSource/dodal/issues/646)

DominicOram (approved):
Great, thank you!

PR #556, opened 2024-05-20
Correct MirrorType declaration

### LINKED ISSUE #537: Running count plan against i22 HFM/VFM gives serialization error
In testing on i22 on 14/05/24, when the Count plan was run via BlueAPI against the HFM and VFM devices the the following error was reported:

MirrorType is not serializable

This appears to be due to the MirrorType class only having an Enum base class, rather than Enum and str in dodal/devices/focusing_mirror.py. 

The test was run via BlueAPI but ought to be reproducable using just a command line Bluesky/Ophyd-Async environment

## Acceptance Criteria
* This error should not be produced when the count plan is run against either focusing mirror.

keithralphs:
Fixes #537

1. Run Tests
2. Alternatively pull change into scratch area of either p38 or i22 and restart BlueAPI, then Run count plan via BlueAPI on vfm or hfm device. No serialization error should be generated

This source of bugs in class declaration will be address as part of https://github.com/bluesky/ophyd-async/issues/310

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/focusing_mirror.py
+++ b/src/dodal/devices/focusing_mirror.py
@@ -19,7 +19,7 @@ VOLTAGE_POLLING_DELAY_S = 0.5
 DEFAULT_SETTLE_TIME_S = 60
 
 
-class MirrorType(Enum):
+class MirrorType(str, Enum):
     """See https://manual.nexusformat.org/classes/base_classes/NXmirror.html"""
 
     SINGLE = "single"


callumforrester:
Test failures are due to incompatibility with the latest ophyd-async, @DominicOram working on it

callumforrester (approved):
Approved, once ophyd-async issues are fixed, can rebase and the CI should pass to allow merge

PR #560, opened 2024-05-20
Add ability to enable dev shm streaming on eiger

DominicOram:
See https://github.com/DiamondLightSource/hyperion/issues/1372

1. Confirm tests pass

### Checks for reviewer


src/dodal/devices/eiger.py
```diff
@@ -144,6 +144,9 @@ def stop(self, *args):
         self.disarm_detector()
         stop_status &= self.disable_roi_mode()
         stop_status.wait(self.GENERAL_STATUS_TIMEOUT)
+        # See https://github.com/DiamondLightSource/hyperion/issues/1395
+        LOGGER.info("Turning off Eiger dev/shm streaming")
```
olliesilvester, line 148: Nit: If we have `enable_dev_shm` set to off, I think we will get a logging message during `change_dev_shm`, and in Hyperion when XRC finishes, and again here - all saying that dev shm is being disabled

olliesilvester (approved):
Minor nit, but looks good

src/dodal/devices/eiger.py
DominicOram: I think that given the consequence of leaving it on accidentally is very bad it's actually good to be very specific in the logs that we're turning it off even if that doesn't actually matter

olliesilvester: yeah, makes sense

PR #563, opened 2024-05-21
Implement describe for fswitch

### LINKED ISSUE #536: Running Count plan against i22 F-Switch gives KeyError
In testing on i22 on 14/05/24, when the Count plan was run via BlueAPI against the F-Switch device the the following error was reported:

KeyError: number_of_lenses

presumably indicating that there was no corrensponding entry in the device attributes for number_of_lenses

The test was run via BlueAPI but ought to be reproducable using just a command line Bluesky/Ophyd-Async environment 

## Acceptance Criteria
* Thte count plan should be runnable aginst the device without producing the error

joeshannon:
This must be implemented to allow the device to be used in plans.

Add a simple test to ensure the count plan runs successfully.

Fixes #536.

1. Run blueapi with i22 devices loaded
2. Run count plan with fswitch and verify that the plan completes without error.

### Checks for reviewer

### DIFF
--- a/tests/devices/i22/test_fswitch.py
+++ b/tests/devices/i22/test_fswitch.py
@@ -1,6 +1,9 @@
 from unittest import mock
 
 import pytest
+from bluesky.plans import count
+from bluesky.protocols import DataKey
+from bluesky.run_engine import RunEngine
 from ophyd_async.core import DeviceCollector, set_mock_value
 
 from dodal.devices.i22.fswitch import FilterState, FSwitch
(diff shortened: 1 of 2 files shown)


tests/devices/i22/test_fswitch.py
```diff
@@ -26,3 +28,7 @@ async def test_reading_fswitch(fswitch: FSwitch):
             "value": 125,  # three filters out
         }
     }
+
+
+def test_fswitch_count_plan(RE: RunEngine, fswitch: FSwitch):
+    RE(count([fswitch]))
```
DiamondJoseph, line 34: Must: Check something about the value of the documens

DiamondJoseph (changes requested):

tests/devices/i22/test_fswitch.py
DiamondJoseph: Here's a way it's been done in ophyd-async, I know there's a ticket either there, in dodal or blueapi to generalise
https://github.com/bluesky/ophyd-async/blob/main/tests/test_flyer_with_panda.py#L137

callumforrester (changes requested):
Could this be done with soft signals, similar to the Undulator?
https://github.com/DiamondLightSource/dodal/pull/511

DiamondJoseph (approved):

PR #567, opened 2024-05-21
Organise testing utils

### LINKED ISSUE #564: Organise testing fixtures and utilities so that they can be made use of by application code
- mock devices should be provided next to real devices
- there should be a `dodal.testing_utils` for generic stuff
- mock devices could be rexported from `dodal.testing_utils` as fixtures

dperl-dls:
Fixes #564

Moves many mock device test fixtures and utility functions to a central place under `src/`, so that they can easily be reused by application code

1. See that no tests are broken

### Checks for reviewer


src/dodal/testing_utils/device_fixtures.py
```diff
@@ -0,0 +1,215 @@
...
+        TEST_APERTURE_POSITIONS
+    )
+    return aperture_positions
+
+
+@pytest.fixture
+async def mock_aperturescatterguard_with_call_log():
+    call_log = MagicMock()
+    ap_sg = await get_mock_ap_sg()
+    with (
```
DominicOram, line 53: Should: `patch_ophyd_async_motor` isn't a context manager

src/dodal/testing_utils/device_fixtures.py
```diff
@@ -0,0 +1,215 @@
...
+        patch_ophyd_async_motor(ap_sg.aperture.x, call_log=call_log),
+        patch_ophyd_async_motor(ap_sg.aperture.y, call_log=call_log),
+        patch_ophyd_async_motor(ap_sg.aperture.z, call_log=call_log),
+        patch_ophyd_async_motor(ap_sg.scatterguard.x, call_log=call_log),
+        patch_ophyd_async_motor(ap_sg.scatterguard.y, call_log=call_log),
```
DominicOram, line 58: Should: Can we do this patching in `get_mock_ap_sg`?

src/dodal/testing_utils/device_fixtures.py
```diff
@@ -0,0 +1,215 @@
...
+    return make_fake_device(DualBacklight)(name=f"backlight: {request.node.name}")
+
+
+@pytest.fixture
+def mock_eiger(request: pytest.FixtureRequest):
+    yield get_mock_eiger(create_new_detector_params(), request.node.name)
+
+
+@pytest.fixture
+def mock_eiger_and_stage(mock_eiger):
```
DominicOram, line 111: Nit: Rename to something like `mock_eiger_with_noop_stage`?

DominicOram (changes requested):
Thank you! I'm not convinced about the structure as whole yet. Could you make the changes in hyperion so we have an example of how this improves code in a downstream app? Then I think maybe it needs a brief in person discussion

src/dodal/testing_utils/device_fixtures.py
dperl-dls: yes it is, it's type-hinted wrong in ophyd-async, fix has been merged but not released, doesn't affect functionality

src/dodal/testing_utils/device_fixtures.py
dperl-dls: no, it leads to circular import errors if we import test utils to device classes and then device classes to test utils

dperl-dls:
never mind, this is too hard, I now think it's better to just copy and paste stuff everywhere

PR #568, opened 2024-05-21
Add configuration signals for i22 fswitch

### LINKED ISSUE #566: Add configuration signals for i22 fswitch
To match the Nexus file requirements, further fields should be available for the fswtich. These are effectively static metadata and can be implemented using `ConfigSignal`.

The required data is:

* `lens_geometry` (NXchar) - `paraboloid`
* `cylindrical` (NXboolean) - `1`
* `lens_material` (NXchar) - `Beryllium`

These are in addition to `number_of_lenses` (NXint) which is already implemented

## Acceptance Criteria
* Signals added as children to i22's fswitch device
* Existing test updated to confirm these fields are present

joeshannon:
These configuration/metadata fields are required to be in the Nexus file for i22's fswitch.

Fixes #566.

Fixes #ISSUE

1. Run fswitch count plan
2. Verify that added configuration fields are in the descriptor document

### Checks for reviewer


tests/devices/i22/test_fswitch.py
```diff
@@ -49,5 +54,11 @@ def subscription(name, doc):
     )
     assert descriptor_doc["data_keys"] == {"number_of_lenses": expected_data_key}
 
+    assert descriptor_doc["configuration"]["fswitch"]["data"] == {
```
joeshannon, line 57: Should I be using things like `assert_configuration` here or is this test simple enough to do it like this?

callumforrester: Maybe leave it like this, but raise an issue with ophyd-async for making partial `assert_reading` and `assert_configuration`s easier?

joeshannon: Done - https://github.com/bluesky/ophyd-async/issues/324

joeshannon:
> Do the definitions in `i22.py` and `p38.py` also need updating?

Yes just pushed a new revision containing the change for i22

I didn't add it to p38 as I assume they don't have an fswitch.

callumforrester:
Could you add a fake one for p38? With `fake_with_ophyd_sim=True`. That way we can generate documents from p38 that contain all the information needed to make the nexus structure we want, even if it's mostly 0s etc in p38's case

joeshannon:
> Could you add a fake one for p38? With `fake_with_ophyd_sim=True`. That way we can generate documents from p38 that contain all the information needed to make the nexus structure we want, even if it's mostly 0s etc in p38's case

Thanks - done now.

DiamondJoseph (approved):

PR #577, opened 2024-05-28
Add linkam3

abbiemery:
Adds the Linkam3 device used for i22.


callumforrester (dismissed):
Needs tests and should be added to i22 and p38 modules

DiamondJoseph (approved):

PR #588, opened 2024-05-30
(hyperion#1307) change DetectorParams detector_size_constants validator to allow the declared type

rtuck99:
Change to address PR comment in DiamondLightSource/hyperion#1307

Tests pass
### Checks for reviewer


DominicOram (approved):
Great, thanks. No one other than i03 is using this so we don't necessarily need to keep both behaviours, I think it's not too much overhead to keep for now though

PR #591, opened 2024-06-03
Add soft signals with required metadata for Synchrotron and TetrAMM devices

DiamondJoseph:
Fixes issues with required fields for NeXus writing for i22.

1. Smoke test i22, p38 to ensure changes have not broken those beamlines
2. Confirm that Hyperion is only using the SynchrotronMode signal from the Synchrotron and that still propagates as required

### Checks for reviewer


DiamondJoseph:
Tagging @dperl-dls  and @DominicOram: in Hyperion I could only find use of the synchrotron-mode signal, which remains propagated in the ConfigSignal: I'd like to take a better look at making a proper `mode` signal later on. I've left the top_up signals (renamed for consistency) for later use with a Suspender (I'm making an issue for one right now)

src/dodal/devices/tetramm.py
```diff
@@ -214,11 +215,20 @@ def __init__(
...
         self.hdf = NDFileHDF(prefix + "HDF5:")
         controller = TetrammController(self.drv)
+        config_signals = [
+            self.drv.values_per_reading,
+            self.drv.averaging_time,
+            self.drv.sample_time,
+        ]
+        if type:
+            type_signal, _ = soft_signal_r_and_setter(str, type)
+            config_signals.append(type_signal)
```
callumforrester, line 231: Following convention elsewhere in dodal:

```suggestion
        if type is not None:
            self.type, _ = soft_signal_r_and_setter(str, type)
        else:
            self.type = None
        config_signals = [
            self.drv.values_per_reading,
            self.drv.averaging_time,
            self.drv.sample_time,
        ]
```

src/dodal/beamlines/i22.py
callumforrester, line 1: Does p38 need similar changes?

callumforrester (changes requested):

src/dodal/beamlines/i22.py
DiamondJoseph: Not need, but I'll go down there Tomorrow and try and ascertain what type is correct.

src/dodal/devices/tetramm.py
DiamondJoseph: That doesn't get it into the config_signals list without then doing the same "if type:" again.

DominicOram:
> I've left the top_up signals (renamed for consistency) for later use with a Suspender (I'm making an issue for one right now)

The topup signals are also used in the wait for topup plans (see failing tests). As such `hyperion` also mocks out `topup_start_countdown` for some of its tests. It could be argued that `hyperion` should be actually mocking the `top_up` plans though. Either way, I'm happy to make the required hyperion changes if you can fix the top up plans

src/dodal/devices/tetramm.py
callumforrester: Good point, you'll need to append to config signals

PR #594, opened 2024-06-04
Update the location off the OAV xml file for I24

noemifrisina:
Following the update to GDA 9.34, the location of the xml files for I24 has changed.


dperl-dls (approved):
LGTM!

PR #629, opened 2024-06-17
Add logic to stop waiting on pin if robot raises pin not found error

DominicOram:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1446 and fixes https://github.com/DiamondLightSource/hyperion/issues/1450

1. Confirm tests pass

### Checks for reviewer


DominicOram:
@coretl and @callumforrester, I'm particularly interested in your thoughts on `pin_mounted_or_no_pin_found`. Is this the best way to solve the problem? Are there helper functions we can put into `ophyd-async` for this? Maybe an `observe_multiple_values` for getting all the updates from multiple signals?

coretl:
I think you've done this in the most concise way at present, but I think you're right about adding a helper to ease this, maybe something like:
```python
    async def pin_mounted_or_no_pin_found(self):
        async for signal, value in observe_signals_value(self.error_code, self.goinio_pin_sensor):
            if signal is self.error_code and value == self.NO_PIN_ERROR_CODE:
                # We didn't get a pin
                raise RobotLoadFailed(self.NO_PIN_ERROR_CODE, "Pin was not detected")
            elif signal is self.goinio_pin_sensor and value == PinMounted.PIN_MOUNTED:
                # We did get a pin
                break
```

DominicOram:
Thanks @coretl, that's exactly the sort of thing I was thinking. I've made https://github.com/bluesky/ophyd-async/issues/402

tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -19,6 +19,24 @@ async def test_bart_robot_can_be_connected_in_sim_mode():
...
+        await sleep(1)
+
+    device._load_pin_and_puck = AsyncMock(side_effect=_sleep)
+
+    set_mock_value(device.error_code, (expected_error_code := 10))
+    set_mock_value(device.error_str, (expected_error_string := "BAD"))
+
+    with pytest.raises(RobotLoadFailed) as e:
+        await device.set(SampleLocation(0, 0))
+        assert e.value.error_code == expected_error_code
```
rtuck99, line 35: These asserts are never called

tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -80,3 +98,22 @@ async def test_given_program_not_running_and_pin_unmounts_then_mounts_when_load_
...
+
+async def test_given_waiting_for_pin_to_mount_when_no_pin_mounted_then_error_raised():
+    device = await _get_bart_robot()
+    status = device.pin_mounted_or_no_pin_found()
+    await sleep(0.01)
+    set_mock_value(device.error_code, 25)
+    await sleep(0.01)
+    with pytest.raises(RobotLoadFailed):
+        await status
+
```
rtuck99, line 111: I don't understand what these sleeps are for - you don't enter the coroutine until `await` is called on it.

rtuck99 (approved):
Looks good now

PR #632, opened 2024-06-18
Add function to save a thumbnail

### LINKED ISSUE #628: Save thumbnail images for snapshots
Currently, for synchweb to display snapshots for OAV images there needs to be a thumbnail image saved next to the full image that is scaled to 192 pixels high and is saved as `xt.png` where `x` is the name of the full image.

Note that this is needed for hyperion robot load testing but after https://jira.diamond.ac.uk/browse/LIMS-1170 (which will be released in a couple of weeks) it will no longer be needed. Therefore it's fine to just put a temporary solution in until that point

## Acceptance Criteria
* Whenever a snapshot is taken a thumbnail is saved with a height of 192 and a name with `t` appended to it

DominicOram:
Fixes #628

Note that this will only be needed for a month or so, no need to review too heavily.

1. Confirm new test passes

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/areadetector/plugins/MJPG.py
+++ b/src/dodal/devices/areadetector/plugins/MJPG.py
@@ -8,6 +8,7 @@ from ophyd import Component, Device, DeviceStatus, EpicsSignal, EpicsSignalRO, S
 from PIL import Image, ImageDraw
 
 from dodal.devices.oav.oav_parameters import OAVConfigParams
+from dodal.devices.oav.utils import save_thumbnail
 from dodal.log import LOGGER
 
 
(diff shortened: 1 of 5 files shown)


src/dodal/devices/areadetector/plugins/MJPG.py
```diff
@@ -49,6 +50,9 @@ def _save_image(self, image: Image.Image):
 
         LOGGER.info(f"Saving image to {path}")
         image.save(path)
+
+        save_thumbnail(Path(path), image)
+
```
rtuck99, line 55: PIL documentation states that this modifies the image, but looks like we are ok in all our usages.

DominicOram: Yh, I thought about doing the copy but didn't seem worth it as we're always calling after saving the original. Thanks for checking though.

rtuck99 (approved):
Looks good

PR #636, opened 2024-06-18
Convert mirror voltage devices to use ophyd async

### LINKED ISSUE #604: Ophyd-async device for mirror voltages
We should convert the mirror voltage devices into `ophyd_async`.

## Acceptance Criteria
* The mirror voltage devices are in ophyd_async
* Tests pass in dodal and in hyperion

rtuck99:
This converts the VFMMirrorVoltages device to use ophyd-async
Fixes #604 

See also corresponding hyperion change TBD

1. Tests pass
2. mirror voltage devices connect and experiment plans continue to work

### Checks for reviewer


DominicOram (commented):
Mainly

tests/devices/unit_tests/test_focusing_mirror.py
```diff
@@ -12,121 +16,194 @@
...
-        vfm_mirror_voltages._channel14_voltage_device._demand_accepted.sim_put(
-            MirrorVoltageDemand.SLEW
+        set_mock_value(
+            vfm_mirror_voltages.voltage_channels[0]._demand_accepted,
+            MirrorVoltageDemand.SLEW,
         )
         return DEFAULT
 
-    vfm_mirror_voltages._channel14_voltage_device._setpoint_v.set = MagicMock(
+    vfm_mirror_voltages.voltage_channels[0]._setpoint_v.set = AsyncMock(
```
DominicOram, line 85: Should: So `set` is already backed by a mock. you can do:
```python
from ophyd_async.core import get_mock_put
mock = get_mock_put(vfm_mirror_voltages.voltage_channels[0]._setpoint_v)
```
`mock` will already be a mock that you can play with and it should fix your type hinting later on

DominicOram (changes requested):
Thank you, I think this is a lot cleaner. Some comments in code

src/dodal/devices/focusing_mirror.py
```diff
@@ -32,93 +36,84 @@ class MirrorStripe(str, Enum):
...
 
-                demand_accepted_status.set_finished()
-            # else timeout handled by parent demand_accepted_status
+        accepted_value = await anext(it_values)
+        assert accepted_value == MirrorVoltageDemand.SLEW
+        LOGGER.debug(
+            f"Demand not accepted for {setpoint_v.name}, waiting for acceptance..."
+        )
+        while MirrorVoltageDemand.SLEW == (accepted_value := await anext(it_values)):
+            pass
```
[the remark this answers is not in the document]
coretl: It passes the `camonitor` updates straight through, so as long as EPICS isn't squashing the updates (via MDEL) then you will get multiple updates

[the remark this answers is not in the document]
DominicOram: Hmm... Then we will need to always check it in a loop

src/dodal/devices/focusing_mirror.py
```diff
@@ -76,21 +76,26 @@ async def set(self, value, *args, **kwargs):
...
+            demand_accepted, timeout=DEFAULT_SETTLE_TIME_S
+        )
+        # discard the current value (OK) so we can await a subsequent change
+        await anext(demand_accepted_iterator)
+        await setpoint_v.set(value)
+
+        # The set should always change to SLEW regardless of whether we are
+        # already at the set point, then change back to OK/FAIL depending on
+        # success
+        accepted_value = await anext(demand_accepted_iterator)
```
DominicOram, line 91: Should: Sorry, I'm wondering if there is another race condition here where if we get an update straight after the set we may still see an OK? I think a better solution might be not to discard the value before the set but to keep reading until the updates show `slew`?

tests/devices/unit_tests/test_focusing_mirror.py
```diff
@@ -45,27 +56,36 @@ def vfm_mirror_voltages_with_set_accepted_fail(
 
 
 def vfm_mirror_voltages_with_set_to_value(
-    vfm_mirror_voltages, new_value: MirrorVoltageDemand
+    vfm_mirror_voltages, new_value: MirrorVoltageDemand, spins: int = 0
```
DominicOram, line 59: Could: `spins` isn't that obvious a name. Maybe `number_of_slew_updates`?

DominicOram (changes requested):
Thank you! Apologies, I think there is still a potential race, maybe I'm getting paranoid now but I feel like if this is the first time we do something like this it will be used as a model for other devices so we should be as accurate as possible.

src/dodal/devices/focusing_mirror.py
rtuck99: I did look at the code in ophyd_async before to confirm that we will always get the current value in the first event, so I don't think that there is any possibility of this race condition and we should be safe.

`observe_value`'s implementation calls `signal.subscribe_value(q.put_nowait)`
before awaiting new values.
`signal.subscribe()` calls `signal_notify()` which calls our `q.put_nowait()` - this posts the event on to the queue which is then collected by the subsequent `await`.

If this implementation ever changes it will probably break calling code as anything that follows this model under this assumption will end up blocking waiting for an event that never comes.

I'm also slightly confused where you think the extra event will come from - we've already validated that the demand will be `OK` prior to entering the function and we should be the only ones applying changes. The discard from the first `anext()` is before the set, so there is no way that the discarded value can be "SLEW" or a subsequent "OK", even if the current value wasn't pushed by the call to `observe`.

DominicOram: Tom's comment https://github.com/DiamondLightSource/dodal/pull/636#discussion_r1649079687 implies that it is possible we get multiple updates with the same value. It depends quite a bit on how the IOC is written, for example it's possible that the PV will "send out" an update on a periodic basis regardless of if the value has changed or not. I think you're right that I'm being overly cautious though, looking at the PV it seems that it should be stable.

DominicOram (approved):
Great, thank you! Sorry for the back and forth, took a while to get my head round all the potential races.

PR #648, opened 2024-06-26
600 Move docs off the wiki

### LINKED ISSUE #600: Move docs off the wiki
There are currently a number of docs that are on the `dodal` GH wiki which should be in the built docs instead (https://diamondlightsource.github.io/dodal/main/index.html). We should move everything off of the wiki

## Acceptance Criteria
* All docs are moved off the GH wiki into https://diamondlightsource.github.io/dodal/main/index.html
* Links are corrected

rtuck99:
Fixes #600 

This moves all pages on the dodal wiki onto the sphinx-generated restructured text documentation.
It also generates API documentation for dodal.

1. View the docs on https://diamondlightsource.github.io/dodal/600_move_docs_off_the_wiki/index.html
The API documentation can be viewed at 
https://diamondlightsource.github.io/dodal/600_move_docs_off_the_wiki/user/reference/api.html

(Accessible from the main landing page -> User Guide -> Reference -> API)

Feedback on the documentation welcome. Some of the wiki pages have been merged with pre-existing pages.
The API documentation deliberately only documents the following modules (and submodules):

* dodal.beamlines
* dodal.common
* dodal.devices
* dodal.plans

but this can be changed if we feel more/less should be documented.

Private members are not documented.
I have tried to generate at least summary documentation for members with no docstring as there are so many, and to deduce types from typehints, although this does lead to some strange formatting sometimes.

There are still quite a few quirks in the documentation, like re-exported functions in modules being listed in the module without a summary table despite not being defined there.

I have disabled the -W option on sphinx as it generates a frankly ridiculous number of warnings building the API, and it is difficult to determine how they originate or how to disable them. I feel this is sensible on the basis that I would rather have up-to-date documentation even if it's not always formatted correctly, than noting at all.


docs/developer/reference/device-standards.rst
```diff
@@ -0,0 +1,87 @@
...
+is preferred over
+
+.. code-block:: python
+
+    class MyDevice(Device):
+        bragg: EpicsMotor = Cpt(EpicsMotor, "-MO-DCM-01:BRAGG")
+
+    device_instantiation(DCM, "dcm", "")
+
+This is so that a new device on say ``-MO-DCM-02`` can be easily created.
```
noemifrisina, line 43: Nit(s). 
- Since we're saying we should write new devices in `ophyd_async` maybe we could update the examples to reflect that?

- I think it would also be useful to add a couple of lines on how to deal with PVs that do not use a standard beamline suffix, in which case `device_instantation` should set `bl_prefix=False`.
It's not super common, but something I run into in the past and it would be nice to have a small example for the docs.

noemifrisina (changes requested):
Thanks, docs look much tidier without all the extra inherited docstrings.

A few comments in code, some admittedly a bit nitpicky...

noemifrisina (approved):
lgtm, thanks

PR #653, opened 2024-06-28
Create util for caching positions before a plan and returning to them afterwards

DominicOram:
Fixes https://github.com/DiamondLightSource/hyperion/issues/1439

1. Confirm unit tests pass

### Checks for reviewer


src/dodal/plans/motor_util_plans.py
```diff
@@ -0,0 +1,100 @@
...
+            raise MoveTooLarge(axis, maximum_move, position)
+        if abs(position - home_position) > smallest_move:
+            positions[axis] = position
+    return positions
+
+
+def home_and_reset_wrapper(
+    plan: MsgGenerator,
+    device: Device,
+    home_position: float,
```
dperl-dls, line 47: are we likely to ever have this not be 0, and if it is ever not 0 isn't it then likely to be different for every motor?

how would you feel about this instead taking a list of motors, and a list of homes or default to 0 for everything?

dperl-dls (changes requested):
This looks very useful but I'm not 100% convinced about the API, I think with a minor tweak which in some ways is simpler it could be more useful still

dperl-dls (approved):
Great, thanks!

PR #659, opened 2024-07-05
Consider filename prefix when incrementing run number

### LINKED ISSUE #643: get_run_number() should consider the filename as well as the run_number
If there are, eg, multiple different samples in one directory, `get_run_number()` still just increases the number at the end of the filename. It should increment the number for the same sample being measured instead.

## Acceptance Criteria
* `get_run_number()` takes a filename argument and only increases the suffix on matches
* `DetectorParams` uses this when setting default run number

dperl-dls:
Fixes #643 
By default, make `DetectorParams` only increment the number for the filename prefix.

1. Run tests
### Checks for reviewer


DominicOram (approved):
Great, thank you!

PR #661, opened 2024-07-08
Add a run_program to the pmac

### LINKED ISSUE #650: PMAC: Have a run program that we can wait for
In I24 SSX we start the PMAC program by setting `bps.abs_set(pmac.pmac_string, f"&2b{prog_num}r", wait=True)` in the plan then later doing a loop waiting for `yield from bps.rd(pmac.scanstatus) == 0`. It would be cleaner if we could do this in the device. i.e. something like:

```python
bps.abs_set(pmac.run_program, prog_num, wait=True) # This will only return when scanstatus becomes 0
```

## Acceptance Criteria
* We can easily run and wait on a pmac program as above

noemifrisina:
Fixes #650 

Add a `run_program` to the PMAC device for I24, that sets the `pmac_string` to the program number and then waits for the collection to be done by checking the `scanstatus` PV value. It only returns when this PV returns 0.

1. Run tests

### Checks for reviewer


dperl-dls (approved):
lgtm

src/dodal/devices/i24/pmac.py
```diff
@@ -95,6 +100,34 @@ async def set(self, enc_string: EncReset):
...
+        self.status = status_sig
+        super().__init__(backend, timeout, name)
+
+    @AsyncStatus.wrap
+    async def set(self, prog_num: int):
+        prog_str = f"&2b{prog_num}r"
+        await self.signal.set(prog_str, wait=True)
+        async for reading in observe_value(self.status):
+            if reading == ScanState.DONE:
+                return
```
DominicOram, line 128: Could: Can you not just use `wait_for_value` here?

PR #664, opened 2024-07-09
Change type of robot error code PV

DominicOram:
Part of https://github.com/DiamondLightSource/hyperion/issues/1465

1. Run `dodal connect i03`
2. Confirm robot is now happy

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/robot.py
+++ b/src/dodal/devices/robot.py
@@ -60,7 +60,8 @@ class BartRobot(StandardReadable, Movable):
         self.program_running = epics_signal_r(bool, prefix + "PROGRAM_RUNNING")
         self.program_name = epics_signal_r(str, prefix + "PROGRAM_NAME")
         self.error_str = epics_signal_r(str, prefix + "PRG_ERR_MSG")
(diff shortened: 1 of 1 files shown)


olliesilvester (approved):
Looks good, but I've noticed we have a few other places where we create epics signals with type int:

in `fast_grid_scan` `smargon`, `tetramm`, `epics_util`, `xspress3`, `xspress3_channel`

If we'd get a same issue for all of these, maybe we should change the rest of the devices too?

DominicOram:
> Looks good, but I've noticed we have a few other places where we create epics signals with type int:
> 
> in `fast_grid_scan` `smargon`, `tetramm`, `epics_util`, `xspress3`, `xspress3_channel`
> 
> If we'd get a same issue for all of these, maybe we should change the rest of the devices too?

For most of these they're not an `ao` record in EPICS, they're either longs or binary. You could argue that we shouldn't be using ints at all for the binary ones I guess but the longs are fine. Basically, if the device connects then it's fine. All these devices connect when we do `dodal connect i03` except the tetramm, which seems fine when I do `dodal connect i22`.

PR #667, opened 2024-07-10
378 Convert source tree to python-copier-template

### LINKED ISSUE #378: Move to copier

rtuck99:
This moves the project to the `python-copier-template` project style

Fixes #378

### Documentation

* Documentation from the template has been changed from .rst Restructured Text to .md Markdown style, and the folder hierarchy has been flattened to remove the separation between `user` and `developer` sections. This is probably a good move for dodal as the distinction between users and developers is somewhat blurred anyway. I have attempted to retain customisation of the documentation where appropriate.
* The API documentation should still work

### Development environment

* Building standalone containers is no longer supported and relevant documentation removed, but there is now support for devcontainers (but I haven't tried it yet)

### CI

* Github workflows are now refactored out into reusable workflows prefixed by an underscore `_` called from the main workflows.
* Warnings in Pytest are now errors and this has resulted in a number of fixes to both production and test code:
   * Where async tasks were not `await`ed and outlived the lifespan of the test
   * Where file resources were not closed after being opened (notably in various logging tests)
   * Where deprecated methods were being called
   * Some warnings generated by code in our dependencies have been suppressed in `pyproject.toml`
* There have been some changes to linter flags in `pyproject.toml` which add various additional checks

### Dependencies

* Nb `python-copier-template` has a dependency on `pydantic` >= 2.0 which conflicts with the rest of dodal dependencies so cannot be included in the list of development dependencies; you have to uninstall it after running it

### Developer Portal

* When merged, the project should appear on the Developer Portal Software Catalog
https://dev-portal.diamond.ac.uk

### TODO 

* Pypi uploads on release (The option appeared partway through the upgrade and I left at the default)

## Checks for reviewer


tests/test_cli.py
```diff
@@ -0,0 +1,12 @@
+import subprocess
+import sys
+
+import pytest
+
+from dodal import __version__
+
+
+@pytest.mark.skip_in_pycharm(reason="subprocess returns tty escape sequences")
```
dperl-dls, line 9: :(

dperl-dls (approved):
This all looks fine, probably too complicated to inspect every individual line - since warnings are now errors I can only assume things are on average better than they were

PR #669, opened 2024-07-11
Fix a number of pyright typing errors

DominicOram:
1. Run `pyright` on here and `main` confirm less errors on here
2. Confirm tests still pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/oav/oav_detector.py
+++ b/src/dodal/devices/oav/oav_detector.py
@@ -46,6 +46,7 @@ class ZoomController(Device):
     sxst = Component(EpicsSignal, "MP:SELECT.SXST")
 
     def set_flatfield_on_zoom_level_one(self, value):
+        self.parent: "OAV"
         flat_applied = self.parent.proc.port_name.get()
         no_flat_applied = self.parent.cam.port_name.get()
         return self.parent.grid_snapshot.input_plugin.set(
--- a/src/dodal/devices/util/epics_util.py
+++ b/src/dodal/devices/util/epics_util.py
@@ -1,5 +1,5 @@
 from functools import partial
-from typing import Callable
+from typing import Callable, Sequence
 
 from bluesky.protocols import Movable
 from ophyd import Component, EpicsSignal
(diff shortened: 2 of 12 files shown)


src/dodal/devices/util/epics_util.py
```diff
@@ -26,7 +26,7 @@ def epics_signal_put_wait(pv_name: str, wait: float = 3.0) -> Component[EpicsSig
 
 
 def run_functions_without_blocking(
-    functions_to_chain: list[Callable[[], StatusBase]],
+    functions_to_chain: Sequence[Callable[[], StatusBase]],
```
rtuck99, line 29: :+1: We probably use `list` in far too many places where we could use Sequence

tests/devices/unit_tests/test_zebra.py
```diff
@@ -1,8 +1,7 @@
-from unittest.mock import AsyncMock
+from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 from bluesky.run_engine import RunEngine
-from mockito import mock, verify
```
rtuck99, line 5: How many places do we still use mockito? Maybe we can get rid of it as a dependency. I like fewer dependencies.

rtuck99 (approved):
approved

tests/devices/unit_tests/test_zebra.py
DominicOram: Yh, I agree, thought about doing it as part of this but ran out of time. Made https://github.com/DiamondLightSource/dodal/issues/670

src/dodal/devices/oav/oav_detector.py
```diff
@@ -46,6 +46,7 @@ class ZoomController(Device):
     sxst = Component(EpicsSignal, "MP:SELECT.SXST")
 
     def set_flatfield_on_zoom_level_one(self, value):
+        self.parent: "OAV"
```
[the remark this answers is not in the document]
DominicOram: Yh, it's not great. Potentially a sign that the structure here is bad, there's a kind of weird relationship between this device and it's parent which I don't really like. Given it's an `ophyd` device I didn't dwell on it too hard though as it'll be hopefully moved to `ophyd-async` soon anyway

PR #672, opened 2024-07-15
Add a simple version to fill the unique needs of slit4

### LINKED ISSUE #587: Create device for i22 slits_4
i22's slits_4 has different PVs than its other slits, a representation of this device is required along with the existing slits.

## Acceptance Criteria

stan-dot:
fixes  #587

### DIFF
--- a/src/dodal/devices/slits.py
+++ b/src/dodal/devices/slits.py
@@ -24,8 +24,8 @@ class SpecialSlitsi22(StandardReadable):
 
     def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
-            self.x = Motor(prefix + "X")
-            self.y = Motor(prefix + "Y")
-            self.horizontal_dso = Motor(prefix + "HDSO")
-            self.vertical_dso = Motor(prefix + "VDSO")
+            self.x = Motor(prefix + "X:RBV")
+            self.y = Motor(prefix + "Y:RBV")
+            self.horizontal_dso = Motor(prefix + "HDSO:RBV")
+            self.vertical_dso = Motor(prefix + "VDSO:RBV")
         super().__init__(name)
--- a/src/dodal/devices/slits.py
+++ b/src/dodal/devices/slits.py
@@ -21,12 +21,13 @@ class SlitWithDefiningSlitOpening(StandardReadable):
     """
     these are unlikely to be moved during a scan
     Only HDSO and VDSO are recorded in the scan
+    DSO means Defining Slit Opening
     """
 
     def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
             self.x = Motor(prefix + "X")
             self.y = Motor(prefix + "Y")
-            self.horizontal_defining_slit_opening = Motor(prefix + "HDSO")
-            self.vertical_definiting_slit_opening = Motor(prefix + "VDSO")
+            self.x_gap = Motor(prefix + "HDSO")
+            self.y_gap = Motor(prefix + "VDSO")
         super().__init__(name)
--- a/src/dodal/devices/slits.py
+++ b/src/dodal/devices/slits.py
@@ -26,8 +26,8 @@ class SlitWithDefiningSlitOpening(StandardReadable):
 
     def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
-            self.x = Motor(prefix + "X")
-            self.y = Motor(prefix + "Y")
             self.x_gap = Motor(prefix + "HDSO")
             self.y_gap = Motor(prefix + "VDSO")
+            self.x_centre = Motor(prefix + "X")
+            self.y_centre = Motor(prefix + "Y")
         super().__init__(name)
--- a/src/dodal/devices/slits.py
+++ b/src/dodal/devices/slits.py
@@ -24,8 +24,8 @@ class SpecialSlitsi22(StandardReadable):
 
     def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
-            self.x = Motor(prefix + "X:RBV")
-            self.y = Motor(prefix + "Y:RBV")
-            self.horizontal_dso = Motor(prefix + "HDSO:RBV")
-            self.vertical_dso = Motor(prefix + "VDSO:RBV")
+            self.x = Motor(prefix + "X")
+            self.y = Motor(prefix + "Y")
+            self.horizontal_dso = Motor(prefix + "HDSO")
+            self.vertical_dso = Motor(prefix + "VDSO")
         super().__init__(name)
--- a/src/dodal/devices/slits.py
+++ b/src/dodal/devices/slits.py
@@ -19,13 +19,13 @@ class Slits(StandardReadable):
 
 class SpecialSlitsi22(StandardReadable):
     """
-    TODO uncertain
+    these are unlikely to be moved during a scan
     """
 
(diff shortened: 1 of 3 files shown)


src/dodal/devices/slits.py
```diff
@@ -15,3 +15,17 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
+    """
+    TODO uncertain
+    """
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        with self.add_children_as_readables():
+            self.x_plus = Motor(prefix + "X:PLUS")
+            self.x_minus = Motor(prefix + "Y:MINUS")
+            self.y_plus = Motor(prefix + "X:PLUS")
+            self.y_minus = Motor(prefix + "Y:MINUS")
```
DiamondJoseph, line 30: Copy Paste error

DiamondJoseph (changes requested):
Let's try it out on the beamline tomorrow, the signals are definitely currently wrong.

callumforrester:
For signal correctness, try running the system tests on the Diamond network: https://github.com/DiamondLightSource/dodal?tab=readme-ov-file#testing-connectivity

stan-dot:
will test on 30.07

tests/common/beamlines/test_beamline_utils.py
```diff
@@ -59,7 +59,7 @@ def test_instantiating_different_device_with_same_name():
         Smargon, "device", "", False, True, None
     )
     assert dev1.name == dev2.name
-    assert type(dev1) is not type(dev2)
+    assert not isinstance(type(dev2), type(dev1))
```
DiamondJoseph, line 62:
```suggestion
    assert not isinstance(dev2, type(dev1))
```
surely?

stan-dot:
not sure what is the desired behavior, which PVs do we want to expose?

the EDM screen shows:
X, Y, HDSO, VDSO, of the 2 latter each has a raw and 'real' variant.
There is an abscence of words known from other slits, such as x centre, or gap

@DiamondJoseph  please advise

DiamondJoseph:
Let's ask the beamline scientist tomorrow, but I assume that we want all of them.
Let's also keep them as Motors, same as the existing Slits, if they're not being moved that's fine but let's keep the option open

stan-dot:
still not right, need to talk to the scientists

stan-dot:
@DiamondJoseph this fits the Hyperion model of 'the changes fit our model of the beamline', and it passes the tests, can we merge this now?

src/dodal/devices/slits.py
```diff
@@ -15,3 +15,17 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
+
+
+class SpecialSlitsi22(StandardReadable):
+    """
+    these are unlikely to be moved during a scan
+    """
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        with self.add_children_as_readables():
+            self.x = Motor(prefix + "X:RBV")
```
DiamondJoseph, line 27: https://github.com/bluesky/ophyd-async/blob/8273a2e0980fc586db38abb15fc6d354e0d40e68/src/ophyd_async/epics/motor.py#L54

This looks wrong, or otherwise incompatible with the Motor class- can you instantiate this on the beamline?
Line 64 of the Motor definition adds an RBV signal, why did you add the RBV suffix here?

stan-dot: I forgot it's in the motor, it did not connect correctly. will investigate further

stan-dot:
dodal connect i22
Attempting connection to i22 (using dodal.beamlines.i22)
16 devices connected :
        oav
        fswitch
        slits_6
        slits_5
        slits_4
        slits_3
        slits_2
        slits_1
        undulator
        hfm
        vfm
        it
        i0
        waxs
        synchrotron
        saxs
        
        this was successful, ready to merge

src/dodal/devices/slits.py
```diff
@@ -15,3 +15,17 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
+    """
+    these are unlikely to be moved during a scan
+    """
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        with self.add_children_as_readables():
+            self.x = Motor(prefix + "X")
+            self.y = Motor(prefix + "Y")
+            self.horizontal_dso = Motor(prefix + "HDSO")
+            self.vertical_dso = Motor(prefix + "VDSO")
```
DominicOram, line 30: Should: As someone not involved in i22, it's not clear what `dso` means here. Could we have a comment or rename the variable to something more human readable

stan-dot: I do not either. that is how it was defined in EPICS. I did inquire on 07.08.2024 and received no response since.

DominicOram: Ok, well I don't feel comfortable us having code in `dodal` where no-one understands what it means. Could you please chase up? Have you asked Scientists+Controls?

stan-dot: I followed up again

stan-dot: fixed!

 @DominicOram

DominicOram: So they stand for `defining_slit_opening`? Can we just use `gap` as that's what we use everywhere else

DiamondJoseph: I'm also assuming that "defining slit opening" -> "slit width" -> "gap"?

DominicOram: Yes, I think they are all equivalent, we use `gap` everywhere else, it's succinct (unlike `defining_slit_opening`) and makes more sense for x and y (unlike `width` which sort of implies just `x`). So I think we should stick to gap at the ophyd level

DiamondJoseph:
```
class Slits(StandardReadable):
            self.x_gap = Motor(prefix + "X:SIZE")
            self.y_gap = Motor(prefix + "Y:SIZE")
            self.x_centre = Motor(prefix + "X:CENTRE")
            self.y_centre = Motor(prefix + "Y:CENTRE")
        super().__init__(name)


class SpecialSlits(StandardReadable):
            self.x = Motor(prefix + "X")
            self.y = Motor(prefix + "Y")
            self.x_gap = Motor(prefix + "HDSO")
            self.y_gap = Motor(prefix + "VDSO")
        super().__init__(name)

```

At which point we've got some very similar looking classes. It'd be nice if we could a. confirm that SpecialSlits.x is defining x centre, not e.g. x_min, and b. treat them all as some sort of generic slit with x_centre, x_gap, y_centre, y_gap

src/dodal/devices/slits.py
stan-dot: Ok I change it into `x_gap` etc, and moving the 'dso' into the class docstring

stan-dot:
yes, it is centre. we can make both classes have the same properties but they do need to have different logic.

@DominicOram and @DiamondJoseph, how would you like this to be implemented?
the ways I see
- (default) keep 2 classes 
- just 1 class with an override option for the channel names - that done either with a ternary inline logic or a big logic branch inside add_children_as_readables
- some interface that both classes implement that defines the same 4 properties

DiamondJoseph:
> 
>     * just 1 class with an override option for the channel names (in the constructor with the 5/6 case as the default and the 1/6 case as the special override)

would be my preference.

DominicOram:
My preference would be:
1. Create an alias at the EPICs layer so that the PVs are the same - we have done this for a set on i03
2. If we're short on time ask for the alias and do the override of channel names

stan-dot:
closing as this will be solved at the epics layer

PR #673, opened 2024-07-15
658 create device for pressure jump cell

### LINKED ISSUE #658: Ophyd-async device for Pressure Jump Cell
P38 and I22's [high pressure X-ray cell](https://confluence.diamond.ac.uk/display/I22/P-jump+Cell) that can be used to create pressure jumps is required for the November Bluesky tests.

## Acceptance Criteria
* Pressure control can be achieved
* Pressure jumps can be achieved
* All key device sensor values can be read
* Fast jump data can be collected

barnettwilliam:
closes #658
Ophyd device for pressure jump cell #658, required for pressure jump experiment.

Initial control PVs, TODO fast adc capture.

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/p38.py
+++ b/src/dodal/beamlines/p38.py
@@ -19,6 +19,7 @@ from dodal.devices.focusing_mirror import FocusingMirror
 from dodal.devices.i22.dcm import DoubleCrystalMonochromator
 from dodal.devices.i22.fswitch import FSwitch
 from dodal.devices.linkam3 import Linkam3
+from dodal.devices.pressure_jump_cell import PressureJumpCell
 from dodal.devices.slits import Slits
 from dodal.devices.tetramm import TetrammDetector
 from dodal.devices.undulator import Undulator
@@ -328,4 +329,17 @@ def ppump(
         "-EA-PUMP-01:",
         wait_for_connection,
         fake_with_ophyd_sim,
+    )
+
+def high_pressure_xray_cell(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> PressureJumpCell:
+    return device_instantiation(
+        PressureJumpCell,
+        "high_pressure_xray_cell",
+        "-EA-HPXC-01:",
+        wait_for_connection,
+        fake_with_ophyd_sim,
+        adc1_prefix=f"{BeamlinePrefix(BL).beamline_prefix}-EA-ADC-01:",
+        adc2_prefix=f"{BeamlinePrefix(BL).beamline_prefix}-EA-ADC-02:",
     )
--- a/src/dodal/beamlines/p38.py
+++ b/src/dodal/beamlines/p38.py
@@ -313,7 +313,7 @@ def linkam(
     return device_instantiation(
         Linkam3,
         "linkam",
-        "-EA-LINKM-02:",
+        f"{BeamlinePrefix(BL).insertion_prefix}-EA-LINKM-02:",
         wait_for_connection,
         fake_with_ophyd_sim,
     )
@@ -337,7 +337,7 @@ def high_pressure_xray_cell(
     return device_instantiation(
         PressureJumpCell,
         "high_pressure_xray_cell",
-        "-EA",
+        f"{BeamlinePrefix(BL).insertion_prefix}-EA",
         wait_for_connection,
         fake_with_ophyd_sim,
         cell_prefix="-HPXC-01:",
--- a/src/dodal/beamlines/p38.py
+++ b/src/dodal/beamlines/p38.py
@@ -337,9 +337,9 @@ def high_pressure_xray_cell(
     return device_instantiation(
         PressureJumpCell,
         "high_pressure_xray_cell",
-        "-EA-HPXC-01:",
+        "-EA",
         wait_for_connection,
         fake_with_ophyd_sim,
-        adc1_prefix=f"{BeamlinePrefix(BL).beamline_prefix}-EA-ADC-01:",
-        adc2_prefix=f"{BeamlinePrefix(BL).beamline_prefix}-EA-ADC-02:",
+        cell_prefix="-HPXC-01:",
+        adc_prefix="-ADC",
     )
--- a/src/dodal/beamlines/p38.py
+++ b/src/dodal/beamlines/p38.py
@@ -331,6 +331,7 @@ def ppump(
         fake_with_ophyd_sim,
     )
 
+
 def high_pressure_xray_cell(
     wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
 ) -> PressureJumpCell:
--- a/src/dodal/devices/pressure_jump_cell.py
+++ b/src/dodal/devices/pressure_jump_cell.py
@@ -216,10 +216,14 @@ class PressureJumpCell(StandardReadable):
         adc_prefix: str = "",
         name: str = "",
     ):
-        self.all_valves_control = AllValvesControl(f"{beamline_prefix}{cell_prefix}", name)
+        self.all_valves_control = AllValvesControl(
+            f"{beamline_prefix}{cell_prefix}", name
+        )
         self.pump = Pump(f"{beamline_prefix}{cell_prefix}", name)
 
-        self.controller = PressureJumpCellController(f"{beamline_prefix}{cell_prefix}", name)
+        self.controller = PressureJumpCellController(
+            f"{beamline_prefix}{cell_prefix}", name
+        )
 
         with self.add_children_as_readables():
             self.pressure_transducers: DeviceVector[PressureTransducer] = DeviceVector(
@@ -233,6 +237,8 @@ class PressureJumpCell(StandardReadable):
                 }
             )
 
-            self.cell_temperature = epics_signal_r(float, f"{beamline_prefix}{cell_prefix}TEMP")
+            self.cell_temperature = epics_signal_r(
+                float, f"{beamline_prefix}{cell_prefix}TEMP"
+            )
 
         super().__init__(name)
--- a/src/dodal/devices/pressure_jump_cell.py
+++ b/src/dodal/devices/pressure_jump_cell.py
@@ -253,15 +253,16 @@ class PressureJumpCell(StandardReadable):
         adc_prefix: str = "",
         name: str = "",
     ):
-        cellFullPrefix = prefix + cell_prefix 
+        cellFullPrefix = prefix + cell_prefix
         adcFullPrefix = prefix + adc_prefix
 
         self.valves = PressureJumpCellControlValves(cellFullPrefix, name)
         self.pump = PressureJumpCellPump(cellFullPrefix, name)
         self.transducers = PressureJumpCellPressureTransducers(
-            cellFullPrefix, name,  
-            adc1_prefix= adcFullPrefix + "-01:", 
-            adc2_prefix= adcFullPrefix + "-02:",
+            cellFullPrefix,
+            name,
+            adc1_prefix=adcFullPrefix + "-01:",
+            adc2_prefix=adcFullPrefix + "-02:",
         )
         self.controller = PressureJumpCellController(cellFullPrefix, name)
 
--- a/src/dodal/devices/pressure_jump_cell.py
+++ b/src/dodal/devices/pressure_jump_cell.py
@@ -1,9 +1,14 @@
+import asyncio
+from dataclasses import dataclass
 from enum import Enum
 
+from bluesky.protocols import HasName, Movable
 from ophyd_async.core import (
+    AsyncStatus,
     ConfigSignal,
     DeviceVector,
     SignalR,
+    SignalRW,
     StandardReadable,
 )
 from ophyd_async.epics.signal import epics_signal_r, epics_signal_rw
@@ -72,7 +77,7 @@ class FastValveState(str, Enum):
     OPEN_ARMED = "Open Armed"
     CLOSED = "Closed"
     CLOSED_ARMED = "Closed Armed"
-    NONE5 = "Unused"
+    NONE = "Unused"
 
 
 class LimitSwitchState(str, Enum):
@@ -80,7 +85,15 @@ class LimitSwitchState(str, Enum):
     ON = "On"
 
 
-class AllValvesControl(StandardReadable):
+@dataclass
+class AllValvesControlState:
+    valve_1: ValveControlRequest | None = None
+    valve_3: ValveControlRequest | None = None
+    valve_5: FastValveControlRequest | None = None
+    valve_6: FastValveControlRequest | None = None
+
+
+class AllValvesControl(StandardReadable, Movable):
     """
     valves 2, 4, 7, 8 are not controlled by the IOC,
     as they are under manual control.
@@ -100,64 +113,61 @@ class AllValvesControl(StandardReadable):
                 )
             )
 
-        with self.add_children_as_readables(ConfigSignal):
-            self.valves_open: DeviceVector[SignalR[int]] = DeviceVector(
-                {i: epics_signal_rw(int, f"{prefix}V{i}:OPENSEQ") for i in [1, 3, 5, 6]}
+        self.fast_valve_control: DeviceVector[SignalRW[FastValveControlRequest]] = (
+            DeviceVector(
+                {
+                    i: epics_signal_rw(FastValveControlRequest, f"{prefix}V{i}:CON")
+                    for i in [5, 6]
+                }
             )
+        )
 
-            self.valve_control: DeviceVector[SignalR[ValveControlRequest]] = (
-                DeviceVector(
-                    {
-                        i: epics_signal_rw(ValveControlRequest, f"{prefix}V{i}:CON")
-                        for i in [1, 3]
-                    }
-                )
-            )
-
-            self.fast_valve_control: DeviceVector[SignalR[FastValveControlRequest]] = (
-                DeviceVector(
-                    {
-                        i: epics_signal_rw(FastValveControlRequest, f"{prefix}V{i}:CON")
-                        for i in [5, 6]
-                    }
-                )
-            )
+        self.valve_control: DeviceVector[SignalRW[ValveControlRequest]] = DeviceVector(
+            {
+                i: epics_signal_rw(ValveControlRequest, f"{prefix}V{i}:CON")
+                for i in [1, 3]
+            }
+        )
 
         super().__init__(name)
+
+    async def set_valve(
+        self, valve: int, value: ValveControlRequest | FastValveControlRequest
+    ):
+        if valve in [1, 3] and isinstance(value, ValveControlRequest):
+            await self.valve_control[valve].set(value)
+        elif valve in [5, 6] and isinstance(value, FastValveControlRequest):
+            await self.fast_valve_control[valve].set(value)
+
+    @AsyncStatus.wrap
+    async def set(self, value: AllValvesControlState):
+        await asyncio.gather(
+            *(
+                self.set_valve(int(i[-1]), value)
+                for i, value in value.__dict__.items()
+                if value is not None
+            )
+        )
 
 
 class Pump(StandardReadable):
     def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
             self.pump_position = epics_signal_r(float, prefix + "POS")
-            self.pump_forward_limit = epics_signal_r(
-                LimitSwitchState, prefix + "D74IN1"
-            )
-            self.pump_backward_limit = epics_signal_r(
-                LimitSwitchState, prefix + "D74IN0"
-            )
             self.pump_motor_direction = epics_signal_r(
                 PumpMotorDirectionState, prefix + "MTRDIR"
             )
-            self.pump_speed_rbv = epics_signal_r(float, prefix + "MSPEED_RBV")
+            self.pump_speed = epics_signal_rw(
+                float, write_pv=prefix + "MSPEED", read_pv="MSPEED_RBV"
+            )
 
         with self.add_children_as_readables(ConfigSignal):
             self.pump_mode = epics_signal_rw(PumpState, prefix + "SP:AUTO")
-            self.pump_speed = epics_signal_rw(float, prefix + "MSPEED")
-            self.pump_move_forward = epics_signal_rw(int, prefix + "M1:FORW")
-            self.pump_move_backward = epics_signal_rw(bool, prefix + "M1:BACKW")
-            self.pump_connection = epics_signal_rw(
-                PumpMotorControlRequest, prefix + "M1:CON"
-            )
 
         super().__init__(name)
 
 
 class PressureTransducer(StandardReadable):
-    """
-    reads pressure
-    """
-
     def __init__(
         self, prefix: str, number: int, name: str = "", adc_prefix: str = ""
     ) -> None:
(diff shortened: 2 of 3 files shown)


src/dodal/beamlines/p38.py
```diff
@@ -318,3 +319,17 @@ def linkam(
...
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> PressureJumpCell:
+    return device_instantiation(
+        PressureJumpCell,
+        "high_pressure_xray_cell",
+        "-EA-HPXC-01:",
+        wait_for_connection,
+        fake_with_ophyd_sim,
+        adc1_prefix=f"{BeamlinePrefix(BL).beamline_prefix}-EA-ADC-01:",
+        adc2_prefix=f"{BeamlinePrefix(BL).beamline_prefix}-EA-ADC-02:",
```
stan-dot, line 334: should:  extract the value the the function body

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,239 @@
...
+
+class PressureJumpCellBusyStatus:
+    IDLE = "Idle"
+    BUSY = "Busy"
+
+
+class PressureJumpCellTimerState:
+    TIMEOUT = "TIMEOUT"
+    COUNTDOWN = "COUNTDOWN"
+
```
stan-dot, line 19: could: other device files such as linkam (https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/linkam3.py) or hutch shutter (https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/hutch_shutter.py) have those classes explicitly extend enum

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,239 @@
...
+            ## Control Pressure ##
+            self.control_timer = epics_signal_r(
+                PressureJumpCellTimerState, prefix + "CTRL:TIMER"
+            )
+            self.control_counter = epics_signal_r(float, prefix + "CTRL:COUNTER")
+            self.control_script_status = epics_signal_r(str, prefix + "CTRL:RESULT")
+            self.control_routine = epics_signal_r(str, prefix + "CTRL:METHOD")
+            self.control_state = epics_signal_r(str, prefix + "CTRL:STATE")
+            self.control_iteration = epics_signal_r(int, prefix + "CTRL:ITER")
```
stan-dot, line 183: should: refactor this into more components such as is in for instance thawer.py https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/thawer.py

stan-dot: those would be pressure transducer, valvecontrol, pressurecontroller,

stan-dot (changes requested):
At the moment it's implemented as one large object hard to debug, should be split into components

src/dodal/devices/pressure_jump_cell.py
DominicOram: Must: For these enums to work with the real PVs they must extend `str` and `enum`

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,239 @@
...
+
+class PressureJumpCellValveControlRequest:
+    OPEN = "Open"
+    CLOSE = "Close"
+    RESET = "Reset"
+    ARM = "Arm"
+    DISARM = "Disarm"
+    # TODO the nones may not be required but FVST and SXST set
+    NONE1 = ""
+    NONE2 = ""
```
DominicOram, line 34: Should: They're not required. In fact they don't really make sense if you read `""` from the device is that a `NONE1` or `NONE2`? What do they each mean?

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,239 @@
...
+            self.valve3_state = epics_signal_r(
+                PressureJumpCellValveState, prefix + "V3:STA"
+            )
+            # V4 - valve manual control
+            self.valve5_state = epics_signal_r(
+                PressureJumpCellFastValveState, prefix + "V5:STA"
+            )
+            self.valve6_state = epics_signal_r(
+                PressureJumpCellFastValveState, prefix + "V6:STA"
+            )
```
DominicOram, line 108: Could: Take a look at `DeviceVector` you can get all these in a list and it will be cleaner I think

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Improved the device structure in a930d192dd22f13bd011d76bedae5af776fef0a0 based on the components.

src/dodal/beamlines/p38.py
barnettwilliam: Moved the adc numbering definition into PressureJumpCell's constructor in 2face7c257b63e3b682f08fbf87266fe1ea7e12c. This would be improved by correcting the PV naming in the device driver itself so the adc would be a sub-section of cell prefix, but I suggest testing first with the unmodified driver.

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Improved as per https://github.com/DiamondLightSource/dodal/pull/673#discussion_r1691655374. There's now the following components: 
- PressureJumpCellControlValves
- PressureJumpCellPump
- PressureJumpCellPressureTransducers
- PressureJumpCellController

DominicOram: Nit: I like this change but I would suggest that you don't need to prefix everything with `PressureJumpCell` it's implied by the fact it's in the `pressure_jump_cell` file

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Fixed enums in 1538a5b04e878c4a927c064a994dc051c1a82eb9

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Removed in 07522bbdf0753fb23ad00510b9f01cee0517c3d3

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Thanks I hadn't noticed that, agree it would help reduce the repetition.

stan-dot: could: make the open + control as a composite device as they seem to occur together

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,236 @@
...
+    OPEN = "Open"
+    OPEN_ARMED = "Open Armed"
+    CLOSED = "Closed"
+    CLOSED_ARMED = "Closed Armed"
+    NONE5 = "Unused"
+
+
+class LimitSwitchState(str, Enum):
+    OFF = "Off"
+    ON = "On"
```
DiamondJoseph, line 79: I think all of the enums should be prepended with the relevant device class, so if you import multiple e.g. `PumpState` it's clearer which is for the `PressureJumpCell` and which is for the `PeristalticPump` if one is `PressureJumpPumpState` and the other is `PeristalticPumpState`

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,236 @@
...
+class Pump(StandardReadable):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        with self.add_children_as_readables():
+            self.pump_position = epics_signal_r(float, prefix + "POS")
+            self.pump_forward_limit = epics_signal_r(
+                LimitSwitchState, prefix + "D74IN1"
+            )
+            self.pump_backward_limit = epics_signal_r(
+                LimitSwitchState, prefix + "D74IN0"
+            )
```
DiamondJoseph, line 131: Are the limits settable and required to be Read signals or are they read-once Config signals?

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,236 @@
...
+            self.pump_speed_rbv = epics_signal_r(int, prefix + "MSPEED_RBV")
+
+        with self.add_children_as_readables(ConfigSignal):
+            self.pump_mode = epics_signal_rw(PumpState, prefix + "SP:AUTO")
+            self.pump_speed = epics_signal_rw(float, prefix + "MSPEED")
+            self.pump_move_forward = epics_signal_rw(bool, prefix + "M1:FORW")
+            self.pump_move_backward = epics_signal_rw(bool, prefix + "M1:BACKW")
+            self.pump_move_backward = epics_signal_rw(
+                PumpMotorControlRequest, prefix + "M1:CON"
```
DiamondJoseph, line 143: duplicated field name

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,236 @@
...
+            self.beckhoff_pressure = epics_signal_r(
+                float, prefix + "STATP:MeanValue_RBV"
+            )
+            self.beckhoff_voltage = epics_signal_r(float, adc_prefix + "CH1")
+            # todo this channel might be liable to change
+
+        super().__init__(name)
+
+
+class PressureJumpCellController(StandardReadable):
```
DiamondJoseph, line 167: Is the Controller just a readable? I'd expect a Controller to have some methods for changing the state of the device.

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,236 @@
...
+    CLOSED_ARMED = "Closed Armed"
+    NONE5 = "Unused"
+
+
+class LimitSwitchState(str, Enum):
+    OFF = "Off"
+    ON = "On"
+
+
+class AllValvesControl(StandardReadable):
```
DiamondJoseph, line 82: Name this after a specific piece of hardware or behaviour, so it's clearer what the intent is. 
From reading the device class, there's 8 valves,  of which 1,3 are slow and controlled by Epics, 5, 6 are fast and controlled by Epics and 2,4,7,8 may be manually (as in physically manipulated) set?

Do we need to distinguish between fast and slow moving valves in the read method? Are the slow valves changing not over the scale of the experiment but only between experiments? Presumably from the _ARMED enum, they're potentially hardware triggered?

Shouldn't the signals of type ValveControlRequest be SignalRW, the ControlRequest is the desired state?

src/dodal/devices/pressure_jump_cell.py
stan-dot: update: no longer necessary

src/dodal/devices/pressure_jump_cell.py
stan-dot: when it comes to it at i22 we can sort it out then

src/dodal/devices/pressure_jump_cell.py
stan-dot: how to check that?

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -197,6 +204,22 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
+#             beckhoff_pressure: NotConnected: ca://-HPXC-01:PP1:STATP:MeanValue_RBV
+#         pressure_transducer_2: NotConnected:
+#             omron_pressure: NotConnected: ca://-HPXC-01:PP2:PRES
+#             omron_voltage: NotConnected: ca://-HPXC-01:PP2:RAW
+#             beckhoff_pressure: NotConnected: ca://-HPXC-01:PP2:STATP:MeanValue_RBV
+#         pressure_transducer_3: NotConnected:
+#             omron_pressure: NotConnected: ca://-HPXC-01:PP3:PRES
+#             omron_voltage: NotConnected: ca://-HPXC-01:PP3:RAW
+#             beckhoff_pressure: NotConnected: ca://-HPXC-01:PP3:STATP:MeanValue_RBV
+
```
stan-dot, line 221: should: we forgot to delete this

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -213,19 +236,22 @@ def __init__(
...
             name="Pressure Transducer 2",
             adc_prefix=f"{prefix}{adc_prefix}-02:",
+            number=2
         )
         self.pressure_transducer_3 = PressureTransducer(
-            prefix + "PP3:",
+            prefix=f"{prefix}{cell_prefix}",
             name="Pressure Transducer 3",
             adc_prefix=f"{prefix}{adc_prefix}-01:",
+            number=3
```
stan-dot, line 254: might: there might be some broader dodal convention on how to handle the prefixes so that the nesting is predictable.

I am not sure
 @DominicOram  what would you do here?

stan-dot (changes requested):
some small bits

src/dodal/devices/pressure_jump_cell.py
DominicOram: > Name this after a specific piece of hardware or behaviour, so it's clearer what the intent is

Apologies @barnettwilliam, this is contradictory to what I said in https://github.com/DiamondLightSource/dodal/pull/673#discussion_r1695332818. @DiamondJoseph and I are also arguing about the same thing over at https://github.com/DiamondLightSource/dodal/pull/713#discussion_r1698083552. I think I'm still not convinced. Is this something that core have generally agreed upon @DiamondJoseph or is it still up for debate?

stan-dot: pressure jump cell is a i22 device we can worry about this specific naming if a moment comes when it is needed elsewhere.

DiamondJoseph: I think this [should be here](https://github.com/DiamondLightSource/dodal/pull/673#discussion_r1696889723), unless it's about AllValvesControl?

So far it's only me being a nuisance: looking at the (nicely newly refactored Ophyd-async) [StandardDetector impls](https://github.com/bluesky/ophyd-async/tree/main/src/ophyd_async/epics) (in `foo/_foo_io.py`) the pattern is consistent but probably because most examples have been copied from the AdAravis which initially had the pattern when it was thought we might need multiple Driver implementations for Mako/Mantis. 

I'm about ready to concede (and wouldn't have blocked merging on it) as examples from that repository are all being exported consistently to work as you have in your example on #713. I can just definitely see someone e.g. setting signal `A` and awaiting signal `A_at_final_position` (of type `A.FinalPositionReached` reaching `B.FinalPositionReached` and then that await times out and it's not clear debugging the scan what went wrong. Or being frustrated that type checking is highlighting an error "TypeError: Using type <long qualified path>.Foo, expected type <long qualified path>.Foo!"

stan-dot: The problem you describe here @DiamondJoseph  is a function of the namespace pollution. 

I think we're fine for now for pressure cell valves and pumps as I haven't seen them in dodal.

You convinced me about the other though:
```
class BusyState(str, Enum):
    IDLE = "Idle"
    BUSY = "Busy"


class TimerState(str, Enum):
    TIMEOUT = "TIMEOUT"
    COUNTDOWN = "COUNTDOWN"


class StopState(str, Enum):
    CONTINUE = "CONTINUE"
    STOP = "STOP"
```
These *should* have the device name in them

barnettwilliam: With regards to the device enums, does the underlying number matter or ever get used by ophyd/bluesky? should we make an effort to keep them aligned them with the  epics field numbers?

stan-dot: which number?

DominicOram: > I think this https://github.com/DiamondLightSource/dodal/pull/673#discussion_r1696889723, unless it's about AllValvesControl?

Yes, my bad, sorry

> These should have the device name in them

I've actually convinced @DiamondJoseph offline that they shouldn't have the device names prepended

>With regards to the device enums, does the underlying number matter or ever get used by ophyd/bluesky? should we make an effort to keep them aligned them with the epics field numbers?

The list of which strings are in the enum must match the list in EPICS but the order does not matter. This is enforced by a check when you run `dodal connect ` anyway.

src/dodal/devices/pressure_jump_cell.py
DominicOram: So I would do:
```python
class PressureJumpCell(StandardReadable):
    """
    High pressure X-ray cell, used to apply pressure or pressure jumps to a sample.
    """

    def __init__(
        self,
        prefix: str = "",
        cell_prefix: str = "",
        adc_prefix: str = "",
        name: str = "",
    ):
        self.all_valves_control = AllValvesControl(f"{prefix}{cell_prefix}", name)
        self.pump = Pump(f"{prefix}{cell_prefix}", name)

        self.pressure_transducers: DeviceVector[PressureTransducer] = DeviceVector(
            {
                i: PressureTransducer(
                    prefix=f"{prefix}{cell_prefix}",
                    number=i,
                    adc_prefix=f"{prefix}{adc_prefix}-{i:02}:",
                )
                for i in range(1, 4)
            }
        )
```
You can omit the name as the `DeviceVector` should fill it in with a sensible name after the fact.

stan-dot: ah of course

src/dodal/devices/pressure_jump_cell.py
stan-dot: deleted

src/dodal/devices/pressure_jump_cell.py
stan-dot: resolved elsewhere

src/dodal/devices/pressure_jump_cell.py
stan-dot: changed the lattern into `pump_connection`

src/dodal/devices/pressure_jump_cell.py
DiamondJoseph: They're `epics_signal_r` read only from the ophyd layer, which means they're read-only from the ophyd layer. They could be derived from other signals, and therefore changing on the rate of every point of a scan, but I would infer from the name that they are hardware limits that don't change over the lifetime of the device, let alone within a scan. I would make them config signals.

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,240 @@
...
+                LimitSwitchState, prefix + "D74IN0"
+            )
+            self.pump_motor_direction = epics_signal_r(
+                PumpMotorDirectionState, prefix + "MTRDIR"
+            )
+            self.pump_speed_rbv = epics_signal_r(float, prefix + "MSPEED_RBV")
+
+        with self.add_children_as_readables(ConfigSignal):
+            self.pump_mode = epics_signal_rw(PumpState, prefix + "SP:AUTO")
+            self.pump_speed = epics_signal_rw(float, prefix + "MSPEED")
```
DiamondJoseph, line 146: While the speed, direction etc. seem likely to change within a scan? It's something that could easily be scanned over, e.g. transmission vs. pump speed if the fluid is being rapidly damaged by the beam.

DiamondJoseph (changes requested):

src/dodal/devices/pressure_jump_cell.py
stan-dot: so you mean that they shouldn't be added with ConfigSignal?

DiamondJoseph: Read signals are read at every point of the scan- if something is changing, or variable or may want to be scanned over, it should be a Read signal.

Config signals are read only at the start of the scan. Something like limits would ideally be metadata on a signal that propagates from the Epics record, but if the data is as additional signals then having it read only once per scan is appropriate.

IMO:

PressureTransducer - perfect

PressureJumpCellController: some of these signals (e.g. GOTOBUSY, TIMER, COUNTER, RESULT, METHOD, ITERATION) seem to be more relevant for the control system itself rather than for reading out the data? I don't know enough about the usage of the device to know which should be read and how often.

pump I think the following is sensible?

```
pump:
        readables:
            self.pump_position = epics_signal_r(float, prefix + "POS")
            self.pump_motor_direction = epics_signal_r(
                PumpMotorDirectionState, prefix + "MTRDIR"
            )
use the epics_signal_rw with different set/read prefixes to have one signal that does both: sets go to set, reads come from _rbv
            self.pump_speed = epics_signal_rw(float, prefix + "MSPEED")

        config:
            self.pump_forward_limit = epics_signal_r(
                LimitSwitchState, prefix + "D74IN1"
            )
            self.pump_backward_limit = epics_signal_r(
                LimitSwitchState, prefix + "D74IN0"
            )
            self.pump_mode = epics_signal_rw(PumpState, prefix + "SP:AUTO")
this is int but the next is bool? Are they just saying which direction the pump is moving? If so, duplicating work of pump_motor_direction?
            self.pump_move_forward = epics_signal_rw(int, prefix + "M1:FORW")
            self.pump_move_backward = epics_signal_rw(bool, prefix + "M1:BACKW")

        super().__init__(name)
```

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: These signals are the state of limit switches for the pump so they will change during a scan if the pump is driven for a some length of time (approx 50s) in one direction. The name could be improved to `pump_..._limit_state`

src/dodal/devices/pressure_jump_cell.py
stan-dot: sorted when reworking with at p38 with @DiamondJoseph  on the 21st of August

DiamondJoseph (changes requested):

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,244 @@
...
+        name: str = "",
+    ):
+        self.all_valves_control = AllValvesControl(
+            f"{beamline_prefix}{cell_prefix}", name
+        )
+        self.pump = Pump(f"{beamline_prefix}{cell_prefix}", name)
+
+        self.controller = PressureJumpCellController(
+            f"{beamline_prefix}{cell_prefix}", name
+        )
```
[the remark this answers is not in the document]
stan-dot: done

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,244 @@
...
+    valve_3: ValveControlRequest | None = None
+    valve_5: FastValveControlRequest | None = None
+    valve_6: FastValveControlRequest | None = None
+
+
+class AllValvesControl(StandardReadable, Movable):
+    """
+    valves 2, 4, 7, 8 are not controlled by the IOC,
+    as they are under manual control.
+    """
```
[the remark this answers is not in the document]
stan-dot: done

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,244 @@
...
+                    i: epics_signal_rw(FastValveControlRequest, f"{prefix}V{i}:CON")
+                    for i in [5, 6]
+                }
+            )
+        )
+
+        self.valve_control: DeviceVector[SignalRW[ValveControlRequest]] = DeviceVector(
+            {
+                i: epics_signal_rw(ValveControlRequest, f"{prefix}V{i}:CON")
+                for i in [1, 3]
```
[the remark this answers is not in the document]
stan-dot: done, looking back this was a great idea!

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -0,0 +1,244 @@
...
+
+class Pump(StandardReadable):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        with self.add_children_as_readables():
+            self.pump_position = epics_signal_r(float, prefix + "POS")
+            self.pump_motor_direction = epics_signal_r(
+                PumpMotorDirectionState, prefix + "MTRDIR"
+            )
+            self.pump_speed = epics_signal_rw(
+                float, write_pv=prefix + "MSPEED", read_pv="MSPEED_RBV"
```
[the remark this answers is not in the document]
stan-dot: thanks for catching this

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -163,6 +150,44 @@ async def set(self, value: AllValvesControlState):
...
+class FastValveControl(StandardReadable):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        with self.add_children_as_readables():
+            self.close = epics_signal_rw(FastValveControlRequest, prefix + ":CON")
+            self.open = epics_signal_rw(ValveOpenSeqRequest, prefix + ":OPENSEQ")
+
+        super().__init__(name)
+
+    def set(self, value: FastValveControlRequest | ValveOpenSeqRequest) -> AsyncStatus:
+        setStatus = None
```
stan-dot, line 181: why not snake_case?

stan-dot (dismissed):

src/dodal/devices/pressure_jump_cell.py
stan-dot: amended

src/dodal/beamlines/p38.py
```diff
@@ -312,7 +313,7 @@ def linkam(
     return device_instantiation(
         Linkam3,
         "linkam",
-        "-EA-LINKM-02:",
+        f"{BeamlinePrefix(BL).insertion_prefix}-EA-LINKM-02:",
```
DiamondJoseph, line 316: While device_factory is reverted, this isn't going to work and is unrelated: it'll cause merge issues when that change goes back in, so let's revert it for now.

```suggestion
        "-EA-LINKM-02:",
```

DiamondJoseph (approved):
Approval conditional on it connects with dodal connect p38 and the prefix is right for both devices, I'm happy to check tomorrow but no time today

PR #674, opened 2024-07-16
(DiamondLightSource/hyperion#1464) Change topup gating to prevent gating with long exposures

rtuck99:
Fixes DiamondLightSource/hyperion#1464

After conversation with beamline staff, rather than implement the same solution as exists currently in GDA, the following behaviour is implemented:

* Topup gating in hyperion is now controlled by the following parameters in `beamlineParameters`:
   * `dodal_topup_threshold_exposure_s` (default 120) controls the exposure time at which hyperion gates topups
   * `dodal_topup_end_delay_s` (default 1) controls an additional adjustable delay after topup end to ensure no accidental overlap with topup due to timing inaccurancies and can be adjusted to allow further time for settling if necessary.

To determine whether a topup is performed:

* If the `COUNTDOWN` PV != -1 and Synchrotron mode is `User` or `Special` then the gating steps are followed as below, otherwise we proceed with the scan as normal.
* The exposure duration, plus an operation-specific setup time (currently default 10s for rotation scans and 30s for grid scans) is compared with the time remaining to topup (`ENDCOUNTDN` PV), if it is more than this then the scan may be delayed, otherwise the scan is not delayed.
* The exposure duration is compared with `dodal_topup_threshold_exposure_s` and if it is less than this the scan is delayed, otherwise it is considered a long duration scan which can be performed during topup and is not delayed.
* If after the above checks have determined a delay is required, we then wait for `ENDCOUNTDN - COUNTDOWN + dodal_topup_end_delay_s` seconds rounded up to the nearest second.

Note:

It is expected that the location of these parameters is temporary until at some later stage they are moved into the config server.

1. Attempt normal and long-duration scans during topup and check the above happens.
2. Tests pass

### Checks for reviewer


src/dodal/plans/check_topup.py
```diff
@@ -23,15 +39,28 @@ def _gating_permitted(machine_mode: SynchrotronMode):
     return False
 
 
-def _delay_to_avoid_topup(total_run_time, time_to_topup):
+def _delay_to_avoid_topup(
```
DominicOram, line 42: Could: We could have a docstring on this function explaining why we need the threshold and why sometimes wait and sometimes don't (basically just summarising this issue).

src/dodal/plans/check_topup.py
```diff
@@ -1,12 +1,28 @@
...
 
+DEFAULT_THRESHOLD_EXPOSURE_S = 120
+DEFAULT_TOPUP_GATE_DELAY_S = 1
+
+
+class TopupConfig:
+    # For planned exposures less than this value, wait for topup to finish instead of
+    # collecting throughout topup.
+    THRESHOLD_EXPOSURE_S = "dodal_topup_threshold_exposure_s"
+    # Additional safety margin to wait after the end of topup.
```
DominicOram, line 23: Should: A little bit more detail on why we need the safety margin would be good.

DominicOram (approved):
Great, thanks. Some optional comments to help with working out why this the behaviour in the future. Otherwise looks good.

Issue #483, opened 2024-04-29
Make `device_factory` decorator to ease making and connecting ophyd-async Devices


coretl:
https://github.com/DiamondLightSource/dodal/issues/415#issuecomment-2079651830 summarizes a discussion on Device connection.
https://github.com/bluesky/ophyd-async/issues/265 will implement the ophyd-async supporting features of idempotent connect. The strategy we would like for dodal is:
- Each beamline module decorates its device factories with a `device_factory`
- This factory makes, names, caches and optionally connects a device
- In script use can pass `connect=True` to the device to connect in the foreground, requires a RunEngine to have been made
- In blueapi use can find all the eager devices and connect them at startup
- Can also connect them in a plan with the `ensure_connected` plan stub

This ticket is for:

## New ADR
Describing how Devices should be connected in dodal, from scripts and under blueapi

## `dodal.beamlines.beamline_utils`
```python
D = TypeVar(bound=Device)
class DeviceFactory(Protocol, Generic[D]):
    def __call__(self, connect: bool = False, timeout: float = DEFAULT_TIMEOUT) -> D:
        ...

_factory_made_devices: Dict[DeviceFactory, Device] = {}
_device_is_lazy: Dict[DeviceFactory, bool] = {}

def device_factory(lazy=False, set_name=True):
    def wrapper(f) -> DeviceFactory[D]:
        def factory(connect=False, timeout: float = DEFAULT_TIMEOUT):
            device = _factory_made_devices.get(f)
            if not device:
                device = f()
                if set_name:
                    device.set_name(f.__name__)
                _factory_made_devices[f] = device
            if connect: 
                call_in_bluesky_event_loop(device.connect(timeout=timeout))
            return device
        _device_is_lazy[factory] = lazy
        factory.__name__ = f.__name__
        return factory
    return wrapper

def get_device_factories() -> Dict[DeviceFactory, bool]:
    return _device_is_lazy.copy()
```
Need to add type hints to this so it knows the type of device it is making from the function it is decorating and check with @callumforrester that `get_device_factories` is what blueapi actually needs

## Acceptance Criteria
* ADR written so we know why this was done
* `device_factory` available to use in beamline modules
* `i22.py` updated to use it

stan-dot:
so the idea is that all `device()...` line in `i22.py` or other beamline gets this decorator, right?

coretl:
> so the idea is that all `device()...` line in `i22.py` or other beamline gets this decorator, right?

yes

[...]

PR #678, opened 2024-07-16
Unify p38 and i22, swap out with env variables

### LINKED ISSUE #502: Automatically keep I22 and P38 in sync

stan-dot:
Fixes #502 

1. Try on i22, try on i38
2. Confirm the devices are differnent yet experiments run correctly

### Checks for reviewer


src/dodal/beamlines/i22.py
```diff
@@ -27,30 +31,41 @@
 set_log_beamline(BL)
 set_utils_beamline(BL)
 
+IS_LAB = BL == "p38"
```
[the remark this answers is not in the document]
stan-dot: please make the use case more clear. a function can be called more than once over life of a program, but I don't think the program will change which cluster it is running while it's there

src/dodal/beamlines/i22.py
```diff
@@ -27,30 +31,41 @@
...
+    )
+    if IS_LAB
+    else StaticVisitDirectoryProvider(
         BL,
         Path("/dls/i22/data/2024/cm37271-2/bluesky"),
         client=DirectoryServiceClient("http://i22-control:8088/api"),
     )
 )
 
+set_directory_provider(directory_provider)
```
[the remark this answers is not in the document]
stan-dot: not sure

stan-dot:
> /workspaces/dodal/tests/preprocessors/test_filesystem_metadata.py:76:9 - error: Method "trigger" overrides class "Triggerable" in an incompatible manner
    Return type mismatch: base method returns type "Status", override returns type "Status"
      "Status" is incompatible with protocol "Status"
        "exception" is an incompatible type
          Type "(timeout: Unknown | None = None) -> (StatusTimeoutError | Exception | type[Exception] | None)" is incompatible with type "(timeout: float | None = 0) -> (BaseException | None)"
            Function return type "StatusTimeoutError | Exception | type[Exception] | None" is incompatible with type "BaseException | None"
        "bluesky.protocols.Status" is incompatible with "ophyd.status.Status"
        "bluesky.protocols.Status" is incompatible with "ophyd.status.Status" (reportIncompatibleMethodOverride)
136 errors, 3 warnings, 0 informations 
ERROR: InvocationError for command /venv/bin/pyright src tests (exited with code 1)

type checking fails due to dodal not catching up to ophyd-async changes

src/dodal/beamlines/i22.py
```diff
@@ -27,30 +31,41 @@
...
+
+# d11 at p38, but disconnected
 def saxs(
     wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
 ) -> PilatusDetector:
     return device_instantiation(
-        NXSasPilatus,
-        "saxs",
-        "-EA-PILAT-01:",
+        AravisDetector if IS_LAB else NXSasPilatus,
```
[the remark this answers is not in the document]
DiamondJoseph: I think because they're returning different types, and the metadataholder (while a temporary construct until we have metadata service) would have completely different information it's clearer to have the obvious branch.

`saxs` should also have a return type of either `PilatusDetector | AravisDetector` or `StandardDetector` (see also for the other device factory functions)

src/dodal/beamlines/i22.py
```diff
@@ -65,6 +80,7 @@ def saxs(
     )
 
 
+# todo this is not available at p38
```
DiamondJoseph, line 83: @skip_device(IS_LAB)

See also https://github.com/DiamondLightSource/dodal/blob/9bfa98f73773725919f91b3d5ee3b3fc7496b9ea/src/dodal/beamlines/i03.py#L105

src/dodal/beamlines/i22.py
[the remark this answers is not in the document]
DiamondJoseph: I think we want the device name to be the same too?

When the device_instantiation is refactored it'll come from the function name, so it's clearer to enforce that now (if this is how we unify p38/i22)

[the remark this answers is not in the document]
stan-dot: I prefer the narrow OR )`|` than the broad `StandardDetector`

src/dodal/beamlines/i22.py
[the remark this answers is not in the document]
callumforrester: I agree it is effectively a constant either way, this was mostly a point of readability. I think this is more readable.

```python
def is_lab() -> bool:
    return BL == "p38"

def my_device():
    if is_lab():
        return ...
    else:
        return ...
```

The one benefit it does give its future proofing. If at some point some new requirement emerged that did require switching to lab mode while running, you just to need to modify the function in one place. But I agree that is unlikely.

But it's a nit, so feel free to ignore.

src/dodal/beamlines/i22.py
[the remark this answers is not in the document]
callumforrester: Would you mind reverting it then? To keep the PR small and focused

src/dodal/beamlines/i22.py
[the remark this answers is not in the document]
stan-dot: > future proofing
that is the inverse of agile

src/dodal/beamlines/i22.py
stan-dot: applied

stan-dot: might need `lambda _: ...`

stan-dot:
codecov for live beamline connectivity does not make that much sense, please let it merge

stan-dot:
I need this for #673

stan-dot:
from @DiamondJoseph  

> parameterised in such a way that you can either mock the IS_LAB call before the device_instantiation is called or otherwise export BEAMLINE=P38, then you should be able to traverse that alternate path

using https://github.com/DiamondLightSource/dodal/blob/502-automatically-keep-i22-and-p38-in-sync/tests/common/beamlines/test_device_instantiation.py

stan-dot:
@coretl would an addition of a 'lab' flag be in the scope for the device instantiation factory deocrator?

stan-dot:
need to delete LAB_NAME from the CLI, and make it all optimized for testability

src/dodal/beamlines/__init__.py
```diff
@@ -14,6 +14,10 @@
     "s03": "i03",
 }
 
+BEAMLINE_LAB_MAPPING: dict[str, str] = {
+    "i22": "p38",
+}
+
```
DiamondJoseph, line 20: Isn't this functionally the same as the _BEAMLINE_NAME_OVERRIDES? $BEAMLINE doesn't match up to expected dodal beamline

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
+LAB_NAME = BEAMLINE_LAB_MAPPING.get("i22", "i22")
+
+BL = LAB_NAME if LAB_FLAG else get_beamline_name("i22")
+print("BL NAME: ", BL)
 set_log_beamline(BL)
 set_utils_beamline(BL)
 
+IS_LAB = LAB_FLAG and BL == LAB_NAME
+print("is this lab? : ", IS_LAB)
+
```
DiamondJoseph, line 41: Revert most of this: current behaviour is as expected: if $BEAMLINE is set to p38, we get that otherwise we get i22. 

IS_LAB can stay, but without LAB_FLAG. We should always try and connect to whatever $BEAMLINE is set to.

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
 # Currently we must hard-code the visit, determining the visit at runtime requires
 # infrastructure that is still WIP.
 # Communication with GDA is also WIP so for now we determine an arbitrary scan number
 # locally and write the commissioning directory. The scan number is not guaranteed to
 # be unique and the data is at risk - this configuration is for testing only.
-set_directory_provider(
+directory_provider = (
```
DiamondJoseph, line 47:
```suggestion
set_directory_provider(
```

We need to call this method for the StandardDetectors to write to disk

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
     StaticVisitDirectoryProvider(
+        BL,
+        Path("/dls/p38/data/2024/cm37282-2/bluesky"),
+        client=LocalDirectoryServiceClient(),
```
DiamondJoseph, line 51: client=DirectoryServiceClient("http://p38-control:8088/api"),

Currently the DirectoryService is GDA, which can be available for p38

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
+    else StaticVisitDirectoryProvider(
         BL,
         Path("/dls/i22/data/2024/cm37271-2/bluesky"),
         client=DirectoryServiceClient("http://i22-control:8088/api"),
     )
 )
 
 
+# d3 at p38
+def oav(
```
DiamondJoseph, line 63: Move this back to where it was in the file to make it clearer what's changed

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> AravisDetector:
+    return device_instantiation(
+        AravisDetector if IS_LAB else NXSasOAV,
+        "d3" if IS_LAB else "oav",
```
DiamondJoseph, line 68: Be consistent with name between both beamlines: always "oav"

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
     )
 )
 
 
+# d3 at p38
+def oav(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> AravisDetector:
+    return device_instantiation(
+        AravisDetector if IS_LAB else NXSasOAV,
```
DiamondJoseph, line 67: Missing required fields for NXSasOAV

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
+            hdf_suffix="HDF5:",
+            # metadata_holder=NXSasMetadataHolder(
+            #     x_pixel_size=(1.72e-1, "mm"),
+            #     y_pixel_size=(1.72e-1, "mm"),
+            #     description="Dectris Pilatus3 2M",
+            #     type="Photon Counting Hybrid Pixel",
+            #     sensor_material="silicon",
+            #     sensor_thickness=(0.45, "mm"),
+            #     distance=(4711.833684146172, "mm"),
+            # ),
```
DiamondJoseph, line 103: Define a `NXSasAravis` that takes the metadataholder  so processing can reliably treat `saxs` as a valid NXsas detector, and find the values.

src/dodal/beamlines/i22.py
```diff
@@ -345,35 +417,23 @@ def panda4(
...
-        hdf_suffix="HDF5:",
-        metadata_holder=NXSasMetadataHolder(
-            x_pixel_size=(3.45e-3, "mm"),  # Double check this figure
-            y_pixel_size=(3.45e-3, "mm"),
-            description="AVT Mako G-507B",
-            distance=(-1.0, "m"),
-        ),
-        directory_provider=get_directory_provider(),
-    )
+LINKAM_IS_IN_LAB = False
```
DiamondJoseph, line 420: Move this to the top for clarity

src/dodal/cli.py
```diff
@@ -28,7 +34,7 @@ def main(ctx: click.Context) -> None:
     "-a",
     "--all",
     is_flag=True,
-    help="Attempt to connect to devices marked as skipped",
+    help="collect_factoriesAttempt to connect to devices marked as skipped",
```
DiamondJoseph, line 37:
```suggestion
    help="Attempt to connect to devices marked as skipped",
```

src/dodal/cli.py
```diff
@@ -39,9 +45,20 @@ def main(ctx: click.Context) -> None:
     "attempt any I/O. Useful as a a dry-run.",
     default=False,
 )
-def connect(beamline: str, all: bool, sim_backend: bool) -> None:
+@click.option(
```
DiamondJoseph, line 48: I think this file can all be reverted in favour of just connecting to whatever $BEAMLINE is.

src/dodal/beamlines/i22.py
```diff
@@ -77,16 +138,17 @@ def synchrotron(
     )
 
 
+# d12 at p38
 def waxs(
```
DiamondJoseph, line 142: I'd rather see this layed out like saxs, it's clearer what's going on

src/dodal/beamlines/i22.py
```diff
@@ -108,17 +170,24 @@ def i0(
     return device_instantiation(
         TetrammDetector,
         "i0",
-        "-EA-XBPM-02:",
+        f"-EA-XBPM-0{1 if IS_LAB else 2}:",
```
DiamondJoseph, line 173: As above, prefered split like saxs

src/dodal/beamlines/i22.py
```diff
@@ -77,16 +138,17 @@ def synchrotron(
...
+# d12 at p38
 def waxs(
     wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
 ) -> PilatusDetector:
     return device_instantiation(
-        NXSasPilatus,
-        "waxs",
-        "-EA-PILAT-03:",
+        AravisDetector if IS_LAB else NXSasPilatus,
+        "d12" if IS_LAB else "waxs",
```
DiamondJoseph, line 147: always "waxs"

src/dodal/beamlines/i22.py
```diff
@@ -23,27 +29,81 @@
...
+# d11 at p38, but disconnected
+@skip_device(lambda: BL == LAB_NAME)
 def saxs(
     wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
-) -> PilatusDetector:
+) -> AravisDetector | NXSasPilatus:
+    if IS_LAB:
+        return device_instantiation(
+            AravisDetector,
+            "d11",
```
DiamondJoseph, line 89:
```suggestion
            "saxs",
```

src/dodal/beamlines/i22.py
```diff
@@ -345,35 +417,23 @@ def panda4(
...
+LINKAM_IS_IN_LAB = False
 
 
-@skip_device()
+@skip_device(
+    lambda: BL == LAB_NAME
+    and not LINKAM_IS_IN_LAB
+    or BL != LAB_NAME
+    and LINKAM_IS_IN_LAB
+)
```
DiamondJoseph, line 428:
```suggestion
@skip_device(
    lambda: IS_LAB != LINKAM_IN_LAB
)
```

src/dodal/beamlines/i22.py
```diff
@@ -345,35 +417,23 @@ def panda4(
...
 def linkam(
     wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
 ) -> Linkam3:
     return device_instantiation(
         Linkam3,
         "linkam",
-        "-EA-TEMPC-05",
+        "-EA-TEMPC-05" if IS_LAB else "-EA-TEMPC-05",
```
DiamondJoseph, line 435: split devices like saxs

tests/common/beamlines/test_device_instantiation.py
```diff
@@ -31,6 +69,29 @@ def test_device_creation(RE, module_and_devices_for_beamline):
     assert len(beamline_utils.ACTIVE_DEVICES) == len(devices)
 
 
+@pytest.mark.parametrize(
```
DiamondJoseph, line 72: I'd make a special p38 test that we can add to when we have additional lab beamlines

DiamondJoseph (changes requested):

src/dodal/beamlines/__init__.py
stan-dot: it's not the same in the grand context of things

src/dodal/beamlines/i22.py
stan-dot: the `--lab` was suggested by @callumforrester and then implemented. I find it neater as is emphasizes the link between the i22 and p38

src/dodal/beamlines/i22.py
stan-dot: it is already called later

src/dodal/beamlines/i22.py
stan-dot: done

src/dodal/beamlines/i22.py
stan-dot: done

src/dodal/beamlines/i22.py
stan-dot: ok

src/dodal/beamlines/i22.py
stan-dot: added

src/dodal/beamlines/i22.py
stan-dot: done

src/dodal/cli.py
stan-dot: alreadu discussed above

src/dodal/beamlines/i22.py
stan-dot: it is not obvious that it would be clearer

src/dodal/beamlines/i22.py
stan-dot: ok

src/dodal/beamlines/i22.py
stan-dot: clever

tests/common/beamlines/test_device_instantiation.py
stan-dot: it's functionally equivalent...

src/dodal/beamlines/i22.py
callumforrester: I suggested a much more granular approach. Rather than a global `--lab` flag (which just sounds like GDA dummy mode, and is likely to have all the same problems), have as many "profiles" as you like per beamline.

What @DiamondJoseph suggests is much closer to the approach currently used in other parts of dodal at the moment, so I suggest we stick with that and revisit the profiles approach as a separate ticket (which, to be honest, is what I thought you were doing).

callumforrester: > I have no idea what dodal profiles are....

They are a thing I made up, they do not actually exist, don't worry :)

stan-dot: pushed the revert

src/dodal/beamlines/i22.py
stan-dot: ???I do not understand

stan-dot:
at the core the still not-agreed-on-implementation detail is where to put the i22-p38 mapping?

when we read in the env in the CLI we need to check if the BEAMLINE belongs to the lab set and then load the `dodal.beamlines.non-lab-beamline`. a dictionary (this time lab: real) sounds like the most Pythonic way to do this and this can easily be hard-coded for the future. 
request for comments @callumforrester , @DiamondJoseph

DiamondJoseph:
> when we read in the env in the CLI we need to check if the BEAMLINE belongs to the lab set ...

Which CLI, dodal's?

What I would expect is if you call `dodal connect i22`, you run i22.py and try and connect to the devices. If the env var `$BEAMLINE` is set to `p38`, that gets picked up by `BL = get_beamline_name("i22")` and mostly you connect to mocked devices. You never try and run `dodal connect p38`

stan-dot:
ok fair rnough

stan-dot:
@callumforrester please weigh in, me and @DiamondJoseph  got stuck

src/dodal/beamlines/i22.py
callumforrester:
```python
if IS_LAB:
    return this_device()
else:
    return that_device()
```

tests/common/beamlines/test_device_instantiation.py
callumforrester: As long as the tests cover both cases I'm not bothered at the moment, can revisit later

src/dodal/beamlines/i22.py
stan-dot: I do not agree that it's more readable, but I won't argue, changing now

stan-dot:
```

@pytest.mark.parametrize(
    "module_and_devices_for_beamline",
    set(all_beamline_modules()),
    indirect=True,
)
def test_device_creation(RE, module_and_devices_for_beamline):
    """
    Ensures that for every beamline all device factories are using valid args
    and creating types that conform to Bluesky protocols.
    """
    _, devices = module_and_devices_for_beamline
    for device_name, device in devices.items():
        assert device_name in beamline_utils.ACTIVE_DEVICES, (
            f"No device named {device_name} was created, devices "
            f"are {beamline_utils.ACTIVE_DEVICES.keys()}"
        )
        assert follows_bluesky_protocols(device)
    assert len(beamline_utils.ACTIVE_DEVICES) == len(devices)


@pytest.mark.parametrize(
    "mappings",
    set({"i22": "p38"}),
    indirect=True,
)
def test_lab_version_of_a_beamline(RE, mappings):
    """
    Ensures that for every lab beamline all device factories are using valid args
    and creating types that conform to Bluesky protocols.
    """
    # todo patch the environment
    with patch.dict(os.environ, {"BEAMLINE": mappings[1]}):
        module = get_beamline_based_on_environment_variable()
        # get the devices file for the beamline namespace
        _, devices = get_module_name_by_beamline_name(mappings[0])
        for device_name, device in devices.items():
            assert device_name in beamline_utils.ACTIVE_DEVICES, (
                f"No device named {device_name} was created, devices "
                f"are {beamline_utils.ACTIVE_DEVICES.keys()}"
            )
            assert follows_bluesky_protocols(device)
        assert len(beamline_utils.ACTIVE_DEVICES) == len(devices)

```

I need some help with the logic for the loading of modules during the test? sometimes it's a module and sometimes it's just a module name

src/dodal/beamlines/i22.py
```diff
@@ -24,26 +28,65 @@
 from dodal.utils import BeamlinePrefix, get_beamline_name, skip_device
 
 BL = get_beamline_name("i22")
+print("BL NAME: ", BL)
```
DiamondJoseph, line 31: Remove these print statements
```suggestion
```

src/dodal/beamlines/i22.py
```diff
@@ -24,26 +28,65 @@
...
 set_log_beamline(BL)
 set_utils_beamline(BL)
 
+_LAB_NAME = "p38"
+
+IS_LAB = BL == _LAB_NAME
+print("is this lab? : ", IS_LAB)
```
DiamondJoseph, line 38:
```suggestion
```

src/dodal/beamlines/i22.py
```diff
@@ -24,26 +28,65 @@
 from dodal.utils import BeamlinePrefix, get_beamline_name, skip_device
 
 BL = get_beamline_name("i22")
+print("BL NAME: ", BL)
 set_log_beamline(BL)
 set_utils_beamline(BL)
 
+_LAB_NAME = "p38"
+
+IS_LAB = BL == _LAB_NAME
```
DiamondJoseph, line 37:
```suggestion
IS_LAB = BL == "p38"
```
closer to i03/s03

src/dodal/beamlines/i22.py
```diff
@@ -24,26 +28,65 @@
...
+    if IS_LAB:
+        return device_instantiation(
+            NXSasAravis,
+            "saxs",
+            "-DI-DCAM-03:",
+            wait_for_connection,
+            fake_with_ophyd_sim,
+            drv_suffix="DET:",
+            hdf_suffix="HDF5:",
+            # todo double check the metadata
```
DiamondJoseph, line 77: Should be able to get most of these figures from the i22 oav which is a similar model. Then can remove the TODO, or otherwise make a ticket to track the TODO and link it here.

src/dodal/beamlines/i22.py
```diff
@@ -65,6 +108,7 @@ def saxs(
     )
 
 
+@skip_device(lambda: BL == _LAB_NAME)
 def synchrotron(
     wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
```
DiamondJoseph, line 113:
```suggestion
def synchrotron(
    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = BL == _LAB_NAME
```
simulate the device for the lab?

src/dodal/beamlines/i22.py
```diff
@@ -77,16 +121,17 @@ def synchrotron(
...
+        AravisDetector if IS_LAB else NXSasPilatus,
         "waxs",
-        "-EA-PILAT-03:",
+        "-DI-DCAM-04:" if IS_LAB else "-EA-PILAT-03:",
         wait_for_connection,
         fake_with_ophyd_sim,
-        drv_suffix="CAM:",
+        drv_suffix="DET:" if IS_LAB else "CAM:",
         hdf_suffix="HDF5:",
         metadata_holder=NXSasMetadataHolder(
```
DiamondJoseph, line 136: This metadata isn't valid for the lab device, and could make analysis wrong (e.g. pixel size, distance used to calculate angles). This is part of why I'd prefer to see structured as a single ternary rather than multiple ternaries inside the device_instantiation function

src/dodal/beamlines/i22.py
```diff
@@ -284,8 +337,10 @@ def fswitch(
     )
 
 
+# Must find which PandA IOC(s) are compatible
 # Must document what PandAs are physically connected to
 # See: https://github.com/bluesky/ophyd-async/issues/284
+@skip_device(lambda: BL == _LAB_NAME)
```
DiamondJoseph, line 343:
```suggestion
# Must document what PandAs are physically connected to
# See: https://github.com/bluesky/ophyd-async/issues/284
```
Panda1 is preset in p38, just need the correct prefix if it's different.

src/dodal/beamlines/i22.py
```diff
@@ -330,7 +385,7 @@ def panda3(
     )
 
 
-@skip_device()
+@skip_device(lambda: BL == _LAB_NAME)
```
DiamondJoseph, line 388: Add to panda2, panda3 also: don't need to skip those devices on i22 anymore, and we're certain that we're reasonably sure only deploying alongside compatible pandas.

src/dodal/beamlines/i22.py
```diff
@@ -345,35 +400,42 @@ def panda4(
...
+        metadata_holder=None
+        if IS_LAB
+        else NXSasMetadataHolder(
+            x_pixel_size=(1.72e-1, "mm"),
+            y_pixel_size=(1.72e-1, "mm"),
+            description="Dectris Pilatus3 2M",
+            type="Photon Counting Hybrid Pixel",
+            sensor_material="silicon",
+            sensor_thickness=(0.45, "mm"),
+            distance=(4711.833684146172, "mm"),
```
DiamondJoseph, line 424: Copy paste error. This is now wrong where it was correct before.

As above re: laying out the devices, again with the rationale about the metadata, although this is one case wehre most of the fields will be the same, the distance will be different once it's measured.

src/dodal/devices/i22/NXSasAravis.py
```diff
@@ -0,0 +1,22 @@
+from ophyd_async.core import DirectoryProvider
+from ophyd_async.epics.areadetector import AravisDetector
+
+from dodal.devices.i22.nxsas import NXSasMetadataHolder
+
+
+class NXSasAravis(AravisDetector):
```
DiamondJoseph, line 7: Can use the existing NXSasOAV class, this implementation doesn't actually read the soft signals.

tests/conftest.py
```diff
@@ -50,7 +50,7 @@ def mock_beamline_module_filepaths(bl_name, bl_module):
 
 
 @pytest.fixture(scope="function")
-def module_and_devices_for_beamline(request):
+def mappings(request):
```
DiamondJoseph, line 53: revert this rename it removes clarity and isn't mapped to the inputs of the tests anymore?

DiamondJoseph (changes requested):

src/dodal/beamlines/i22.py
stan-dot: done

src/dodal/beamlines/i22.py
stan-dot: done

tests/conftest.py
stan-dot: reverted

src/dodal/beamlines/i22.py
callumforrester: You've done

```python
if IS_LAB:
    return this_device()
return that_device()
```

Which is functionally the same, but harder to read.

stan-dot: > Which is functionally the same, but harder to read.

that's a purely stylistic opinion, usefulness of `else` is a debated point among software engineers

src/dodal/beamlines/i22.py
stan-dot: here it's just this one bit that is different. if the device is the same physically imo inline ternary just at the point of difference is better

stan-dot:
this might need a re-work once #483  is merged

PR #682, opened 2024-07-17
(DiamondLightSource/hyperion#1169) remove test feature flags from dodal

rtuck99:
Removes test feature flags from dodal that were inserted in order to try to test the fix for #1169


dperl-dls (approved):
seems fine, just fix the linting please

Issue #679, opened 2024-07-16
Move to pydantic v2 


DominicOram:
We should make sure that all our dependencies etc. work on pydantic v2

## Acceptance Criteria
* We pin `pydantic>2`
* Unpin zocalo < 1.0.0
* We remove all the ignores around pydantic in `filterwarnings`
* Dependencies are happy to install on `pip install`
* Unit tests work

DominicOram:
Currently blocked on:
* https://github.com/bluesky/scanspec/issues/122
* https://github.com/DiamondLightSource/python-zocalo/issues/247

dperl-dls:
Once we can, we should then update some of the way we use pydantic features, e.g.: https://github.com/DiamondLightSource/dodal/blob/fe3f48cc7e65e05ca50025d22a2161bef4eadba3/src/dodal/devices/detector/detector.py#L59

dperl-dls:
We are now blocked by scanspec requiring numpy>=2.0 and p4p requiring numpy<2.0

blueapi and nexgen also still pinned to pydantic<2

DominicOram:
Waiting for another release of p4p

DiamondJoseph:
I believe we're currently *not* blocked by p4p- only Numpy 2 support is. I'm going to make the required changes and check

dperl-dls:
@DiamondJoseph yes, but unfortunately the version of scanspec which supports pydantic 2 also requires numpy 2, so we are indirectly still blocked by that

DiamondJoseph:
Not since [ScanSpec 0.7.2](https://github.com/bluesky/scanspec/releases/tag/0.7.2)

PR #683, opened 2024-07-17
Pin ophyd_async due to pydantic

DominicOram:
See https://github.com/DiamondLightSource/dodal/issues/679

1. Re-install `dodal` and confirm dependencies are happy, particularly we're still on `pydantic<2.0`

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ description = "Ophyd devices and other utils that could be used across DLS beaml
 dependencies = [
     "click",
     "ophyd",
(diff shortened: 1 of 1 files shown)


DiamondJoseph (approved):

dperl-dls (changes requested):
I think that dodal is not actually compatible with ophyd-async 0.3.1, should probably remove the lower bound or at least increase it to the version which works

dperl-dls (approved):
lgtm, thanks

PR #694, opened 2024-07-22
Make more use of AsyncStatus.wrap

DominicOram:
1. Confirm tests still pass and code is cleaner

### Checks for reviewer


olliesilvester (commented):
Changes look good, but CI is complaining

src/dodal/devices/zebra.py
```diff
@@ -110,10 +110,9 @@ async def _set_armed(self, demand: ArmDemand):
             if reading == demand.value:
                 return
 
-    def set(self, demand: ArmDemand) -> AsyncStatus:
-        return AsyncStatus(
-            asyncio.wait_for(self._set_armed(demand), timeout=self.TIMEOUT)
-        )
+    @AsyncStatus.wrap
```
noemifrisina, line 113: lgtm, thanks.

I think we could move the code in `_set_armed` in this `set` here instead of keeping in split into two functions but I don't feel that strongly about it

noemifrisina (approved):

src/dodal/devices/zebra.py
noemifrisina: Yes, but I don't mind either way

PR #705, opened 2024-07-25
Fix misleading zocalo results message

rtuck99:
Fixes a minor issue with logging found during investigation of i03 "gonio slump" - a log message from `zocalo_results.py` which suggests there are always 2 found crystals.

1. Tests pass
2. Everything works normally
3. Log reports correct number of found crystals


dperl-dls (approved):
lgtm, thank you!

PR #713, opened 2024-07-29
Convert cyrostream to ophyd-async

### LINKED ISSUE #686: Convert to ophyd-async: cryostream

DominicOram:
Fixes #686

1. Confirm unit tests pass
2. Confirm `dodal connect i03` passes

### Checks for reviewer


src/dodal/devices/cryostream.py
```diff
@@ -1,9 +1,21 @@
...
-class Cryo(Device):
-    course = Cpt(EpicsSignal, "-EA-CJET-01:COARSE:CTRL")
-    fine = Cpt(EpicsSignal, "-EA-CJET-01:FINE:CTRL")
-    temp = Cpt(EpicsSignalRO, "-EA-CSTRM-01:TEMP")
-    backpress = Cpt(EpicsSignalRO, "-EA-CSTRM-01:BACKPRESS")
+
+class InOut(str, Enum):
+    IN = "In"
+    OUT = "Out"
+
```
DiamondJoseph, line 10: I've been pushing for prefixing all signal value enums with the device class they're for, e.g. from ophyd-async

```python
class AravisTriggerMode(str, Enum):
    """GigEVision GenICAM standard: on=externally triggered"""

    on = "On"
    off = "Off"
```

It feels a bit overkill for In/Out, On/Off, etc., but it means won't get caught up trying to compare 2 different enums when inevitable some other device version of InOut has In, Out and Moving. And means can be clearer what In/Out mean (sometimes in beam, sometimes in box, sometimes direction...)

DominicOram: Is it not clear by where it lives though? e.g. `from cyrostream import InOut`. If you have multiple `InOut`s you might want in a plan then you will have to import them differently anyway:
```python
from cyrostream import InOut as CryostreamInOut
from other import InOut as OtherInOut

cryosteam.set(OtherInOut) # This would get caught by the type checker
```

stan-dot: then we could have a shared `InOut`? 

That the values are in/out is a happenstance of a given channel, what matters it where they go - the key. On/Off as string is a bit of an odd case as it logically maps to boolean, we might want to handle it in some special way.

Perhaps in OphydAsync?

`from cyrostream import InOut as CryostreamInOut`
this might not be the most popular opinion but for me this would be a code smell as there is no guarantee the 'CryostreamInOut` would be consistent across different plans. Another plan might have `CryoInOut`.

`numpy as np` or `bluesky.plan_stubs as bps` are an exception as they just shorten the import but both are in widespread use so misspelling is less likely

DominicOram: I dislike a shared `InOut` because I don't think we can rely on the assumption that different devices share enum values.

> this might not be the most popular opinion but for me this would be a code smell as there is no guarantee the `CryostreamInOut` would be consistent across different plans. Another plan might have `CryoInOut`.

I don;t really see any issue with different plans using different names like this, if they want to.

I have discussed with @DiamondJoseph offline and he is happy for us to not prepend the device name on enums

olliesilvester (approved):

PR #710, opened 2024-07-29
Add i18 beamline file

stan-dot:
Fixes #709 

### Instructions to reviewer on how to test:
1. `dodal connect i181
2. Confirm thing all the listed devices connect

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

PR #721, opened 2024-07-31
706 make undulator gap writeable i18

### LINKED ISSUE #706: Make Undulator gap writeable - i18
This is for beamline alignment lookup tables at i18
https://github.com/DiamondLightSource/i18-bluesky/issues/12

At the moment it only implements the StandardReadable interface

## Acceptance Criteria
* can be run in a plan

stan-dot:
Fixes #706

needed for i18 #710 and https://github.com/DiamondLightSource/i18-bluesky/issues/4

1. run the undulator_dcm routine for lookup table update
2. Confirm nothing breaks happens

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -308,7 +308,7 @@ def undulator(
         wait_for_connection,
         fake_with_ophyd_sim,
         bl_prefix=False,
-        id_gap_lookup_table_path="/dls_sw/i03/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt",
+        id_gap_lookup_table_path="/dls_sw/i04/software/gda/config/lookupTables/BeamLine_Undulator_toGap.txt",
     )
 
 
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -308,6 +308,7 @@ def undulator(
         wait_for_connection,
         fake_with_ophyd_sim,
         bl_prefix=False,
+        id_gap_lookup_table_path="/dls_sw/i03/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt",
     )
 
 
--- a/src/dodal/beamlines/i22.py
+++ b/src/dodal/beamlines/i22.py
@@ -199,6 +199,7 @@ def undulator(
         bl_prefix=False,
         poles=80,
         length=2.0,
+        id_gap_lookup_table_path="/dls_sw/i22/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt",
     )
 
 
--- a/src/dodal/devices/undulator.py
+++ b/src/dodal/devices/undulator.py
@@ -106,9 +106,7 @@ class Undulator(StandardReadable, Movable):
         Args:
             value: energy in keV
         """
-        await asyncio.gather(
-            self._set_undulator_gap(value),
-        )
+        await self._set_undulator_gap(value)
 
     async def _set_undulator_gap(self, energy_kev: float) -> None:
         access_level = await self.gap_access.get_value()
--- a/src/dodal/devices/undulator.py
+++ b/src/dodal/devices/undulator.py
@@ -100,7 +100,7 @@ class Undulator(StandardReadable, Movable):
     @AsyncStatus.wrap
     async def set(self, value: float):
         """
-        set the undulator gap to a given ENERGY
+        Set the undulator gap to a given energy in keV
 
         Args:
             value: energy in keV
@@ -114,22 +114,22 @@ class Undulator(StandardReadable, Movable):
         if access_level is UndulatorGapAccess.DISABLED and not TEST_MODE:
             raise AccessError("Undulator gap access is disabled. Contact Control Room")
         LOGGER.info(f"Setting undulator gap to {energy_kev:.2f} kev")
-        gap_to_match_dcm_energy = await self._gap_to_match_dcm_energy(energy_kev)
+        target_gap = await self._get_gap_to_match_energy(energy_kev)
 
         # Check if undulator gap is close enough to the value from the DCM
         current_gap = await self.current_gap.get_value()
         tolerance = await self.gap_discrepancy_tolerance_mm.get_value()
-        difference = abs(gap_to_match_dcm_energy - current_gap)
+        difference = abs(target_gap - current_gap)
         if difference > tolerance:
             LOGGER.info(
                 f"Undulator gap mismatch. {difference:.3f}mm is outside tolerance.\
-                Moving gap to nominal value, {gap_to_match_dcm_energy:.3f}mm"
+                Moving gap to nominal value, {target_gap:.3f}mm"
             )
             if not TEST_MODE:
                 # Only move if the gap is sufficiently different to the value from the
                 # DCM lookup table AND we're not in TEST_MODE
                 await self.gap_motor.set(
-                    gap_to_match_dcm_energy,
+                    target_gap,
                     timeout=STATUS_TIMEOUT_S,
                 )
             else:
@@ -140,8 +140,11 @@ class Undulator(StandardReadable, Movable):
                 f"{energy_kev}, no need to ask it to move"
             )
 
-    async def _gap_to_match_dcm_energy(self, energy_kev: float) -> float:
-        # from lookup table a get a 2d np.array converting energies to undulator gap distance
+    async def _get_gap_to_match_energy(self, energy_kev: float) -> float:
+        """
+        get a 2d np.array from lookup table that
+        converts energies to undulator gap distance
+        """
         energy_to_distance_table: np.ndarray = await energy_distance_table(
             self.id_gap_lookup_table_path
         )
--- a/src/dodal/devices/undulator.py
+++ b/src/dodal/devices/undulator.py
@@ -93,9 +93,15 @@ class Undulator(StandardReadable, Movable):
         super().__init__(name)
 
     @AsyncStatus.wrap
-    async def set(self, energy_value_kev: float):
+    async def set(self, value: float):
+        """
+        set the undulator gap to a given ENERGY
+
+        Args:
+            value: energy in keV
+        """
         await asyncio.gather(
-            self._set_undulator_gap(energy_value_kev),
+            self._set_undulator_gap(value),
         )
 
     async def _set_undulator_gap(self, energy_kev: float) -> None:
--- a/src/dodal/devices/undulator.py
+++ b/src/dodal/devices/undulator.py
@@ -93,13 +93,13 @@ class Undulator(StandardReadable, Movable):
         super().__init__(name)
 
     @AsyncStatus.wrap
-    async def set(self, value: float):
+    async def set(self, energy_value_kev: float):
         await asyncio.gather(
-            self._set_undulator_gap(value),
+            self._set_undulator_gap(energy_value_kev),
         )
 
     async def _set_undulator_gap(self, energy_kev: float) -> None:
-        LOGGER.info(f"Setting DCM energy to {energy_kev:.2f} kev")
+        LOGGER.info(f"Setting undulator gap to {energy_kev:.2f} kev")
         gap_to_match_dcm_energy = await self._gap_to_match_dcm_energy(energy_kev)
 
         # Check if undulator gap is close enough to the value from the DCM
--- a/src/dodal/devices/undulator.py
+++ b/src/dodal/devices/undulator.py
@@ -25,6 +25,8 @@ class AccessError(Exception):
 
 # Enable to allow testing when the beamline is down, do not change in production!
 TEST_MODE = False
+# will be made more generic in https://github.com/DiamondLightSource/dodal/issues/754
+
 
 # The acceptable difference, in mm, between the undulator gap and the DCM
 # energy, when the latter is converted to mm using lookup tables
--- a/src/dodal/devices/undulator.py
+++ b/src/dodal/devices/undulator.py
@@ -1,12 +1,30 @@
+import asyncio
 from enum import Enum
 
-from ophyd_async.core import ConfigSignal, StandardReadable, soft_signal_r_and_setter
+import numpy as np
+from bluesky.protocols import Movable
+from numpy import argmin, ndarray
+from ophyd_async.core import (
+    AsyncStatus,
+    ConfigSignal,
+    StandardReadable,
+    soft_signal_r_and_setter,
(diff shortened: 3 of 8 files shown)


DiamondJoseph:
needs to be added to all of the undulator device instantiations

src/dodal/beamlines/i04.py
```diff
@@ -303,6 +303,7 @@ def undulator(
         wait_for_connection,
         fake_with_ophyd_sim,
         bl_prefix=False,
+        id_gap_lookup_table_path="/dls_sw/i03/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt",
```
stan-dot, line 306: @DominicOram I added the path from i03 as I don't see the undulator_dcm at i04.

if dcm is not present at i04 this parameter kind of does not make sense, we could make the lookup table path optional and then not allow for the method to update it proceed if the object does not have that path property

tests/devices/unit_tests/test_undulator_dcm.py
```diff
@@ -10,11 +10,14 @@
 )
 
 from dodal.devices.dcm import DCM
-from dodal.devices.undulator import Undulator, UndulatorGapAccess
+from dodal.devices.undulator import (
+    Undulator,
+    UndulatorGapAccess,
+    _get_closest_gap_for_energy,
+)
```
DominicOram, line 17: Should: I don't think it makes sense for the `test_undulator_dcm.py` to be testing the `undulator` itself. This logic should be tested in `test_undulator.py`

src/dodal/devices/undulator_dcm.py
```diff
@@ -78,7 +63,7 @@ def __init__(
     async def set(self, value: float):
         await asyncio.gather(
             self._set_dcm_energy(value),
-            self._set_undulator_gap_if_required(value),
+            self.undulator._set_undulator_gap(value),
```
DominicOram, line 66: Should: This should use the `set` in the undulator rather than private methods.

src/dodal/devices/undulator.py
```diff
@@ -63,3 +90,49 @@ def __init__(
...
         super().__init__(name)
+
+    @AsyncStatus.wrap
+    async def set(self, value: float):
+        await asyncio.gather(
+            self._set_undulator_gap(value),
+        )
+
+    async def _set_undulator_gap(self, energy_kev: float) -> None:
+        LOGGER.info(f"Setting DCM energy to {energy_kev:.2f} kev")
```
DominicOram, line 101: Should: This is now decoupled from the DCM, we shouldn't be mentioning it in the logs/comments as it will cause confusion

src/dodal/devices/undulator.py
```diff
@@ -63,3 +90,49 @@ def __init__(
                 self.length = None
 
         super().__init__(name)
+
+    @AsyncStatus.wrap
+    async def set(self, value: float):
```
DominicOram, line 95: Should: This needs a docstring to make it clear you are setting energy and in what units, otherwise people may think you should be setting a gap here

src/dodal/beamlines/i04.py
DominicOram: The correct location is `/dls_sw/i04/software/gda/config/lookupTables/BeamLine_Undulator_toGap.txt`

DominicOram (changes requested):
The undulator is used in `i04_1`, `i22` and `p38` so we would need to add lookup paths there too.

The undulator was originally like this but @callumforrester changed it at https://github.com/DiamondLightSource/dodal/pull/461#discussion_r1578191446 to "make the read-only case simpler". Do we still have a requirement for this to be read-only in some cases? Is this to do `Adjust undulator gap to using current lookup table for initial Bragg angle` in https://github.com/DiamondLightSource/i18-bluesky/issues/12#issuecomment-2242684375? In which case you could leave it in the `undulator_dcm` and just set the undulator_dcm device to the current energy and it should correct the undulator with the correct gap?

src/dodal/beamlines/i04.py
stan-dot: applied

tests/devices/unit_tests/test_undulator_dcm.py
stan-dot: done

stan-dot:
>              raise AccessError("Undulator gap access is disabled. Contact Control Room")

that is the not covered line

src/dodal/devices/undulator.py
```diff
@@ -23,6 +53,7 @@ class Undulator(StandardReadable):
     def __init__(
         self,
         prefix: str,
+        id_gap_lookup_table_path: str,
```
DiamondJoseph, line 56: i22 undulator still doesn't have this required constructor arg

DominicOram:
> @DominicOram perhaps you see something I don't about the behavior of the device here?

Odd, I agree it looks like it's tested. I think it's fine to not worry about it. You're still at high coverage

src/dodal/devices/undulator.py
```diff
@@ -63,3 +95,58 @@ def __init__(
                 self.length = None
 
         super().__init__(name)
+
+    @AsyncStatus.wrap
+    async def set(self, value: float):
+        """
+        set the undulator gap to a given ENERGY
```
DominicOram, line 102: Nit: 
```suggestion
        Set the undulator gap to a given energy in keV.
```

src/dodal/devices/undulator.py
```diff
@@ -63,3 +95,58 @@ def __init__(
...
+                )
+            else:
+                LOGGER.debug("In test mode, not moving ID gap")
+        else:
+            LOGGER.debug(
+                "Gap is already in the correct place for the new energy value "
+                f"{energy_kev}, no need to ask it to move"
+            )
+
+    async def _gap_to_match_dcm_energy(self, energy_kev: float) -> float:
```
DominicOram, line 142: Should: As before, why are we referencing the `DCM` at all now? We wanted to decouple them and this can now run happily even if there is no DCM

src/dodal/devices/undulator.py
```diff
@@ -63,3 +95,58 @@ def __init__(
...
+        # from lookup table a get a 2d np.array converting energies to undulator gap distance
```
DominicOram, line 143: Nit:
```suggestion
        # get a 2d np.array from the lookup table that converts energies to undulator gap distance
```

tests/devices/unit_tests/test_undulator.py
```diff
@@ -95,6 +108,27 @@ async def test_length_not_propagated_if_not_supplied():
...
             name="undulator",
             poles=80,
+            id_gap_lookup_table_path=ID_GAP_LOOKUP_TABLE_PATH,
         )
     assert undulator.length is None
     assert "undulator-length" not in (await undulator.read_configuration())
+
+
+@pytest.mark.parametrize(
+    "dcm_energy, expected_output", [(5730, 5.4606), (7200, 6.045), (9000, 6.404)]
```
DominicOram, line 118: Nit: This isn't `DCM` energy, it's just energy

tests/devices/unit_tests/test_undulator.py
```diff
@@ -95,6 +108,27 @@ async def test_length_not_propagated_if_not_supplied():
...
+        == expected_output
+    )
+
+
+async def test_when_gap_access_is_disabled_set_energy_then_error_is_raised(
+    undulator,
+):
+    set_mock_value(undulator.gap_access, UndulatorGapAccess.DISABLED)
+    with pytest.raises(AccessError):
+        # AccessError("Undulator gap access is disabled. Contact Control Room")
```
DominicOram, line 133: Nit: What does this comment give us?

DominicOram (approved):
Great, thanks. Minor points, mostly take them or leave them other than @DiamondJoseph's point that `i22`/`p38` still need a lookup table

DominicOram (changes requested):
Sorry, meant request changes based on the i22 issue

src/dodal/devices/undulator.py
stan-dot: fixed

tests/devices/unit_tests/test_undulator.py
stan-dot: I left it temporarily for reference to show when the test wasn't passing. it passes now, so I deleted it

tests/devices/unit_tests/test_undulator.py
stan-dot: ok

src/dodal/devices/undulator.py
stan-dot: done

src/dodal/devices/undulator.py
stan-dot: done

src/dodal/devices/undulator.py
stan-dot: done

.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         runs-on: ["ubuntu-latest"] # can add windows-latest, macos-latest
-        python-version: ["3.10", "3.11"]
+        python-version: ["3.10", "3.11"] # add 3.12 when p4p #145 is fixed
```
DiamondJoseph, line 24: Already being tracked by an issue, let's keep this neat

src/dodal/beamlines/i22.py
```diff
@@ -198,6 +198,7 @@ def undulator(
         bl_prefix=False,
         poles=80,
         length=2.0,
+        id_gap_lookup_table_path="/dls_sw/i22/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt",
```
DiamondJoseph, line 201: Are all these files in the same format?

DiamondJoseph: As in both file extension and internal format

DiamondJoseph: This will require making blueapi mount at least `/dls_sw/i22/software/daq_configuration/lookup/` or move that file into a Kubernetes ConfigMap and mounting it- how complex is the file and how often is it changing?

src/dodal/devices/undulator.py
```diff
@@ -1,20 +1,50 @@
...
+
+from .util.lookup_tables import energy_distance_table
+
+
+class AccessError(Exception):
+    pass
+
+
+# Enable to allow testing when the beamline is down, do not change in production!
+TEST_MODE = False
```
DiamondJoseph, line 26: Can we define 2 different implementations of MovableUndulator and switch between them in the dodal module, rather than having this global and the branching behaviour within this class' set method? I think it's clearer what the intention is.

How does TEST_MODE differ from backing everything with simulated signals? TEST_MODE isn't mentioned on the linked tickets, what're the requirements and where are they defined?

src/dodal/devices/undulator.py
```diff
@@ -1,20 +1,50 @@
...
-class Undulator(StandardReadable):
+def _get_closest_gap_for_energy(
+    dcm_energy_ev: float, energy_to_distance_table: ndarray
+) -> float:
+    table = energy_to_distance_table.transpose()
+    idx = argmin(np.abs(table[0] - dcm_energy_ev))
+    return table[1][idx]
+
+
+class Undulator(StandardReadable, Movable):
```
DiamondJoseph, line 47: Does i18 not have a DCM and that's why we can't use the UndulatorDCM? I don't know the context

DiamondJoseph (changes requested):

src/dodal/devices/undulator.py
stan-dot: I just copied the TEST_MODE over from the dcm_undulator. I don't have the historical knowledge how it came to be

src/dodal/devices/undulator.py
stan-dot: the context is that undulatordcm is moved as a set in i04 etc but at i18 they are moved independently when doing alignment

src/dodal/beamlines/i22.py
stan-dot: the one for i22 is empty. the one for i03 was made using harmonics 3 5 7 and the date indicated inside is 2022

src/dodal/devices/undulator.py
stan-dot: @DominicOram  is it ok to keep this TEST_MODE like that in the undulator.py?

.github/workflows/ci.yml
stan-dot: I think this got here because of rebase was not possible directly, so I did some merge?

src/dodal/devices/undulator.py
DominicOram: https://github.com/DiamondLightSource/dodal/issues/754 describes the usecase and tracks fixing it more nicely. Can we just leave the code as it is, with a link to this issue?

stan-dot: splendid, keeping as-is

src/dodal/beamlines/i22.py
stan-dot: I think the Undulator device won't be used in bluesky anytime soon. I am not familiar with how alignment works there at the moment, and in the near/mid term future it will use the new redis based lookup tables. 

for this PR I suggest we just leave it as-is, I can watch at i22 when could we change this.

@DiamondJoseph can we resolve this thread?

src/dodal/devices/undulator.py
DiamondJoseph: Can we then get a link to the github issue as part of the comment @stan-dot?

stan-dot: ah right, I forgot

stan-dot: solved now

DominicOram (approved):
Thanks, one minor change, otherwise looks good.

src/dodal/devices/undulator.py
```diff
@@ -63,3 +98,61 @@ def __init__(
...
+    async def set(self, value: float):
+        """
+        Set the undulator gap to a given energy in keV
+
+        Args:
+            value: energy in keV
+        """
+        await asyncio.gather(
+            self._set_undulator_gap(value),
+        )
```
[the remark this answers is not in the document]
stan-dot: must have been by mistake, just fixed this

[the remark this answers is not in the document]
DominicOram: Great, thanks, feel free to merge then

PR #725, opened 2024-08-06
Changes to zocalo interaction to enable containerisation

rtuck99:
Removes dependency on the system hostname from zocalo so that we can be run in a container

See DiamondLightSource/hyperion#1511

### Checks for reviewer


dperl-dls (approved):
lgtm

PR #726, opened 2024-08-06
convert det motion to ophyd-async

### LINKED ISSUE #692: Convert to ophyd-async: detector_motion

dperl-dls:
Fixes #692

1. Run tests and beamline connect scripts

### Checks for reviewer


src/dodal/devices/detector/detector_motion.py
```diff
@@ -10,28 +13,34 @@ class ShutterState(IntEnum):
...
+        )  # on limit = 1, off = 0
+        self.z_disabled = epics_signal_r(
+            float, f"{prefix}{device_prefix}Z:DISABLED"
+        )  # robot interlock, 0=ok to move, 1=blocked
+        self.crate_power = epics_signal_r(
+            float, f"{prefix}{device_prefix}CRATE2_HEALTHY"
+        )  # returns 0 if no power
+        self.in_robot_load_safe_position = epics_signal_r(
+            float, f"{prefix}{device_prefix}GPIO_INP_BITS.B2"
+        )  # returns 1 if safe
```
DominicOram, line 44: Must: These are actually PVs on the `pmac` and so `dodal connect i03` is failing

DominicOram, line 44: Could: It's annoying these are `ai` records and not `bi`/`mbbi` records so we can actually write enums against them. Would be good to raise a (low priority) ticket in controls to get them cleaned up but also we don't currently use them really so I'm not that fussed

src/dodal/devices/detector/detector_motion.py
```diff
@@ -10,28 +13,34 @@ class ShutterState(IntEnum):
...
+        self.downstream_x = Motor(f"{prefix}{device_prefix}DOWNSTREAMX")
+        self.x = Motor(f"{prefix}{device_prefix}X")
+        self.y = Motor(f"{prefix}{device_prefix}Y")
+        self.z = Motor(f"{prefix}{device_prefix}Z")
+        self.yaw = Motor(f"{prefix}{device_prefix}YAW")
+        self.shutter = Motor(f"{prefix}{device_prefix}SET_SHUTTER_STATE")
+
+        self.shutter = epics_signal_rw(
+            float, "-MO-DET-01:SET_SHUTTER_STATE"
+        )  # 0=closed, 1=open monitors
```
DominicOram, line 28: Must: I think you're right with the second one, it's a `bo`, not a motor. This also means we should be using an enum to read/set it.

src/dodal/devices/detector/detector_motion.py
```diff
@@ -1,7 +1,10 @@
 from enum import IntEnum
 
-from ophyd import Component as Cpt
-from ophyd import Device, EpicsMotor, EpicsSignal, EpicsSignalRO
+# from ophyd import Component as Cpt
+# from ophyd import Device, EpicsMotor, EpicsSignal, EpicsSignalRO
```
DominicOram, line 4: Should: Remove commented out code

DominicOram (changes requested):
Thanks, not sure if this was actually ready for review but saw it in the list so have put some bits on it.

DominicOram (approved):
Thanks. Unfortunately we're still needing support 3.10 until next year (https://numpy.org/neps/nep-0029-deprecation_policy.html) so had to remove the `StrEnum`. Otherwise all good.

PR #733, opened 2024-08-07
Mx bluesky 135 stream to murko

DominicOram:
Fixes https://github.com/DiamondLightSource/mx_bluesky/issues/135

1. Compare against docs in mx_bluesky and confirm behavior makes sense
2. Confirm unit tests pass

### Checks for reviewer


dperl-dls (approved):
This seems fine for now. It could easily be a lot more generic if e.g. the stream url was configurable, and I wonder if the performance will always be good enough, but I think those are questions for later

DominicOram:
> the stream url was configurable, 

Not sure what you mean here the URL is taken from the OAV. Do you mean make it generic to work with any given URL, even if it doesn't come from an OAV?

> and I wonder if the performance will always be good enough, but I think those are questions for later

Yh, I've done no performance testing. Works for this use case though

PR #742, opened 2024-08-12
(DiamondLightSource/hyperion#1517) Fix panda subdirectory not created

rtuck99:
Fixes part of DiamondLightSource/hyperion#1517

The 'PandASubdirectoryProvider' now allows the filename suffix to be specified, and it also ensures the 'panda' subdirectory is created.

See also DiamondLightSource/hyperion#1520

### Checks for reviewer

### DIFF
--- a/src/dodal/common/udc_directory_provider.py
+++ b/src/dodal/common/udc_directory_provider.py
@@ -8,7 +8,7 @@ from dodal.log import LOGGER
 
 class PandASubdirectoryProvider(UpdatingDirectoryProvider):
     """Directory provider for the HDFPanda. Points to a panda subdirectory within the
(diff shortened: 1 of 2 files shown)


src/dodal/common/udc_directory_provider.py
```diff
@@ -8,7 +8,7 @@
 
 class PandASubdirectoryProvider(UpdatingDirectoryProvider):
     """Directory provider for the HDFPanda. Points to a panda subdirectory within the
-    directory path provided, which must exist before attempting to arm the PCAP block"""
```
olliesilvester, line 11: Isn't this still true?

olliesilvester (approved):
Great! Thanks

PR #749, opened 2024-08-15
Fix exception calling getpass() when inside container

rtuck99:
The previous fix #725 did not work because getpass() was still called and failed; this ensures it is not called unless necessary.

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/zocalo/zocalo_interaction.py
+++ b/src/dodal/devices/zocalo/zocalo_interaction.py
@@ -39,7 +39,12 @@ class ZocaloStartInfo:
 
 
 def _get_zocalo_headers() -> tuple[str, str]:
-    user = os.environ.get("ZOCALO_GO_USER", getpass.getuser())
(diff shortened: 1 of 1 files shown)


src/dodal/utils.py
```diff
@@ -249,9 +249,9 @@ def collect_factories(
         if (
             callable(var)
             and is_any_device_factory(var)
-            and (include_skipped or not _is_device_skipped(var))
+            and (include_skipped or not _is_device_skipped(var))  # type: ignore
         ):
-            factories[var.__name__] = var
+            factories[var.__name__] = var  # type: ignore
```
DominicOram, line 254: Should: Thanks for fixing this but I think https://github.com/DiamondLightSource/dodal/pull/783 is the better fix, could you review and merge that in and revert this?

DominicOram (changes requested):
Great, thanks, one comment

DominicOram (approved):
Great, thank you!

PR #755, opened 2024-08-20
ApertureScatterguard now set using a well defined enum

### LINKED ISSUE #740: Clean up aperture scatterguard interface
The interface for the aperture scatterguard is currently very confusing as we have to build up `AperturePositions` from a file and so it looks like you may expect it to be an enum but it's not. This means that you always need to use the `AperturePositions` object owned by the `ApertureScatterguard` object (which may be `None`).

e.g. To currently set the position you need to do:
```python
    assert composite.aperture_scatterguard.aperture_positions is not None
    yield from bps.abs_set(
        composite.aperture_scatterguard,
        composite.aperture_scatterguard.aperture_positions.ROBOT_LOAD,
        group=OAV_SNAPSHOT_SETUP_GROUP,
    )
```

It would be cleaner to have a common enum that we don't need to initialise e.g.
```python
    yield from bps.abs_set(
        composite.aperture_scatterguard,
        AperturePositions.ROBOT_LOAD,
        group=OAV_SNAPSHOT_SETUP_GROUP,
    )
```

The `ApertureScatterguard` would then internally map this to a position. This more closely matches what we have for other devices and will lead to less potential error.

## Acceptance Criteria
* The changes above are made

DominicOram:
Fixes #740

1. Confirm interface to `ApertureScatterguard` is cleaner
1. Confirm tests still pass

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -1,5 +1,6 @@
 from ophyd_async.panda import HDFPanda
 
+from dodal.common.beamlines.beamline_parameters import get_beamline_parameters
 from dodal.common.beamlines.beamline_utils import (
     device_instantiation,
     get_directory_provider,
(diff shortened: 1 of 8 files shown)


dperl-dls (changes requested):
See comment on the other PR, otherwise looks good

PR #763, opened 2024-08-29
ZocaloResults: add parameter to use results from GPU

### LINKED ISSUE #559: Prepare `ZocaloResults` to use multiple results sources
I03 will soon have both GPU-based and traditional xray-centring results available.  `ZocaloResults`  should fetch both of these from the queue and pick which one to use based on input from hyperion.

olliesilvester:
Fixes #559 

See related `mx_bluesky` PR: https://github.com/DiamondLightSource/mx-bluesky/pull/445

Check this PR adds and tests the following behavior:

If the GPU is toggled on, then:
- Wait for results twice
- If CPU arrives before GPU, then warn and continue
- if CPU results differ from GPU within a tolerance, log the two results and continue
- if GPU times out but CPU results arrive, warn and continue
- if GPU results arrive but CPU results timeout, error
- Always use CPU results after comparing the two
If it's off:
- Ignore results from GPU

### Checks for reviewer


src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -165,7 +167,10 @@ async def trigger(self):
             )
 
             raw_results = self._raw_results_received.get(timeout=self.timeout_s)
-            LOGGER.info(f"Zocalo: found {len(raw_results['results'])} crystals.")
+            source_of_results = "CPU" if raw_results["ispyb_ids"]["gpu"] == None else "GPU"
```
ndevenish, line 170: is `raw_results["ispyb_ids"]` a defaultdict? If not then won't this cause an `IndexError`? Possibly:
```suggestion
            source_of_results = "CPU" if not raw_results["ispyb_ids"].get("gpu") else "GPU"
```
(Although `"GPU" if raw_results["ispyb_ids"].get("gpu") else "CPU"` is more terse, the focus being on the presumed default of "CPU" might be justified)

olliesilvester: According to https://docs.python.org/3/library/stdtypes.html#dict.get , an ordinary dict returns None rather than raising an IndexError if the key doesn't exist.

ndevenish: Also now out-of-date, but `== None` is normally not what you want, generally you want to `is None`. But the falseish comparison beats both.

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -237,9 +242,18 @@ def _receive_result(
...
-                {"results": results, "ispyb_ids": recipe_parameters}
-            )
+
+            if self.use_fastest_zocalo_result:
+                self._raw_results_received.put(
+                    {"results": results, "ispyb_ids": recipe_parameters}
+                )
+            else:
+                #Only add to queue if results are from CPU
+                if recipe_parameters.get('gpu') == None:
```
ndevenish, line 252: Also here
```suggestion
                if recipe_parameters.get('gpu') is None:
```

ndevenish: In fact again truthiness of None is probably better
```suggestion
                if recipe_parameters.get('gpu'):
```

olliesilvester: Yeah, I always prefer to explicitly compare it to None if it can be True/False or None

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -56,12 +56,25 @@ def bbox_size(result: XrcResult):
     ]
 
 
+# TODO this may give uneccessary warnings if dicts are different within rounding errors
+def get_dict_differences(
```
olliesilvester, line 60: This will be sensitive to rounding errors, so I could change the function to something a bit more complicated to avoid that

tests/devices/unit_tests/test_zocalo_results.py
```diff
@@ -250,3 +260,183 @@ async def test_when_exception_caused_by_zocalo_message_then_exception_propagated
...
+    mock_connection, RE: RunEngine
+):
+    zocalo_results = ZocaloResults(
+        name="zocalo", zocalo_environment="dev_artemis", use_cpu_and_gpu_zocalo=True
+    )
+    await zocalo_results.connect()
+    await zocalo_results.stage()
+    zocalo_results._raw_results_received.get = MagicMock(side_effect=Empty)
+    with pytest.raises(NoResultsFromZocalo):
+        await zocalo_results.trigger()
```
olliesilvester, line 393: There was a strange thing here where this test would fail if I used `RE(bps.trigger(zocalo_results)` instead, even though debugging the pytest showed the exception was being raised

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -79,6 +92,7 @@ def __init__(
         self._prefix = prefix
         self._raw_results_received: Queue = Queue()
         self.transport: CommonTransport | None = None
+        self.use_cpu_and_gpu_zocalo = use_cpu_and_gpu_zocalo
```
DominicOram, line 95: Nit: This device is already has `zocalo` in the name so having this variable with `zocalo` in is a bit redundant. It would also be good to have a docstring on it

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -165,7 +179,40 @@ async def trigger(self):
             )
 
             raw_results = self._raw_results_received.get(timeout=self.timeout_s)
-            LOGGER.info(f"Zocalo: found {len(raw_results['results'])} crystals.")
+            source_of_first_results, source_of_second_results = (
+                ("CPU", "GPU")
+                if not raw_results["ispyb_ids"].get("gpu")
+                else ("GPU", "CPU")
+            )
```
DominicOram, line 186: Could: There's a bit of an inbuilt assumption here that there's not two from the same source, which is probably fine but I also don't really like the multiple use of strings and stuff. What about something like:
```python
class ZocaloSource(enum):
    CPU = "CPU"
    GPU = "GPU"

def source_from_results(results):
    return ZocaloSource.GPU if results["ispyb_ids"].get("gpu") else ZocaloSource.CPU
```
then each time we get results call `source_from_results(results)`

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -237,9 +284,17 @@ def _receive_result(
...
-                {"results": results, "ispyb_ids": recipe_parameters}
-            )
+
+            if self.use_cpu_and_gpu_zocalo:
+                self._raw_results_received.put(
+                    {"results": results, "ispyb_ids": recipe_parameters}
+                )
+            else:
+                # Only add to queue if results are from CPU
+                if not recipe_parameters.get("gpu"):
```
DominicOram, line 294: Should: A unit test covering this would be good.

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -237,9 +284,17 @@ def _receive_result(
             self.transport.ack(header)  # type: ignore # we create transport here
 
             results = message.get("results", [])
-            self._raw_results_received.put(
-                {"results": results, "ispyb_ids": recipe_parameters}
-            )
+
+            if self.use_cpu_and_gpu_zocalo:
+                self._raw_results_received.put(
+                    {"results": results, "ispyb_ids": recipe_parameters}
```
DominicOram, line 290: Could: `ispyb_ids` isn't a great name any more as it contains other stuff.

tests/devices/unit_tests/test_zocalo_results.py
```diff
@@ -204,14 +206,20 @@ async def test_zocalo_results_trigger_log_message(
     mock_wrap_subscribe, mock_logger, RE: RunEngine
 ):
     zocalo_results = ZocaloResults(
-        name="zocalo", zocalo_environment="dev_artemis", timeout_s=2
+        name="zocalo",
+        zocalo_environment="dev_artemis",
+        timeout_s=2,
```
DominicOram, line 211: Could: It's not great that we're waiting on this timeout for the test. Ideally we would mock it out so there's no wait on it but if not could we at least drop it down to as low as possible?

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -165,7 +179,40 @@ async def trigger(self):
...
+            source_of_first_results, source_of_second_results = (
+                ("CPU", "GPU")
+                if not raw_results["ispyb_ids"].get("gpu")
+                else ("GPU", "CPU")
+            )
+
+            # Wait for results from CPU and GPU, warn and continue if one timed out, error if both time out
+            if self.use_cpu_and_gpu_zocalo:
+                if source_of_first_results == "CPU":
+                    LOGGER.warning("Recieved zocalo results from CPU before GPU")
```
DominicOram, line 191: Nit:
```suggestion
                    LOGGER.warning("Received zocalo results from CPU before GPU")
```

DominicOram (changes requested):
Great, thank you. Couple of comments in the code.

src/dodal/devices/zocalo/zocalo_results.py
olliesilvester: I couldn't find any function which did something like math.isclose which worked for this kind of dictionary, so this was me being lazy instead. I think iterating through each key and treating lists differently to floats is the way to do this properly

DominicOram: It looks like `deepdiff` https://zepworks.com/deepdiff/current/diff.html might do this with `math_epsilon` being the tolerance. I think on balance it's fine to add another dependency but given this check is probably temporary add a comment next to the dependency pointing to removing it in a ticket that around cleaning up GPU stuff

olliesilvester: Ah, nice!

src/dodal/devices/zocalo/zocalo_results.py
olliesilvester: I like the function, but I think it's useful to assume that they're not for the same source so that the logging message

```
LOGGER.warning(
                        f"Zocalo results from {source_of_second_results} timed out. Using results from {source_of_first_results}"
                    )
```
Is more useful.

src/dodal/devices/zocalo/zocalo_results.py
DominicOram: So have I tbf

DominicOram (approved):
Great, thank you!

PR #765, opened 2024-08-31
Fixes the OAVToRedisForwarder based on beamline tests

DominicOram:
See https://github.com/DiamondLightSource/mx-bluesky/issues/195

1. Confirm `dodal connect i04` works
2. Confirm new system and unit tests work

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -410,9 +410,11 @@ def oav_to_redis_forwarder(
     """
     return device_instantiation(
         OAVToRedisForwarder,
-        "oav_to_redis",
+        "oav_to_redis_forwarder",
         "",
         wait_for_connection,
         fake_with_ophyd_sim,
-        params=OAVConfigParams(ZOOM_PARAMS_FILE, DISPLAY_CONFIG),
+        redis_host=REDIS_HOST,
+        redis_password=REDIS_PASSWORD,
+        redis_db=7,
     )
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -411,7 +411,7 @@ def oav_to_redis_forwarder(
     return device_instantiation(
         OAVToRedisForwarder,
         "oav_to_redis_forwarder",
-        "",
+        "-DI-OAV-01:",
         wait_for_connection,
         fake_with_ophyd_sim,
         redis_host=REDIS_HOST,
--- a/src/dodal/devices/oav/oav_to_redis_forwarder.py
+++ b/src/dodal/devices/oav/oav_to_redis_forwarder.py
@@ -81,7 +81,7 @@ class OAVToRedisForwarder(StandardReadable, Flyable):
         self.uuid_setter(image_uuid := str(uuid.uuid4()))
         img = Image.open(io.BytesIO(jpeg_bytes))
         image_data = pickle.dumps(np.asarray(img))
-        sample_id = await self.sample_id.get_value()
+        sample_id = str(await self.sample_id.get_value())
         await self.redis_client.hset(sample_id, image_uuid, image_data)  # type: ignore
         await self.redis_client.expire(sample_id, timedelta(days=self.DATA_EXPIRY_DAYS))
         LOGGER.debug(f"Sent frame to redis key {sample_id} with uuid {image_uuid}")
--- a/src/dodal/devices/oav/oav_to_redis_forwarder.py
+++ b/src/dodal/devices/oav/oav_to_redis_forwarder.py
@@ -2,12 +2,12 @@ import asyncio
 import io
 import pickle
 import uuid
-from collections.abc import Callable
+from collections.abc import Awaitable, Callable
 from datetime import timedelta
 
 import numpy as np
 from aiohttp import ClientResponse, ClientSession
-from bluesky.protocols import Flyable
+from bluesky.protocols import Flyable, Stoppable
 from ophyd_async.core import AsyncStatus, StandardReadable
 from ophyd_async.core.signal import soft_signal_r_and_setter, soft_signal_rw
 from ophyd_async.epics.signal import epics_signal_r
@@ -26,7 +26,7 @@ async def get_next_jpeg(response: ClientResponse) -> bytes:
             return line + await response.content.readuntil(JPEG_STOP_BYTE)
 
 
-class OAVToRedisForwarder(StandardReadable, Flyable):
+class OAVToRedisForwarder(StandardReadable, Flyable, Stoppable):
     """Forwards OAV image data to redis. To use call:
 
     > bps.kickoff(oav_forwarder)
@@ -65,6 +65,8 @@ class OAVToRedisForwarder(StandardReadable, Flyable):
             host=redis_host, password=redis_password, db=redis_db
         )
 
+        self._stop_flag = False
+
         self.sample_id = soft_signal_rw(int, initial_value=0)
 
         # The uuid that images are being saved under, this should be monitored for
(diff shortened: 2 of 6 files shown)


src/dodal/beamlines/i04.py
```diff
@@ -410,9 +410,11 @@ def oav_to_redis_forwarder(
     """
     return device_instantiation(
         OAVToRedisForwarder,
-        "oav_to_redis",
+        "oav_to_redis_forwarder",
         "",
```
olliesilvester, line 414: Do we not need a prefix here?

src/dodal/devices/oav/oav_to_redis_forwarder.py
```diff
@@ -79,20 +81,32 @@ async def _get_frame_and_put_to_redis(self, response: ClientResponse):
         self.uuid_setter(image_uuid := str(uuid.uuid4()))
         img = Image.open(io.BytesIO(jpeg_bytes))
         image_data = pickle.dumps(np.asarray(img))
-        await self.redis_client.hset(self.redis_key, image_uuid, image_data)  # type: ignore
-        LOGGER.debug(f"Sent frame to redis key {self.redis_key} with uuid {image_uuid}")
+        sample_id = str(await self.sample_id.get_value())
```
olliesilvester, line 84: Which part of the code actually puts things into the `sample_id` signal?

src/dodal/devices/oav/oav_to_redis_forwarder.py
```diff
@@ -79,20 +81,32 @@ async def _get_frame_and_put_to_redis(self, response: ClientResponse):
...
+        await self.redis_client.hset(sample_id, image_uuid, image_data)  # type: ignore
+        await self.redis_client.expire(sample_id, timedelta(days=self.DATA_EXPIRY_DAYS))
+        LOGGER.debug(f"Sent frame to redis key {sample_id} with uuid {image_uuid}")
 
-    async def _stream_to_redis(self):
+    async def _open_connection_and_(self, function_to_do: Callable):
```
olliesilvester, line 89: Nit: Function name looks a bit weird to me, maybe `_open_connection_and_do_function`?

tests/devices/system_tests/test_oav_to_redis_system.py
```diff
@@ -0,0 +1,51 @@
...
+    oav_to_redis_forwarder.stream_url = AsyncMock()
+    oav_to_redis_forwarder.stream_url.get_value.return_value = (
+        "http://www.this_is_not_a_valid_webpage.com/"
+    )
+    with pytest.raises(ClientConnectorError):
+        await oav_to_redis_forwarder.kickoff()
+
+
+@pytest.mark.s03  # Doesn't actually depend on s03, just on DLS network. See https://github.com/DiamondLightSource/mx-bluesky/issues/183
+async def test_given_stream_url_is_real_webpage_but_not_mjpg_when_kickoff_then_error(
```
olliesilvester, line 33: This test isn't passing for me device the PV's haven't been connected. Using a mock device with the `DeviceCollector` will fix it

tests/devices/system_tests/test_oav_to_redis_system.py
```diff
@@ -0,0 +1,51 @@
...
+    URL = "https://www.google.com/"
+    oav_to_redis_forwarder.stream_url = AsyncMock()
+    oav_to_redis_forwarder.stream_url.get_value.return_value = URL
+    with pytest.raises(ValueError) as e:
+        await oav_to_redis_forwarder.kickoff()
+    assert URL in str(e.value)
+
+
+@pytest.mark.s03  # Doesn't actually depend on s03, just on DLS network. See https://github.com/DiamondLightSource/mx-bluesky/issues/183
+async def test_given_valid_stream_when_kickoff_then_multiple_images_written_to_redis(
```
olliesilvester, line 45: Similar to last comment, this is using the real device with mock=False, so it could cause the real device to do things. Also, it isn't passing as the signals haven't been connected to

olliesilvester (changes requested):
Looks mostly okay, but I can't see how the PV prefix is set, ~or how the `sample_id` is being set~. I could have missed something there though. Also some comments on the system tests not using mock=True

src/dodal/devices/oav/oav_to_redis_forwarder.py
olliesilvester: nvm, just saw the `mx_bluesky` PR!

tests/devices/system_tests/test_oav_to_redis_system.py
DominicOram: This one is designed to use the real device I have made the comment clearer

src/dodal/beamlines/i04.py
DominicOram: > I can't see how the PV prefix is set

`device_instantiation` automatically adds the beamline prefix to the device so if the device part is in the device we can just leave it blank

src/dodal/devices/oav/oav_to_redis_forwarder.py
```diff
@@ -86,7 +86,7 @@ async def _get_frame_and_put_to_redis(self, response: ClientResponse):
         await self.redis_client.expire(sample_id, timedelta(days=self.DATA_EXPIRY_DAYS))
         LOGGER.debug(f"Sent frame to redis key {sample_id} with uuid {image_uuid}")
 
-    async def _open_connection_and_(self, function_to_do: Callable):
+    async def _open_connection_and_do_function(self, function_to_do: Callable):
```
olliesilvester, line 89: Could: Sorry I should've said this before, but it might be nice to type hint `function_to_do` so that it must accept `ClientResponse` as a parameter, and an optional `str`

olliesilvester, line 89:
```suggestion
    async def _open_connection_and_do_function(self, function_to_do: Callable[[ClientResponse, Optional[str]], Awaitable]):
```

Along with 
from typing import Awaitable, Optional, Callable

tests/devices/system_tests/test_oav_to_redis_system.py
olliesilvester: Ah ok, and I assume that's fine because `oav_forwarder.redis_client.hset` is mocked

src/dodal/devices/oav/oav_to_redis_forwarder.py
```diff
@@ -103,9 +103,9 @@ async def _confirm_mjpg_stream(self, response, stream_url):
 
     @AsyncStatus.wrap
     async def kickoff(self):
-        await self._open_connection_and_(self._confirm_mjpg_stream)
+        await self._open_connection_and_do_function(self._confirm_mjpg_stream)
         self.forwarding_task = asyncio.create_task(
```
olliesilvester, line 107: If we error between `kickoff` and `complete` then this task will never be cancelled. I think either this part of code or the plan should make sure the task is cancelled upon exceptions

olliesilvester (approved):
A few more comments, but I will approve to avoid blocking anything

tests/devices/system_tests/test_oav_to_redis_system.py
DominicOram: Yep, even if it wasn't mocked it doesn't have the credentials to do anything anyway. Ultimately it would be nice to add a system test that also tested the redis part but that's probably for a later date

src/dodal/devices/oav/oav_to_redis_forwarder.py
DominicOram: Good spot. I think if the RE fails it will call stop on all the devices and so by implementing that we fix this issue

DominicOram: In doing this I also realised we're using cancel poorly, have refactored into a `stop` flag

olliesilvester (approved):
Looks good

PR #767, opened 2024-09-02
Make sure backlight is moved in/out when set

DominicOram:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/198

1. Confirm this would make sure we move the backlight out properly

### Checks for reviewer


olliesilvester (approved):
Maybe should write tests here but the logic is so simple so I doubt I'd have bothered either. How confident are we about the 1 second?

olliesilvester (approved):
Thanks!

PR #768, opened 2024-09-02
Supress warnings from setuptools_dso

DominicOram:
Currently we're getting depreciation warnings on `setuptools_dso`, we're now treating these as errors so running tests etc. will fail. This will temporarily suppress the warning.

1. On `main` run `pytest .` and see it fails with a depreciation warning on `setuptools_dso`
2. Do the same on this branch and it passes

### Checks for reviewer


noemifrisina (approved):
lgtm, thanks.
Do we have an issue here in Dodal to remove this once the issue in epics-base is merged? If not, can we open one please?

DominicOram:
I've made https://github.com/DiamondLightSource/mx-bluesky/issues/451 to cover both here and `mx-bluesky` as we'll need to do both

PR #769, opened 2024-09-02
Refactor ApertureScatterguard to return individual signals

DiamondJoseph:
Fixes an issue discovered as part of #679, https://github.com/bluesky/ophyd-async/issues/550#issuecomment-2324227988 : SoftConverter is no longer exported from ophyd_async as of 0.5.0, and @coretl @danielballan think that https://github.com/bluesky/ophyd-async/issues/310 should not include being able to pass arbitrary objects, only those that are tableable [tabulable? tabulatable?]. This change therefore adjusts how the ApertureScatterguard returns its signals to be nested under its child devices but as invidual signals.

1. Ensure that the ApertureScatterguard can still connect and be read
2. The documents for the signals contained inside the device should only use primitive float types.

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/aperturescatterguard.py
+++ b/src/dodal/devices/aperturescatterguard.py
@@ -118,6 +118,7 @@ class ApertureScatterguard(StandardReadable, Movable):
             ],
             wrapper=HintedSignal,
         )
+        super().__init__(name)
 
     def get_position_from_gda_aperture_name(
         self, gda_aperture_name: str
--- a/src/dodal/devices/aperturescatterguard.py
+++ b/src/dodal/devices/aperturescatterguard.py
@@ -1,12 +1,16 @@
+from __future__ import annotations
+
 import asyncio
-from collections import namedtuple
-from dataclasses import asdict, dataclass
 from enum import Enum
 
-from bluesky.protocols import Movable, Reading
-from event_model.documents.event_descriptor import DataKey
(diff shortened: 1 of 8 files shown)


DiamondJoseph:
This then becomes a requirement for #679

DiamondJoseph:
Making a Hyperion patchset now

src/dodal/devices/aperturescatterguard.py
```diff
@@ -17,266 +21,214 @@ class InvalidApertureMove(Exception):
...
-    ],
-)
-
+class AperturePosition(BaseModel):
+    aperture_x: float
+    aperture_y: float
+    aperture_z: float
+    scatterguard_x: float
+    scatterguard_y: float
+    radius: float | None = Field(json_schema_extra={"units": "µm"}, default=None)
```
dperl-dls, line 30: :heart:

dperl-dls (approved):
This is great, thanks

DominicOram:
A side effect of this is that we're no longer reading the state of the hardware whenever we read `selected_aperture`. This means that the device itself is holding on to the state of the beamline based on what state it last moved to. This means that the state could be wrong if:
* The device has just been initialised
* The hardware has been moved via EPICS/manually
* There was a failure in moving the device

Was this a deliberate design decision @DiamondJoseph?

PR #773, opened 2024-09-04
Wait for scan status to change in the PMAC ProgramRunner

### LINKED ISSUE #771: PMAC ProgramRunner should first wait for status to change
In the `PMAC` device, the `ProgramRunner` sends the program number to the pmac_string and then monitors the value of the status PV `BL24I-MO-STEP-14:signal:P2401`, waiting until it goes to 0 signaling the end of the collection. However, from testing on the beamilne it looks like it takes a bit of time for this value to be set to 1 in the first place, meaning that the code exits right away.

## Acceptance Criteria
* Once the pmac_string is set, the program first check that the status PV goes to 1 - idicating collection has started - and then monitors it until is goes back to 0.

noemifrisina:
Fixes #771 

Also adds a command to abort a program.

1. Run tests

### DIFF
--- a/src/dodal/devices/i24/pmac.py
+++ b/src/dodal/devices/i24/pmac.py
@@ -136,6 +136,30 @@ class ProgramRunner(SignalRW):
         await wait_for_value(self.status, ScanState.DONE, timeout)
 
 
+class ProgramAbort(Triggerable):
+    """Abort a data collection by setting the PMAC string and then wait for the \
+        status value to go back to 0.
+    """
+
+    def __init__(
+        self,
+        pmac_str_sig: SignalRW,
+        status_sig: SignalR,
+    ) -> None:
+        self.signal = pmac_str_sig
+        self.status = status_sig
+
+    @AsyncStatus.wrap
+    async def trigger(self):
+        await self.signal.set("A", wait=True)
+        await self.signal.set("P2401=0", wait=True)
+        await wait_for_value(
+            self.status,
+            ScanState.DONE,
+            timeout=DEFAULT_TIMEOUT,
+        )
+
+
 class PMAC(StandardReadable):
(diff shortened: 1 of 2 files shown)


src/dodal/devices/i24/pmac.py
```diff
@@ -141,11 +142,42 @@ async def set(self, value: int, wait=True, timeout=None):
...
-        ), f"ProgramRunner does not support calculating timeout itself, {timeout=}"
+        ), f"ProgramRunner does not support calculating timeout itself, {timeout}"
         await self.signal.set(prog_str, wait=wait)
+        # First wait for signal to go to 1, then wait for the scan to finish.
+        await wait_for_value(
+            self.status,
+            ScanState.RUNNING,
+            timeout=DEFAULT_TIMEOUT,
+        )
         await wait_for_value(self.status, ScanState.DONE, timeout)
```
DominicOram, line 153: Should: The timeout is now wrong. At it's worst this function as a whole won't timeout until `DEFAULT_TIMEOUT` + `timeout`. You can fix this by putting them in their own `asyncio.wait_for`, see for example the `BartRobot`

DominicOram (changes requested):
Great, thanks, one comment

src/dodal/devices/i24/pmac.py
noemifrisina: Added a comment explaining the timeout. In short, the name is misleading but forced by the signature of `set`. The second timeout should be the total time the scan should take so that it doesn't timeout while the motion program is still running. The first one is just a check that the scan actually starts so it doesn't need to be as long.

DominicOram (approved):
Great, thanks. We should probably use kickoff and complete, I have made https://github.com/DiamondLightSource/dodal/issues/788

PR #789, opened 2024-09-14
Read hardware on every read of the aperture scatterguard 

### LINKED ISSUE #782: Read aperture/scatterguard hardware on read
Following https://github.com/DiamondLightSource/dodal/pull/769 a side effect of this is that we're no longer reading the state of the hardware whenever we read `selected_aperture`. This means that the device itself is holding on to the state of the beamline based on what state it last moved to. This means that the state could be wrong if:
* The device has just been initialised
* The hardware has been moved via EPICS/manually
* There was a failure in moving the device

## Acceptance Criteria
* Hardware is read when the device is read
* The device holds no assumptions about the state of the hardware

DominicOram:
Fixes #782

1. Confirm tests now all pass
2. Confirm docs make sense

### Checks for reviewer


src/dodal/common/signal_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        await self._update_value()
+        return await super().get_value()
+
+
+def create_hardware_backed_soft_signal(
+    datatype: type[T],
+    get_from_hardware_func: Callable[[], Coroutine[Any, Any, T]],
+    units: str | None = None,
+):
+    metadata = SignalMetadata(units=units, precision=None)
```
DiamondJoseph, line 44: nit: for completeness
```suggestion
    units: str | None = None,
    precision: int | None = None,
):
    metadata = SignalMetadata(units=units, precision=precision)
```

src/dodal/common/signal_utils.py
```diff
@@ -0,0 +1,49 @@
...
+        super().__init__(*args, **kwargs)
+
+    async def _update_value(self):
+        new_value = await self.get_from_hardware_func()
+        await self.put(new_value)
+
+    async def get_datakey(self, source: str) -> DataKey:
+        await self._update_value()
+        return await super().get_datakey(source)
+
```
DiamondJoseph, line 29: nit: The datakey should be consistent before/after reading, unless there's some SoftSignal before-first-read weirdness?

```suggestion
```

src/dodal/devices/aperturescatterguard.py
```diff
@@ -119,7 +121,9 @@ def __init__(
             ],
         )
         with self.add_children_as_readables(HintedSignal):
-            self.selected_aperture = soft_signal_rw(ApertureValue)
+            self.selected_aperture = create_hardware_backed_soft_signal(
+                ApertureValue, self.get_current_aperture_position
```
DiamondJoseph, line 125: nit: `get_current_aperture_position` should be made private on the device class (I should have done this in the prior change): its value is currently exposed through device.read() [intended design] and would allow future changes to the method if required.

nit because I don't know how simple the change would be in mx-bluesky and I don't want to hold up broken device

DiamondJoseph (approved):

coretl:
This is better than what was there originally, but I'm concerned that you don't override `subscribe` so you have a Signal where `get_value` gives you an enum, but `subscribe` gives you a float.

I wonder if there is a case for supplying both a `forward` and `inverse` transform, that operate on values rather than Signals, and making a `DerivedSignalBackend` that applies the transforms on get, put and subscribe?

DominicOram:
> This is better than what was there originally, but I'm concerned that you don't override `subscribe` so you have a Signal where `get_value` gives you an enum, but `subscribe` gives you a float.

Yh, I agree but this is going to be hard. I guess the correct solution is to subscribe to all the signals you derive from then recalculate the derivation and push an update when any of them change.

> I wonder if there is a case for supplying both a `forward` and `inverse` transform, that operate on values rather than Signals, and making a `DerivedSignalBackend` that applies the transforms on get, put and subscribe?

Yep, I think we will need these in the generic case.

Can we push the full solution to https://github.com/bluesky/ophyd-async/issues/525 and treat this as a first pass example of how you could do it but with it's limitations? I would be happy to help contribute to the full solution but given it's a reasonably large feature, with quite a few stakeholders it will take a while and I would rather not have this bug in `hyperion` in the mean time.

coretl:
> Can we push the full solution to [bluesky/ophyd-async#525](https://github.com/bluesky/ophyd-async/issues/525) and treat this as a first pass example of how you could do it but with it's limitations? I would be happy to help contribute to the full solution but given it's a reasonably large feature, with quite a few stakeholders it will take a while and I would rather not have this bug in `hyperion` in the mean time.

Sounds good to me

noemifrisina (approved):
Looks like comments have been addressed.
Thanks, this is useful!

PR #793, opened 2024-09-19
Demark plan vs stub, move orphaned plans into dodal

DiamondJoseph:
Bluesky distinguishes between [plans:](https://github.com/bluesky/bluesky/blob/main/src/bluesky/plans.py) complete experimental proceedures, which open and close data collection runs, which may be part of a larger plan that collect data multiple times, but that might also be run alone to collect data, and [plan_stubs:](https://github.com/bluesky/bluesky/blob/main/src/bluesky/plan_stubs.py) which do not create & complete data collection runs and are either isolated behaviours or building blocks for plans.

In order to make it clearer when a MsgGenerator can be safely used without considering the enclosing run (as opening a run whilst in a run without explicitly passing a RunID is likely to cause both runs to fail), when it is required to manage a run and when running a procedure will create data documents, we should adopt this standard.

This change also adds the count and spec_scan plan previously from dls_bluesky_core, as remove this as a requirement for blueapi is long overdue and this is the last remnants of that module. `count` is a basic collection of a set of devices, equivalent to Mapping GDA's StaticScan. `spec_scan` is analogous to GDA's `mscan` command, constructing a generic N-dimensional trajectory for scanning.

After these plans have been moved, the default blueapi context can be configured to import as a plans module dodal.plans and dodal.plan_stubs.wrapped: this will only expose those stubs that should be runnable by a user from the client.

### Checks for reviewer


src/dodal/plans/scanspec.py
```diff
@@ -0,0 +1,64 @@
...
+def spec_scan(
+    detectors: Annotated[
+        set[Readable],
+        Field(
+            description="Set of readable devices, will take a reading at each point, \
+            in addition to any Movables in the Spec",
+        ),
+    ],
+    spec: Annotated[
+        Spec[Movable],  # type: ignore
```
DiamondJoseph, line 27: @coretl @callumforrester do you have any idea why this is currently required? Previously we had a Spec[str] in blueapi and this did not cause issues- here in dodal, it seems we cannot have a Spec[str] or Spec[Movable].

`"__getitem__" method not defined on type "type"Pylance[reportIndexIssue](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportIndexIssue)`

This is an issue in Pylance in the IDE and also in the type-checking step of the CI job.

DiamondJoseph: This was due to the discriminated_union decorator's type hinting not being adequate.
This change is therefore blocked until ScanSpec 0.7.3 can be released, which is blocked until time is spent to handle dataclasses.

src/dodal/plans/scanspec.py
```diff
@@ -0,0 +1,63 @@
...
+            in addition to any Movables in the Spec",
+        ),
+    ],
+    spec: Annotated[
+        Spec[Movable],
+        Field(description="ScanSpec modelling the path of the scan"),
+    ],
+    metadata: dict[str, Any] | None = None,
+) -> MsgGenerator:
+    """Generic plan for reading `detectors` at every point of a ScanSpec `spec`."""
```
callumforrester, line 31: Could: For consistency, could we make this have a docstring like the others, rather than use `Annotated`? Ditto for `count`

src/dodal/plan_stubs/__init__.py
callumforrester, line 1: Should: I'm not actually sure we gain much from hiding complexity in this file. For specific plans such as `check_topup_and_wait_if_necessary`, we've just saved a small amount of typing

```diff
-from dodal.plan_stubs.check_topup import check_topup_and_wait_if_necessary
+from dodal.plan_stubs import check_topup_and_wait_if_necessary"
```

But we've hidden away some potentially important namespacing, given ourselves some extra bookkeeping to do, and done something that isn't really done elsewhere in dodal (generally dodal is happy with long import strings).

Meanwhile, we've also had to add the  `__export__` which is meaningless here but holds a special meaning for blueapi, which is a leaky abstraction. I would rather the plans were here, available, and sensibly organised and the blueapi user's job was to select what they want.

tests/plans/test_scanspec.py
```diff
@@ -0,0 +1,97 @@
...
+        assert metadata_doc in docs
+        assert len(docs[metadata_doc]) == 1
+
+    start: RunStart = docs["start"][0]  # type: ignore (cannot type check TypedDict)
+    assert (hints := start.get("hints")) and (
+        hints.get("dimensions") == [([x_axis.user_readback.name], "primary")]
+    )
+    assert start.get("shape") == (3,)
+
+    descriptor: EventDescriptor = docs["descriptor"][0]  # type: ignore
```
callumforrester, line 74: Could: use `cast` rather than `# type: ignore`

callumforrester (changes requested):

src/dodal/plan_stubs/wrapped.py
```diff
@@ -0,0 +1,151 @@
...
+from bluesky.utils import MsgGenerator
+
+"""
+Wrappers for Bluesky built-in plan stubs with type hinting
+"""
+
+Group = Annotated[str, "String identifier used by 'wait' or stubs that await"]
+
+
+# https://github.com/bluesky/bluesky/issues/1821
```
DominicOram, line 16: Nit: From an outside perspective it's not clear from this comment or from the issue itself how this function relates to the issue. Will the function be removed, changed in some other way etc.

src/dodal/plan_stubs/wrapped.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Returns:
+        MsgGenerator: Plan
+
+    Yields:
+        Iterator[MsgGenerator]: Bluesky messages
+    """
+
+    return (
+        # https://github.com/bluesky/bluesky/issues/1809
```
DominicOram, line 108: Nit: As above, how does that issue change this code?

src/dodal/plans/scanspec.py
```diff
@@ -0,0 +1,65 @@
...
+from pydantic import Field, validate_call
+from scanspec.specs import Spec
+
+from dodal.common import MsgGenerator
+from dodal.plan_stubs.data_session import attach_data_session_metadata_decorator
+
+
+@attach_data_session_metadata_decorator()
+@validate_call(config={"arbitrary_types_allowed": True})
+def spec_scan(
```
DominicOram, line 17: FYI: I added https://github.com/DiamondLightSource/mx-bluesky/issues/596 as I think we have overlap here

src/dodal/plans/wrapped.py
```diff
@@ -0,0 +1,42 @@
...
+            the delays are between each point, if value or None is the delay for every \
+            gap",
+            json_schema_extra={"units": "s"},
+        ),
+    ] = 0.0,
+    metadata: dict[str, Any] | None = None,
+) -> MsgGenerator:
+    """Reads from a number of devices.
+    Wraps bluesky.plans.count(det, num, delay, md=metadata) exposing only serializable
+    parameters and metadata."""
```
DominicOram, line 35: Nit: I assume we've discussed with bluesky to try and get this into the core codebase and couldn't? May be worth a link to that discussion in case anyone asks the question again

tests/plan_stubs/test_wrapped_stubs.py
```diff
@@ -0,0 +1,145 @@
...
+    assert msgs[4] == Msg("wait", group="foo")
+
+
+def test_sleep():
+    assert list(sleep(1.5)) == [Msg("sleep", None, 1.5)]
+
+
+def test_wait():
+    # Waits for all groups
+    # move_on and wait are antonyms: what does move_on do?
```
DominicOram, line 128: Should: I agree that `move_on` is bizarre. It should be documented in the `bps.wait` docstring rather than this comment here, which doesn't really add anything for the reader other than more confusion. Can we add an issue into the bluesky repo to document it properly then either link here or remove the comment?

tests/plans/test_compliance.py
```diff
@@ -0,0 +1,69 @@
+import inspect
+from types import ModuleType
+from typing import Any, get_type_hints
+
+from bluesky.utils import MsgGenerator
+
+from dodal import plan_stubs, plans
+from dodal.common.types import PlanGenerator
+
+
```
DominicOram, line 10: Nit: This is sort of a structural/architectural test rather than a unit test. I think it could do with a docstring at the top preparing the reader for that fact and explaining it's general perpose

tests/plans/test_scanspec.py
```diff
@@ -0,0 +1,98 @@
...
+
+@pytest.fixture
+def path_provider(static_path_provider: PathProvider):
+    # Prevents issue with leftover state from beamline tests
+    with patch("dodal.plan_stubs.data_session.get_path_provider") as mock:
+        mock.return_value = static_path_provider
+        yield
+
+
+def test_output_of_simple_spec(
```
DominicOram, line 54: Nit: I'm not a big fan of big tests like this that are testing loads of things at once as I think they are poor documentation and make it harder to know what exactly the problem is when they break. I would prefer smaller tests like `test_scan_spec_gives_run_start_with_expected_shape` etc.

tests/plans/test_wrapped.py
```diff
@@ -0,0 +1,139 @@
...
+        ValidationError, match="Input should be greater than or equal to 1"
+    ):
+        RE(count({det}, num=-1))
+    with pytest.raises(
+        ValidationError, match="Input should be greater than or equal to 1"
+    ):
+        RE(count({det}, num=0))
+
+
+def test_count_output(det: StandardDetector, path_provider, RE):
```
DominicOram, line 76: Nit: As above with long, complex tests

tests/plans/test_wrapped.py
```diff
@@ -0,0 +1,139 @@
...
+
+    # Check that dets are all triggered
+    assert "stream_resource" in docs
+    assert len(docs["stream_resource"]) == 2  # det, det.sum
+
+    assert "stream_datum" in docs
+    assert len(docs["stream_datum"]) == 1 * 2  # each point per resource
+
+
+def test_multi_count_output(det: StandardDetector, path_provider, RE):
```
DominicOram, line 109: Nit: There's a lot of shared code here with the test above. Splitting them up into small, specific tests would help or I would also consider adding some helper functions

DominicOram (approved):
I'm happy with all the changes that affect MX, thanks for doing those. I have added some comments to the other code but given it's just a copy of existing code and we do not make use of it in MX please consider them optional.

src/dodal/plans/scanspec.py
callumforrester: This plan has now moved repositories so many times that the issue has now been lost to time, but it does need improvement. The `axes_to_move` parameter was a temporary fix and we should try to get rid of it.

callumforrester: https://github.com/DiamondLightSource/dodal/issues/866

DiamondJoseph: I think we no longer need it, I just need to find time to start up blueapi with these plans and ensure we can run a spec just passing the name of the device and no axes_to_move

src/dodal/plans/wrapped.py
callumforrester: It was a series of mostly offline discussions and should probably be written up as an ADR

tests/plans/test_scanspec.py
callumforrester: My $0.02: It's fine as long as you do both, a big integration test is sometimes good too as it catches issues that only arise when you put everything together, however a small unit test is _more_ useful for the reasons you say so if you can only do one, do that.

On the other hand I think we do have to keep a close eye on how long it takes to run the dodal test suite. As we add more and more beamlines I'm expecting it to increase significantly and we might need to start thinking about solutions like [pytest-parallel](https://pypi.org/project/pytest-parallel/). I'd like to avoid that as long as possible though because we might get lazy and complacent if we know we can always get more resources (see every web browser).

src/dodal/plans/wrapped.py
DominicOram: Does someone involved in the discussion have time to write an issue with some back of the envelope bullet point reasons and saying to put them in an ADR, just before it's lost to time?

tests/plans/test_scanspec.py
DominicOram: I agree that both is best if possible, we have some tests in `mx-bluesky` that are larger integration ones alongside the ones just testing small bits of logic. I also agree with the idea that they shouldn't take too long and I do regularly add PRs that just speed them up a bit. However, in my experience long running tests are mostly a consequence of poor mocking rather than quantity and it's actually large integration tests that end up taking longer than many small ones as people tend to have done more mocking in the smaller tests. Tests currently take ~30s on `dodal`, which seems reasonable. I would suggest if they start hitting ~60s we should think harder about this.

src/dodal/plans/scanspec.py
DiamondJoseph: In theory, all plans that use a ScanSpec should be constructing their specific spec and passing it to this for software scans) or an ophyd plan to-be-produced for hardware.

src/dodal/plans/wrapped.py
DiamondJoseph: I believe we need https://github.com/DiamondLightSource/blueapi/issues/474 for the @attach_metadata to be handled in blueapi consistently, then that decorator gets moved back into blueapi and then blueapi can just use bluesky.plans as a plans module (or... the ones we actually want).

tests/plan_stubs/test_wrapped_stubs.py
DiamondJoseph: Tracked down the initial issue that added it, raised an upstream ticket to document better.
https://github.com/bluesky/bluesky/issues/1824

src/dodal/plans/wrapped.py
DominicOram: In which case can we have a comment here saying we can remove this when that has been done?

tests/plans/test_compliance.py
```diff
@@ -0,0 +1,78 @@
...
+is likely to cause both runs to fail), when it is required to manage a run and when
+running a procedure will create data documents, we have adopted this standard.
+
+We further impose other requirements on both plans and stubs exported from these modules
+to enable them to be exposed in UIs in a consistent way:
+- They must have a docstring
+- They must not use variadic arguments (*args, **kwargs)
+- Plans must and Stubs may have an optional argument for metadata, named "metadata"
+- The metadata argument, where present, must be optional with a default of None
+"""
```
DominicOram, line 29: :heart:

DominicOram (approved):
One minor comment but otherwise looks really good, much cleaner tests, thank you!

src/dodal/plans/wrapped.py
DiamondJoseph: I think actually we'll need to maintain them for at least a while afterwards- bp.count has a an argument of type `PerShot` which is not serialisable. We may be able to ignore all fields that are optional and of a non-serialisable type? Especially where they are `Callable`s. I'll add something to that effect.

```python
PerShot = Callable[[Sequence[Readable], Optional[bps.TakeReading]], MsgGenerator]
``` 

https://github.com/bluesky/bluesky/blob/92d2633127e33e51aa1d1b109c80503e74bbe0e0/src/bluesky/plans.py#L66

callumforrester (approved):
LGTM

PR #799, opened 2024-09-23
use devicevector

stan-dot:
Fixes #ISSUE

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


DominicOram (changes requested):
I think this is a worse interface for the users. The grouping of devices in the `beamline` file should be logical to a scientist as that is their interface with the devices. They do not think of the slits on the beamline as one large device with multiple slits but rather as individual devices. So they would want to define a plan like `def slit_scan(slit: Slits = inject("slits_1"))`. This also brings the issue of how we deal with disconnected slits. It is very possible that the beamline may have one slit set offline but still want to use the others. With `SixSlits` the whole device will fail to connect and throw errors.

stan-dot:
@DominicOram at the moment I only see the `Slits` device used at i22 and p38, where I know the scientists and their ways of working and for now they don't write tests themselves. And the device is just in the `devices/i22` folder and any beamline can use the single `Slits` device however they want. 

>  With SixSlits the whole device will fail to connect and throw errors.

that is a valid point, might need a decorator 'allow fail'? I am not sure if the slits are used independently like what you're describing now at i22

DiamondJoseph:
If we go this way I don't think it should be called six_slits: we'd still be able to address individual slits with e.g. `inject` by doing `inject = ("slits.1")` (NB: Might not work for device vectors? Would this work with `inject("slits[1]")`?)

stan-dot:
at the latest version it's a `SlitsCollection`. testing out ipython behavior of device factor shortly

stan-dot:
@DominicOram what is a plan that you'd like to define in ipython and then run it that the inject behavior would be used?

DominicOram:
> @DominicOram what is a plan that you'd like to define in ipython and then run it that the inject behaviour would be used?

Something like:
```python
def slit_scan(slit: Slits = inject("slits_1"), i0: TetrammDetector = inject("i0"):
    yield from bps.scan([i0], slit.x, 0, 100)
```

stan-dot:
ok how do I get the beamline module to be referred to with `inject`? 

I don't see this in [`tutorials`](https://diamondlightsource.github.io/dodal/main/tutorials.html)


search also does not provide information on the use of dodal in ipython https://diamondlightsource.github.io/dodal/main/search.html?q=ipython

DiamondJoseph:
> ok how do I get the beamline module to be referred to with `inject`?

It's a blueapi-ism, assuming that there is a device with that name in the blueapi context that will be able to be fetched when the pydantic model is deserialised: regardless of what beamline module is loaded. It's not particularly well developed or documented, I'm hoping there's something in Pydantic 2 that we'll be able to leverage for the same effect but I will have to look after I get back.

stan-dot:
@DominicOram  on re-reading your initial comment I see that the granularity of `skip_device` decorator is decreased as a result of this PR, closing the PR then

PR #806, opened 2024-09-25
add eiger timeouts availability

### LINKED ISSUE #256: Parameterise ALL_FRAMES_TIMEOUT in eiger
The `ALL_FRAMES_TIMEOUT` variable in the eiger device is currently set to value suited to i03. We should move this to a parameter to other beamlines (eg VMXm) can use this ophyd device

## Acceptance Criteria
* `ALL_FRAMES_TIMEOUT` moved to a suitable parameter
* Test

stan-dot:
Fixes #256

1. Dodal connect i03 
2. should be successful

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -65,10 +65,8 @@ class EigerDetector(Device):
     def __init__(self, beamline: str = "i03", *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.beamline = beamline
-        if beamline in AVAILABLE_TIMEOUTS:
-            self.timeouts = AVAILABLE_TIMEOUTS[beamline]
-        else:
-            raise ValueError(f"Unknown beamline_name: {beamline}")
+        # using i03 timeouts as default
+        self.timeouts = AVAILABLE_TIMEOUTS.get(beamline, AVAILABLE_TIMEOUTS["i03"])
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/dodal/devices/eiger.py
```diff
@@ -34,27 +55,29 @@ def set(self, value, *, timeout=None, settle_time=None, **kwargs):
...
     arming_status = Status()
     arming_status.set_finished()
 
+    def __init__(self, beamline: str = "i03", *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.beamline = beamline
+        if beamline in AVAILABLE_TIMEOUTS:
+            self.timeouts = AVAILABLE_TIMEOUTS[beamline]
+        else:
+            raise ValueError(f"Unknown beamline_name: {beamline}")
```
olliesilvester, line 71: Could: I think it's better to use i03's timeout values as defaults rather than raising an error here. Then we could log a message to say that default values are being used

olliesilvester (approved):
Thanks, looks good. Just one comment

src/dodal/devices/eiger.py
olliesilvester: Please could you also add a test for this line

stan-dot: ok will do today

PR #807, opened 2024-09-25
805 move p99 detectors into dodal

### LINKED ISSUE #805: Move p99 detectors into dodal
Move the rest of the p99 [active devices](https://github.com/DiamondLightSource/p99-bluesky/blob/main/src/p99_bluesky/beamlines/p99.py) into dodal.  

## Acceptance Criteria

- [x] all devices are in dodal
- [x] all devices passes test
- [x]  all device connect.

Relm-Arrowny:
Fixes #805 

1. Run unit test
2. Confirm connections.

### Checks for reviewer


src/dodal/devices/areadetector/andor2_epics/andor2_controller.py
```diff
@@ -0,0 +1,79 @@
...
+    Andor2TriggerMode,
+    ImageMode,
+)
+
+
+class Andor2Controller(DetectorControl):
+    """
+    Andor 2 controller
+
+    """
```
DominicOram, line 23: Could: This docstring isn't very informative. I would either expand to give more detail about how the Andor is controlled or remove this.

src/dodal/devices/areadetector/andor2_epics/andor2_controller.py
```diff
@@ -0,0 +1,79 @@
...
+    ) -> None:
+        if good_states is None:
+            good_states = set(DEFAULT_GOOD_STATES)
+        self._drv = driver
+        self.good_states = good_states
+
+    def get_deadtime(self, exposure: float | None) -> float:
+        if exposure is None:
+            return 0.1
+        return exposure + 0.1
```
DominicOram, line 43: Nit: I would pull the 0.1 out into a constant of `DEAD_TIME` or something, aids in readability + maintainability.

src/dodal/devices/areadetector/andor2_epics/andor2_controller.py
```diff
@@ -0,0 +1,79 @@
...
+
+    async def prepare(self, trigger_info: TriggerInfo):
+        if trigger_info.livetime is not None:
+            await adcore.set_exposure_time_and_acquire_period_if_supplied(
+                self, self._drv, trigger_info.livetime
+            )
+        await asyncio.gather(
+            self._drv.trigger_mode.set(self._get_trigger_mode(trigger_info.trigger)),
+            self._drv.num_images.set(
+                999_999 if trigger_info.number == 0 else trigger_info.number
```
DominicOram, line 53: Nit: I would pull `999_999` out into a constant for readability

src/dodal/devices/areadetector/andor2.py
```diff
@@ -0,0 +1,47 @@
...
+    Andor 2 area detector device. Andor model:DU897_BV
+
+    Parameters
+    ----------
+    prefix: str
+        Epic Pv,
+    path_provider: PathProvider
+        Path provider for hdf writer
+    name: str
+        Name of the device
```
DominicOram, line 19: Should: This docstring is on the class but is talking about the params the init. From https://peps.python.org/pep-0257 `The class constructor should be documented in the docstring for its __init__ method. `

src/dodal/beamlines/p99.py
```diff
@@ -53,9 +62,44 @@ def sample_lab_xyz_stage(
...
+        Andor2,
+        prefix="-EA-DET-03:",
+        name="andor2_det",
+        path_provider=andor_data_path,
+        wait=wait_for_connection,
+        fake=fake_with_ophyd_sim,
+    )
+
+
+def andor2_point(
```
DominicOram, line 92: Could: As someone not too familiar with p99 it's not clear to me what the difference is between `andor2_det` and `andor2_point` and when I should use one over the other. I think putting some docstrings on them would be good

tests/devices/unit_tests/areadetector/test_andor2.py
```diff
@@ -0,0 +1,151 @@
...
+
+    with patch("ophyd_async.core.wait_for_value", return_value=None):
+        await Andor.disarm()
+    with pytest.raises(ValueError):
+        Andor._get_trigger_mode(DetectorTrigger.edge_trigger)
+
+    assert await driver.acquire.get_value() is False
+
+
+# area detector that is use for testing
```
DominicOram, line 89: Nit: I don't think this comment adds much

src/dodal/devices/areadetector/andor2_epics/andor2_io.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ADBaseDataType(str, Enum):
+    UInt16 = "UInt16"
+    UInt32 = "UInt32"
+    b1 = ""
+    b2 = ""
+    b3 = ""
+    b4 = ""
+    b5 = ""
+    b6 = ""
```
DominicOram, line 35: Should: I don't think you need all these empty string ones, it should be fine without them

tests/devices/unit_tests/areadetector/test_andor2.py
```diff
@@ -0,0 +1,151 @@
...
+@pytest.fixture
+async def Andor() -> Andor2Controller:
+    async with DeviceCollector(mock=True):
+        drv = Andor2DriverIO("DRIVER:")
+        controller = Andor2Controller(drv)
+
+    return controller
+
+
+async def test_Andor_controller(RE, Andor: Andor2Controller):
```
DominicOram, line 60: Should: Capitilisation is normally reserved for class names in Python so `Andor` looks like you're passing the class around, not the specific instance. I think it should be `andor`

DominicOram (changes requested):
Some comments in code. Additionally, I would recommend more tests that are smaller and more descriptive. For example:
* `test_get_deadtime_returns_expected` <- just test deadtime
* `test_given_trigger_number_of_0_then_number_of_images_set_to_max`
* `test_given_no_live_time_then_esposure_time_not_set`
The advantage of this is it very clearly lays out what is expected of the device and if one of them fail in the future it should be obvious to someone what went wrong. In general I would expect a device like this to have maybe a dozen small tests like this.

src/dodal/devices/areadetector/andor2_epics/andor2_io.py
Relm-Arrowny: It fail the data type check when I try to connect without matching them. 
image

DominicOram: Odd, from https://github.com/bluesky/ophyd-async/blob/55f3552e667569d872d4f4b5ffc1df4d64601a17/src/ophyd_async/epics/signal/_aioca.py#L230 it looks like the options should be filtered to unique ones. For reference the PV:
```
[ffv81422@p99-ws001 ~]$ caget BL99P-EA-DET-03:CAM:DataType_RBV
BL99P-EA-DET-03:CAM:DataType_RBV UInt16
[ffv81422@p99-ws001 ~]$ caget -d31 BL99P-EA-DET-03:CAM:DataType_RBV
BL99P-EA-DET-03:CAM:DataType_RBV
    Native data type: DBF_ENUM
    Element count:    1
    Value:            UInt16
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            (10)
                      [ 0] UInt16
                      [ 1] UInt32
                      [ 2] 
                      [ 3] 
                      [ 4] 
                      [ 5] 
                      [ 6] 
                      [ 7] 
                      [ 8] Float32
                      [ 9] Float64
```
@coretl, is this a bug in ophyd-async?

coretl: Does it go away if you add `"_": ""` to the enum members? The error is coming from  [here](https://github.com/bluesky/ophyd-async/blob/55f3552e667569d872d4f4b5ffc1df4d64601a17/src/ophyd_async/epics/signal/_common.py#L45) which seems to do a set

Relm-Arrowny: Yes it work with `"_" : "" `, as long as we have an empty string, I guess the set on the right [here](https://github.com/bluesky/ophyd-async/blob/55f3552e667569d872d4f4b5ffc1df4d64601a17/src/ophyd_async/epics/signal/_common.py#L45) will have one "" and we need to match it to pass the check. Would it be better to .remove it in ophyd or add a blank when we define the data class?

coretl: I'm not sure either is particularly good, but that's what EPICS gives us. And it's technically possible for the PV to be set to "", so I think adding it to the Enum subclass is more technically correct.

PR #808, opened 2024-09-25
Allow OAV to Redis forwarder to use different OAV streams

DominicOram:
Needed for https://github.com/DiamondLightSource/mx-bluesky/issues/511

### Instructions to reviewer on how to test:
1. Confirm tests still pass

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

PR #809, opened 2024-09-27
530 only stream oav on updates

DominicOram:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/530 and https://github.com/DiamondLightSource/mx-bluesky/issues/147

Built on top of https://github.com/DiamondLightSource/dodal/pull/808 so review/merge that first

Additionally this PR:
* Changes the format of data we store in redis to be just the raw JPEG bytes. This was done after discussion with @aragaod to reduce the memory usage of the redis database.
* Adds a timeout of 30s and so will not stream data for longer than this, FYI @aragaod. We can update this timeout later if needed

1. Confirm tests still pass
2. Confirm device still connects with `dodal connect i04`

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/oav/oav_to_redis_forwarder.py
+++ b/src/dodal/devices/oav/oav_to_redis_forwarder.py
@@ -1,7 +1,6 @@
 import asyncio
 import io
 import pickle
-import uuid
 from collections.abc import Awaitable, Callable
 from datetime import timedelta
 from enum import Enum
@@ -13,6 +12,7 @@ from ophyd_async.core import (
     AsyncStatus,
     DeviceVector,
     StandardReadable,
+    observe_value,
     soft_signal_r_and_setter,
     soft_signal_rw,
 )
@@ -35,6 +35,17 @@ async def get_next_jpeg(response: ClientResponse) -> bytes:
 class Source(Enum):
     FULL_SCREEN = 0
     ROI = 1
+
+
+class OAVSource(StandardReadable):
+    def __init__(
+        self,
+        prefix: str,
+        name: str,
+    ):
+        self.url = epics_signal_r(str, f"{prefix}MJPG_URL_RBV")
+        self.image_uuid = epics_signal_r(str, f"{prefix}UniqueId_RBV")
+        super().__init__(name=name)
 
 
 class OAVToRedisForwarder(StandardReadable, Flyable, Stoppable):
@@ -68,16 +79,11 @@ class OAVToRedisForwarder(StandardReadable, Flyable, Stoppable):
         """
         self._sources = DeviceVector(
             {
-                Source.FULL_SCREEN.value: epics_signal_r(
-                    str, f"{prefix}MJPG:MJPG_URL_RBV"
-                ),
-                Source.ROI.value: epics_signal_r(str, f"{prefix}XTAL:MJPG_URL_RBV"),
+                Source.FULL_SCREEN.value: OAVSource(f"{prefix}MJPG:", "fullscreen"),
+                Source.ROI.value: OAVSource(f"{prefix}XTAL:", "ROI"),
             }
         )
         self.selected_source = soft_signal_rw(Source)
-
-        with self.add_children_as_readables():
-            self.uuid, self.uuid_setter = soft_signal_r_and_setter(str)
 
         self.forwarding_task = None
         self.redis_client = StrictRedis(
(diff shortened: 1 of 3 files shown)


src/dodal/devices/oav/oav_to_redis_forwarder.py
```diff
@@ -66,68 +75,70 @@ def __init__(
...
 
-    async def _stream_to_redis(self, response, _):
-        while not self._stop_flag:
-            await self._get_frame_and_put_to_redis(response)
-            await asyncio.sleep(0.01)
+    async def _stream_to_redis(self, response: ClientResponse, source: OAVSource):
+        done_status = AsyncStatus(
+            asyncio.wait_for(self._stop_flag.wait(), timeout=self.TIMEOUT)
+        )
+        async for image_uuid in observe_value(self.counter, done_status=done_status):
```
rtuck99, line 131: this counter signal, is it really a counter? It looks like it's actually uuids? Or are they not really uuids and are just sequence numbers?

rtuck99: Q: How frequently does this signal update, is it guaranteed that we will see every frame id that we need to collect?

rtuck99 (approved):
Looks good apart from maybe change the name of the counter and also query about whether we might get all the frames.

src/dodal/devices/oav/oav_to_redis_forwarder.py
DominicOram: This is the frame counter on the device. We can use it as a UUID as it increments with every frame and we're only using the one camera so will be unique. I suspect there is a roll-over at something around the max int value but given we don't keep data more than a week I think it's unlikely to affect us. 

It updates at the rate of the camera (~15Hz). You're right that we may end up missing some, though experimentally we haven't yet, but even if we do it's not critical we catch every frame.

Legitimate questions though, thanks for raising them, I will add some comments in the code to this affect.

aragaod: Re: uniqueness of the UUID

Unsure if it is overkill but can envisage during some operational issue a few camera IOC restarts could be needed. It is possible the counter always starts at one from IOC restart. In a combination of IOC restarts plus same sampleID being loaded and data collected on the same frame number (very unlikely combinations) then a UUID collision could technically happen. A possible solution would be to add a prefix of some kind randomly generated.

But unsure if it is worth the effort.

DominicOram: Good point, thanks. I will add a proper UUID section in too as a precaution.

src/dodal/devices/oav/oav_to_redis_forwarder.py
```diff
@@ -66,68 +75,77 @@ def __init__(
...
-        """Converts the data that comes in as a jpeg byte stream into a numpy array of
-        RGB values, pickles this array then writes it to redis.
-        """
+    async def _get_frame_and_put_to_redis(
+        self, redis_uuid: str, response: ClientResponse
+    ):
         jpeg_bytes = await get_next_jpeg(response)
-        self.uuid_setter(image_uuid := str(uuid.uuid4()))
-        img = Image.open(io.BytesIO(jpeg_bytes))
-        image_data = pickle.dumps(np.asarray(img))
```
aragaod, line 104: Before we move forward and forget would like 5 minutes look and maybe A comment left in the code why we make this decision now (allo5wing to change in the future if necessary). 

First we decided to not store numpy arrays of the raw images (jpg decompressed) after realising it would occupy enormous amounts of DB space with no real benefit on speed. 

Question on current implementation:  there 3 ways to store binary jpg into redis

I) directly as an bytesIO object using data.getvalues()
II) serialising it (as we make here with pickle)
III) using a redis optimize structure called redis bitmap for such objects

Happy we stay with (II) as I don't fully know disadvantages of the other ones Vs this one but maybe leave a comment this could be relooked if necessary in the future?

DominicOram: Issue at https://github.com/DiamondLightSource/mx-bluesky/issues/592 to investigate

DominicOram: Ah, I actually think this PR already removes the pickling and just stores the raw bytes. I will add a comment though.

PR #814, opened 2024-10-04
(DiamondLightSource/mx-bluesky#475) Method to allow access to full zocalo results

rtuck99:
Extends ZocaloResults to get access to everything instead of just the position and bounding box

Fixes DiamondLightSource/mx-bluesky#475

### DIFF
--- a/src/dodal/devices/zocalo/__init__.py
+++ b/src/dodal/devices/zocalo/__init__.py
@@ -4,14 +4,14 @@ from dodal.devices.zocalo.zocalo_results import (
     NoZocaloSubscription,
     XrcResult,
     ZocaloResults,
-    get_processing_result,
+    get_full_processing_results,
 )
 
 __all__ = [
     "ZocaloResults",
     "XrcResult",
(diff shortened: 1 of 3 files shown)


DominicOram (changes requested):
The consensus in `ophyd_async` is to move away from arbitrary types on signals. I have made https://github.com/bluesky/ophyd-async/issues/601 to actually try and advertise this fact. We should probably refactor this to use individual signals rather than `XRCResults`, like we did for https://github.com/DiamondLightSource/dodal/pull/769.

Additionally, I think after this PR `get_processing_result` is no longer used, can we remove it?

dperl-dls (approved):
Looks good, approving, not merging until https://github.com/DiamondLightSource/mx-bluesky/pull/553 is ready

rtuck99:
This PR now obsoleted by #858

DominicOram:
With https://github.com/DiamondLightSource/dodal/pull/858 merged I think this is now redundant so closing

Issue #716, opened 2024-07-29
Convert the OAV to ophyd-async 


DominicOram:
We should try and get all our devices into `ophyd-async`. The OAV is one of the last ones left

## Note
This is a reasonably large device, it may make sense to split it into multiple issues

## Acceptance Criteria
* The OAV is converted into `ophyd-async`
* Tests all still pass

DominicOram:
https://github.com/DiamondLightSource/dodal/issues/801 should be done first

DominicOram:
As part of this we should probably look at how we do the `OAVConfigParams`. Particularly how they get updated from the device. It may make more sense for them to be derived signals https://github.com/bluesky/ophyd-async/issues/525

noemifrisina:
As this is a fairly large device and we probably can't have an ophyd-async device with ophyd components (or viceversa), the first step will be to make a new very basic ophyd-async device meeting the minimum needs for i24 serial and then we can add the rest of the functionality as we go along.

PR #815, opened 2024-10-04
Start an ophyd-async OAV

noemifrisina:
Starts the work for #716 

Creates a first version of an ophyd-async OAV, for the moment including only the zoom controller and the code to get the microns per pixels and beam position out of the parameter files.

1. Run tests

### Checks for reviewer


src/dodal/devices/oav/oav_async.py
```diff
@@ -0,0 +1,159 @@
...
+        self,
+        mock: bool = False,
+        timeout: float = DEFAULT_TIMEOUT,
+        force_reconnect: bool = False,
+    ):
+        self.micronsPerXPixel = create_hardware_backed_soft_signal(
+            float,
+            lambda: self._get_microns_per_pixel("x"),
+        )
+        self.micronsPerYPixel = create_hardware_backed_soft_signal(
```
DominicOram, line 146: Should: Can we make these camel case please?

src/dodal/devices/oav/oav_async.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        beam_x = await self.beam_centre_i.get_value()
+        beam_y = await self.beam_centre_j.get_value()
+
+        return (
+            beam_x - horizontal_pixels,
+            beam_y - vertical_pixels,
+        )
+
+    async def connect(
```
DominicOram, line 136: Should: I think it might be ok to do all this on `init` rather than `connect`

DominicOram (changes requested):
Great start, thanks. Some comments in code. I also wondered if it was a bit cleaner to treat everything as tuples, see https://github.com/DiamondLightSource/dodal/tree/716_ophyd_async_oav_suggestion. I think it makes the functions in `OAV` a bit cleaner but I'm not 100% sure I like the implicit assumption that 0 == x and 1 == y. What do you think?

Additionally, it would be good to have some extra tests more specific to `oav_parameters` that directly test `OAVConfig` rather than test it through the device.

DominicOram (approved):
Great, thank you!

PR #819, opened 2024-10-07
Return no results if CPU and GPU agree on no results

DominicOram:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/545

1. Confirm test fails on main and passes here

### Checks for reviewer


olliesilvester (approved):
Looks good - not sure if you linked the correct issue though?

PR #820, opened 2024-10-08
Add the diamond filter and qbpm devices

DominicOram:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/544

1. Run `dodal connect i03` and `dodal connect i04` and confirm they pass

### Checks for reviewer


src/dodal/devices/diamond_filter.py
```diff
@@ -0,0 +1,46 @@
...
+    EMPTY = "Empty"
+    TWO_HUNDRED = "200um"
+    FIFTY = "50um"
+    OUT = "Out"
+
+
+T = TypeVar("T", bound=Filters)
+
+
+class DiamondFilter(StandardReadable, Generic[T]):
```
DominicOram, line 29: @coretl or @callumforrester. Do we like this pattern for the different enums on each beamline or do you think better to just drop it and leave it as a `str`?

coretl: Depends, do you in code write `filter.thickness.set(I03Filters.FIFTY)` anywhere? If so then it's useful to check that the specified filter value exists at connect too. If not then drop to string.

DominicOram: I'm not sure I understand. We don't write `filter.thickness.set(I03Filters.FIFTY)` anywhere because there is no 50um filter on I03. The filters are different on each beamline but otherwise the hardware is the same. What I'm trying to do is allow people to write:
```python
i03_filter = i03.filter()
i04_filter = i04.filter()

i03_filter.thickness.set(I03Filters.TWO_HUNDRED)  # All good
i03_filter.thickness.set(I04Filters.FIFTY) # Type system gives an error here, there is no 50 on i03
```
My concern with `str` was that you can then do:
```python 
i03_filter.thickness.set("200 um")  # All good
i03_filter.thickness.set("50 um") # No warning this is not valid until runtime
```
I wanted to make sure this pattern wasn't overly complicating things or doing something unexpected for `ophyd_async` though

src/dodal/devices/diamond_filter.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def __init__(
+        self,
+        prefix: str,
+        data_type: type[T],
+        name: str = "",
+    ) -> None:
+        with self.add_children_as_readables():
+            self.y_motor = Motor(prefix + "Y")
+            self.thickness = epics_signal_rw(data_type, f"{prefix}Y:MP:SELECT")
```
olliesilvester, line 44: Nit: It's not immediately obvious (to me) that selecting a thickness is just selecting a motor value

olliesilvester (approved):
Looks good, and devices behaved on i03. I think what you've done with generics is good too

src/dodal/devices/diamond_filter.py
DominicOram: Added a bit more in the docstring

src/dodal/devices/diamond_filter.py
coretl: The procedural case will remain compatible with this, but making the declarative approach work for this would be overkill. I'd say it makes sense for the usecase you mention.

PR #832, opened 2024-10-10
Defer implementation of MsgGenerator to prevent strange behaviour

DiamondJoseph:
Previously it has been observed in blueapi that defining the same type alias in different modules has stopped being the same object, and then later stopped even being equal to each other. As checking that the return type of a function is `MsgGenerator` is used by blueapi to determine which plans to expose, this has prevented plans and stubs from being visible. As of bluesky 1.13, bluesky now defines its own implementation of MsgGenerator, which also meets the requirements of @rtuck99 of having a return type.

Prevents re-occurence of these issues and similar:
https://github.com/DiamondLightSource/blueapi/pull/533
https://github.com/DiamondLightSource/blueapi/issues/643

1. Run type checking and ensure that functions that return a MsgGenerator are correctly typed
2. Confirm that 
```python
from dodal.common import MsgGenerator as m1
from bluesky.utils import MsgGenerator as m2

assert m1 is m2
assert m1 == m2
```

### Checks for reviewer

### DIFF
--- a/src/dodal/common/types.py
+++ b/src/dodal/common/types.py
@@ -1,16 +1,11 @@
 from abc import ABC, abstractmethod
-from collections.abc import Callable, Generator
-from typing import (
-    Any,
-)
+from collections.abc import Callable
 
-from bluesky.utils import Msg
+from bluesky.utils import MsgGenerator
 from ophyd_async.core import PathProvider
 
 # String identifier used by 'wait' or stubs that await
 Group = str
-# A true 'plan', usually the output of a generator function
-MsgGenerator = Generator[Msg, Any, None]
 # A function that generates a plan
 PlanGenerator = Callable[..., MsgGenerator]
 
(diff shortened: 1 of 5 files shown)


DiamondJoseph:
https://github.com/bluesky/bluesky/blob/b6828c53a68252caeefc3712ea50c49e507d8db2/src/bluesky/utils/__init__.py#L98

Here is the current implementation of MsgGenerator in bluesky.

callumforrester (changes requested):
Unless you have a good reason to avoid it I would prefer this change were just a big find and replace where all imports within dodal are replaced with `from bluesky.utils import MsgGenerator`.

DiamondJoseph:
> Unless you have a good reason to avoid it I would prefer this change were just a big find and replace where all imports within dodal are replaced with `from bluesky.utils import MsgGenerator`.

Done

And PRs opened to allow eventual removal of dodal pass-through in other libraries too
https://github.com/DiamondLightSource/mx-bluesky/pull/564
https://github.com/DiamondLightSource/blueapi/pull/645

i22 etc. to follow

callumforrester (approved):

PR #833, opened 2024-10-10
Limit i03 beam width to 80um

dperl-dls:
- `beam_size_from_apterture` now limits the returned width to the maximum for the beamline


DominicOram (changes requested):
Thanks, Can we have a test on this please?

DominicOram (approved):
Thanks!

PR #597, opened 2024-06-05
make device factory decorator

stan-dot:
Fixes #483 

### Instructions to reviewer on how to test:
1. Do thing run i22 beamline
2. Confirm lazy loading is possible

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

Issue #780, opened 2024-09-09
Expose help and docs from device factory functions.


DiamondJoseph:
After #597 but also relevant for the current state:

Device classes have docstrings about their capabilities/use/expected behaviour. Specific device factory functions may also have docstrings about their expected behaviour, other usage information. Some combination of these should be presented to the user when they use the python builtin `help` function on the device_factory_function. 

## Acceptance Criteria
* After importing `ixx.device`, `help(device)` gives useful information about using the device, which may include
* a description of any parameters that are added by our decorator and either their default or how to find their default
* (if provided) docstring of any parameters that may be passed into the decorated function
* (if provided) docstring of the device class that is being returned
* (if provided) docstring of the device instance

DiamondJoseph:
Note that the `help` method uses the docstrings generated via pydoc, which [may require that we are using numpydoc](https://stackoverflow.com/a/21948562) style docstrings. @coretl has opinions about the verbosity.

coretl:
> Note that the `help` method uses the docstrings generated via pydoc, which [may require that we are using numpydoc](https://stackoverflow.com/a/21948562) style docstrings. @coretl has opinions about the verbosity.

Really? That SO post seems to be about Spyder. `help` seems to just print what you wrote:

```python
class A:
    def google_meth(self, a: int):
        """A google style doc method.
        
        Args:
            a: The parameter
        """
    def numpy_meth(self, b: str):
        """A numpy style doc method

        Parameters
        ----------
        b
            The parameter
        """
help(A)
```
```
class A(builtins.object)
 |  Methods defined here:
 |  
 |  google_meth(self, a: int)
 |      A google style doc method.
 |      
 |      Args:
 |          a: The parameter
 |  
 |  numpy_meth(self, b: str)
 |      A numpy style doc method
 |      
 |      Parameters
 |      ----------
 |      b
 |          The parameter
```

PR #841, opened 2024-10-14
Isolated device factory

### LINKED ISSUE #483: Make `device_factory` decorator to ease making and connecting ophyd-async Devices

DiamondJoseph:
Fixes #483 

This is a re-implementation of #597 isolated to just the changes necessary to implement the `@device_factory()` decorator and have functions decorated with that decorator picked up by `make_all_devices`: respecting any configuration passed into the decorator.

NB: this is also required for [blueapi #647](https://github.com/DiamondLightSource/blueapi/issues/647)

1. System test the i22 beamline to ensure that devices are instantiated, connected and named appropriately.

### Checks for reviewer

### DIFF
--- a/src/dodal/utils.py
+++ b/src/dodal/utils.py
@@ -309,6 +309,16 @@ def extract_dependencies(
             yield name
 
 
+def get_device_factories(
+    module: ModuleType,
+) -> dict[str, DeviceInitializationController]:
+    return {
+        name: controller
+        for name, controller in module.__dict__.items()
+        if isinstance(controller, DeviceInitializationController)
+    }
+
+
 def collect_factories(
     module: ModuleType, include_skipped: bool = False
 ) -> dict[str, AnyDeviceFactory]:
--- a/src/dodal/utils.py
+++ b/src/dodal/utils.py
@@ -98,7 +98,7 @@ class BeamlinePrefix:
 
 T = TypeVar("T", bound=AnyDevice)
 D = TypeVar("D", bound=OphydV2Device)
-skip_type = bool | Callable[[], bool]
+SkipType = bool | Callable[[], bool]
 
 
 def skip_device(precondition=lambda: True):
@@ -122,7 +122,7 @@ class DeviceInitializationController(Generic[D]):
         use_factory_name: bool,
         timeout: float,
         mock: bool,
-        skip: skip_type,
+        skip: SkipType,
     ):
         self._factory: Callable[[], D] = factory
         self._cached_device: D | None = None
@@ -155,6 +155,9 @@ class DeviceInitializationController(Generic[D]):
         """Returns an instance of the Device the wrapped factory produces: the same
         instance will be returned if this method is called multiple times, and arguments
         may be passed to override this Controller's configuration.
+        Once the device is connected, the value of mock must be consistent, or connect
+        must be False.
+
 
         Args:
             connect_immediately (bool | None, optional): whether to call connect on the
@@ -174,7 +177,8 @@ class DeviceInitializationController(Generic[D]):
               ophyd_async's DEFAULT_TIMEOUT.
             mock (bool | None, optional): overrides whether to connect to Mock signal
               backends, if connect is called. Defaults to None, which uses the mock
-              parameter of this Controller.
+              parameter of this Controller. This value must be used consistently when
+              connect is called on the Device.
 
         Returns:
             D: a singleton instance of the Device class returned by the wrapped factory.
--- a/src/dodal/utils.py
+++ b/src/dodal/utils.py
@@ -6,13 +6,14 @@ import socket
 import string
 from collections.abc import Callable, Iterable, Mapping
 from dataclasses import dataclass
-from functools import wraps
+from functools import update_wrapper, wraps
 from importlib import import_module
 from inspect import signature
 from os import environ
 from types import ModuleType
 from typing import (
     Any,
+    Generic,
     Protocol,
     TypeGuard,
     TypeVar,
@@ -35,6 +36,7 @@ from bluesky.protocols import (
     Triggerable,
     WritesExternalAssets,
 )
+from bluesky.run_engine import call_in_bluesky_event_loop
 from ophyd.device import Device as OphydV1Device
 from ophyd_async.core import Device as OphydV2Device
 
@@ -95,6 +97,8 @@ class BeamlinePrefix:
 
 
 T = TypeVar("T", bound=AnyDevice)
+D = TypeVar("D", bound=OphydV2Device)
+skip = bool | Callable[[], bool]
 
 
 def skip_device(precondition=lambda: True):
@@ -108,6 +112,68 @@ def skip_device(precondition=lambda: True):
         return wrapper
 
     return decorator
+
+
+class DeviceInitializationController(Generic[D]):
+    def __init__(
+        self,
+        factory: Callable[[], D],
+        eager_connect: bool,
+        use_factory_name: bool,
+        timeout: float,
+        mock: bool,
+        skip: skip,
+    ):
+        self._factory: Callable[[], D] = factory
+        self._cached_device: D | None = None
+        self._callbacks: list[Callable[[D], None]] = []
+        self._eager_connect = eager_connect
+        self._use_factory_name = use_factory_name
+        self._timeout = timeout
+        self._mock = mock
+        self._skip = skip
+        update_wrapper(self, factory)
+
+    @property
+    def skip(self) -> bool:
+        return self._skip() if callable(self._skip) else self._skip
+
+    @property
+    def device(self) -> D | None:
+        return self._cached_device
+
+    def add_callback(self, callback: Callable[[D], None]):
+        self._callbacks.append(callback)
+
+    def __call__(
+        self,
+        connect: bool | None = None,
+        name: str | None = None,
+        timeout: float | None = None,
+        mock: bool | None = None,
+    ) -> D:
+        if self.device is not None:
+            device = self.device
+        else:
+            device = self._factory()
+
+        if name:
+            device.set_name(name)
+        elif not device.name and self._use_factory_name:
+            device.set_name(self._factory.__name__)
+
+        if connect or connect is None and self._eager_connect:
+            call_in_bluesky_event_loop(
+                device.connect(
+                    timeout=timeout if timeout is not None else self._timeout,
+                    mock=mock if mock is not None else self._mock,
+                )
+            )
+
+        self._cached_device = device
+        for callback in self._callbacks:
+            callback(device)
+        return device
 
 
 def make_device(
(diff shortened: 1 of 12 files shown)


callumforrester:
I don't necessarily expect parallel connect to be implemented here but do we have tickets for achieving it?

DiamondJoseph:
> I don't necessarily expect parallel connect to be implemented here but do we have tickets for achieving it?

https://github.com/DiamondLightSource/blueapi/issues/440

callumforrester (approved):

stan-dot (approved):
nice functools use

src/dodal/utils.py
```diff
@@ -110,6 +114,68 @@ def wrapper(*args, **kwds) -> T:
...
+
+    def add_callback(self, callback: Callable[[D], None]):
+        self._callbacks.append(callback)
+
+    def __call__(
+        self,
+        connect: bool | None = None,
+        name: str | None = None,
+        timeout: float | None = None,
+        mock: bool | None = None,
```
DominicOram, line 153: Could: This is basically the interface for when I do `i22.dcm(...)` from a script. As such it would be better to make the arg names a bit cleaner e.g. `connect_immediately`? and `connection_timeout`

DominicOram: Also docstrings but happy for that to be covered by https://github.com/DiamondLightSource/dodal/issues/780

src/dodal/utils.py
```diff
@@ -95,6 +97,8 @@ def __post_init__(self):
 
 
 T = TypeVar("T", bound=AnyDevice)
+D = TypeVar("D", bound=OphydV2Device)
+skip = bool | Callable[[], bool]
```
DominicOram, line 101: Should: Can we call this something like `skip_type` to avoid confusion?

src/dodal/utils.py
```diff
@@ -202,7 +268,14 @@ def invoke_factories(
         dependent_name = leaves.pop()
         params = {name: devices[name] for name in dependencies[dependent_name]}
         try:
-            devices[dependent_name] = factories[dependent_name](**params, **kwargs)
+            factory = factories[dependent_name]
+            if isinstance(factory, DeviceInitializationController):
+                mock = kwargs.get("mock", False) or kwargs.get(
+                    "fake_with_ophyd_sim", False
+                )
```
DominicOram, line 275: Could: Is `fake_with_ophyd_sim` purely in there for backwards compatibility downstream? I think from an MX POV we can break compatibility here and it will be fine, we will need to do a bit of migration work anyway.

src/dodal/utils.py
```diff
@@ -236,6 +309,16 @@ def device_one(): and def device_two(device_one: Readable):, indicate taht
             yield name
 
 
+def get_device_factories(
```
DominicOram, line 312: Should: Can we have a test around this please?

DominicOram: Could: It's not 100% clear what the difference is between this and `collect_factories` and when one should use this or the other. It would be good to have some more in the docstrings or change the function names to make it clearer

src/dodal/utils.py
```diff
@@ -264,6 +347,8 @@ def collect_factories(
 
 
 def _is_device_skipped(func: AnyDeviceFactory) -> bool:
+    if isinstance(func, DeviceInitializationController):
+        return func.skip
```
DominicOram, line 351: Should: Can we have a test here please?

DominicOram (changes requested):
Great, thank you!

Some comments in code, mostly minor. I think a few more tests would also be good. e.g. we can test (assuming these are all desired behavior?):
* A device gets renamed if called again with a different name 
* If we call with `connect=False, mock=True` we can then call again with `mock=False`

Additionally I think there's a potential pitfall in:
```python
dcm(mock=True)
.... # Lots of code
dcm() # This now fails because it tries to turn the cached mock one into a real one
```
I'm not 100% sure if this is expected or if you would rather give the cached mocked one, maybe with a warning. We should at least document it but can do so in #780.

src/dodal/utils.py
DiamondJoseph: Yes that's there for backwards compat- is dodal still wanting a "make all devices" method (just for `dodal connect`?) or is that API going to move into blueapi? I was going to make a PR after this got merged to remove the kwargs and pass explicitly the args that connect wants (after moving more beamlines over).

src/dodal/utils.py
callumforrester: As an aside: I've been in enough debates about verbosity vs. "why say many word when few do trick" now to wonder if we need a style guide. Java is very pro-verobosity and has that culture around it so we never needed one with GDA, the Python world is a bit more fragmented.

src/dodal/utils.py
callumforrester: Also maybe should be in PascalCase, similar to most other types and type aliases.

src/dodal/utils.py
callumforrester: I'm also happy to break compatibility, however this change is required to fix some ongoing blueapi build issues so maybe we shouldn't block it/do that separately?

https://github.com/DiamondLightSource/blueapi/issues/665
https://github.com/DiamondLightSource/blueapi/issues/647

DominicOram: Yes, we still want `make_all_devices` I think. If it's going to be addressed in another issue that's fine.

src/dodal/utils.py
callumforrester: Hmm, they are very similar, maybe we should just mark `collect...` as deprecated and have done with it?

src/dodal/utils.py
DominicOram: Ooof, it'll be controversial and I think hard to codify into a style guide. I feel like the answer is always "As verbose as is required to make it clear given the context" which is so subjective.  But I would be happy to weigh in on a discussion

src/dodal/utils.py
DominicOram: Marking as depreciated works for me, with an explaining comment/link to issue

src/dodal/utils.py
DominicOram: > Also maybe should be in PascalCase, similar to most other types and type aliases.

I also agree with this but thought it maybe too opinionated to add. I would rather PascalCase though.

src/dodal/utils.py
callumforrester: It is quite subjective, but it's actually quite useful. What you said is a nice middle ground between "you should minimise the amount of lines/characters you write" (C) or "You should write function signatures as full sentences" (Java)

src/dodal/utils.py
DiamondJoseph: https://github.com/DiamondLightSource/dodal/issues/844

src/dodal/utils.py
DiamondJoseph: I just nixxed this implementation and reworded the ADR: the only difference was that it returned a dict[str, DeviceInitilizationController] rather than a dict[str, Callable[[], AnyDevice]: as beamlines move over this can be converted to all return DIC, and when we call it from blueapi we can check if it's a DIC and handle appropriately

src/dodal/utils.py
DiamondJoseph: :sunglasses: :+1:

src/dodal/utils.py
DiamondJoseph: :sunglasses: :+1:

src/dodal/utils.py
DiamondJoseph: :sunglasses: :+1:

DominicOram (approved):
Thanks. I'm still a little concerned about the not 100% intuitive behavior of calling the factory twice with different permutations of connect/mock but I thin we can document/fix that later.

DiamondJoseph:
> Assuming these are all desired behavior?

@coretl for comments?

>     * A device gets renamed if called again with a different name

[A device can be renamed after having a name](https://github.com/bluesky/ophyd-async/blob/3e74761b76f337bf0e36dcdbdd7fdc0dd814105b/src/ophyd_async/core/_device.py#L55), so I think it should be allowed. Currently the device_factory name is only used if the device is currently unnamed.

>     * If we call with `connect=False, mock=True` we can then call again with `mock=False`

Only when connect is called is the device being mock or not considered. Calling with connect=False, mock=True then later connecting with mock=False would connect to the real signal backend.

> Additionally I think there's a potential pitfall in:
> ```python
> dcm(mock=True)
> .... # Lots of code
> dcm() # This now fails because it tries to turn the cached mock one into a real one
> ```

[Ophyd-async throws an exception if you try and connect a 2nd time with a different value of `mock`.
](https://github.com/bluesky/ophyd-async/blob/3e74761b76f337bf0e36dcdbdd7fdc0dd814105b/src/ophyd_async/core/_device.py#L96). I will amend to cache the value of `mock` after `connect` has been called at least one, and defer to it rather than the `mock` value set up with the decorator, so that we can still try and explicitly connect to [non-]mock, but will throw the exception only when we do so.

DominicOram:
> I will amend to cache the value of mock after connect has been called at least one, and defer to it rather than the mock value set up with the decorator, so that we can still try and explicitly connect to [non-]mock, but will throw the exception only when we do so.

I'm not even sure that is the best answer either. You then end up with the above code connecting you to a `mock` when you actually want a real device and potentially you fail to realise. I'm not sure there's a good solution here. I think the best thing to do is just document that calling the factory multiple times can lead to odd behavior and is not recommended.

callumforrester:
I agree there is no good solution to be honest, caching singletons is hard to do without some hidden magic. Maybe add more logging (but in a separate change).

PR #722, opened 2024-08-01
add i18 beamline definition - WITH device_factory

stan-dot:
Fixes #709 

### Instructions to reviewer on how to test:
1. Do `dodal connect i181
2. Confirm all the devices expected to work, work

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://github.com/DiamondLightSource/dodal/wiki/Device-Standards)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

Issue #592, opened 2024-06-03
Reconcile I22 and I03 DCMs


callumforrester:
As a developer I would like to work with a minimal number of `DCM` classes.

Currently the PV naming conventions vary across beamlines, even when the hardware configuration is relatively similar.
In the case of I22 and I03, the DCMs are quite similar. According to @JamesOHeaDLS:

> The main difference is that I03 only has roll and pitch on one of their crystals, whereas I22 has roll and pitch on both their first and second crystals

Which suggests that the PVs could at least adopt the same naming _conventions_, even if one DCM has extra motors, and we could have a shared base class. 

## Acceptance Criteria
* One `DCM` base class
* I03 and I22 implementations with only the extra I22 motor added

callumforrester:
Hi James, currently we've been doing it ad-hoc. When we encounter a discrepancy between two beamlines, we try asking the relevant controls engineer to change the PV names. This is quite nice and agile, and it means we don't need a giant work package to rename all PVs across Diamond simultaneously, just that the level of standardisation that is needed for DAQ purposes should creep in over time. I'll defer to @coretl though as I know he has opinions.

coretl:
I'm not a big fan of the current approach of "make som PV names, and rely on DAQ to tell us if they're right", so I'm keen to find a better solution.

One possibility discussed with @gilesknap was that we make some DLS record alias templates that live in a single support module that we can include for existing beamlines. This could point to dodal as the canonical description of what PV names are required for which type of device.

olliesilvester:
As of https://github.com/DiamondLightSource/dodal/pull/652 , we now have an i24 DCM which has some common parts too. So the DCM base class should now contain common PV's between i03, i22 and i24. And I guess any PV's which only differ in syntax should be renamed to a common name

callumforrester:
@olliesilvester Somewhere, I can't find it, you outlined a nice compositional idea where you build a DCM object out of a bunch of modules, so we can cope with design differences etc.

@coretl RE: finding a better solution than relying on DAQ to tell you if the PVs are correct. I think you should be relying on CI rather than DAQ, I think any solution that relies on CI to tell DAQ and/_or_ controls if the PVs are wrong is a huge improvement on the status quo. We were hoping https://github.com/DiamondLightSource/dodal/issues/507 would provide at least some of that, but open to other ideas.

olliesilvester:
[...]

PR #843, opened 2024-10-16
Crystal metadata change

stan-dot:
Code extracted from https://github.com/DiamondLightSource/dodal/pull/722/files

I don't quite understand why do we set this as a soft signal - that is one failing test at the moment

```python

        # Soft metadata
        # If supplied include crystal details in output of read_configuration
        with self.add_children_as_readables(ConfigSignal):
            if crystal_1_metadata.usage is not None:
                self.crystal_1_usage, _ = soft_signal_r_and_setter(
                    str, initial_value=crystal_1_metadata.usage
                )
            else:
                self.crystal_1_usage = None
                
```

And I deleted one test too

### DIFF
--- a/src/dodal/common/crystal_metadata.py
+++ b/src/dodal/common/crystal_metadata.py
@@ -1,4 +1,3 @@
-import math
 from dataclasses import dataclass
 from enum import Enum
 from typing import Literal
@@ -29,7 +28,6 @@ class CrystalMetadata:
     usage: Literal["Bragg", "Laue"] | None = None
     type: str | None = None
     reflection: tuple[int, int, int] | None = None
-    d_spacing: tuple[float, str] | None = None
 
     def __init__(
         self,
@@ -44,22 +42,4 @@ class CrystalMetadata:
         ), "Reflection plane indices must be positive integers"
         object.__setattr__(self, "type", material.value.name)
         object.__setattr__(self, "reflection", reflection_plane)
-        d_spacing = self.calculate_d_spacing(
-            material.value.lattice_parameter, reflection_plane
-        )
-        object.__setattr__(self, "d_spacing", d_spacing)
         object.__setattr__(self, "usage", usage)
-
-    @staticmethod
-    def calculate_d_spacing(
-        lattice_parameter: float, reflection: tuple[int, int, int]
-    ) -> tuple[float, str]:
-        """
-        Calculates the d-spacing value in nanometers based on the given lattice parameter and reflection indices.
-        """
-        h_index, k_index, l_index = reflection
-        d_spacing_m = lattice_parameter / math.sqrt(
-            h_index**2 + k_index**2 + l_index**2
-        )
-        d_spacing_nm = d_spacing_m * 1e9  # Convert meters to nanometers
-        return round(d_spacing_nm, 5), "nm"
--- a/src/dodal/common/crystal_metadata.py
+++ b/src/dodal/common/crystal_metadata.py
@@ -26,10 +26,10 @@ class CrystalMetadata:
     see https://manual.nexusformat.org/classes/base_classes/NXcrystal.html
     """
 
-    usage: Literal["Bragg", "Laue"] | None = None
-    type: str | None = None
-    reflection: tuple[int, int, int] | None = None
-    d_spacing: tuple[float, str] | None = None
+    usage: Literal["Bragg", "Laue"]
+    type: str
+    reflection: tuple[int, int, int]
+    d_spacing: tuple[float, str]
 
     @staticmethod
     def calculate_default_d_spacing(
--- a/src/dodal/common/crystal_metadata.py
+++ b/src/dodal/common/crystal_metadata.py
@@ -32,7 +32,10 @@ class CrystalMetadata:
     d_spacing: tuple[float, str] | None = None
 
     def __init__(
-        self, material: MaterialsEnum, reflection_plane: tuple[int, int, int]
+        self,
+        material: MaterialsEnum,
+        reflection_plane: tuple[int, int, int],
+        usage: str = "Bragg",
     ) -> None:
         # Determine material from reflection plane prefix
         # Set attributes using object.__setattr__ since the class is frozen
@@ -45,7 +48,7 @@ class CrystalMetadata:
             material.value.lattice_parameter, reflection_plane
         )
         object.__setattr__(self, "d_spacing", d_spacing)
-        object.__setattr__(self, "usage", "Bragg")  # Assuming "Bragg" usage by default
+        object.__setattr__(self, "usage", usage)
 
     @staticmethod
     def calculate_d_spacing(
--- a/src/dodal/common/crystal_metadata.py
+++ b/src/dodal/common/crystal_metadata.py
@@ -1,0 +1,62 @@
+import math
+from dataclasses import dataclass
+from enum import Enum
+from typing import Literal
+
+
+@dataclass(frozen=True)
+class Material:
+    """
+    Class representing a crystalline material with a specific lattice parameter.
+    """
+
+    name: str
+    lattice_parameter: float  # Lattice parameter in meters
+
+
+class MaterialsEnum(Enum):
+    Si = Material(name="silicon", lattice_parameter=5.4310205e-10)
+    Ge = Material(name="germanium", lattice_parameter=5.6575e-10)
+
+
+@dataclass(frozen=True)
+class CrystalMetadata:
+    """
+    Metadata used in the NeXus format,
+    see https://manual.nexusformat.org/classes/base_classes/NXcrystal.html
+    """
+
+    usage: Literal["Bragg", "Laue"] | None = None
+    type: str | None = None
+    reflection: tuple[int, int, int] | None = None
+    d_spacing: tuple[float, str] | None = None
+
+    def __init__(
+        self, material: MaterialsEnum, reflection_plane: tuple[int, int, int]
+    ) -> None:
+        # Determine material from reflection plane prefix
+        # Set attributes using object.__setattr__ since the class is frozen
+        assert all(
+            i > 0 for i in reflection_plane
+        ), "Reflection plane indices must be positive integers"
+        object.__setattr__(self, "type", material.value.name)
+        object.__setattr__(self, "reflection", reflection_plane)
+        d_spacing = self.calculate_d_spacing(
+            material.value.lattice_parameter, reflection_plane
+        )
+        object.__setattr__(self, "d_spacing", d_spacing)
+        object.__setattr__(self, "usage", "Bragg")  # Assuming "Bragg" usage by default
+
+    @staticmethod
+    def calculate_d_spacing(
+        lattice_parameter: float, reflection: tuple[int, int, int]
+    ) -> tuple[float, str]:
+        """
+        Calculates the d-spacing value in nanometers based on the given lattice parameter and reflection indices.
+        """
+        h_index, k_index, l_index = reflection
+        d_spacing_m = lattice_parameter / math.sqrt(
+            h_index**2 + k_index**2 + l_index**2
+        )
+        d_spacing_nm = d_spacing_m * 1e9  # Convert meters to nanometers
+        return round(d_spacing_nm, 5), "nm"
--- a/src/dodal/common/crystal_metadata.py
+++ b/src/dodal/common/crystal_metadata.py
@@ -1,3 +1,4 @@
+import math
 from dataclasses import dataclass
 from enum import Enum
 from typing import Literal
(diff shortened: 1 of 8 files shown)


stan-dot:
the error has to do with crystal spacing values

callumforrester:
What is the reason for this change?

stan-dot:
@callumforrester this is a refactor as `d_spacing` can be directly traced from the reflection values and the material properties.

Neater for the `i18` lookup service

src/dodal/common/crystal_metadata.py
```diff
@@ -0,0 +1,65 @@
...
+        object.__setattr__(self, "usage", usage)
+
+    @staticmethod
+    def calculate_d_spacing(
+        lattice_parameter: float, reflection: tuple[int, int, int]
+    ) -> tuple[float, str]:
+        """
+        Calculates the d-spacing value in nanometers based on the given lattice parameter and reflection indices.
+        """
+        h_index, k_index, l_index = reflection
```
rtuck99, line 60: Do we have sources for these values? I believe the values used up until now have ultimately been derived from literature as they are experimentally measured properties, and I think the feeling is that we should cite sources for these values and use them directly rather than calculate them.

stan-dot: I heard from @iain-hall  that there is a formula for the d-spacing from the reflection parameters

if we're talking about a lattice parameter sure it's a good idea to include a link
https://physics.nist.gov/cgi-bin/cuu/Value?asil

callumforrester: You should probably consult with a controls engineer to make sure the formula is general and not only applicable to I18's DCM, otherwise you could break I22's. The DCMs are all subtly different and were designed in blocks. @JamesOHeaDLS might know more.

JamesOHeaDLS: Hello,

There is a general formula for deriving d-spacing from a particular crystals unit cell length and miller indices (d = a/sqrt(h^2 + k^2 + l^2) but the unit cell length is temperature dependent, and at the motor controller level we specify the d-spacing, rather than derive it.

For our coordinate system calculations we take the d-spacing from literature.  For example the 'Accel DCM' coordinate system has, as a default, the d-spacing of Silicon 111 (3.1355 A).  This would have been taken from reference literature.

This coordinate system parameter is overridable, so if the beamline wanted to use a different value, or the mono in question uses a different type of crystal, it could be specified.  It is also settable and gettable from the EPICS layer.

For example there is the X-Ray Data Booklet (you see copies of this orange booklet all over DLS. I have a copy myself). 
 See section 4.1  - https://xdb.lbl.gov/Section4/Sec_4-1.html which tabulates d-spacings for a range of crystals.  Here Silicon 111 has a value of 6.2712 in its 2d column (ie a d of 3.1356).

Note that the stated d-spacing can vary with temperature.  The above is for room temperature.  It is worth noting that the Diamond quad crystal mono coordinate system uses a hard-coded d-spacing for Si111 at liquid nitrogen temperature (3.13475).

So I think it makes more sense to just get the beamline scientists to specify the d-spacing they wish to use, and then ensure that the software (at all levels) uses the same value.  

James

DominicOram: >  It is also settable and gettable from the EPICS layer.

If this is true for DCMs on all the beamlines then I strongly support just always reading it from EPICS and using that as the one source of truth.

JamesOHeaDLS: Yes this is true for all dual crystal monos, but not for quad (but could be very easily)

stan-dot: thanks @JamesOHeaDLS for the in-depth explanation.

now I agree with @DominicOram that EPICS is the best source.


Revising the PR now

stan-dot:
thanks @rtuck99 , merged

src/dodal/devices/dcm.py
```diff
@@ -36,4 +43,16 @@ def __init__(
...
+            )
+            self.crystal_metadata_type, _ = soft_signal_r_and_setter(
+                str, initial_value=crystal_metadata.type
+            )
+            self.crystal_metadata_reflection, _ = soft_signal_r_and_setter(
+                Sequence[int],
+                initial_value=list(crystal_metadata.reflection),  # type: ignore
+            )
+            # todo note this fails at i18 for now, waiting for an EPICS fix
+            self.crystal_metadata_d_spacing = epics_signal_r(float, "DSPACING")
```
DominicOram, line 57: Must: We need to read `DSPACING:RBV` as this contains the read value, `DSPACING` is currently 0

src/dodal/common/crystal_metadata.py
```diff
@@ -0,0 +1,65 @@
...
+        object.__setattr__(self, "type", material.value.name)
+        object.__setattr__(self, "reflection", reflection_plane)
+        d_spacing = self.calculate_d_spacing(
+            material.value.lattice_parameter, reflection_plane
+        )
+        object.__setattr__(self, "d_spacing", d_spacing)
+        object.__setattr__(self, "usage", usage)
+
+    @staticmethod
+    def calculate_d_spacing(
```
DominicOram, line 54: Should: So now for the i22 version of the DCM we have to calculate the `d-spacing` ourselves but for the i03/i18 version it's on a PV? Could you reach out to the i22 controls guys to get the PV put there and add a comment to https://github.com/DiamondLightSource/dodal/issues/592 that this is a pre-requisite

src/dodal/common/crystal_metadata.py
```diff
@@ -0,0 +1,65 @@
...
+        assert all(
+            i > 0 for i in reflection_plane
+        ), "Reflection plane indices must be positive integers"
+        object.__setattr__(self, "type", material.value.name)
+        object.__setattr__(self, "reflection", reflection_plane)
+        d_spacing = self.calculate_d_spacing(
+            material.value.lattice_parameter, reflection_plane
+        )
+        object.__setattr__(self, "d_spacing", d_spacing)
+        object.__setattr__(self, "usage", usage)
```
DominicOram, line 51: Should: Using `__setattr__` here is a bit of a code-smell and by doing this "by the backdoor" you're causing potential issues:
* I can make a typo in the attr name and I don't know until runtime
* I can set attributes to any type and the type checker won't notice

I think there are some cleaner ways to do this, e.g. my immediate thought would be a factory function:
```python
@dataclass(frozen=True)
class CrystalMetadata:
    """
    Metadata used in the NeXus format,
    see https://manual.nexusformat.org/classes/base_classes/NXcrystal.html
    """

    usage: Literal["Bragg", "Laue"] | None = None
    type: str | None = None
    reflection: tuple[int, int, int] | None = None
    d_spacing: tuple[float, str] | None = None


def calculate_d_spacing(
    lattice_parameter: float, reflection: tuple[int, int, int]
) -> tuple[float, str]:
    """
    Calculates the d-spacing value in nanometers based on the given lattice parameter and reflection indices.
    """
    h_index, k_index, l_index = reflection
    d_spacing_m = lattice_parameter / math.sqrt(h_index**2 + k_index**2 + l_index**2)
    d_spacing_nm = d_spacing_m * 1e9  # Convert meters to nanometers
    return round(d_spacing_nm, 5), "nm"


def make_crystal_metadata_from_material(
    material: MaterialsEnum,
    reflection_plane: tuple[int, int, int],
    usage: Literal["Bragg", "Laue"] = "Bragg",
):
    d_spacing = calculate_d_spacing(material.value.lattice_parameter, reflection_plane)
    return CrystalMetadata(usage, material.value.name, reflection_plane, d_spacing)
```

But you could probably also do it by making use of `pydantic` rather than `dataclasses`.

src/dodal/common/crystal_metadata.py
```diff
@@ -0,0 +1,65 @@
...
+class CrystalMetadata:
+    """
+    Metadata used in the NeXus format,
+    see https://manual.nexusformat.org/classes/base_classes/NXcrystal.html
+    """
+
+    usage: Literal["Bragg", "Laue"] | None = None
+    type: str | None = None
+    reflection: tuple[int, int, int] | None = None
+    d_spacing: tuple[float, str] | None = None
```
DominicOram, line 32: Should: Are these really all `None`? Are there any times we only know some of these params and not others?

DominicOram (changes requested):
Thanks Stan. Looks like i18 has the PVs now. Some comments in code though

stan-dot:
```python


<<<<<<< HEAD
async def test_crystal_metadata_not_propagated_when_not_supplied():
        dcm = DoubleCrystalMonochromator(
            prefix="FOO-MO",
            temperature_prefix="FOO-DI",
            crystal_1_metadata=None,
            crystal_2_metadata=None,
        )

    configuration = await dcm.read_configuration()
    expected_absent_keys = {
        "crystal-1-usage",
        "crystal-1-type",
        "crystal-1-reflection",
        "crystal-1-d_spacing",
        "crystal-2-usage",
        "crystal-2-type",
        "crystal-2-reflection",
        "crystal-2-d_spacing",
    }
    assert expected_absent_keys.isdisjoint(configuration)


=======
>>>>>>> e8ac54553 (add to the beamline files)


```


@DominicOram  all is fixed besides this test. I am not sure what is the purpose of the behavior here to have absent keys. For it to pass the metadata param in the DCM __init__ would need to be `CrystalMetadata | None` and not `CrystalMetadata`. 

I'm happy to delete the test, git blame points last modification to the test, for May 2024 by @callumforrester  - what is the reason for the test? I'm equally happy to keep the test and change the behavior provided I note the reason for allowing `None`  in the code.

DominicOram:
>  all is fixed besides this test...

Ok, I'm happy to retract https://github.com/DiamondLightSource/dodal/pull/843/files#r1820621900 so that we can keep this usecase. I'm still not 100% sure when it would be necessary but that's scope creep.

src/dodal/common/crystal_metadata.py
DominicOram: As discussed, this is scope creep

src/dodal/devices/dcm.py
stan-dot: I thought that :RBV is added automatically in epics_signal_r but ok, will amend this

src/dodal/common/crystal_metadata.py
stan-dot: I'd prefer to avoid __setattr__ too!

Not sure how this came to be. Will look into the pydantic part

src/dodal/common/crystal_metadata.py
stan-dot: added the EPICS source, resolving the conversation

stan-dot:
@DominicOram  I'm afraid I don't understand your last comment

src/dodal/common/crystal_metadata.py
DominicOram: I think remove it, I don't see why anyone would want to regularly calculate it for comparison so it'll just be extra code that's never used.

stan-dot: good idea - so both that and the frozen class now need changing

src/dodal/devices/dcm.py
stan-dot: changed

stan-dot:
```python
 self.crystal_metadata_reflection, _ = soft_signal_r_and_setter(
                Sequence[int],
                initial_value=list(cm.reflection),  # type: ignore
            )
            
```


I think this might be due to ophyd-async change

src/dodal/common/crystal_metadata.py
stan-dot: changed into a dataclass setup

src/dodal/common/crystal_metadata.py
```diff
@@ -0,0 +1,61 @@
...
+    see https://manual.nexusformat.org/classes/base_classes/NXcrystal.html
+    """
+
+    usage: Literal["Bragg", "Laue"]
+    type: str
+    reflection: tuple[int, int, int]
+    d_spacing: tuple[float, str]
+
+    @staticmethod
+    def calculate_default_d_spacing(
```
DominicOram, line 35: Should: Sorry, maybe I've misunderstood, I thought the conclusion of the discussion at https://github.com/DiamondLightSource/dodal/pull/843#discussion_r1827660919 was to remove all this calculation and use the PV? Or is it not yet on i22?

src/dodal/devices/i22/dcm.py
```diff
@@ -74,63 +61,45 @@ def __init__(
 
         # Soft metadata
         # If supplied include crystal details in output of read_configuration
-        crystal_1_metadata = crystal_1_metadata or CrystalMetadata()
-        crystal_2_metadata = crystal_2_metadata or CrystalMetadata()
         with self.add_children_as_readables(ConfigSignal):
             if crystal_1_metadata.usage is not None:
```
DominicOram, line 65: Should: If we've now removed the option for these to be `None` then I don't think we need to check them all?

tests/devices/unit_tests/test_dcm.py
```diff
@@ -13,13 +13,24 @@ async def dcm() -> DCM:
     return dcm
 
 
+async def test_metadata_reflection(dcm: DCM):
+    signal = dcm.crystal_metadata_reflection
+    v = await signal.read()
+    print(v)
```
DominicOram, line 19: Should: It's not best practice to leave print statements in code

src/dodal/common/crystal_metadata.py
DominicOram: If the PV isn't on i22 yet then I said in that comment, can we get the ball rolling on that?

DominicOram (changes requested):
Thanks @stan-dot. Couple of minor queries, I think I may have misunderstood a previous discussion.

src/dodal/devices/i22/dcm.py
stan-dot: removed

src/dodal/common/crystal_metadata.py
stan-dot: there was a discussion was to add the PV on i18, which ultimately just was a slight name change.

regarding the static method
>             self.crystal_metadata_d_spacing = epics_signal_r(float, "DSPACING:RBV")

at the moment there is the PV value used. I might have mistaken with the line above where the 'initial value' is used.

at the moment the only place when the calculation is used is in the `make_crystal_metadata_from_material` where if the d_spacing is not supplied by the user, it's calculated. 

we could remove it and make it required

DominicOram: Ok, so we still need it in there for i22. That's fine, I'll make a note on the conversion issue.

DominicOram (approved):
Looks good to me, thank you!

PR #851, opened 2024-10-18
Catch device creation exceptions in test_device_instantiation

### LINKED ISSUE #850: I22 connection tests did not fail when they weren't mocking
#847 appears to have been caused by #841 not passing `mock=True` to the `DeviceInitializationController`, that should have caused a timeout failure when the CI tried to connect to PVs that don't exist on its network, instead the tests passed after a very long time, suggesting that something was timing out but not failing the test, perhaps an exception was being swallowed.

This issue is to investigate and prevent recurrence

## Acceptance Criteria
* The cause of the non-failure is known
* A regression test/tests are in place to prevent recurrence

callumforrester:
Fixes #850 

1. Run unit tests
2. Break a device and confirm tests fail

### Checks for reviewer

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,6 +13,7 @@ import pytest
 from bluesky.run_engine import RunEngine
 from ophyd.status import Status
 from ophyd_async.core import (
+    NotConnected,
     PathInfo,
     PathProvider,
 )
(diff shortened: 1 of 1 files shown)


callumforrester:
Unsurprisingly, since these tests have been swallowing exceptions for 5 months, there are now previously hidden errors that we need to decide what to do about. Most of them seem to revolve around `/dls_sw` not being available in the CI environment, which I think is fair, our device instantiation tests should be able to run without it. For one thing, it is down right now.

It appears this wasn't a problem previously because `fake_with_ophyd_sim=True` was being used to create fake ophyd device classes via [ophyd.sim.make_fake_device](https://github.com/bluesky/ophyd/blob/84d9e3dfb0e3c1743f211f21954eaa17953b4c1c/ophyd/sim.py#L1157), which strips out all of the logic from the ophyd device class and essentially makes a mock. Now many of those devices have been migrated to ophyd-async, whose mock feature only affects signal backends, not the logic in the class constructor/methods.

One fix I thought of was to make sure that `/dls_sw` is not accessed in any of the device constructors or `connect` methods, in which case `test_device_instantation.py` and `dodal connect -s` would never touch it, but I am open to alternatives. Tagging @DominicOram, @DiamondJoseph and @noemifrisina for opinions.

DominicOram:
I think @rtuck99 has fixed some of these tests in https://github.com/DiamondLightSource/dodal/pull/858

callumforrester:
Excellent! I'll rebase once that's in and see where we are, thanks

callumforrester:
Completely superseded by #868

PR #853, opened 2024-10-18
Fix p38 init call

DiamondJoseph:
Fixes an issue with p38 that was discovered in debugging #850 
Prevents further issues with constructors not reflecting changes to the device class by switching that beamline to @device_factory.
As #706 added functionality to the Undulator that is not required/used on i22 and p38, switched those devices to fetch a blank lookup table (functionally identical to what was set for i22) that should be available regardless of the DLS filesystem.

1. Ensure that p38 connect test with mock devices has no exceptions

### Checks for reviewer


DiamondJoseph:
Removed Dom since this isn't passing due to other issues from #850 so we can discuss what to do about i22 and p38 until they are resolved.

stan-dot (approved):
lgtm

DiamondJoseph:
Probably a victim of reverting then reinstating the device_factory.

PR #839, opened 2024-10-14
Remove Inject

DiamondJoseph:
Fixes #599, #845 current issues with `inject` being broken currently. 

Instead, we should pass in the  instance of the device and decorate our plans to ensure devices are connected prior to running. 

### Instructions to reviewer on how to test:
1. Do thing x
2. Confirm thing y happens

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://diamondlightsource.github.io/dodal/main/reference/device-standards.html)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

PR #854, opened 2024-10-18
Add deprecation warning to inject

### LINKED ISSUE #845: Deprecate `inject` method
`inject` only worked by coincidence of Pydantic coercion: now that we can have device singletons created without connect being called until appropriate, we can use the device instances as the default arguments to plans.

## Acceptance Criteria
* inject has been marked as deprecated
* plans and stubs that currently make use of it in dodal have been replaced by a call to the device factory of the appropriate name called with `(connect_immediately=False)`
* documentation has been updated to reflect this new recommended way of injecting default arguments into plans
* plans in our plan repositories (mx-bluesky, i22-bluesky, etc.) have a pathway (issues or PRs) for migrating to the new method

callumforrester:
Fixes #845 
Supersedes #839 

1. Confirm tests pass
2. Optional: write a plan that uses `inject` and confirm it raises a warning when run.

### Checks for reviewer

### DIFF
--- a/docs/how-to/include-devices-in-plans.md
+++ b/docs/how-to/include-devices-in-plans.md
@@ -1,0 +1,60 @@
+# Include Devices in Plans
+
+There are three main ways to include dodal devices in plans
+
+## 1. Pass as Argument
+
+```python
+import bluesky.plans as bp
+
+from bluesky.protocols import Readable
+from bluesky.utils import MsgGenerator
+from dodal.beamlines import i22
+
+def my_plan(detector: Readable) -> MsgGenerator:
+    yield from bp.count([detector])
+
+RE(my_plan(i22.saxs()))
+```
+
+This is useful for generic plans that can run on a variety of devices and are not designed with any specific device in mind.
+
+## 2. Pass as Default Argument
+
+```python
+import bluesky.plans as bp
+
+from bluesky.protocols import Readable
+from bluesky.utils import MsgGenerator
+from dodal.beamlines import i22
+
+def my_plan(detector: Readable = i22.saxs(connect_immediately=False)) -> MsgGenerator:
+    yield from bp.count([detector])
+
+RE(my_plan()))
+```
+
+This is useful for plans that will usually, but not exclusively, use the same device.
+
+## 3. Instantiate Within the Plan
+
+```python
+import bluesky.plans as bp
+import ophyd_async.plan_stubs as ops
+
+from bluesky.protocols import Readable
+from bluesky.utils import MsgGenerator
+from dodal.beamlines import i22
+
+def my_plan() -> MsgGenerator:
+    detector = i22.saxs(connect_immediately=False)
+    # We need to connect via the stub instead of directly 
+    # so that the RunEngine performs the connection in the
+    # correct event loop
+    yield from ops.ensure_connected(detector)
+    yield from bp.count([detector])
+
+RE(my_plan()))
+```
+
+This is useful for plans that are designed to only ever work with a specific device.
--- a/docs/how-to/include-devices-in-plans.md
+++ b/docs/how-to/include-devices-in-plans.md
@@ -1,6 +1,6 @@
 # Include Devices in Plans
 
-There are three main ways to include dodal devices in plans
+There are two main ways to include dodal devices in plans
 
 ## 1. Pass as Argument
 
@@ -34,27 +34,4 @@ def my_plan(detector: Readable = i22.saxs(connect_immediately=False)) -> MsgGene
 RE(my_plan()))
 ```
 
-This is useful for plans that will usually, but not exclusively, use the same device.
-
-## 3. Instantiate Within the Plan
-
-```python
-import bluesky.plans as bp
-import ophyd_async.plan_stubs as ops
-
-from bluesky.protocols import Readable
-from bluesky.utils import MsgGenerator
-from dodal.beamlines import i22
-
-def my_plan() -> MsgGenerator:
-    detector = i22.saxs(connect_immediately=False)
-    # We need to connect via the stub instead of directly 
-    # so that the RunEngine performs the connection in the
-    # correct event loop
-    yield from ops.ensure_connected(detector)
-    yield from bp.count([detector])
-
-RE(my_plan()))
-```
-
-This is useful for plans that are designed to only ever work with a specific device.
+This is useful for plans that will usually, but not exclusively, use the same device or that are designed to only ever work with a specific device.
--- a/src/dodal/common/coordination.py
+++ b/src/dodal/common/coordination.py
@@ -1,4 +1,6 @@
 import uuid
+import warnings
+from textwrap import dedent
 from typing import Any
 
 from dodal.common.types import Group
@@ -37,4 +39,20 @@ def inject(name: str) -> Any:  # type: ignore
 
     """
 
+    warnings.warn(
+        dedent("""
+        Inject is deprecated, users are now expected to call the device factory 
+        functions in dodal directly, these will cache devices as singletons after
+        they have been called once. For example:
+
+        import i22
+
+        + def my_plan(detector: Readable = i22.saxs()) -> MsgGenerator:
+            ...
+
+        Where previously the default would have been inject("saxs")
+        """),
+        DeprecationWarning,
+        stacklevel=2,
+    )
     return name
--- a/src/dodal/common/coordination.py
+++ b/src/dodal/common/coordination.py
@@ -41,7 +41,7 @@ def inject(name: str) -> Any:  # type: ignore
 
     warnings.warn(
         dedent("""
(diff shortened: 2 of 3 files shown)


callumforrester:
@DominicOram is there anything else from your comments in #841 that you feel would be appropriate to document here?

docs/how-to/include-devices-in-plans.md
```diff
@@ -0,0 +1,60 @@
...
+
+def my_plan() -> MsgGenerator:
+    detector = i22.saxs(connect_immediately=False)
+    # We need to connect via the stub instead of directly 
+    # so that the RunEngine performs the connection in the
+    # correct event loop
+    yield from ops.ensure_connected(detector)
+    yield from bp.count([detector])
+
+RE(my_plan()))
```
DiamondJoseph, line 57: I don't want to encourage this, I want to always have the device as a default argument. 
This prevents any metrics/tracing/logging/documenting on the arguments of the plan, leads to N different plans that are just using a different device, makes it easier for the scope of the device to be managed incorrectly and cause issues (conditionally instantiating my device as mock/connecting my device?).
I think the Finder in GDA was a mistake that made every momentary decision to use it slightly easier and every subsequent attempt to extract what the hell was going on a lot harder.

I think it's risking making debugging a lot harder for a small benefit in verbosity, so I would like more attention paid to the above and comments about the benefits of doing it that way.

I'm prepared to be told I'm getting in the way of writing plans for this standpoint. I really don't like it.

DiamondJoseph: It also means we can call ensure_connected at the top of the plan and fail fast if any device won't be available, rather than after completing the first iteration of the fastest axis: and we can call connect on all of the devices in parallel rather than on each one as we encounter it.

callumforrester: To be honest I broadly agree. Dependency injection is a good thing and, in my opinion, having lots of defaulted parameters is not a particularly terrible thing. Maybe we take this out for now and add it later if demand calls for it?

DiamondJoseph (approved):

DominicOram:
> @DominicOram is there anything else from your comments in #841 that you feel would be appropriate to document here?

Sorry, been in and out of the office. Maybe we should have a chat about this when I'm back. I'm concerned that by encouraging the use of the factory function in the default args we're:
* Causing the device to be initialized on import of the plan (even though it's not connecting)
* Tying the plan to a specific beamline, you could pass in other devices I guess but it's not obvious and you've already started initialising a different beamline's device

DiamondJoseph:
With aims of making plans more generic and more available, I think it encourages having modules like `mx-bluesky` with i04.py [with i04 defaults injected] and plans.py without: blueapi can import both, plans get written into i04.py first then genericised and moved into plans.py when ready.
And moves us away from e.g. `i22-bluesky` individual beamline module.

callumforrester:
One pattern to avoid side-effect imports could be:

```python
def my_plan(det: Factory[Readable] = i22.saxs) -> MsgGenerator:
    yield from bp.count([det()])
```

As far as tying the plan to the beamline goes, there is a tradeoff between that and having too many parameters that you have to fill out every time. You always still have the option of making the plan generic via

```python
def my_plan(det: Readable) -> MsgGenerator: ...
```

But this helps you in the cases that really are beamline-specific. My thinking on plans is that there will be a layer that forms a "beamline API" (see [this comment](https://github.com/DiamondLightSource/blueapi/issues/652#issuecomment-2399823095)). So I can imagine wrapping generic plans in beamline specific plans e.g.

```python
from saxs_waxs import linkam_experiment as generic_linkam_experiment  # Multi-beamline plan

# I22 specific version with fewer parameters
def linkam_experiment(start: float, stop: float, stages: int = 1, saxs: Factory[Pilatus] = i22.saxs, waxs: Factory[Pilatus] = i22.waxs) -> MsgGenerator:
    yield from  generic_linkam_experiment(start=start, stop=stop, stages=stages, min_temp=-160, detectors=[saxs, waxs])
```

DiamondJoseph:
```python
def my_plan(det: Factory[Readable] = i22.saxs):
```

Then blueapi's context now passes references to the DFF instead of the device? It seems very clunky to ask people to write plans that look like that.

callumforrester:
I agree it's not ideal, was just exploring ways of avoiding import side effects. I wonder if @coretl has any thoughts, since he's the main proponent of the use case of writing a plan offline and migrating it to blueapi with minimal effort.

coretl:
Is there anything wrong with initializing devices as default args in beamline plans?
```python
from saxs_waxs import linkam_experiment as generic_linkam_experiment  # Multi-beamline plan

# I22 specific version with fewer parameters
def linkam_experiment(start: float, stop: float, stages: int = 1, saxs = i22.saxs(), waxs = i22.waxs()) -> MsgGenerator:
    yield from  generic_linkam_experiment(start=start, stop=stop, stages=stages, min_temp=-160, detectors=[saxs, waxs])
```
Then when they are made generic they will lose their beamline defaults and be wrapped with a bl specific plan

olliesilvester:
> Is there anything wrong with initializing devices as default args in beamline plans?
>
It's been causing some issues with unit tests - importing the plan instantiates the device with its default arguments, leading to typing errors when the test tries to create a fake device for this plan

DiamondJoseph:
> > Is there anything wrong with initializing devices as default args in beamline plans?
> 
> It's been causing some issues with unit tests - importing the plan instantiates the device with its default arguments, leading to typing errors when the test tries to create a fake device for this plan

Do you have an example of a plan that's causing issues? [There was discussion around passing *args, **kwargs to the device factory function](https://github.com/DiamondLightSource/dodal/pull/597#discussion_r1740700701) and the ability to pass args was removed- this would only have helped if your test function called `device(args)` before importing `plans.py`, but importing a module during a test is something that can be done: and with intentions of dodal linting imports to ensure that devices never import from plans, something that should ensure the correct order?

callumforrester:
@coretl A more general answer: https://docs.astral.sh/ruff/rules/function-call-in-default-argument/#why-is-this-bad

DiamondJoseph:
@olliesilvester just had a chat with me, think we can resolve what's going on much easier if it's just whether the device is `mock` or not that needs to be set, not extra init arguments:

**A device is not "mock" or "real" until connect() has been called on the device**. 
[I've made an issue because we already think that the device should not eagerly connect by default](https://github.com/DiamondLightSource/dodal/issues/856), default arguments of plans should always definitely have `connect_immediately=False` to prevent connecting being a side effect of importing plans.

e.g.

```python
@device_factory()
def saxs()
   return PilatusDetector()


def plan(det: saxs(connect_immediately=False)):
    yield from ensure_connected(saxs)
    yield from {}


def test_that_only_gets_mock_saxs():
    saxs(connect_immediately=True, mock=True)
    # s is the same instance of device in the default arg, and only now connects and only now is mock or not
    yield from plan()
```

coretl:
> @coretl A more general answer: https://docs.astral.sh/ruff/rules/function-call-in-default-argument/#why-is-this-bad

> If the use of a singleton is intentional, assign the result call to a module-level variable, and use that variable in the default argument:

Ok fine...

```python
from saxs_waxs import linkam_experiment as generic_linkam_experiment  # Multi-beamline plan

SAXS = i22.saxs()
WAXS = i22.waxs()

# I22 specific version with fewer parameters
def linkam_experiment(start: float, stop: float, stages: int = 1, saxs = SAXS, waxs = WAXS) -> MsgGenerator:
    yield from  generic_linkam_experiment(start=start, stop=stop, stages=stages, min_temp=-160, detectors=[saxs, waxs])
```

DiamondJoseph:
Example from i22 using a `default_devices.py` 

https://github.com/DiamondLightSource/i22-bluesky/pull/99/files#diff-88a0805a81dca16d95038e63da5ca21e5ad4976539050ca654c3bfa006f4e6b0

dperl-dls:
> default arguments of plans should always definitely have connect_immediately=False to prevent connecting being a side effect of importing plans

I really don't think you should have to _add code_ to prevent there being side effects on import

DiamondJoseph:
I don't know of a way to know that the call to ixx.device() is in the defaults of a method or not, so if a device can be connected on creation, [even if I make it less likely to be the default](https://github.com/DiamondLightSource/dodal/issues/856), I'd prefer to be explicit.

The alternative is that we **can't ever** connect a device on creation and always have to call ensure_connected in a plan or device.connect()

dperl-dls:
yes, maybe that would be better

DiamondJoseph:
I've opted for the latter: blueapi will call connect on all devices concurrently on startup, any users on the command line should ensure their devices are connected before use by awaiting `device.connect()`, and any plans with defaults should be written with `yield from ensure_connected(*devices)` to make the case simpler for both of the above.

PR #860, opened 2024-10-23
Prevent devices connecting as side effect of use elsewhere

### LINKED ISSUE #856: device_factory should default to eager_connect=False
While calling connect multiple times is idempotent, it cannot be called with differing values of mock, and a device cannot be disconnected once connected. The primary use of the device_factory will be to be called for default arguments with connect=False, with ensure_connected called on all devices when they are first passed into plans.

## Acceptance Criteria
* The decorator passes False by default to the Controller's init method.

DiamondJoseph:
Fixes #856, issues discussed in the comments of #854
With this change, devices are *always* created not connected, and it is the responsibility of the user to ensure their device is connected before trying to use it:

e.g. on an ipython terminal:
```python
from bluesky.utils import MsgGenerator
from dodal.beamlines.i22 import saxs
from ophyd_async.core import ensure_connected

s = saxs()
mock = int(time.time()) % 2 == 0

def plan(det: Readable) -> MsgGenerator:
    yield from ensure_connected(det)
    yield from {}

# Connects by any of the below
await s.connect(mock=mock)
RE(ensure_connected(s, mock=mock))
RE(plan(s))
s = saxs(connect_immediately=True)
```

or in blueapi, we will call get_all_device_factories() and call connect on all devices in parallel, reporting on any device connection issues.

### Checks for reviewer

### DIFF
--- a/src/dodal/common/coordination.py
+++ b/src/dodal/common/coordination.py
@@ -49,7 +49,7 @@ def inject(name: str) -> Any:  # type: ignore
         from bluesky.utils import MsgGenerator
         from dodal.beamlines import i22
 
-        def my_plan(detector: Readable = i22.saxs(connect_immediately=False)) -> MsgGenerator:
+        def my_plan(detector: Readable = i22.saxs()) -> MsgGenerator:
             ...
 
         Where previously the default would have been inject("saxs")
--- a/docs/how-to/include-devices-in-plans.md
+++ b/docs/how-to/include-devices-in-plans.md
@@ -28,7 +28,7 @@ from bluesky.protocols import Readable
 from bluesky.utils import MsgGenerator
 from dodal.beamlines import i22
 
-def my_plan(detector: Readable = i22.saxs(connect_immediately=False)) -> MsgGenerator:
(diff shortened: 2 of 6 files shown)


dperl-dls (approved):
Thanks, I think this is much safer

src/dodal/common/coordination.py
```diff
@@ -49,7 +49,7 @@ def scan(x: Movable = inject("stage_x"), start: float = 0.0 ...)
         from bluesky.utils import MsgGenerator
         from dodal.beamlines import i22
 
-        def my_plan(detector: Readable = i22.saxs(connect_immediately=False)) -> MsgGenerator:
+        def my_plan(detector: Readable = i22.saxs()) -> MsgGenerator:
```
callumforrester, line 52: Must: Can you also update the docs and do a grep for `connect_immediately=False` in case we missed it anywhere else?

callumforrester (dismissed):
Looks good, one small thing

src/dodal/common/coordination.py
callumforrester: Never mind, you've already done it, my bad, sorry!

PR #861, opened 2024-10-23
add license check

### LINKED ISSUE #825: Add a github action for licensing compliance checker (testing for the python-copier-template purposes)
To prevent using libraries in our projects that are not compatible with our project licenses. For instance using a library with a more restrictive license in our library with a more permissive license. The likelihood of legal action for many of those libraries might not be that great, but multiplied by many libraries and time might be costly, at least in the developer time to migrate to a compatible alternative

https://github.com/marketplace/actions/dependencies-license-compliance-checker

See this

https://github.com/DiamondLightSource/python-copier-template/pull/199/files

stan-dot:
Fixes #825 

package used initially:
- https://github.com/pilosus/action-pip-license-checker

### DIFF
--- a/.github/workflows/_license.yml
+++ b/.github/workflows/_license.yml
@@ -2,9 +2,34 @@ on:
   workflow_call:
 
 jobs:
-  artifacts:
+  license_check:
     runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - name: Get explicit and transitive dependencies
+        run: |
+          pip install
+          pip freeze > requirements-all.txt
+      - name: Check python
+        id: license_check_report
+        uses: pilosus/action-pip-license-checker@v2
+        with:
+          requirements: "requirements-all.txt"
+          fail: "Copyleft"
+          exclude: "(?i)^(pylint|aio[-_]*).*"
+      - name: Print report
+        if: ${{ always() }}
+        run: echo "${{ steps.license_check_report.outputs.report }}"
 
+  get_report:
+    runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


DiamondJoseph:
This is only run on release- shouldn't we be checking during the PR to add the dependency?

stan-dot:
good question @DiamondJoseph 

I'd say yes the footwork should be done beforehand. The license check in the release is more just for a sanity check, it does not need to happen at the early stage.

two layers > one

DiamondJoseph:
What I mean @stan-dot is that this action should be run during the PR, not that it should be part of the review process. If we add a dependency that is not licensed, then we make use of it it'll be difficult to extract after the fact. I'd expect this PR to have triggered a new check so that we could see that we are not already non-compliant.

stan-dot:
fair enough @DiamondJoseph , where exactly do you consider it is best to run it then? before the lint? after the test?

DiamondJoseph:
It's independent of those, it's just another workflow to be run. Make it a new `_license.yml` in `.github/workflows` and add it to run alongside lint and test in `ci.yml`

coretl:
This doesn't appear to be working, there is no license.json uploaded...

DiamondJoseph:
image

stan-dot:
looking into it

stan-dot:
the package isn't that great.

maybe this is not a fire to put out https://swizec.com/blog/let-small-fires-burn/

stan-dot:
not on critical path now, unassigning

DiamondJoseph:
I think if we're breaking the law and using packages we are not allowed to, that's a pretty big fire. 
I'm back in the office this week if you want a 2nd pair of eyes to look at it, or I can make a ticket for myself to investigate other packages/tools.

stan-dot:
other possible packages
- https://github.com/raimon49/pip-licenses
- https://github.com/graphcore/license-checker
- https://github.com/dhatim/python-license-check

stan-dot:
this PR no longer relevant, comment in the issue - which would need to be solved in a different way

PR #862, opened 2024-10-24
migrate p38 to i22

### LINKED ISSUE #502: Automatically keep I22 and P38 in sync

stan-dot:
Fixes #502

1. `BEAMLINE=p38 dodal connect i22`
2. expect all to connect

### Checks for reviewer


stan-dot:
```python
@pytest.fixture(scope="function")
def module_and_devices_for_beamline(request):
    beamline = request.param
    with patch.dict(os.environ, {"BEAMLINE": beamline}, clear=True):
        bl_mod = importlib.import_module("dodal.beamlines." + beamline)
        importlib.reload(bl_mod)
        mock_beamline_module_filepaths(beamline, bl_mod)
        devices, _ = make_all_devices(
            bl_mod,
            include_skipped=True,
            fake_with_ophyd_sim=True,
        )
        yield (bl_mod, devices)
        beamline_utils.clear_devices()
        del bl_mod



```

the conftest is very opinionated and overriding it 'from the other file' seems impossible,

the quick and easy way to fix is it to use another conftest,

 or add some param to the fixture to skip the os environ patch if it's a lab beamline


@DominicOram , @callumforrester please share your perspectives

src/dodal/beamlines/i22.py
```diff
@@ -25,10 +25,18 @@
...
+LINKAM_IS_IN_LAB = False
 PREFIX = BeamlinePrefix(BL)
 set_log_beamline(BL)
 set_utils_beamline(BL)
 
+
+VISIT_PATH: Path = Path(
+    f"/dls/${'p38' if IS_LAB else 'i22'}/data/2024/cm37271-2/bluesky"
+)
+API_URL: str = f"http://{'p38' if IS_LAB else 'i22'}-control:8088/api"
```
DominicOram, line 38: Should: I think `BL` contains `p38` or `i22` string can't we use that rather than the if statement here?

DominicOram:
> the conftest is very opinionated and overriding it 'from the other file' seems impossible,

Sorry, I'm not sure I understand. You can override the `BEAMLINE` env variable by passing in the beamline you want like in `@pytest.mark.parametrize("module_and_devices_for_beamline", ["i22"], indirect=True)`. Or are you trying to override something else?

On a side-note I would be careful about merging this until we have fixed https://github.com/DiamondLightSource/dodal/issues/864 as it could be masking errors.

src/dodal/beamlines/i22.py
```diff
@@ -25,10 +25,18 @@
 from dodal.utils import BeamlinePrefix, get_beamline_name
 
 BL = get_beamline_name("i22")
+IS_LAB = BL == "p38"
+LINKAM_IS_IN_LAB = False
```
callumforrester, line 29: Should: Our current design for beamlines where some equipment may or may not be there is to try to initialise every device, report the failures and move on, so we should always try to connect the linkam.

src/dodal/beamlines/i22.py
stan-dot: good catch

src/dodal/beamlines/i22.py
stan-dot: ok

src/dodal/beamlines/i22.py
stan-dot: addressed

src/dodal/beamlines/i22.py
stan-dot: deleted that part

stan-dot:
now probably this is just waiting for the connect and device_factory applied to the main branch again

https://github.com/DiamondLightSource/dodal/issues/864

stan-dot:
waiting for https://github.com/DiamondLightSource/dodal/pull/921

stan-dot:
out of date, closing this attempt. once #921 is merged this PR will be much easier

PR #867, opened 2024-10-25
Fix unit test borkage due to loading from /dls_sw

rtuck99:
Prevents unit tests loading files from places where they shouldn't, this is a port from code that's already in the mx-bluesky unit tests.

1. Do the unit tests pass
2. Any attempt to load files from `BANNED_PATHS` inside a unit test will result in an exception (at least inside those files that are in submodules of where the affected `conftest.py` module is.

### Checks for reviewer

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,7 +17,7 @@ from ophyd_async.core import (
     PathProvider,
 )
 
-from dodal.common.beamlines import beamline_utils
+from dodal.common.beamlines import beamline_parameters, beamline_utils
 from dodal.common.visit import (
     DirectoryServiceClient,
     LocalDirectoryServiceClient,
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -37,8 +37,29 @@
     "s03": mock_paths,
     "i04": mock_paths,
     "s04": mock_paths,
+    "i24": mock_paths,
 }
 
+BANNED_PATHS = [Path("/dls"), Path("/dls_sw")]
```
DiamondJoseph, line 43: Can we just ban absolute paths altogether? /dev/null we can still get with `os.devnull`, otherwise everything should be included with the module

rtuck99: I didn't want to exclude /tmp as there are lots of things that write to the tmp directory. As it's a patch on the `open` builtin I didn't want to be too overreaching in case some dependency of ours tries to do something reasonable.
I think this will catch 99% of things, I don't know what other directories we are likely to be touching ourselves.

DominicOram (commented):
I'm not really sure why we need this? The unit tests should fail in CI if we're doing this accidentally anyway or are we just trying to be extra cautious in catching things locally?

rtuck99:
> I'm not really sure why we need this? The unit tests should fail in CI if we're doing this accidentally anyway or are we just trying to be extra cautious in catching things locally?

In theory, yes CI should catch it, however it can be difficult to debug where things are going wrong if it only happens in CI, especially if you don't get all the exception info. We do already have this in mx-bluesky where I introduced it a while back.

DominicOram (approved):
I don't think I personally find it useful to have this locally but I have no strong objection to it if it helps others so all good :)

Issue #760, opened 2024-08-22
Support J and K beamline prefixes 


callumforrester:
Attempts to integrate I13-1 were hampered by its PVs having `BL13J-...` prefixes. Ideally we should be able to configure dodal so `device_instantiation` can pick this up from I13-1. 

Since I04-1 and I20-1 are already in dodal and their PVs resolve to `BL04I-...` and `BL20I-...`, I assume those are the actual prefixes for those beamlines. I will check when able.

## Acceptance Criteria
* We can put some configuration in the `i13_1` dodal module such that `device_instantiation` automatically inserts `BL13J` prefixes at the front of PV names
* We can remove the `bl_prefix=False` argument currently present for all devices
* There are regression tests to ensure we do not break existing beamlines

DominicOram:
Hmm... `i20-1` and `i04-1` are currently also broken because of this. It did work at one point and it's why `BeamlinePrefix` allows you to specify a suffix but clearly there was a regression. Running `dodal connect i20-1` confirms that it's broken but annoyingly `dodal connect i04-1` looks like it works because the hardware in dodal so far is the same for `i04` and `i04-1`

PR #872, opened 2024-10-25
Remove i04-1 as it's unused and broken

### LINKED ISSUE #871: Remove i04-1.py
We currently run no bluesky code in production on `i04-1` and even if we tried it would be broken anyway because of https://github.com/DiamondLightSource/dodal/issues/760. Given upcoming changes in device instantiation this is just dead code that we're having to support and so we should remove it

## Acceptance Criteria
* `i04-1.py` is removed

DominicOram:
Fixes #871 

1. Confirm tests still pass

### Checks for reviewer


olliesilvester (approved):
Looks okay, there's a reference to i04_1 in the docs which I'll change now. What was broken about `i04_1.py` ?

DominicOram:
> Looks okay, there's a reference to i04_1 in the docs which I'll change now. What was broken about `i04_1.py` ?

Any `_1` beamlines don't work as we don't pick up the different prefix i.e. https://github.com/DiamondLightSource/dodal/issues/760. It's not obvious because it's actually connecting to the i04 PVs when you run `dodal connect i04-1`. We don't actually do anything on i04_1 anyway so it seems a bit silly keeping it around when it's broken and not used.

Issue #569, opened 2024-05-22
Make CLI beamline mappings work without mutating environment


callumforrester:
Setting `os.environ["BEAMLINE"]` caused unintended side effects, will make an issue to find a better way

_Originally posted by @callumforrester in https://github.com/DiamondLightSource/dodal/pull/519#discussion_r1609836803_

callumforrester:
The CLI mutates the `$BEAMLINE` environment variable to keep the mappings between the beamline name and the module name consistent. For example if a user types `dodal connect s03`, dodal knows it must use the `i03` module, but the `i03` module depends on `$BEAMLINE` being `s03` to connect to the `s03` PVs.

We need to find a cleaner way of doing this that doesn't force us to keep the CLI argument and the environment variable in sync.

May be related to #502, as a solution to that could supersede the `$BEAMLINE` variable.

Another possible solution is to have a `BEAMLINE` global that is informed by the CLI argument if supplied, and the environment variable if not, and have that passed universally to all parts of the code including the beamline module.

DominicOram:
> Setting `os.environ["BEAMLINE"]` caused unintended side effects

Can you elaborate on the side effects? In general though I agree that forcing the CLI arg and the env variable to be in sync is messy so happy to look for a better way.

callumforrester:
The main one I saw was when I called it in the tests, it could change the behaviour of other tests that called code that read `${BEAMLINE}`, for example: https://github.com/DiamondLightSource/dodal/blob/4f1376592e8e2a155b903147e869efe82c3b8288/tests/unit_tests/test_log.py#L88

I had to make sure `os.environ` was patched for every CLI test, which is a reasonable solution, but as you say it is messy and generally good practice to avoid mutating global state because things like that pop up and can be hard to diagnose.

PR #877, opened 2024-10-31
Implement `dodal connect ALL` to test connection to all beamlines

### LINKED ISSUE #875: Dodal connect --all is broken
When I run `dodal connect --help` I am told I should be able to run `dodal connect --all` to try and connect to all known beamlines. However, when I try this I get `Error: Missing argument '{i04|i24|b01_1|s03|i03|i04-1|i04_1|i13-1|i13_1|i20-1|i20_1|i23|p99|p49|p46|training_rig|p48|p47|i22|p38|p45}'`.

## Acceptance Criteria
* `dodal connect --all` and `dodal connect -a` work as expected

rtuck99:
Fixes #875

Running `dodal connect ALL` will result in the special ALL beamline name being expanded into the list of all known beamlines. `dodal connect` will then attempt to connect to the beamlines in order and print out the connect results for each.

By default, training (in the `training_rig` module) and sim (starting with `s`) beamlines are skipped but can be included with the `--include-training` and `--include-sim` options respectively.

1. `dodal connect` works as advertised
2. Unit tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/cli.py
+++ b/src/dodal/cli.py
@@ -8,6 +8,7 @@ from dodal.beamlines import all_beamline_names, module_name_for_beamline
 from dodal.utils import make_all_devices
 
 from . import __version__
+from .common.beamlines.beamline_utils import clear_devices
 
 
 @click.group(invoke_without_command=True)
@@ -21,7 +22,7 @@ def main(ctx: click.Context) -> None:
 @main.command(name="connect")
 @click.argument(
     "beamline",
-    type=click.Choice(list(all_beamline_names())),
+    type=click.Choice(list(all_beamline_names()) + ["ALL"]),
     required=True,
 )
 @click.option(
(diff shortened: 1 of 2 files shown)


callumforrester:
I'm a bit concerned that I would find this confusing as a user, since a lot of our beamline modules depend on the `${BEAMLINE}` variable anyway. Some example cases I can see that are not obvious:
- Following #862 it is not clear to me if this will connect to I22 and/or P38
- If I include the training rigs I would expect it to try and connect to all of them but it will just use the default (p47)
- As an outside user I may not understand why P46 is a training rig but P45 isn't

I'm leaning away from `--all` option at this time, I would prefer a script (or even 30 parallel CI jobs) for running the connection tests per-beamline.

DominicOram:
I don't think I feel particularly strongly about having this functionality in the CLI or having it at all. I was originally trying to use it to test new connection logic in `dodal` and misread the CLI to think that it was there when actually it wasn't. I then probably pushed to get it in out of frustration. In hindsight I'm happy to drop this if @callumforrester thinks it's not worth doing at this time

callumforrester:
If we fix #569 I could see this being a useful feature. Or alternative some tagging system e.g. `dodal connect --tag mx`, where we tagged all MX beamline modules somehow. But I think we need sort out the role of `${BEAMLINE}` first.

DominicOram:
Wont do for now

Issue #864, opened 2024-10-24
 Dodal connect on i22 is not actually connecting


DominicOram:
If I run `dodal connect i22` I get all of the beamline connected suspiciously quickly. It turns out that if I edit one of the PV prefixes in `i22.py`to be incorrect then `dodal connect i22` will still pass.

## Acceptance Criteria
* If I deliberately break one of the PV prefixes in i22.py then `dodal connect i22` actually fails

DominicOram:
@callumforrester and @DiamondJoseph am I right that this is different to https://github.com/DiamondLightSource/dodal/issues/850?

DiamondJoseph:
This is different than 850: 850 is that when make_all_devices is called, any exceptions in the `device_instantiation` way of creating and connected devices are swallowed and do not cause test failures.
This is that make_all_devices doesn't ever call connect on the devices given by `@device_factory` on the returned devices after they are made- make_all_devices may catch exceptions in instantiating the device, but not in the connect which is what's more useful...

callumforrester:
Additional acceptance criteria: If we could have a regression unit test that would be great

DiamondJoseph:
#865 which is intended to fix this issue (#864) is also running into the same broken connects that were hidden by #850 so they are very much related

DominicOram:
With https://github.com/DiamondLightSource/dodal/pull/869 reverted I don't believe this issue is on main now. However, we will need to address it before reinstating the device factory

callumforrester:
The cause of this issue that `dodal connect <beamline>` does not attempt to connect the devices, but expects `device_instantiation` to do it, which is does as long as `wait_for_connection=True` is passed. In practice this is true often enough that no one has really noticed, but it is technically still a bug in main because `dodal connect` does not work for lazy devices. It is just made worse in the world of `DeviceInitializationController` because all devices are lazy by default there.

So the fix is to make sure the dodal CLI actually calls `ensure_connected` on the devices it connects to.

PR #858, opened 2024-10-22
Fixes for ophyd async types mega merge

rtuck99:
Fixes breakages in dodal for bluesky/ophyd-async#594

~Note this fix depends on #814~
Note this fix now incorporates #814
See also mx-bluesky PR DiamondLightSource/mx-bluesky#593

### Instructions to reviewer on how to test:
1. Unit tests pass
2. Check the following changes:
   * Usage of `StrictEnum`/`SubsetEnum` is appropriate
   * Conversion of signals to numpy array types is appropriate

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://diamondlightsource.github.io/dodal/main/reference/device-standards.html)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

PR #881, opened 2024-11-01
Reinstate `DeviceInitializationController`

### LINKED ISSUE #878: Reinstate new device factory and its use for I22
From @DominicOram, the plan:

> - Reinstate new device factory and it's use for i22
> - In the short term leave inject as it is i.e. currently broken (we can work around this by just specifying the device name in GDA anyway)
> - Fix any outstanding issues with the new factory found from i22 testing
> - Move all beamlines to the new factory
> - Come up with the best way to fix inject, as discussed on the linked issue

This issue covers the first and third points

## Acceptance Criteria
* https://github.com/DiamondLightSource/dodal/pull/841 (reverted in https://github.com/DiamondLightSource/dodal/pull/869) is restored
* #874, #849 and #860 are not present and there are tests to verify this

callumforrester:
Fixes #878 

1. Run unit tests
2. Run system tests for I22
3. Run system tests for at least 1 other beamline

### Checks for reviewer


tests/common/beamlines/test_beamline_utils.py
```diff
@@ -134,3 +136,69 @@ def test_wait_for_v2_device_connection_passes_through_timeout(kwargs, expected_t
...
+def test_device_controller_eager_connect(RE):
+    @beamline_utils.device_factory(mock=True)
+    def device() -> FocusingMirror:
+        return dummy_mirror()
+
+    mirror = device(connect_immediately=True)
+    assert mirror.connect.call_count == 1  # type: ignore
+
+
+def test_multiple_layers_of_lru_caching_does_not_affect_device():
```
DominicOram, line 181: Should: I think it would be good to test that the returned devices are the same without having the extra cache here too

DominicOram (changes requested):
Thanks, some comments in code. Additionally, I don;t think the system tests are working. When I run `dodal connect i22` it passes but if I change a device to be the wrong PV it still passes, I would expect it to fail

callumforrester:
@DominicOram I'm aware the system tests are unhappy, should have commented here, sorry about that.

src/dodal/utils.py
```diff
@@ -110,6 +115,96 @@ def wrapper(*args, **kwds) -> T:
...
+class DeviceInitializationController(Generic[D]):
+    def __init__(
+        self,
+        factory: Callable[[], D],
+        use_factory_name: bool,
+        timeout: float,
+        mock: bool,
+        skip: SkipType,
+    ):
+        self._factory: Callable[[], D] = functools.lru_cache(factory)
```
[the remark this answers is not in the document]
callumforrester: Here is where it uses `ACTIVE_DEVICES`, albeit in a write-only capacity: https://github.com/DiamondLightSource/dodal/blob/ed41f3b4770b252cb766d241b93dfe774e477139/src/dodal/common/beamlines/beamline_utils.py#L160

So maybe you're right, can we get away without that and have a clean separation?

tests/common/beamlines/test_beamline_utils.py
callumforrester: That's implicitly covered by `test_device_factory_can_rename` but I think you may be right that it deserves its own test

src/dodal/beamlines/i22.py
```diff
@@ -46,237 +43,149 @@
...
+        type="silicon",
+        reflection=(1, 1, 1),
+        d_spacing=(3.13475, "nm"),
+    )
+    crystal_2_metadata = CrystalMetadata(
+        usage="Bragg",
+        type="silicon",
+        reflection=(1, 1, 1),
+        d_spacing=(3.13475, "nm"),
+    )
```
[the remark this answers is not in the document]
callumforrester: Good catch

src/dodal/utils.py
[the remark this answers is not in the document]
DominicOram: Yes, I think there's no need for the new instantiation method to write to it

DominicOram:
> @DominicOram I'm aware the system tests are unhappy, should have commented here, sorry about that

No worries, the `Instructions to test` imply they are ok. Will this be addressed in another issue or part of this one?

callumforrester:
It will be addressed here, I don't want to merge this without addressing this problem, I just thought it had already been fixed in #858, but apparently not...

src/dodal/utils.py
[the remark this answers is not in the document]
callumforrester: Done, although I had to change [this test](https://github.com/DiamondLightSource/dodal/blob/b2c947537f605a9e84bf4622e1a59c4ebf86291c/tests/common/beamlines/test_device_instantiation.py#L19) so it no longer relies on `ACTIVE_DEVICES`, see what you think

tests/common/beamlines/test_device_instantiation.py
```diff
@@ -21,15 +21,10 @@ def test_device_creation(RE, module_and_devices_for_beamline):
...
-        assert device_name in beamline_utils.ACTIVE_DEVICES, (
-            f"No device named {device_name} was created for {module}, "
-            f"devices are {beamline_utils.ACTIVE_DEVICES.keys()}"
-        )
-        assert follows_bluesky_protocols(device)
-    assert len(beamline_utils.ACTIVE_DEVICES) == len(devices)
+    _, _, exceptions = module_and_devices_for_beamline
+    if len(exceptions) > 0:
+        raise NotConnected(exceptions)
+    # Pass test if there were no exceptions
```
DominicOram, line 27: Could: This change is fine by me. You could also still check `follows_bluesky_protocols` for all returned devices but I'm not sure it gives you loads. If we're not using `follows_bluesky_protocols` now though we should remove it.

callumforrester:
> It will be addressed here, I don't want to merge this without addressing this problem, I just thought it had already been fixed in https://github.com/DiamondLightSource/dodal/pull/858, but apparently not...

@DominicOram I've changed my mind about this having realised that this bug also affects `device_instantiation` in `main` (albeit less), see #864.

I'm going to take over and finish #865 to address.

tests/common/beamlines/test_device_instantiation.py
```diff
@@ -21,15 +21,17 @@ def test_device_creation(RE, module_and_devices_for_beamline):
...
+    if len(exceptions) > 0:
+        raise NotConnected(exceptions)
+    devices_not_following_bluesky_protocols = [
+        name
+        for name, device in devices.items()
+        if not follows_bluesky_protocols(device)
+    ]
+    assert (
+        len(devices_not_following_bluesky_protocols) == 0
+    ), f"{devices_not_following_bluesky_protocols} do not follow blueesky protocols"
```
DominicOram, line 34: Nit:
```suggestion
    ), f"{devices_not_following_bluesky_protocols} do not follow bluesky protocols"
```

src/dodal/beamlines/i22.py
```diff
@@ -286,107 +191,61 @@ def fswitch(
...
-    return device_instantiation(
-        Linkam3,
-        "linkam",
-        "-EA-TEMPC-05",
-        wait_for_connection,
-        fake_with_ophyd_sim,
-    )
+@device_factory()
+def linkam() -> Linkam3:
+    return Linkam3(prefix=f"{PREFIX.beamline_prefix}-EA-TEMPC-05")
```
DominicOram, line 245: Could: I suspect we need a `:` after the prefix here. Hard to tell 100% as it doesn't connect either way and was previously being skipped so we would have missed it anyway

```suggestion
    return Linkam3(prefix=f"{PREFIX.beamline_prefix}-EA-TEMPC-05:")
```

src/dodal/beamlines/i22.py
```diff
@@ -47,237 +48,141 @@
...
-    return device_instantiation(
-        FocusingMirror,
-        "hfm",
-        "-OP-KBM-01:HFM:",
-        wait_for_connection,
-        fake_with_ophyd_sim,
+@device_factory()
+def hfm() -> FocusingMirror:
+    return FocusingMirror(
+        prefix=f"{PREFIX.beamline_prefix}|-OP-KBM-01:HFM:",
```
DominicOram, line 124: Must: Accidental `|` after the prefix here.
```suggestion
        prefix=f"{PREFIX.beamline_prefix}-OP-KBM-01:HFM:",
```

DominicOram (changes requested):
I found a few issues from running `dodal connect i22` and comparing to `main`

PR #885, opened 2024-11-04
Add constant for zocalo environment constant with new names

olliesilvester:
Fixes [#ISSUE](https://github.com/DiamondLightSource/mx-bluesky/issues/612)
Constant now defined in Dodal instead of `mx_bluesky` since it's needed by the device. Alternatively the Zocalo devices could take the environment as a parameter, but for now there's no need as it'll always use the same string

1. Tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/zocalo/zocalo_constants.py
+++ b/src/dodal/devices/zocalo/zocalo_constants.py
@@ -1,0 +1,3 @@
+from dodal.utils import is_test_mode
+
+ZOCALO_ENV = "dev_bluesky" if is_test_mode() else "bluesky"
--- a/src/dodal/devices/zocalo/zocalo_interaction.py
+++ b/src/dodal/devices/zocalo/zocalo_interaction.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 import zocalo.configuration
 from workflows.transport import lookup
 
+from dodal.devices.zocalo.zocalo_constants import ZOCALO_ENV
 from dodal.log import LOGGER
 
 
@@ -56,7 +57,7 @@ class ZocaloTrigger:
 
     see https://github.com/DiamondLightSource/dodal/wiki/How-to-Interact-with-Zocalo"""
 
-    def __init__(self, environment: str = "artemis"):
+    def __init__(self, environment: str = ZOCALO_ENV):
         self.zocalo_environment: str = environment
 
     def _send_to_zocalo(self, parameters: dict):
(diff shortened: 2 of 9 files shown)


DominicOram (changes requested):
Thanks. I'm not sure it's picking up whether it's in test mode correctly. If I run the system tests no data gets sent to fake zocalo and I get:
```
bluesky.utils.FailedStatus: <AsyncStatus, device: zocalo, task: <coroutine object ZocaloResults.trigger at 0x7f46be97d150>, errored: NoResultsFromZocalo('Timed out waiting for Zocalo results')>
```
But if I change them to be:
```
@pytest.fixture
async def zocalo_device():
    zd = ZocaloResults(zocalo_environment="dev_bluesky")
    await zd.connect()
    return zd


@pytest.mark.s03
async def test_read_results_from_fake_zocalo(zocalo_device: ZocaloResults):
    RE = RunEngine()
    zocalo_device._subscribe_to_results()
    zc = ZocaloTrigger("dev_bluesky")
    zc.run_start(ZocaloStartInfo(0, None, 0, 100, 0))
    zc.run_end(0)
    zocalo_device.timeout_s = 5

    def plan():
        yield from bps.open_run()
        yield from bps.trigger_and_read([zocalo_device])
        yield from bps.close_run()

    RE(plan())

    results = await zocalo_device.read()
    assert results["zocalo-results"]["value"][0] == TEST_RESULT_LARGE
```
I get a data in fake zocalo (and a different error). Additionally, I know it's a bit of scope creep as they're probably broken on `main` but could you fix the system test?

olliesilvester:
OK, there was quite a long wrong with the system tests but I think I have it working now.

- Dodal conftests need to move up a directory so that they are used by system tests
- Local rabbitmq wasn't configured correctly - I needed to manually add the 'results' exchange and create a binding to it to the i03.xrc queue
- `zocalo_device.read()` gives a dict where they keys are different `ZocaloDevice` readables and the values are their `Reading` (special bluesky time). The system test expects a single `Reading`, probably it was made before we had `add_readables` in `ZocaloResults`
I'll make a separate PR which fixes the Zocalo system tests

olliesilvester:
I am also keen to obliterate S03 as you suggested, so then we can properly run the working system tests daily

dperl-dls (approved):
lgtm

olliesilvester:
@dperl-dls @DominicOram Could one you look at https://github.com/DiamondLightSource/dodal/pull/893 to check that this fixes the tests? Then I'll merge things. Thanks!

DominicOram (approved):
This works for me now too, thanks for accepting my scope creep

PR #895, opened 2024-11-08
Test ophyd async speedup

DominicOram:
Following changes in https://github.com/bluesky/ophyd-async/pull/641 the sped up tests uncovered some places where our tests were failing due to references to devices that weren't wrapped in `Reference`. This fixes the tests again.

1. Ensure tests work

### Checks for reviewer


DominicOram:
@Relm-Arrowny - I added the references and fixed the tests but I'm not 100% sure how else the device is used so I don't know if I've broken things elsewhere. I also noticed you're using quite a few `References`, we tend not to need them much in MX, do you want to have a chat to compare approaches?

Relm-Arrowny:
> @Relm-Arrowny - I added the references and fixed the tests but I'm not 100% sure how else the device is used so I don't know if I've broken things elsewhere. I also noticed you're using quite a few `References`, we tend not to need them much in MX, do you want to have a chat to compare approaches?

Yes, but I am away for the next two weeks may be we can make time when I get back.

PR #897, opened 2024-11-11
Fix pin tip detect getting stuck in loop

olliesilvester:
Potential fix for https://github.com/DiamondLightSource/mx-bluesky/issues/634. See https://github.com/DiamondLightSource/mx-bluesky/issues/634#issuecomment-2468011536 for explanation

To test:
1. Run `pip install pytest-timeouts`
2. Confirm new test fails without the added `asyncio.sleep(0)`
3. Confirm test passes with it

### Checks for reviewer


src/dodal/devices/oav/pin_image_recognition/__init__.py
```diff
@@ -153,6 +153,7 @@ async def _set_triggered_tip():
             """
             async for value in observe_value(self.array_data):
                 try:
+                    await asyncio.sleep(0)
                     location = await self._get_tip_and_edge_data(value)
                     self._set_triggered_values(location)
                 except Exception as e:
```
rtuck99, line 159: Would be good if the `InvalidPinException` was raised with a message as currently the log message doesn't print any information about the exception.

rtuck99 (approved):
Approved.

I'm not sure how much sense the explanation makes - since there are plenty of awaits. However I suppose none of them actually do any IO so perhaps the thread never yields.

This does seem like quite a significant thing to omit from the documentation for `wait_for()` though. Unless that's implied by the caveat regarding having to wait for the cancellation of the future.

I guess this means this could be something to watch out for in future where we are waiting for something which is only dependent on the values of soft signals and doesn't involve real IO.

src/dodal/devices/oav/pin_image_recognition/__init__.py
olliesilvester: In the longs, don't we see the
`LOGGER.warning(f"Failed to detect pin-tip location, will retry with next image: {e}")`
message each time this exception is raised?

coretl:
I've tried to reproduce this in ophyd-async but can't so far. This passes without needing the `asyncio.sleep`:
```python
async def test_observe_value_times_out_with_busy_sleep():
    sig, setter = soft_signal_r_and_setter(float)

    async def tick():
        for i in range(5):
            await asyncio.sleep(0.1)
            setter(i + 1)

    recv = []

    async def watch():
        async for val in observe_value(sig, timeout=0.5):
            time.sleep(0.15)
            recv.append(val)

    t = asyncio.create_task(tick())
    start = time.time()
    try:
        with pytest.raises(asyncio.TimeoutError):
            await asyncio.wait_for(watch(), timeout=0.2)
        assert recv == [0, 1]
        assert time.time() - start == pytest.approx(0.3, abs=0.05)
    finally:
        t.cancel()
```

coretl:
Ok, manages to reproduce it in the same manner as your test:
```python
async def test_observe_value_times_out_with_no_external_task():
    sig, setter = soft_signal_r_and_setter(float)

    recv = []

    async def watch():
        async for val in observe_value(sig):
            # If you put "await asyncio.sleep(0.00001)" here it times out
            recv.append(val)
            setter(val + 1)

    start = time.time()
    with pytest.raises(asyncio.TimeoutError):
        await asyncio.wait_for(watch(), timeout=0.2)
    assert recv
    assert time.time() - start == pytest.approx(0.3, abs=0.05)
```

However I question if this is your actual failure mode on I03, as this depends on the busy task prompting the callback in the same asyncio coro, while on I03 you have to yield already to an already busy queue. I'll see if I can reproduce this with an actual EPICS signal...

DominicOram:
> However I question if this is your actual failure mode on I03, as this depends on the busy task prompting the callback in the same asyncio coro, while on I03 you have to yield already to an already busy queue. I'll see if I can reproduce this with an actual EPICS signal...

It's worth noting that we only saw this issue once a couple of weeks ago and have not seen it again since so it's likely some timing race condition that's quite rare.

olliesilvester:
Closing in favour of https://github.com/bluesky/ophyd-async/pull/648#event-15298821917. Let's reopen this if we see the issue again (after the ophyd-async change is live on i03)

PR #910, opened 2024-11-20
Aithre lasershaping

DominicOram:
PR for laser shaping work, makes it easier for people to comment on the code

### Checks for reviewer


src/dodal/devices/aithre_lasershaping/cameras.py
```diff
@@ -0,0 +1,59 @@
...
+from dodal.log import LOGGER
+
+# put this into a config file
+laserOAVconfig = {"microns_per_pixel_x": 6,
+                  "microns_per_pixel_y": 6,
+                  "beam_centre_x": 1000,
+                  "beam_centre_y": 1028,
+                }
+
+class OAV(StandardReadable)
```
DominicOram, line 20: There's a missing colon here, you can tell at first glance in github because the `__init__` below it has no highlighting unlike other functions
```suggestion
class OAV(StandardReadable):
```

src/dodal/devices/aithre_lasershaping/cameras.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self, prefix: str, config: OAVConfig, name: str = ""):
+        self.oav_config = config
```
DominicOram, line 22: `OAVConfig` is mainly there to handle how beam centre/microns per pixel changes with zoom. If we don't have a zoom controller here then I'm guessing these are fixed, in which case we can just remove it.
```suggestion
    def __init__(self, prefix: str, name: str = ""):
```

src/dodal/devices/aithre_lasershaping/cameras.py
```diff
@@ -0,0 +1,59 @@
...
+        # don't need zoom controller
+
+        self.cam = Cam(f"{prefix}CAM:", name=name)
+
+        with self.add_children_as_readables():
+            self.grid_snapshot = SnapshotWithGrid(f"{prefix}MJPG:", name)
+            self.microns_per_pixel_x = laserOAVconfig["microns_per_pixel_x"]
+            self.microns_per_pixel_y = laserOAVconfig["microns_per_pixel_y"]
+            self.beam_centre_x = create_hardware_backed_soft_signal(int, lambda: self._get_beam_position("X"))
+            self.beam_centre_y = create_hardware_backed_soft_signal(int, lambda: self._get_beam_position("Y"))
```
DominicOram, line 35: In the other OAV we have these as signals as they may change. Signals let you easily monitor things. In the case of a fixed microns/pixel and beam centre they don't __have__ to be signals but I think it's nicer for consistency. Making fixed signals is simpler, you don't need the `create_hardware_backed_soft_signal` stuff and so can remove `_get_beam_position` below too:
```python
from ophyd_async.core import soft_signal_r_and_setter

self.microns_per_pixel_x, _ = soft_signal_r_and_setter(float, laserOAVconfig["microns_per_pixel_x"])
self.microns_per_pixel_y, _ = soft_signal_r_and_setter(float, laserOAVconfig["microns_per_pixel_x"])
self.beam_centre_x, _ = soft_signal_r_and_setter(float, laserOAVconfig["beam_centre_x"])
self.beam_centre_y, _ = soft_signal_r_and_setter(float, laserOAVconfig["beam_centre_y"])
```

src/dodal/devices/aithre_lasershaping/cameras.py
```diff
@@ -0,0 +1,59 @@
...
+        elif coord == "Y":
+            value = laserOAVconfig["beam_centre_y"]
+        else:
+            LOGGER.error("NO AXIS DEFINED FOR BEAM POSITION")
+            value = 1
+        return value
+
+
+# LA18L-DI-OAV-01: is prefix for lab 18 OAV laser shaping
+# not sure if I need to use this or just create a new json and use the standard
```
DominicOram, line 59: I think removing the zoom reduces the complexity so much that it makes sense to rewrite it like this. We could probably do some more code re-use where this and the other OAV have a shared parent class or something but we can leave that to another day once it's working.

DominicOram (commented):
So the different files used in configuring the OAV are confusing and horrible and a bit of a hold over from GDA, I have written https://github.com/DiamondLightSource/dodal/issues/911 to document them.

I think the long and short of it though is that you only need them if your beam centre/microns per pixel changes with zoom so in your case (I assume) you don't. With the changes I have suggested in the code I can do:
```python
from bluesky import plan_stubs as bps
from ophyd_async.core import DeviceCollector

RE = RunEngine()

with DeviceCollector():
    my_oav = OAV("BL03I-DI-OAV-01:", "my_oav")


def take_an_image():
    yield from bps.mv(my_oav.snapshot.directory, "/scratch/")
    yield from bps.mv(my_oav.snapshot.filename, "test_image")
    yield from bps.trigger(my_oav.snapshot, wait=True)


RE(take_an_image())
```
and I get the snapshot I expect (from i03's OAV).

PR #912, opened 2024-11-21
Create a common base for attenuator

### LINKED ISSUE #889: Create an `Attenuator` base class
The current `Attenuator` device has a couple of PVs and filter specifications that do not match all beamlines, but most of it is common to most and could be reused. It would be useful to have one common attenuator base class with common signals that can be re-used to implement beamline specific devices.

## Acceptance Criteria
* There is an `Attenuator` base class with common features across beamlines

noemifrisina:
Fixes #889 

Creates an AttenuatorBase with a minimum set of common PVs that beamline-specific devices can inherit from. 

Used in [MX_bluesky#627](https://github.com/DiamondLightSource/mx-bluesky/pull/627)

1. Check attenuator devices for i03 and i04 still connect
2. Check new device for i24 connects
3. Run tests

### Checks for reviewer


src/dodal/beamlines/i24.py
```diff
@@ -29,6 +30,21 @@
 set_utils_beamline(BL)
 
 
+def attenuator(
+    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
+) -> AttenuatorBase:
+    """Get a basic (without filter wheels) attenuator device for i24, instantiate it
```
DominicOram, line 36: Should: This is a bit misleading, there are filter wheels on the I24 attenuator:
image.

src/dodal/devices/attenuator_base.py
```diff
@@ -0,0 +1,15 @@
+from ophyd_async.core import StandardReadable
+from ophyd_async.epics.core import epics_signal_r, epics_signal_rw, epics_signal_x
+
+
+class AttenuatorBase(StandardReadable):
+    """An attenuator base class with a minimum set of common PVs to phase1 beamlines."""
```
DominicOram, line 6: Should: I think it's worth making clear here that the read transmission will be fractional i.e. 0-1 not a percentage

src/dodal/devices/attenuator_base.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self._desired_transmission = epics_signal_rw(float, prefix + "T2A:SETVAL1")
+        self._use_current_energy = epics_signal_x(prefix + "E2WL:USECURRENTENERGY.PROC")
```
DominicOram, line 10: Should: Either this attenuator is only used for reading and we remove these PVs or we want to add the setting too, in which case `AttenuatorBase` should have a `set`. 

If we're going to do setting as well I actually think you might be able to get away with exactly the same `set` between the two. In the i24 case:
* `_filters_in_position` contains only `BL24I-OP-ATTN-01:MP1:INPOS` and `BL24I-OP-ATTN-01:MP2:INPOS`
* `_calculated_filter_states` are soft signals that are all high
*  `_change` is a soft signal that's a no-op

If we don't do this in this issue can you put this in a new issue please?

DominicOram (changes requested):
Thanks, a couple of comments

src/dodal/beamlines/i24.py
noemifrisina: Yes, but for the moment there's no filter wheel device (Made a ticket to write one #927 , which will also be needed soon for i19) and I24 doesn't need it at this stage... Beyond making it clearer in the comment that it's only temporary and will be added in the future I'm not sure what could be added here?

src/dodal/devices/attenuator_base.py
noemifrisina: Made an issue here #928

DominicOram (changes requested):
I think it would be clearer about what's going on if we just reframe it all as read-only vs settable Attenuators. I've done this in https://github.com/DiamondLightSource/dodal/pull/931, let me know what you think. Sorry to jump on it and re-write it but it seemed easier than trying to articulate what I meant

rtuck99 (approved):
Approved

PR #917, opened 2024-11-26
Update example create-beamline docs for new preferred method

DiamondJoseph:
Update docs for new device_factory decorator

1. Read docs
2. Ensure that docs make sense

### Checks for reviewer

### DIFF
--- a/docs/how-to/create-beamline.rst
+++ b/docs/how-to/create-beamline.rst
@@ -27,12 +27,12 @@ The following example creates a fictitious beamline ``w41``, with a simulated tw
 
     from dodal.common.beamlines.beamline_utils import device_instantiation
     from dodal.common.beamlines.beamline_utils import set_beamline as set_utils_beamline
-    from dodal.devices.areadetector.adaravis import AdAravisDetector
     from dodal.devices.synchrotron import Synchrotron
     from dodal.log import set_beamline as set_log_beamline
     from dodal.utils import get_beamline_name, skip_device
 
     BL = get_beamline_name("s41")  # Default used when not on a live beamline
+    PREFIX = BeamlinePrefix(BL)
     set_log_beamline(BL)  # Configure logging and util functions
     set_utils_beamline(BL)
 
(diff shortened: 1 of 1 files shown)


docs/how-to/create-beamline.rst
```diff
@@ -68,12 +81,17 @@ The following example creates a fictitious beamline ``w41``, with a simulated tw
             bl_prefix=False,
         )
 
-    def d11(name: str = "D11") -> AdAravisDetector:
-        """
-        Also a valid Device factory function, but as multiple calls would instantiate
-        multiple copies of a device, discouraged.
-        """
-        return AdAravisDetector(name=name, prefix=f"{BL}-DI-DCAM-01:")
```
DominicOram, line 76: Should: Now that you've removed the detector the text above about having a synchrotron and a detector doesn't make sense. If we don't want to use the `AdAravisDetector` as an example we should have some example of a device instantiated w/o `skip_device`

docs/how-to/create-beamline.rst
```diff
@@ -68,12 +81,17 @@ The following example creates a fictitious beamline ``w41``, with a simulated tw
...
+    A valid factory function which:
+    - may be instantiated automatically
+        - cannot distinguish between simulated beamline and live
+        - each call instantiates a new copy
+    - must be explicitly connected (may be automated by tools)
+        - when connected may connect to a simulated backend
+        - may be connected concurrently (when automated by tools)
+    """
+    def synchrotron() -> Synchrotron:
+        return Synchrotron()
```
DominicOram, line 94: Should: Why do we want this in the docs? Presumably we would like to encourage people to always be using `@device_factory`?

docs/how-to/create-beamline.rst
```diff
@@ -43,13 +42,27 @@ The following example creates a fictitious beamline ``w41``, with a simulated tw
...
-    - can optionally be connected concurrently by not waiting for connect to complete
+    A valid factory function which:
+    - may be instantiated automatically, selectively on live beamline
+        - a maximum of once
+    - must attempt to connect when first called
+        - can optionally be faked with ophyd simulated axes
+        - can optionally be connected concurrently by not waiting for connect to complete
     - if constructor took a prefix, could optionally exclude the BLIXX prefix
     """"
     @skip_device(lambda: BL == "s41")  # Conditionally do not instantiate this device
```
DominicOram, line 68: Should: This is the old way of doing things. I think it's fine that the behavior is still there whilst we're depreciating it but we shouldn't be documenting it in a how to otherwise people might use it for new stuff

docs/how-to/create-beamline.rst
```diff
@@ -43,13 +42,27 @@ The following example creates a fictitious beamline ``w41``, with a simulated tw
     to one or more Bluesky Protocols.
     """
 
+    """
+    A valid factory function which:
+    - may be instantiated automatically, selectively on live beamline
+        - a maximum of once
```
DominicOram, line 48: Should: At a first glance this implies you should only call `synchrotron()` once, not that it's cached

DominicOram (changes requested):
Thanks, couple of comments.

docs/how-to/create-beamline.rst
DiamondJoseph: I've restored the Detector. I had originally thought of showing the pros and cons of all of the methods currently available, and using the Synchrotron for all of them to make it clear what was going on, but I think you're right and we don't need to highlight the outdated ways of doing things.

docs/how-to/create-beamline.rst
DiamondJoseph: How's the new wording, other than wordier?

DominicOram (approved):
Great, thank you!

PR #924, opened 2024-11-26
Update training-rigs to use device_factory

DiamondJoseph:
1. test beamline still connects using dodal connect from beamline workstation

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/training_rig.py
+++ b/src/dodal/beamlines/training_rig.py
@@ -3,15 +3,16 @@ from pathlib import Path
 from ophyd_async.epics.adaravis import AravisDetector
 
 from dodal.common.beamlines.beamline_utils import (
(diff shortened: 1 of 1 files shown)


src/dodal/beamlines/training_rig.py
```diff
@@ -25,6 +26,7 @@
 #
 
 BL = get_beamline_name("p47")
```
ZohebShaikh, line 28: nit: can this be changed to `p46`

ZohebShaikh (approved):

src/dodal/beamlines/training_rig.py
DiamondJoseph: v. good point, if p46 is going to be our default

PR #932, opened 2024-11-27
Move aperture scatterguard independently

DominicOram:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/267

1. Confirm tests pass
2. Check the added API for aperture/scatterguard makes sense if given the docstring

### Checks for reviewer


src/dodal/devices/aperturescatterguard.py
```diff
@@ -107,47 +111,210 @@ def load_positions_from_beamline_parameters(
...
 
+async def _safe_move_whilst_in_beam(
+    aperture: Aperture,
+    scatterguard: Scatterguard,
+    position: AperturePosition,
+    aperture_z_tolerance: float,
+):
+    """
+    Move the aperture and scatterguard combo safely to a new position.
+    See https://github.com/DiamondLightSource/hyperion/wiki/Aperture-Scatterguard-Collisions
```
rtuck99, line 122: We will need to update this link when we do DiamondLightSource/mx-bluesky#578

tests/devices/unit_tests/test_aperture_scatterguard.py
```diff
@@ -73,11 +75,11 @@ def aperture_tolerances():
 
 def get_all_motors(ap_sg: ApertureScatterguard):
     return [
-        ap_sg.aperture.x,
-        ap_sg.aperture.y,
-        ap_sg.aperture.z,
-        ap_sg.scatterguard.x,
-        ap_sg.scatterguard.y,
```
rtuck99, line 80: `_assert_position_in_reading()` also needs corresponding changes

src/dodal/devices/aperturescatterguard.py
```diff
@@ -107,47 +111,210 @@ def load_positions_from_beamline_parameters(
...
+    @AsyncStatus.wrap
+    async def trigger(self):
+        """Moves the assembly out of the beam."""
+        await self.aperture_y().set(self.out_y)
+
+
 class ApertureScatterguard(StandardReadable, Movable):
+    """Move the aperture and scatterguard assembly in a safe way. There are two ways to
+    interact with the device depending on if you want simplicity or move flexibility.
+
```
rtuck99, line 240: These could be in an Examples: section and napoleon will format them nicely if they are in a double-colon code block.

src/dodal/devices/aperturescatterguard.py
```diff
@@ -107,47 +111,210 @@ def load_positions_from_beamline_parameters(
...
+    However, the aperture Y axis is faster than the others. In some cases we may want to
+    move the assembly out of the beam with this axis without moving others:
+
+        await aperture_scatterguard.move_out.trigger()
+
+    We may then want to keep the assembly out of the beam whilst asynchronously preparing
+    the other axes for the aperture that's to follow:
+
+        await aperture_scatterguard.aperture_outside_beam.set(ApertureValue.LARGE)
+
```
rtuck99, line 257: I think this is confusingly named. "outside beam" suggests that the movement is always outside the beam (or even that moving it will put the aperture out), but we are in beam, then it will still move the aperture, but not move it out (unless you explicitly set it to ROBOT_LOAD).

I feel like something akin to `prepare_aperture` would be better. But supporting the in-beam movement as well using this method still makes me uncomfortable because the whole point of this is to be able to call it earlier when it's not clear to move it back in yet, so why would you want this method to be able to do that anyway?

src/dodal/devices/aperturescatterguard.py
```diff
@@ -107,47 +111,210 @@ def load_positions_from_beamline_parameters(
...
+            await _safe_move_whilst_in_beam(
+                self.aperture(),
+                self.scatterguard(),
+                self.loaded_positions[value],
+                self.aperture_z_tolerance,
+            )
+
+
+class OutTrigger(StandardReadable, Triggerable):
+    """Allows for moving just the Y stage of the assembly out of the beam."""
```
rtuck99, line 220: Do the `ApertureSelector` and `OutTrigger` need to be `StandardReadable`? I don't think they have anything to read and they could just be `Device`

src/dodal/devices/aperturescatterguard.py
DominicOram: > I think this is confusingly named. "outside beam" suggests that the movement is always outside the beam (or even that moving it will put the aperture out), but we are in beam, then it will still move the aperture, but not move it out (unless you explicitly set it to ROBOT_LOAD).

Happy to rename it if we can think of a better name.

>  I think an important feature of this function is that until you do `move_out.trigger()` the aperture is in an undefined position which you can't make assumptions about.

I'm unsure what you mean by this. There are 6 positions, which are implicitly defined:
* In + Large (maps to LARGE_APERTURE in the file)
* In + Medium (maps to MEDIUM_APERTURE in the file)
* In + Small (maps to SMALL_APERTURE in the file)
* Out + Large (maps to ROBOT_LOAD in the file)
* Out + Medium
* Out + Small

`move_out.trigger()` actually moves you to one of the `Out` positions, which are less explicitly defined.

> The more I think about this, the more I feel that perhaps OutTrigger and ApertureSelector (or perhaps even the whole ApertureScatterguard) should be one device which implements Preparable and Triggerable, and that you should just call prepare() and trigger() on it.

So `aperture_scatterguard.trigger()` would move it out or move it in? I'm not sure I'm understanding. I like the idea of using prepare though, we could maybe replace `aperture_outside_beam` with just `prepare`

In general I'm very open to changing the structure but I have gone round this loop 3 or 4 times now and it's not easy. I'll put in some time to whiteboard it.

In general my thought I do like keeping the interface of `await aperture_scatterguard.set(ApertureValue.LARGE)` moving the aperture in though, this seems intuitively what you would expect

DominicOram: As discussed in person. The proposed solution is to have:
* `aperture_scatterguard.set(value: ApertureValueWithOut)` where `ApertureValueWithOut` can be S/M/L/Out. For S/M/L the selected aperture gets moved into the beam. For Out we move the assembly out of the beam without moving the scatterguard
* `aperture_scatterguard.prepare(value: ApertureValue)` where `ApertureValue` can be S/M/L. This will move the scatterguard to the selected aperture without moving in

There are some gotchas with this:
* What do we do when prepare is called and we're already in - in this case we just still move to the location and stay in i.e. we just call `set`
* There is a potential problem with calling `prepare` then `set` in quick succession. In this case we should make the `set` wait for the `prepare` to finish

src/dodal/devices/aperturescatterguard.py
DominicOram: Removed them so no longer relevant :tada:

tests/devices/unit_tests/test_aperture_scatterguard.py
DominicOram: I've gone back to no underscore, it was getting messy and out of scope

rtuck99 (approved):
Approved

PR #936, opened 2024-11-29
Wait for eiger to disarm before stopping it

### LINKED ISSUE #935: Stopping the Eiger whilst it's disarming causes issues
When a bluesky plan errors the RE will call `stop` on all the associated ophyd devices. In the Eiger stop/unstage are a very similar operation and touch basically the same PVs. We have an issue where we:
1. Start unstaging the Eiger
2. Have a failure in the plan (the failure in this case is no diffraction found so can be quite common)
3. This tries and stops the Eiger but the disarm is in progress causing a new exception

## Acceptance Criteria
* When stop is called on the eiger, if a disarm is already in progress it just waits for the disarm

DominicOram:
Fixes #935

1. Confirm new test passes

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -62,6 +62,9 @@ class EigerDetector(Device):
     arming_status = Status()
     arming_status.set_finished()
 
+    disarming_status = Status()
+    disarming_status.set_finished()
+
     def __init__(self, beamline: str = "i03", *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.beamline = beamline
@@ -140,6 +143,7 @@ class EigerDetector(Device):
     def unstage(self) -> bool:
         assert self.detector_params is not None
         try:
+            self.disarming_status = Status()
             self.wait_on_arming_if_started()
             if self.detector_params.trigger_mode == TriggerMode.FREE_RUN:
                 # In free run mode we have to manually stop odin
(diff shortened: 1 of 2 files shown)


callumforrester (approved):

src/dodal/devices/eiger.py
```diff
@@ -158,23 +162,28 @@ def unstage(self) -> bool:
...
-        stop_status &= self.disable_roi_mode()
-        stop_status.wait(self.timeouts.general_status_timeout)
-        # See https://github.com/DiamondLightSource/hyperion/issues/1395
-        LOGGER.info("Turning off Eiger dev/shm streaming")
-        self.odin.fan.dev_shm_enable.set(0).wait()
-        LOGGER.info("Eiger has successfully been stopped")
+        if not self.disarming_status.done:
+            LOGGER.info("Eiger still disarming, waiting on disarm")
+            self.disarming_status.wait(self.timeouts.arming_timeout)
+        else:
```
rtuck99, line 174: Perhaps this fixes the issue, but I'm unclear on what exactly we should do in `stop()`, looking at bluesky `run_engine.py` it seems `stop()` is invoked in a variety of scenarios, including at the end of a plan, during a suspend and when `pause()` is invoked. The impression I get is that being stopped is at least somewhat resumable.

DominicOram: Sure but I think it's outside of the scope of this issue, I've made https://github.com/DiamondLightSource/dodal/issues/994.

PR #941, opened 2024-12-03
Convert transfocator to ophyd_async

### LINKED ISSUE #748: Transfocator: Convert to ophyd-async
We should convert the i04 transfocator to ophyd-async. As discussed at https://github.com/DiamondLightSource/dodal/issues/399 the transfocator and f-switch are basically the same device. However, the i22 one does not have some of the features that the i04 one has. In doing this conversion we should try and make sure the two share as much code as reasonable. This may involve a bit of discussion with controls/scientists to get the features on i04's added to i22's IOC as well.

## Acceptance Criteria
* I04's transfocator no longer uses `ophyd` but instead uses `ophyd-async`

shihab-dls:
Fixes #748 

1. Check if signal components converted to analogous ophyd_async types.
2. Check if setting device results in expected behaviour.

### Checks for reviewer


olliesilvester:
CI initially failed due to a flakey unit test unrelated to this PR, it passed after re-running. See https://github.com/DiamondLightSource/dodal/issues/943

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,80 @@ class Transfocator(Device):
...
+                self.latest_pred_vertical_num_lenses = value
+                have_we_done_it = True
+
+        # The value hasn't changed so assume the device is already set up correctly
+        async for value in observe_value(self.predicted_vertical_num_lenses):
+            await set_based_on_predicition(value)
+            if have_we_done_it:
+                break
+
+    async def polling_wait_on_start_rbv(self, for_value):
```
olliesilvester, line 68: This entire function looks like it could just be replaced with `ophyd_async.core.wait_for_value`. From the comment, it looks like there were some issues with monitoring this PV in ophyd, but I wonder if it works in `ophyd-async?` 

If we do try `wait_for_value` here instead, we should go to i04 and test it before merging

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,80 @@ class Transfocator(Device):
...
+        self.latest_pred_vertical_num_lenses = (
+            await self.predicted_vertical_num_lenses.get_value()
+        )
 
         LOGGER.info(f"Transfocator setting {beamsize_microns} beamsize")
-        if self.beamsize_set_microns.get() != beamsize_microns:
-            subscriber = self.predicted_vertical_num_lenses.subscribe(
-                set_based_on_predicition, run=False
+
+        if await self.beamsize_set_microns.get_value() != beamsize_microns:
```
olliesilvester, line 97: Sorry, I think I changed my mind about this since we spoke about it

The original function lets you do a `set` which will run in the background and return a `status`, so we can do other stuff while we wait for this status to finish. Using `await asyncio.gather` here (which I know I originally suggested) will block until it's finished. We should use `asyncio.create_task` instead.

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,80 @@ class Transfocator(Device):
...
+            await asyncio.gather(
+                self.beamsize_set_microns.set(beamsize_microns),
```
olliesilvester, line 99: Should: Add a comment here to explain that changing that at the IOC level, `beamsize_set_microns` will cause `predicted_vertical_num_lenses` to change

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,80 @@ class Transfocator(Device):
...
+                await self.start.set(1)
+                await self.polling_wait_on_start_rbv(1)
+                await self.polling_wait_on_start_rbv(0)
+                self.latest_pred_vertical_num_lenses = value
+                have_we_done_it = True
+
+        # The value hasn't changed so assume the device is already set up correctly
+        async for value in observe_value(self.predicted_vertical_num_lenses):
+            await set_based_on_predicition(value)
+            if have_we_done_it:
```
olliesilvester, line 65: `have_we_done_it` isn't the most descriptive name :laughing:

tests/devices/i04/test_transfocator.py
```diff
@@ -1,66 +1,75 @@
...
     def lens_number_is_half_beamsize(value, *args, **kwargs):
-        transfocator.predicted_vertical_num_lenses.sim_put(int(value / 2))  # type: ignore
+        set_mock_value(transfocator.predicted_vertical_num_lenses, int(value / 2))
 
-    transfocator.beamsize_set_microns.subscribe(lens_number_is_half_beamsize)
+    transfocator.beamsize_set_microns.subscribe_value(lens_number_is_half_beamsize)
 
 
-def test_given_beamsize_already_set_then_when_transfocator_set_then_returns_immediately(
+async def set_beamsize_to_same_value_as_mock_signal(
```
olliesilvester, line 28: I think the functions here are in the wrong order. In the test below, we want to check that doing the set returns immediately AFTER `beamsize_set_microns` has been set - not before.

tests/devices/i04/test_transfocator.py
```diff
@@ -1,66 +1,75 @@
...
+
+
+async def test_given_beamsize_already_set_then_when_transfocator_set_then_returns_immediately(
     fake_transfocator: Transfocator,
 ):
-    fake_transfocator.beamsize_set_microns.sim_put(100)  # type: ignore
-    set_status = fake_transfocator.set(100)
-    set_status.wait(0.01)
-    assert set_status.done and set_status.success
+    await asyncio.wait_for(
```
olliesilvester, line 38: Should: Assert that none of the other logic, (eg `set_based_on_predicition`) was ever ran. I think that would have caught the above function having sets in the wrong order

tests/devices/i04/test_transfocator.py
```diff
@@ -1,66 +1,75 @@
...
-    transfocator_statuses &= SubscriptionStatus(
-        fake_transfocator.start,
-        lambda *, old_value, value, **kwargs: value == 1,
-    )
-
+    fake_transfocator.start_rbv.get_value = AsyncMock(side_effect=[1, 0])
     set_status = fake_transfocator.set(315)
 
-    transfocator_statuses.wait(0.1)
     assert not set_status.done
```
olliesilvester, line 52: This assertion isn't useful anymore, since we haven't awaited on `set_status`, so there's no chance of it being done at this stage. 

In one of my earlier comments I suggested using `asyncio.create_task` in the `set` method. If we do this, then we should be able to do the same kind of assertion here

tests/devices/i04/test_transfocator.py
```diff
@@ -1,66 +1,75 @@
...
-    set_status.wait(0.1)
+    await set_status
     assert set_status.done and set_status.success
 
-    try:
-        transfocator_statuses.wait(0.1)
-    except BaseException:
-        pass
+
+async def test_if_timeout_exceeded_and_start_rbv_not_equal_to_set_value_then_timeout_exception(
```
olliesilvester, line 64: Should: use [`with pytest.raises`](https://www.google.com/search?client=firefox-b-e&q=with+pytest.raises) to assert that a specific assertion was rasied

olliesilvester (changes requested):
Thanks, this looks great so far!

I've left a few comments, but the main one is something I incorrectly suggested to you, which was to not bother running `set` in the background by using `asyncio.create_task`. The original code ran `set` asynchronously, so we should try to keep that.

By the way, if you're curious, this was the recently added policy for reviews in dodal: https://diamondlightsource.github.io/dodal/main/explanations/reviews.html

.vscode/launch.json
```diff
@@ -8,7 +8,7 @@
             "name": "Debug Unit Test",
             "type": "debugpy",
             "request": "launch",
-            "justMyCode": false,
```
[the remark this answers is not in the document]
shihab-dls: Hi yes, no problem; will be amended

src/dodal/devices/i04/transfocator.py
olliesilvester: As discussed it's fine to use `asyncio.gather` the the entire function is already running in the background via `AsyncStatus`

tests/devices/i04/test_transfocator.py
olliesilvester: As discussed, this assert statement would be useful if there were any awaits inbetween this and the `set` being called. Eg an `asyncio.sleep`

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,80 @@ class Transfocator(Device):
...
+    def __init__(self, prefix: str, name: str = ""):
+        with self.add_children_as_readables():
+            self.beamsize_set_microns = epics_signal_rw(float, prefix + "VERT_REQ")
+            self.predicted_vertical_num_lenses = epics_signal_rw(
+                float, prefix + "LENS_PRED"
+            )
+            self.number_filters_sp = epics_signal_rw(int, prefix + "NUM_FILTERS")
+
+            self.start = epics_signal_rw(int, prefix + "START.PROC")
+            self.start_rbv = epics_signal_r(int, prefix + "START_RBV")
```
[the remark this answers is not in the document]
olliesilvester: As discussed this doesn't actually work since the prefix doesn't match

src/dodal/devices/i04/transfocator.py
shihab-dls: Added a comment to describe what happens at the IOC level 👍

olliesilvester (changes requested):
Looks mostly good now, just a couple more things. When you're happy with this, we can create a short bluesky plan to test this device on the beamline

tests/devices/i04/test_transfocator.py
```diff
@@ -1,66 +1,74 @@
...
+    fake_transfocator: Transfocator,
+) -> None:
+    with patch.object(fake_transfocator, "TIMEOUT", 0):
+        given_predicted_lenses_is_half_of_beamsize(fake_transfocator)
+        fake_transfocator.start.get_value = AsyncMock(side_effect=[0, 1])
+        with pytest.raises(TimeoutError):
```
olliesilvester, line 73: This test fails in python 3.10 as, in this version, the standard TimeoutError is different to asyncio.TimeoutError. Doing `with pytest.raises((TimeoutError, asyncio.exceptions.TimeoutError)):` should make it work for all versions

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,70 @@ class Transfocator(Device):
...
+        with self.add_children_as_readables():
+            self.beamsize_set_microns = epics_signal_rw(float, prefix + "VERT_REQ")
+            self.predicted_vertical_num_lenses = epics_signal_rw(
+                float, prefix + "LENS_PRED"
+            )
+            self.number_filters_sp = epics_signal_rw(int, prefix + "NUM_FILTERS")
 
-    start = Cpt(EpicsSignal, "START.PROC")
-    start_rbv = Cpt(EpicsSignalRO, "START_RBV")
+            self.start = epics_signal_rw(
```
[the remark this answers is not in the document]
shihab-dls: Sounds good, reverted to separate PV's now.

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,70 @@ class Transfocator(Device):
...
+            await self.predicted_vertical_num_lenses.get_value()
+        )
 
         LOGGER.info(f"Transfocator setting {beamsize_microns} beamsize")
-        if self.beamsize_set_microns.get() != beamsize_microns:
-            subscriber = self.predicted_vertical_num_lenses.subscribe(
-                set_based_on_predicition, run=False
+
+        if await self.beamsize_set_microns.get_value() != beamsize_microns:
+            """Setting beam_set_microns results in a change to predicted_vertical_num_lenses.
```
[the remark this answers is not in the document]
shihab-dls: Yeah that is way nicer; amended.

olliesilvester (changes requested):
This seemed to work well during testing. A few minor features were requested by David on I04:

- Add the signals for the read-back values for the number of lenses and the vertical beam size, `BL04I-MO-FSWT-01:NUM_FILTERS` and `BL04I-MO-FSWT-01:VER`
- After the set is done, log the readbacks for the above

src/dodal/devices/i04/transfocator.py
shihab-dls: Looks like these changes worked fine on i04; resolving for now.

src/dodal/devices/i04/transfocator.py
```diff
@@ -85,3 +85,8 @@ async def set(self, value: float):
                 self.beamsize_set_microns.set(value),
                 self._observe_beamsize_microns(),
             )
+        number_filters_rbv = await self.number_filters_sp.get_value()
+        vertical_lens_size_rbv = await self.vertical_lens_rbv.get_value()
```
olliesilvester, line 89: Better to put these in an `asyncio.gather`

src/dodal/devices/i04/transfocator.py
```diff
@@ -85,3 +85,8 @@ async def set(self, value: float):
...
+        LOGGER.info(
+            f"Transfocator set complete. \nNumber of filters is: {number_filters_rbv} \nVertical beam size is: {vertical_lens_size_rbv}"
```
olliesilvester, line 91: Personal preference but I'd put all this on a single line, or two lines at most

olliesilvester (approved):
Thanks! Looks neater than the old device :partying_face:

src/dodal/devices/i04/transfocator.py
shihab-dls: Oh yeah, definitely

src/dodal/devices/i04/transfocator.py
shihab-dls: One line sounds good!

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,70 @@ class Transfocator(Device):
...
         LOGGER.info(f"Transfocator setting {beamsize_microns} beamsize")
-        if self.beamsize_set_microns.get() != beamsize_microns:
-            subscriber = self.predicted_vertical_num_lenses.subscribe(
-                set_based_on_predicition, run=False
+
+        if await self.beamsize_set_microns.get_value() != beamsize_microns:
+            """Setting beam_set_microns results in a change to predicted_vertical_num_lenses.
+            In the following logic, we observe this change, and set it in number_filters_sp.
+            """
+            await asyncio.gather(
```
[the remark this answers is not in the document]
DominicOram: I agree, did you chase this?

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,75 @@ class Transfocator(Device):
...
+                await self.start.set(1)
+                LOGGER.info("Waiting for start_rbv to change to 1")
+                await wait_for_value(self.start_rbv, 1, self.TIMEOUT)
+                LOGGER.info("Waiting for start_rbv to change to 0")
+                await wait_for_value(self.start_rbv, 0, self.TIMEOUT)
+                self.latest_pred_vertical_num_lenses = value
+                is_set_filters_done = True
+
+        # The value hasn't changed so assume the device is already set up correctly
+        async for value in observe_value(self.predicted_vertical_num_lenses):
```
DominicOram, line 62: Must: I think there's a potential infinite loop here. See https://github.com/DiamondLightSource/dodal/issues/995

src/dodal/devices/i04/transfocator.py
```diff
@@ -18,65 +22,75 @@ class Transfocator(Device):
...
+        is_set_filters_done = False
 
-    def set(self, beamsize_microns: float) -> DeviceStatus:
+        async def set_based_on_prediction(value: float):
+            if not math.isclose(
+                self.latest_pred_vertical_num_lenses, value, abs_tol=1e-8
+            ):
+                # We can only put an integer number of lenses in the beam but the
+                # calculation in the IOC returns the theoretical float number of lenses
+                nonlocal is_set_filters_done
```
DominicOram, line 49: Could: I always think `nonlocal` is a bit of a code smell. For example, in this case we could just return the value but I'm also not 100% sure of the purpose of `is_set_filters_done`

Issue #943, opened 2024-12-05
 Fix flakey bart robot unit test


olliesilvester:
One of our tests for the bart robot occasionally fails. We should fix this

`tests/devices/unit_tests/test_bart_robot.py::test_given_program_not_running_and_pin_unmounts_then_mounts_when_load_pin_then_pin_loaded - 
AssertionError: Expected 'trigger' to have been called once. Called 0 times.`

Or see https://github.com/DiamondLightSource/dodal/actions/runs/12165707719/job/33930319285

## Acceptance Criteria
* Test made more robust

paulorochaoliveira:
FYI, I'm taking a look at this flaky test right now.

paulorochaoliveira:
It seems like what makes this test flaky is the sleep time sometimes being to short. Increasing the sleep time would most likely fix the issue.
Another more robust solution is to replace arbitrary sleeps with an event-driven approach. For example, use asyncio.Event to synchronize the test with the actual behavior.
What do you think @olliesilvester ?

olliesilvester:
I think increasing the sleep is good enough for this issue, however I like the idea of using `asyncio.Event` instead. I will make an issue to replace all of our arbitrary sleeps in tests instead of this! Thanks!

paulorochaoliveira:
Should I open a PR to at least close this issue?

olliesilvester:
Yes, go ahead, thank you

DominicOram:
I think we need to properly mock it and rethink how the test works rather than just change the sleeps. e.g. we could patch sleep like in https://github.com/DiamondLightSource/dodal/blob/f93183fb50d3012c2c4074508e09792c43660fbe/tests/devices/unit_tests/test_thawer.py#L29 so that we can be very specific on when things should be happening

PR #947, opened 2024-12-05
fix flaky test in test_bart_robot.py

arikaran-13:
#943 Fix flakey bart robot unit test

1. Run the test tests/devices/unit_tests/test_bart_robot.py::test_given_program_not_running_and_pin_unmounts_then_mounts_when_load_pin_then_pin_loaded
2. Confirm that the test passes consistently, even if run multiple times.

### Checks for reviewer

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,5 +1,4 @@
 import asyncio
-import importlib
 import logging
 import os
 import sys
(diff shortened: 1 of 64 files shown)


olliesilvester (changes requested):
I don't think these extra statements guarantee that the task will have finished in time - if we change the sleep times then the tests still fail.

A good short term the solution would just be to increase `await sleep(0.01)` slightly, and to increase `device.LOAD_TIMEOUT = 0.03` to account for this.

arikaran-13:
> I don't think these extra statements guarantee that the task will have finished in time - if we change the sleep times then the tests still fail.
> 
> A good short term the solution would just be to increase `await sleep(0.01)` slightly, and to increase `device.LOAD_TIMEOUT = 0.03` to account for this.

I thought that since we are changing the state from PIN_MOUNTED to NO_PIN_MOUNTED and then back to PIN_MOUNTED, if we await the state change, the trigger() method will be called correctly. But if that doesn't fix the test, as a short-term solution, I have updated the timing of the LOAD_TIMEOUT

olliesilvester (approved):
Thanks!

arikaran-13:
> Thanks!

Thanks you
This is my first open source contribution :)

PR #956, opened 2024-12-11
Prevent deserialisation error when fetching data from GDA embedded visit service

DiamondJoseph:
During beamline testing on i22 December 10th, it was found that the current visit service impl uses the wrong method to convert the return value of the GDA numTracker

1. Run a scan that uses GDA as the numTracker service
2. Confirm that the data_session can be deserialised

### Checks for reviewer

### DIFF
--- a/src/dodal/common/visit.py
+++ b/src/dodal/common/visit.py
@@ -78,7 +78,7 @@ class RemoteDirectoryServiceClient(DirectoryServiceClient):
         ):
             response.raise_for_status()
             json = await response.json()
-            return DataCollectionIdentifier.model_validate_json(json)
+            return DataCollectionIdentifier.model_validate(json)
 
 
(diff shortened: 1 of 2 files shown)


src/dodal/common/visit.py
```diff
@@ -78,7 +78,7 @@ async def _identifier_from_response(
         ):
             response.raise_for_status()
             json = await response.json()
-            return DataCollectionIdentifier.model_validate_json(json)
+            return DataCollectionIdentifier.model_validate(json)
```
DiamondJoseph, line 81: `model_validate_json` is for turning a JSON string into the object. `model_validate` is for turning a Python object. `response.json()` seems to convert the JSON string returned by the numTracker service into a dict.

rtuck99:
Please can you fix the tests

olliesilvester (approved):

rtuck99 (approved):
Approved

PR #957, opened 2024-12-12
Use Reference to prevent Device name mangling of undulator and dcm

### LINKED ISSUE #929: Use references in undulator_dcm
The `undulator_dcm` has a reference to both the `undulator` and `dcm` but has a note about doing them after the `super.__init__` to avoid name mangling. This is not the way to do this any more, instead we should use `Reference`

## Acceptance Criteria
* Device uses `Reference`
* Tests still pass

shihab-dls:
Fixes #929

1. Check for correct usage of Reference

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/undulator_dcm.py
+++ b/src/dodal/devices/undulator_dcm.py
@@ -1,7 +1,7 @@
 import asyncio
 
 from bluesky.protocols import Movable
-from ophyd_async.core import AsyncStatus, StandardReadable
+from ophyd_async.core import AsyncStatus, Reference, StandardReadable
 
 from dodal.common.beamlines.beamline_parameters import get_beamline_parameters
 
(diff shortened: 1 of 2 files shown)


rtuck99:
Thanks for this Shihab!
The code changes here look fine, however before merging this, you will need to make changes and raise another PR in `mx-bluesky`, because the changes here are exposed to calling code in mx-bluesky, and this PR as it stands can't currently be merged without breaking it.

If you run the unit tests and search around in mx-bluesky you will find that there are several places where there are similar dereferences of `undulator_dcm.dcm`/`undulator_dcm.undulator`.

You might also want to alter the name of these to make it clear that they are reference wrappers rather than the actual devices - elsewhere where I have added them I have suffixed the names with `_ref` to make this clear.

A second consideration is whether to make them private by prefixing an underscore, however I think in this case it's ok to leave them public as it's probably both expected and useful to expose the undulator and DCM devices that are backing this composite.

rtuck99 (changes requested):
See previous comment

olliesilvester (approved):
Looks good!

PR #960, opened 2024-12-13
Make ophyd devices for the diagonstics for i10

### LINKED ISSUE #949: Make Ophyd devices for the Diagonstics, it consist of screen camera and diode.

Relm-Arrowny:
Fixes #949 

1. Pass test.
2. Check connection.

### Checks for reviewer


src/dodal/devices/i10/diagnostics.py
```diff
@@ -0,0 +1,307 @@
...
+                stage_read_enum,
+                read_pv=prefix + stage_read_suffix,
+            )
+        super().__init__(name=name)
+
+
+class I10AravisDriverIO(AravisDriverIO):
+    """This is the standard webcam that can be found in ophyd_async
+    with four extra centroid signal. There is also a change in data_type due to it being
+    an older/different model"""
```
callumforrester, line 125: I would like @coretl's opinion on this approach, whether we should make a configurable "super detector" in ophyd-async or accept variations like these

coretl: This should be rolled into ophyd-async AravisDetector. Since https://github.com/bluesky/ophyd-async/pull/606 this should now be possible. We pass `stats` as a plugin to the init. I'm unsure why the datatype change was needed.

src/dodal/devices/i10/diagnostics.py
```diff
@@ -0,0 +1,307 @@
...
+    async def stage(self) -> None:
+        await asyncio.gather(
+            self.drv.image_mode.set(ImageMode.SINGLE),
+            self.drv.wait_for_plugins.set(True),
+        )
+        await super().stage()
+
+    @AsyncStatus.wrap
+    async def trigger(self) -> None:
+        await self.drv.acquire.set(True)
```
DominicOram, line 181: Should: I think you could just inherit from `SingleTriggerDetector` then? Something like:
```suggestion
class I10CentroidDetector(SingleTriggerDetector):
    """Detector to read out the centroid position,
    this is based off the SingleTriggerDetector in ophyd_async with added
    readable defaults"""

    def __init__(
        self,
        prefix: str,
        name: str = "",
    ) -> None:
        driver = I10AravisDriverIO(prefix=prefix)
        super().__init__(driver, [driver.centroid_x, driver.centroid_y], name)
```

src/dodal/devices/i10/diagnostics.py
```diff
@@ -0,0 +1,307 @@
...
+    ) -> None:
+        with self.add_children_as_readables(Format.HINTED_SIGNAL):
+            self.stage_drop_down_set = epics_signal_rw(
+                stage_write_enum,
+                read_pv=prefix + stage_write_suffix,
+            )
+            self.stage_drop_down_readback = epics_signal_r(
+                stage_read_enum,
+                read_pv=prefix + stage_read_suffix,
+            )
```
DominicOram, line 118: Should: As above with `drop_down` -> `position`

src/dodal/devices/i10/diagnostics.py
```diff
@@ -0,0 +1,307 @@
...
+class ScreenCam(Device):
+    """Compound device of pneumatic stage(fluorescent screen) and webcam"""
+
+    def __init__(
+        self,
+        prefix: str,
+        stage_write_enum: type[StrictEnum] = InOutTable,
+        stage_read_enum: type[StrictEnum] = InOutReadBackTable,
+        stage_read_suffix: str = "STA",
+        stage_write_suffix: str = "CON",
```
DominicOram, line 193: Should: I can't see anywhere in the code currently where we use different values for `stage_write_enum`, `stage_read_enum`, `stage_read_suffix` or `stage_write_suffix` than the defaults. I would just hardcode them in below until you have a case where they are different as it leads to lots of boilerplate

tests/devices/i10/test_diagnostics.py
```diff
@@ -0,0 +1,66 @@
...
+async def mock_centroid_detector(
+    prefix: str = "BLXX-EA-DET-007:",
+) -> I10CentroidDetector:
+    async with DeviceCollector(mock=True):
+        mock_centroid_detector = I10CentroidDetector(prefix)
+    assert mock_centroid_detector.name == "mock_centroid_detector"
+    return mock_centroid_detector
+
+
+async def test_I10CentroidDetector_with_blueSky(
```
DominicOram, line 29: Nit: My preference is to name tests very verbosely and explicitly so it's clear what you're testing. For example here I would name it something like `test_when_I10CentroidDetector_is_counted_in_plan_then_emitted_document_contains_expected_values`. If you have vague names I think it leads to the temptation to make one big test that tests loads of different stuff, that test then becomes long to run and hard to debug

DominicOram (changes requested):
Great, thanks! I've added a bunch of suggestions based on how I would do a similar device in MX but obviously I don't know your usecase so feel free to reject suggestions if they don't work for you. Do you have some example plans that use the devices? I was expecting to see some in https://github.com/DiamondLightSource/i10-bluesky but couldn't, maybe I'm just missing them.

src/dodal/devices/i10/slits.py
```diff
@@ -20,7 +26,38 @@ def __init__(self, prefix: str, name: str = "") -> None:
         )
 
 
+class BladeDrainCurrents(Device):
```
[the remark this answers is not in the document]
Relm-Arrowny: Just a very quick one, I will add the dostring later, The way we use drain current on mirror the absolute flux does not matter we use it to normalise the detector so it is always relative and it has the added advantage of you do not put anything in the beam path that is not needed.  As for the slits/blade drain, we use it to correct correct the beam position(feedback). E.g. depending of the drain current values on the left and right blades we will move the mirror pitch to make sure it always in the right place.     
I am not sure I would call it a technique, it is mostly  [photoelectric effect](https://www.nobelprize.org/prizes/physics/1921/summary/). In theory you can convert them into flux but there are a lot assumptions/corrections you have to make, for example the beam distribution, as we are not intercepting all of the beam, furthermore drain current also suffer form charging up and has a strong energy dependence, it will be a very big job to do all the corrections.

src/dodal/devices/i10/diagnostics.py
```diff
@@ -0,0 +1,307 @@
...
+        self.d7 = DropDownStage(
+            prefix=prefix + "PHDGN-07:",
+            positioner_enum=D7DropDown,
+            positioner_suffix="Y",
+        )
+        super().__init__(name)
+
+
+class I10Diagnotic5ADet(Device):
+    """Diagnotic 5a detection with drain current and photo diode"""
```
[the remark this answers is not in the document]
Relm-Arrowny: I do but every time I rebuild container, I lost it and forgot to reinstall. :(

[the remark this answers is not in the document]
coretl: You can put them in `.config/Code/User/settings.json` to make them always installed in a dev container:
https://code.visualstudio.com/docs/devcontainers/containers#_always-installed-extensions

src/dodal/devices/i10/slits.py
```diff
@@ -1,15 +1,21 @@
...
+from ophyd_async.core._device import DeviceConnector
+from ophyd_async.epics.core import epics_signal_r
 from ophyd_async.epics.motor import Motor
 
-from dodal.devices.slits import Slits
+from dodal.devices.slits import MinimalSlits, Slits
 
 
-class I10Slits(Slits):
+class I10SlitsBlades(Slits):
```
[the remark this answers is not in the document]
Relm-Arrowny: I guess I do not definitely need it to be able to, it is an old habit of mine when I align slit, I like to scan one blade at a time, where I completely block the beam with just 1 blade and define the half cut position of as 0 so that 0,0,0,0 for the blade is the close/centre position, but I have not done that for a very long time.
The other time I would use it is for clean up slits where I would like to block more on one side. But the same effect can be achieve by closing the gap and moving the centre at the same time, which is slightly painful.

src/dodal/devices/i10/diagnostics.py
Relm-Arrowny: The issue for datatype is without changing it I with get the follow error as datatype is strict enum `       drv: NotConnected:
                data_type: TypeError: BL10I-DI-PHDGN-01:DCAM:CAM:DataType_RBV has choices ('UInt8', 'UInt16'), but <enum 'ADBaseDataType'> requested ['Int8', 'UInt8', 'Int16', 'UInt16', 'Int32', 'UInt32', 'Int64', 'UInt64', 'Float32', 'Float64'] to be strictly equal to them.`

src/dodal/devices/i10/slits.py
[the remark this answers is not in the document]
DominicOram: Thank you, that is clearer

DominicOram (changes requested):
Thanks! Sorry it's taken me so long to get back round to this. Mostly comments that I think you missed from last review plus the `connect` now doesn't work

DominicOram (approved):
Great, thank you! I think we can make the positioner more generic but I can do this in https://github.com/DiamondLightSource/dodal/issues/1176

Issue #927, opened 2024-11-27
 Create a filter wheel attenuator device 


noemifrisina:
Some beamlines, eg I24 and soon I19, have filter wheels on their attenuators, for which there is not yet a device

## Acceptance Criteria
* There is a fairly general filter wheel device with common features that beamline-specific attenuators can use

DiamondJoseph:
Is the implementation anything similar to p99's [FilterMotor](https://github.com/DiamondLightSource/dodal/blob/cb7717c2e1f73dc696227e98cb44acb16c4fd7f8/src/dodal/devices/p99/sample_stage.py#L33), with an enum of possible positions?

olliesilvester:
> Is the implementation anything similar to p99's [FilterMotor](https://github.com/DiamondLightSource/dodal/blob/cb7717c2e1f73dc696227e98cb44acb16c4fd7f8/src/dodal/devices/p99/sample_stage.py#L33), with an enum of possible positions?

Yes, same thing here - but the enum options are different

Issue #926, opened 2024-11-26
 BeamlinePrefix does not support branch beamlines (e.g. i13-1 does not resolve to `BL13J`)


DiamondJoseph:
get_beamline_name returns what is set in the environment variable BEAMLINE, or a default string. 
The return of get_beamline_name is (often) used to create the BeamlinePrefix, which is used by both device_instantiation and device_factory to create the device prefix.

On e.g. i13-1-ws001, BEAMLINE=i13-1, and BeamlinePrefix("i13-1").beamline_prefix = "BL13I" however devices on i13-1 are actually prefixed `BL13J`.

Is using the environment variable to connect, other than in the cases of a simulated twin, a common behaviour that we need to support? This is done by the training_rigs, which share a dodal beamline module, and so must be distinguished.

## Acceptance Criteria
* Specific criteria that will be used to judge if the issue is fixed

noemifrisina:
IIRC, in a case such as `i13-1`, specifying the suffix in BeamlinePrefix should do the trick - even if it's not a very nice solution (and we might want something else in the long run):

```python
from dodal.utils import BeamlinePrefix

bl = BeamlinePrefix("i13-1", suffix="J")
```
This should return `bl.beamline_prefix = "BL13J"

Btw, just to make things more interesting, I'll note that the beamline prefix for both i19-1 and i19-2 is `BL19I` but that might be a single edge case.

DominicOram:
FYI, this is a duplicate of https://github.com/DiamondLightSource/dodal/issues/760. I will close the older one

DominicOram:
I think `bl = BeamlinePrefix("i13-1", suffix="J")` is a good enough solution to this for now and seems to be working on a few beamlines

PR #969, opened 2024-12-19
Make attenuator with enum filter values

olliesilvester:
Fixes some of #927 

This will require a change in `mx_bluesky` before merging

1. Check changes are sensible
2. `dodal connect p99` should still work assuming devices are connected
3. `dodal connect i02_1` passes

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i02_1.py
+++ b/src/dodal/beamlines/i02_1.py
@@ -1,0 +1,34 @@
+"""Beamline i02-1 is also known as VMXm, or I02J"""
+
+from dodal.common.beamlines.beamline_utils import (
+    device_factory,
+)
+from dodal.common.beamlines.beamline_utils import set_beamline as set_utils_beamline
+from dodal.devices.attenuator.attenuator import EnumFilterAttenuator
+from dodal.devices.attenuator.filter_selections import (
+    I02_1FilterFourSelections,
+    I02_1FilterOneSelections,
+    I02_1FilterThreeSelections,
+    I02_1FilterTwoSelections,
+)
+from dodal.log import set_beamline as set_log_beamline
+from dodal.utils import BeamlinePrefix, get_beamline_name
+
+BL = get_beamline_name("i22")
+PREFIX = BeamlinePrefix(BL)
+set_log_beamline(BL)
+set_utils_beamline(BL)
+
+
+@device_factory()
+def attenuator() -> EnumFilterAttenuator:
+    return EnumFilterAttenuator(
+        num_filters=4,
+        filter_selection=[
+            I02_1FilterOneSelections,
+            I02_1FilterTwoSelections,
+            I02_1FilterThreeSelections,
+            I02_1FilterFourSelections,
+        ],
+        prefix="BL02J-OP-ATTN-01:",
+    )
--- a/src/dodal/beamlines/i02_1.py
+++ b/src/dodal/beamlines/i02_1.py
@@ -14,7 +14,7 @@ from dodal.devices.attenuator.filter_selections import (
 from dodal.log import set_beamline as set_log_beamline
 from dodal.utils import BeamlinePrefix, get_beamline_name
 
-BL = get_beamline_name("i22")
+BL = get_beamline_name("i02-1")
 PREFIX = BeamlinePrefix(BL)
 set_log_beamline(BL)
 set_utils_beamline(BL)
@@ -22,6 +22,10 @@ set_utils_beamline(BL)
 
 @device_factory()
 def attenuator() -> EnumFilterAttenuator:
+    """Get the i02-1 attenuator device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i02-1, it will return the existing object.
+    """
+
     return EnumFilterAttenuator(
         num_filters=4,
         filter_selection=[
--- a/src/dodal/beamlines/i02_1.py
+++ b/src/dodal/beamlines/i02_1.py
@@ -27,12 +27,11 @@ def attenuator() -> EnumFilterAttenuator:
     """
 
     return EnumFilterAttenuator(
-        num_filters=4,
-        filter_selection=[
+        f"{PREFIX.beamline_prefix}-OP-ATTN-01:",
+        (
             I02_1FilterOneSelections,
             I02_1FilterTwoSelections,
             I02_1FilterThreeSelections,
             I02_1FilterFourSelections,
-        ],
-        prefix="BL02J-OP-ATTN-01:",
+        ),
     )
--- a/src/dodal/beamlines/i02_1.py
+++ b/src/dodal/beamlines/i02_1.py
@@ -15,7 +15,7 @@ from dodal.log import set_beamline as set_log_beamline
 from dodal.utils import BeamlinePrefix, get_beamline_name
 
 BL = get_beamline_name("i02-1")
-PREFIX = BeamlinePrefix(BL)
+PREFIX = BeamlinePrefix(BL, suffix="J")
 set_log_beamline(BL)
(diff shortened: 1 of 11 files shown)


src/dodal/beamlines/i02_1.py
```diff
@@ -0,0 +1,34 @@
...
+
+@device_factory()
+def attenuator() -> EnumFilterAttenuator:
+    return EnumFilterAttenuator(
+        num_filters=4,
+        filter_selection=[
+            I02_1FilterOneSelections,
+            I02_1FilterTwoSelections,
+            I02_1FilterThreeSelections,
+            I02_1FilterFourSelections,
```
olliesilvester, line 31: Ugly, but maybe it has to be. Open to suggestions

src/dodal/devices/attenuator/attenuator.py
```diff
@@ -83,3 +86,29 @@ async def set(self, value: float):
...
+        name: str = "",
+    ):
+        with self.add_children_as_readables():
+            self.filters: DeviceVector[FilterMotor] = DeviceVector(
+                {
+                    i: FilterMotor(filter_selection[i], f"{prefix}MP{i+1}:", name)
+                    for i in range(num_filters)
+                }
+            )
+        super().__init__(name=name, prefix=prefix)
```
DiamondJoseph, line 114:
```suggestion
    def __init__(
        self,
        prefix: str,
        filter_selection: tuple[type[FilterSelection]],
        name: str = "",
    ):
        with self.add_children_as_readables():
            self.filters: DeviceVector[FilterMotor] = DeviceVector(
                {
                    index: FilterMotor(filter , f"{prefix}MP{index+1}:", name)
                    for index, filter in enumerate(filter_selection)
                }
            )
        super().__init__(prefix, name=name)
```

Should: If num_filters ever doesn't match len(filter_selection) [e.g. I am making a new device and aren't careful] it will cause issue, and is extra book-keeping. 
Should: Making the filter_selection immutable means we can pass in a default if it ends up saving time later, and generally is good practice if the contents of the Sequence don't need to change (if they do, we can adjust typing to Sequence and support tuple/list as appropriate)
Nit: device_factory doesn't require args to be in a particular order, but we have a general practice of prefix being the first argument, just for usability I'd stick to it.

src/dodal/beamlines/i02_1.py
```diff
@@ -0,0 +1,38 @@
...
+        ],
+        prefix="BL02J-OP-ATTN-01:",
+    )
```
DiamondJoseph, line 38:
```suggestion
def attenuator() -> EnumFilterAttenuator:
    """Get the i02-1 attenuator device, instantiate it if it hasn't already been.
    If this is called when already instantiated in i02-1, it will return the existing object.
    """

    return EnumFilterAttenuator(
        f"{PREFIX.beamline_prefix}-OP-ATTN-01:",
        (
            I02_1FilterOneSelections,
            I02_1FilterTwoSelections,
            I02_1FilterThreeSelections,
            I02_1FilterFourSelections,
        ),
    )
```

propagating through my suggested changes, can just add all to batch

src/dodal/beamlines/i02_1.py
```diff
@@ -0,0 +1,38 @@
...
+    I02_1FilterFourSelections,
+    I02_1FilterOneSelections,
+    I02_1FilterThreeSelections,
+    I02_1FilterTwoSelections,
+)
+from dodal.log import set_beamline as set_log_beamline
+from dodal.utils import BeamlinePrefix, get_beamline_name
+
+BL = get_beamline_name("i02-1")
+PREFIX = BeamlinePrefix(BL)
```
DiamondJoseph, line 18:
```suggestion
PREFIX = BeamlinePrefix(BL, suffix="J")
```

#926

src/dodal/beamlines/p99.py
```diff
@@ -17,7 +19,9 @@ def angle_stage() -> SampleAngleStage:
 
 @device_factory()
 def filter() -> FilterMotor:
-    return FilterMotor(f"{PREFIX.beamline_prefix}-MO-STAGE-02:MP:SELECT")
+    return FilterMotor(
+        p99StageSelections, f"{PREFIX.beamline_prefix}-MO-STAGE-02:MP:SELECT"
+    )
```
DiamondJoseph, line 24:
```suggestion
    return FilterMotor(
        f"{PREFIX.beamline_prefix}-MO-STAGE-02:MP:SELECT", p99StageSelections
    )
```

src/dodal/devices/attenuator/filter.py
```diff
@@ -0,0 +1,13 @@
...
+from dodal.devices.attenuator.filter_selections import FilterSelection
+
+
+class FilterMotor(StandardReadable):
+    def __init__(
+        self, filter_selections: type[FilterSelection], prefix: str, name: str = ""
+    ):
+        with self.add_children_as_readables():
+            self.user_setpoint = epics_signal_rw(filter_selections, f"{prefix}SELECT")
+        super().__init__(name=name)
```
DiamondJoseph, line 13:
```suggestion
class FilterMotor(StandardReadable):
    def __init__(
        self, prefix: str, filter_selections: type[FilterSelection], name: str = ""
    ):
        with self.add_children_as_readables():
            self.user_setpoint = epics_signal_rw(filter_selections, f"{prefix}SELECT")
        super().__init__(name=name)
```

src/dodal/devices/attenuator/filter_selections.py
```diff
@@ -0,0 +1,75 @@
+from ophyd_async.core import SubsetEnum
+
+
+class FilterSelection(SubsetEnum): ...
```
DiamondJoseph, line 4: I'd remove this intermediary class, and just allow FilterMotor to have any SubsetEnum, and just make sure that type is well documented in ophyd_async. That then also allows each of these filter types to be defined in ixx.py beamline module, directly alongside the device that is using them, and may allow us to move the concept of the FilterMotor out of dodal in future.

tests/devices/unit_tests/p99/test_p99_stage.py
```diff
@@ -26,7 +26,11 @@ async def sim_sampleAngleStage():
 @pytest.fixture
 async def sim_filter_wheel():
     async with DeviceCollector(mock=True):
-        sim_filter_wheel = FilterMotor("p99-MO-TABLE-01:", name="sim_filter_wheel")
+        sim_filter_wheel = FilterMotor(
+            filter_selections=p99StageSelections,
+            prefix="p99-MO-TABLE-01:",
+            name="sim_filter_wheel",
+        )
```
DiamondJoseph, line 33:
```suggestion
        sim_filter_wheel = FilterMotor(
            "p99-MO-TABLE-01:",
            p99StageSelections,
        )
```

nit: prefix first, name comes automatically from DeviceCollector

DiamondJoseph (changes requested):
Generally looks good, I have opinionated change requests.

src/dodal/devices/attenuator/filter_selections.py
```diff
@@ -0,0 +1,72 @@
+from ophyd_async.core import SubsetEnum
+
+
+class p99StageSelections(SubsetEnum):
```
DiamondJoseph, line 4:
```suggestion
class P99FilterSelections(SubsetEnum):
```

DiamondJoseph: I think if we can make these enums consistently named then I think it's ready to merge

DiamondJoseph (changes requested):

DiamondJoseph (approved):

PR #970, opened 2024-12-19
Add dodal device for i03 beamstop

rtuck99:
Part of fix for DiamondLightSource/mx-bluesky#698

This adds a beamstop device to the i03 beamline.

See also mx-bluesky PR [698](https://github.com/DiamondLightSource/mx-bluesky/pull/725)

1. `dodal connect i03` connects the beamstop
2. tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/i03/beamstop.py
+++ b/src/dodal/devices/i03/beamstop.py
@@ -36,7 +36,7 @@ class Beamstop(StandardReadable):
         x: beamstop x position in mm
         y: beamstop y position in mm
         z: beamstop z position in mm
-        pos_select: Get the current position of the beamstop as an enum. Currently this
+        selected_pos: Get the current position of the beamstop as an enum. Currently this
             is read-only.
     """
 
@@ -47,18 +47,18 @@ class Beamstop(StandardReadable):
         name: str = "",
     ):
         with self.add_children_as_readables():
-            self.x = Motor(prefix + "X")
-            self.y = Motor(prefix + "Y")
-            self.z = Motor(prefix + "Z")
-            self.pos_select = create_hardware_backed_soft_signal(
+            self.x_mm = Motor(prefix + "X")
+            self.y_mm = Motor(prefix + "Y")
+            self.z_mm = Motor(prefix + "Z")
+            self.selected_pos = create_hardware_backed_soft_signal(
                 BeamstopPositions, self._get_selected_position
             )
 
-        self._in_beam_xyz = [
+        self._in_beam_xyz_mm = [
             float(beamline_parameters[f"in_beam_{axis}_STANDARD"])
             for axis in ("x", "y", "z")
         ]
-        self._xyz_tolerance = [
+        self._xyz_tolerance_mm = [
             float(beamline_parameters[f"bs_{axis}_tolerance"])
             for axis in ("x", "y", "z")
         ]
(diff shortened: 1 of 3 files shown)


src/dodal/devices/i03/beamstop.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(
+        self,
+        prefix: str,
+        beamline_parameters: GDABeamlineParameters,
+        name: str = "",
+    ):
+        with self.add_children_as_readables():
+            self.x = Motor(prefix + "X")
+            self.y = Motor(prefix + "Y")
+            self.z = Motor(prefix + "Z")
```
olliesilvester, line 52: Maybe better to call these, eg, `x_mm`

src/dodal/devices/i03/beamstop.py
```diff
@@ -0,0 +1,82 @@
...
+            self.pos_select = create_hardware_backed_soft_signal(
```
olliesilvester, line 53: nit: `pos_select` kind of sounds like we are selecting the position rather than checking it. Maybe better to call it `selected_pos`

src/dodal/devices/i03/beamstop.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class BeamstopPositions(StrictEnum):
+    """
+    Beamstop positions.
+    GDA supports Standard/High/Low resolution positions, as well as parked and
+    robot load however all 3 resolution positions are the same. We also
+    do not use the robot load position in Hyperion.
+
+    Until we support moving the beamstop it is only necessary to check whether the
```
olliesilvester, line 18: We could reference https://github.com/DiamondLightSource/mx-bluesky/issues/484 in this comment too

olliesilvester (approved):
Looks good, more optional comments but feel free to ignore

src/dodal/devices/i03/beamstop.py
rtuck99: I originally modelled this after the I24 beamstop which has identically named PVs but maybe we should change them here and also look at whether we should update that one as well.

src/dodal/devices/i03/beamstop.py
rtuck99: `pos_select` does actually select the position on I24 beamstop as there it is rw, perhaps later we could make ours rw but for now it is ro

src/dodal/devices/i03/beamstop.py
olliesilvester: Yeah, personally I always prefer to include the units in names, so in I24's too

DiamondJoseph: [When emitting bluesky documents] the units will come through as part of the descriptor document Configuration, straight from the EPICS motor record if they're configured there- is there a reason you want the units in the names here?

olliesilvester: It's just much easier to write plans when you can immediately see the units in code

rtuck99: We've found quite a few cases where unit conversions have been the cause of a lot of confusion, areas of the code where we thought things were in one unit but actually in another. It's much easier to spot when a conversion up or down the unit scale is missing if signals and attributes are named clearly, and it's not a big overhead - if the unit changes, you have to change the code everywhere anyway so it's not a big deal.
It's a general preference across the team.

Issue #846, opened 2024-10-17
 Update MX beamlines to use new device factory


DominicOram:
Following https://github.com/DiamondLightSource/dodal/pull/841 we have a new device factory decorator for making devices. We should update all the MX beamlines to use this.

## Acceptance Criteria
* MX beamlines use the new device decorator
* Tests in dodal and in mx-bluesky pass

DominicOram:
Confirm this has been run on i22 first

PR #975, opened 2025-01-07
Update MX beamlines to use new device factory - part 1

rtuck99:
Fixes 
* #846

for I03, other beamlines to follow in separate PR on top of this one.

Also fixes the following issues:
* `diamond_filter` was not being instantiated because python generic device types were not recognised as valid device factories.
* ~`device_factory` did not support usage with sync v1 ophyd devices~ These changes now in follow on PR #984 

See also mx-bluesky PR
* DiamondLightSource/mx-bluesky#735

1. Dodal connect i03, i04 should still work

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -322,35 +322,27 @@ def xspress3mini() -> Xspress3:
     )
 
 
-def panda(
-    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
-) -> HDFPanda:
+@device_factory()
+def panda() -> HDFPanda:
     """Get the i03 panda device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i03, it will return the existing object.
     """
-    return device_instantiation(
-        HDFPanda,
-        "panda",
-        "-EA-PANDA-01:",
-        wait_for_connection,
-        fake_with_ophyd_sim,
+    return HDFPanda(
+        f"{PREFIX.beamline_prefix}-EA-PANDA-01:",
         path_provider=get_path_provider(),
+        name="panda",
     )
 
 
+@device_factory()
 @skip_device(lambda: BL == "s03")
-def sample_shutter(
-    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
-) -> ZebraShutter:
+def sample_shutter() -> ZebraShutter:
     """Get the i03 sample shutter device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i03, it will return the existing object.
     """
-    return device_instantiation(
-        ZebraShutter,
+    return ZebraShutter(
+        f"{PREFIX.beamline_prefix}-EA-SHTR-01:",
         "sample_shutter",
-        "-EA-SHTR-01:",
-        wait_for_connection,
-        fake_with_ophyd_sim,
     )
 
 
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -360,79 +360,59 @@ def flux(wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False) ->
     )
 
 
-def xbpm_feedback(
-    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
-) -> XBPMFeedback:
+@device_factory()
+def xbpm_feedback() -> XBPMFeedback:
     """Get the i03 XBPM feeback device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i03, it will return the existing object.
     """
-    return device_instantiation(
-        XBPMFeedback,
+    return XBPMFeedback(
+        PREFIX.beamline_prefix,
         "xbpm_feedback",
-        "",
-        wait_for_connection,
-        fake_with_ophyd_sim,
     )
 
 
-def zocalo(
-    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
-) -> ZocaloResults:
+@device_factory()
+def zocalo() -> ZocaloResults:
     """Get the i03 ZocaloResults device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i03, it will return the existing object.
     """
-    return device_instantiation(
-        ZocaloResults,
-        "zocalo",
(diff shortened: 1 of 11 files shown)


olliesilvester (changes requested):
Looks good, but I feel like we should put the part which fixes `device_factory` for v1 devices into a separate PR. 

Also, for this issue we might as well convert all of the devices in i04, i24 and i23 to use the new instantiation

src/dodal/beamlines/i03.py
```diff
@@ -57,494 +57,367 @@
...
+        f"{PREFIX.beamline_prefix}-MO-DCM-01:",
         "dcm",
-        "-MO-DCM-01:",
-        wait_for_connection,
-        fake_with_ophyd_sim,
     )
 
 
+@device_factory()
 @skip_device(lambda: BL == "s03")
```
DiamondJoseph, line 97: Depending on how mx-bluesky/hyperion is handling this:

```suggestion
@device_factory(skip = BL == "s03")
```

DiamondJoseph: or `skip = (BL == "s03")` if it's any clearer what is meant

src/dodal/beamlines/i03.py
```diff
@@ -57,494 +57,367 @@
...
     If this is called when already instantiated in i03, it will return the existing object.
     """
     return device_instantiation(
         Flux,
         "flux",
         "-MO-FLUX-01:",
-        wait_for_connection,
-        fake_with_ophyd_sim,
+        wait=False,
+        fake=mock,
```
DiamondJoseph, line 321: Seems to be mixing the two device factory types?

src/dodal/common/beamlines/beamline_utils.py
```diff
@@ -141,8 +140,8 @@ def device_factory(
         SkipType,
         "mark the factory to be (conditionally) skipped when beamline is imported by external program",
     ] = False,
-) -> Callable[[Callable[[], D]], DeviceInitializationController[D]]:
-    def decorator(factory: Callable[[], D]) -> DeviceInitializationController[D]:
+) -> Callable[[Callable[[], T]], DeviceInitializationController[T]]:
+    def decorator(factory: Callable[[], T]) -> DeviceInitializationController[T]:
```
DiamondJoseph, line 144: This has to be D, Ophyd-Async only device, the connection logic does not work for Ophyd devices.

DiamondJoseph (changes requested):

src/dodal/common/beamlines/beamline_utils.py
rtuck99: I have made modifications so that the device factory will now work with ophyd v1 devices.

DiamondJoseph: So you have, if you can't tell that I stopped looking when I reached this point ;)

I thought MX only had the Eiger left on ophyd v1?

rtuck99: Plus a few other minor devices, s4_slits, etc. There is currently a GH ticket to move those across but the Eiger is the main thing,  it will probably be a little while yet before that gets moved over to ophyd_async. I thought better to make a start on this.

Once they are moved over, we will be able to get rid of a whole bunch of code including the `device_instantiation` stuff.

src/dodal/beamlines/i03.py
rtuck99: At this point, I think we are likely going to remove all this `s03` stuff soon as I think we have pretty much given up on the simulator.

rtuck99: There is a separate issue to remove the s03 system tests and I have updated that to remove s03 completely

* DiamondLightSource/mx-bluesky#631

rtuck99:
> Looks good, but I feel like we should put the part which fixes `device_factory` for v1 devices into a separate PR.

These changes now in 
* #984 

> Also, for this issue we might as well convert all of the devices in i04, i24 and i23 to use the new instantiation

These changes to follow in subsequent PRs

src/dodal/beamlines/i04.py
```diff
@@ -427,17 +429,13 @@ def oav_to_redis_forwarder(
     )
 
 
-def diamond_filter(
-    wait_for_connection: bool = True, fake_with_ophyd_sim: bool = False
-) -> DiamondFilter[I04Filters]:
+@device_factory()
+def diamond_filter() -> DiamondFilter[I04Filters]:
```
olliesilvester, line 433: Doesn't really matter, but did you mean to take out this change in favour for your i04 PR?

olliesilvester (changes requested):
The i03 changes look good, but I think there are some lingering changes which you have also made in some other PRs, would you mind taking them out of this one so that we don't end up merging the wrong features from this PR?

src/dodal/utils.py
```diff
@@ -410,7 +412,27 @@ def is_any_device_factory(func: Callable) -> TypeGuard[AnyDeviceFactory]:
 
 
 def is_v2_device_type(obj: type[Any]) -> bool:
```
DominicOram, line 414: Should: Would be nice to have a test that explicitly checks this works for generics.
Could: Also testing the edge cases of not passing a class etc.

DominicOram (changes requested):
Thanks, just one comment

src/dodal/beamlines/i04.py
rtuck99: If these lines aren't updated, then `test_device_creation` fails without further modifications because `device_instantiation` also doesn't work with generics, so then I need to fix `device_instantiation` as well.

tests/test_utils.py
```diff
@@ -19,30 +21,59 @@
     get_beamline_based_on_environment_variable,
     get_hostname,
     get_run_number,
+    is_v2_device_type,
     make_all_devices,
     make_device,
 )
 
+# Duplicated here because of top-level import issues
```
olliesilvester, line 29: Is there an issue for this we can link to in the comment?

olliesilvester (approved):
I'm happy previous comments have been addressed, although I don't fully understand what's going in `is_v2_device_type`

PR #976, opened 2025-01-07
Add beamline-specific zebra mappings to zebra device

### LINKED ISSUE #772: Come up with a system for beamline specific zebra constants
Zebras on different beamlines tend to be wired differently. It is useful to refer to e.g. which output different hardware is on via a nicely named constant rather that the output number. In `zebra.py` we have some constants defined e.g. `EIGER_TTL` but these are only valid for i03. Ideally we would have these defined for other beamlines in a way that we can do:
```
def my_plan(zebra):
    yield from abs_set(zebra.in_1, DETECTOR_TTL)
```
and this works for any beamline. For example we could do this by when instantiating the zebra we also pass in the constants mapping.

## Acceptance Criteria
* Nicely named Zebra constants are defined for all MX beamlines where we currently use bluesky
* It's clear which beamline, and which zebra, these constants are for
* We can easily write generic plans that will use these constants

olliesilvester:
Fixes #772 

Associated mx-bluesky change: https://github.com/DiamondLightSource/mx-bluesky/pull/741

1. Check changes are sensible and address issue
2. Confirm unit tests pass

### Checks for reviewer


src/dodal/devices/zebra/zebra_constants_mapping.py
```diff
@@ -0,0 +1,100 @@
...
+    TTL_FAST_SHUTTER: int | None = None
+    TTL_DETECTOR: int | None = None
+    TTL_SHUTTER: int | None = None
+    TTL_XSPRESS3: int | None = None
+    TTL_PANDA: int | None = None
+
+    # Zebra signal sources. The integers refer to where the different sources are located
+    # on the zebra IOC. Can be mapped to a zebra output by doing, eg,
+    # bps.abs_set(zebra.output.out_pvs[TTL_DETECTOR], DISCONNECT)
+    DISCONNECT: int | None = 0
```
noemifrisina: I'm not sure these should have `| None` ? Afaik these values are set and don't change, is there a use case when you need to set one to None instead of the value on the IOC?

olliesilvester: Yeah you might be right, I wasn't sure if non mx Zebras used the same IOC

PR #981, opened 2025-01-10
Update the lut linear interpolation to be used by different lookup table files

noemifrisina:
Split the linear_interpolation_lut into two different functions so that it can be used for lookup tables that have more than 2 columns in them (eg. DetDistConverter)

Needed for [Mx#708](https://github.com/DiamondLightSource/mx-bluesky/pull/708)

1. Run tests

### Checks for reviewer


src/dodal/devices/util/lookup_tables.py
```diff
@@ -32,28 +32,32 @@ async def energy_distance_table(lookup_table_path: str) -> np.ndarray:
     return loadtxt(StringIO(raw_table), comments=["#", "Units"])
 
 
-def linear_interpolation_lut(filename: str) -> Callable[[float], float]:
+def parse_lookup_table(filename: str) -> list[Sequence]:
+    """Parse a generic lookup table with a number of columns >= 2 and return a list \
+        of all the values in it."""
```
rtuck99, line 37: Can you reword this - the description makes it sound as though it returns a list in row major order (which is what you'd naively expect) but it actually returns it in column-major order.

src/dodal/devices/util/lookup_tables.py
```diff
@@ -32,28 +32,32 @@ async def energy_distance_table(lookup_table_path: str) -> np.ndarray:
...
         LOGGER.info(
-            f"Configuration file {filename} values are not ascending, trying reverse order..."
+            "Configuration file values are not ascending, trying reverse order..."
         )
         s_values = list(reversed(s_values))
         t_values = list(reversed(t_values))
         if not np.all(np.diff(s_values) > 0):
             raise AssertionError(
-                f"Configuration file {filename} lookup table does not monotonically increase or decrease."
+                "Configuration file lookup table does not monotonically increase or decrease."
```
rtuck99, line 60: It's not a file anymore, it's just a table

PR #982, opened 2025-01-10
Add 'not external_tests' option to pytest

### LINKED ISSUE #954: Create pytest marker to easily run local unit tests
To run unit tests locally, we now need to type `pytest -m "not (s03 or adim)"`, and the list of things which we are skipping is going to grow. We should add a `unit tests only` flag which does all the 'nots' for us

## Acceptance Criteria
* Add pytest flag to run local unit tests

olliesilvester:
Fixes #954 
Enables us to run unit tests with `pytest -m "not external_tests"` rather than `pytest -m "not (s03 or adsim)"`

1. Confirm that running `pytest "not external_tests"` correctly skips `s03` and `adsim` tests but nothing else

### Checks for reviewer

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -25,6 +25,9 @@ from dodal.common.visit import (
 )
 from dodal.log import LOGGER, GELFTCPHandler, set_up_all_logging_handlers
 
+# List of pytest markers for tests which won't work without external configuration
+EXTERNAL_TEST_MARKERS = {"s03", "adsim"}
+
 MOCK_DAQ_CONFIG_PATH = "tests/devices/unit_tests/test_daq_configuration"
 mock_paths = [
     ("DAQ_CONFIGURATION_PATH", MOCK_DAQ_CONFIG_PATH),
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -81,11 +84,15 @@ def pytest_internalerror(excinfo: pytest.ExceptionInfo[Any]):
 
 
 def pytest_runtest_setup(item):
+    if item.config.getoption("-m") == "not external_tests":
+        if EXTERNAL_TEST_MARKERS.intersection(
+            mark.name for mark in item.iter_markers()
+        ):
+            pytest.skip()
```
rtuck99, line 91: Would be nice if you could also run `pytest -m external_tests`. Can this be done? 
Also, should it be called `system_tests` instead of `external_tests` since that's what the folder is called?

BTW you can also do this already with `tox -e tests`

olliesilvester:
Closing in favour of using `tox -e tests` !

PR #985, opened 2025-01-13
Remove ophyd references from motion devices

### LINKED ISSUE #168: Convert all basic motion devices to ophyd-async
We should move all the devices in this repo to `ophyd-async` (formally ophyd v2) rather than `ophyd`. The easiest place to start is devices that contain no logic but are just collections of mostly motors.

## Acceptance Criteria
* A list of such devices is put here
* We go through all devices and convert them
* All Hyperion/dodal tests should pass with no changes

shihab-dls:
Fixes #168 

1. Check no references to ophyd for motion devices 
2. Check that p45 devices are correctly instantiated

### Checks for reviewer


rtuck99:
I believe the ticket originally referred to i03 devices, so I wasn't expecting to see them for `p45`. However I suppose it doesn't hurt!

I think the ones to include are:

i03, i04:
S4SlitGaps
Flux

Then there is another device:
InfoLoggingDevice (seems like this is unused, don't know what it is for)

I think the only other ones are in `eiger.py` and `eiger_odin.py` which can be done as part of the Eiger device.

src/dodal/devices/p45.py
```diff
@@ -1,44 +1,51 @@
...
+class SampleY(StandardReadable):
     """
     Motors for controlling the sample's y position and stretch in the y axis.
     """
 
-    base = Cpt(EpicsMotor, "CS:Y")
-    stretch = Cpt(EpicsMotor, "CS:Y:STRETCH")
-    top = Cpt(EpicsMotor, "Y:TOP")
-    bottom = Cpt(EpicsMotor, "Y:BOT")
+    def __init__(self, prefix: str, name="") -> None:
```
rtuck99, line 10: Q: As these are `StandardReadable`s, do we want to add these motors with `self.add_children_as_readables()`? (see e.g. `Smargon`),  this enables the values to be read in callback documents.

as I don't really know much about the p45 beamline, I can't say for sure but would assume this is the case.

src/dodal/devices/p45.py
```diff
@@ -1,44 +1,51 @@
...
+        self.base = Motor("CS:Y")
```
rtuck99, line 11: Generally speaking, in all these devices, you will need to include the prefix in the constructors of the child devices + signals

shihab-dls: Ah, of course; amended.

rtuck99 (approved):
Marvellous. Approved

PR #992, opened 2025-01-16
Bimorph mirrors

dan-fernandes:
Fixes writing to bimorph when bimorph is not IDLE.

Removes unnecessary readback tolerance.

1. Check tests pass
2. Run a bimorph mirror plan (against the simulated bimorph on BL02J or a real bimorph) and check plan behaves as expected

### Checks for reviewer


tests/devices/unit_tests/test_bimorph_mirror.py
```diff
@@ -29,20 +30,35 @@ def valid_bimorph_values(mirror: BimorphMirror) -> dict[int, float]:
 
 
 @pytest.fixture
-def mock_vtrgt_vout_propogation(mirror: BimorphMirror):
+def bimorph_functionality(mirror: BimorphMirror):
```
callumforrester, line 33: Should: I would suggest this instead

```python
@pytest.fixture
def mirror_with_mocked_put(mirror: BimorphMirror) -> BimorphMirror:
    # Do things to mirror
    return mirror
```

Then depend on `mirror_with_mocked_put` _instead of_ `mirror` in tests which need this logic. It reads better.

src/dodal/devices/bimorph_mirror.py
```diff
@@ -118,22 +118,27 @@ async def set(self, value: Mapping[int, float], tolerance: float = 0.0001) -> No
             )
 
         # Write target voltages:
-        await asyncio.gather(
-            *[
-                self.channels[i].target_voltage.set(target, wait=True)
-                for i, target in value.items()
-            ]
-        )
+        for i, target in value.items():
```
callumforrester, line 121: Should: Doing this in sequence will significantly increase latency. If this is intentional I think it deserves a comment to stop someone in the future thinking they can speed it by parallelising it.

callumforrester (changes requested):

src/dodal/devices/bimorph_mirror.py
dan-fernandes: This is intentional, I have added a comment.

tests/devices/unit_tests/test_bimorph_mirror.py
```diff
@@ -29,117 +30,113 @@ def valid_bimorph_values(mirror: BimorphMirror) -> dict[int, float]:
...
-def mock_vtrgt_vout_propogation(mirror: BimorphMirror):
+def mirror_with_mocked_put(mirror: BimorphMirror):
+    async def busy_idle():
+        await asyncio.sleep(0)
+        set_mock_value(mirror.status, BimorphMirrorStatus.BUSY)
+        await asyncio.sleep(0)
+        set_mock_value(mirror.status, BimorphMirrorStatus.IDLE)
+
+    async def status(*_, **__):
+        asyncio.create_task(busy_idle())
```
DiamondJoseph, line 41:
```suggestion
    async def status(value: Any, wait: bool):
        asyncio.create_task(busy_idle())
```

This should be the correct type annotation and args, I think?

`Any` could probably be `float | BimorphMirrorStatus` but it's fine

tests/devices/unit_tests/test_bimorph_mirror.py
```diff
@@ -29,117 +30,113 @@ def valid_bimorph_values(mirror: BimorphMirror) -> dict[int, float]:
...
+
+    for signal in walk_rw_signals(mirror).values():
+        callback_on_mock_put(signal, status)
+
     for channel in mirror.channels.values():
 
-        def effect(value: float, wait=False, signal=channel.output_voltage):
+        def vout_propogation_and_status(
+            value: float, wait=False, signal=channel.output_voltage
+        ):
```
DiamondJoseph, line 50:
```suggestion
        def vout_propogation_and_status(value: float, wait: bool = False):
```

I don't think the third param was needed here?

DiamondJoseph (changes requested):

src/dodal/devices/bimorph_mirror.py
```diff
@@ -117,23 +117,29 @@ async def set(self, value: Mapping[int, float], tolerance: float = 0.0001) -> No
...
+        # Write target voltages in serial
+        # Voltages are written in serial as bimorph PSU cannot handle simultaneous sets
+        for i, target in value.items():
+            await wait_for_value(
+                self.status, BimorphMirrorStatus.IDLE, timeout=DEFAULT_TIMEOUT
+            )
+            await self.channels[i].target_voltage.set(target, wait=True)
+            await wait_for_value(
+                self.status, BimorphMirrorStatus.BUSY, timeout=DEFAULT_TIMEOUT
+            )
```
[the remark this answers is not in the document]
DominicOram: If you really want to avoid this race condition then there is a pattern in https://github.com/DiamondLightSource/dodal/blob/5fe046416a70105106f1dfcfa1c8ecb6a95a9a95/src/dodal/devices/focusing_mirror.py#L89 where we start watching for the change before we do the set. I have added XXX to document this better, if you do use it it may make sense to pull the pattern into a helper. TBH, the focussing mirrors are basically the same as this device. We should probably check if we can use some commonality between the two.

>Andy is pretty sure it may need the extra 0.1s sleep after returning to IDLE before being ready to take the next set
Please no, if this is really the case can Andy add this into the EPICS logic that it doesn't return IDLE until 0.1s is up. Otherwise we just end up having the same issue if we ever have a client that isn't going through `ophyd`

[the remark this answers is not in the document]
DominicOram: Equally, I would say that the `await asyncio.sleep(0.1)` already here is bad, it at least needs a comment as to why we're waiting at best it needs more discussion with Andy on why we have to wait and if we can actually monitor something. We have been bitten in MX with lots of places in GDA where something doesn't work and people have just added an arbitrary sleep to avoid it and that sleep just grows or masks an actual underlying issue where we're not setting the right PVs etc. It's one of the big reasons Hyperion is faster than GDA. Sorry, rant over.

tests/devices/unit_tests/test_bimorph_mirror.py
```diff
@@ -29,117 +30,113 @@ def valid_bimorph_values(mirror: BimorphMirror) -> dict[int, float]:
...
+    async def status(*_, **__):
+        asyncio.create_task(busy_idle())
+
+    for signal in walk_rw_signals(mirror).values():
+        callback_on_mock_put(signal, status)
+
     for channel in mirror.channels.values():
 
-        def effect(value: float, wait=False, signal=channel.output_voltage):
+        def vout_propogation_and_status(
```
DominicOram, line 48: Nit: `vout` confused me as a word when I first read this

tests/devices/unit_tests/test_bimorph_mirror.py
dan-fernandes: Added `Any` type hint to both parameters following in-office discussion.

tests/devices/unit_tests/test_bimorph_mirror.py
dan-fernandes: Renamed all `vout` references to `output_voltage`.

This is also in line with variable names in the `BimorphMirror` itself.

src/dodal/devices/bimorph_mirror.py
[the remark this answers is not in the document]
dan-fernandes: I will organise a meeting with Andy Foster to discuss whether these changes could be made on the IOC / hardware level.

dan-fernandes:
I have had an in-office conversation with Andrew Foster. We are going to write to the bimorph through setting to `VOUT` rather than `VTRGT`.

tests/devices/unit_tests/test_bimorph_mirror.py
dan-fernandes: Resolved in person. Late binding closure issues fixed by definition an additional function within the loop.

src/dodal/devices/bimorph_mirror.py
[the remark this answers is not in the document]
dan-fernandes: I have implemented Andy's suggestion (writing directly to `VOUT` (output voltage) rather than to `VTRGT` (target voltage) then proccing `ALLTRGT.PROC` (set target voltages).

I have not introduced additional sleeps.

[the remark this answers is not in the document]
DominicOram: Great, I think that's a lot nicer. Thank you!

callumforrester (approved):
LGTM and I'm satisfied previous conversations have been resolved

PR #996, opened 2025-01-21
Return immediately if transfocator number of lenses is already correct

### LINKED ISSUE #995: Transfocator: Return immediately if number of lenses is already correct
If the transfocator is asked for a beamsize that is not exactly the same as the previous size but close enough that it results in the same number of lenses it gets stuck and waits for ever. Instead it should return immediately and do nothing.

This can be confirmed with the following test:
```python
async def test_given_number_of_lenses_is_already_correct_then_transfocator_set_returns_immediately(
    fake_transfocator: Transfocator,
):
    set_mock_value(fake_transfocator.predicted_vertical_num_lenses, 10)
    await fake_transfocator.set(20)
```

## Acceptance Criteria
* The test above is added and passes

shihab-dls:
Fixes #995 

1. Check that transfocator set operates as intended
2. Check that tests pass, and appropriately test an early return if predicted number of lenses is already set

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/i04/transfocator.py
+++ b/src/dodal/devices/i04/transfocator.py
@@ -37,32 +37,18 @@ class Transfocator(StandardReadable):
 
         super().__init__(name=name)
 
-    async def _observe_beamsize_microns(self):
-        is_set_filters_done = False
-
-        async def set_based_on_prediction(value: float):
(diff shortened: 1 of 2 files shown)


DominicOram (approved):
Great, thank you!

PR #1018, opened 2025-01-24
Remove logic for checking if zoom level is available

### LINKED ISSUE #884: Remove check for zoom levels in ZoomController
In the `ZoomController` we currently raise a `ZoomLevelNotFoundError` if the zoom levels set by the user do not match the levels specified on the device. These could be different as we use a string signal but the device actually has an `enum`. The only purpose of this check is to give a nice exception to the user that lists the possible values that the zoom controller has. However, this behavior should probably be in `ophyd-async` (if it's not already).

## Acceptance Criteria
* Test what `ophyd-async` does if you try and set an enum with in invalid string
* If `ophyd-async` responds with an error containing the actually valid strings then remove this check in `dodal`
* If not add this behavior to `ophyd-async`

shihab-dls:
Fixes #884 

EPICS enum records should error if provided an unspecified value; thus, this PR removes logic checking for unspecified values in the `ZoomController`.

1. Test if setting the `str` signal `level` to an unspecified zoom level raises an error.

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/oav/oav_detector.py
+++ b/src/dodal/devices/oav/oav_detector.py
@@ -1,12 +1,12 @@
 from enum import IntEnum
 
+from bluesky.protocols import Movable
 from ophyd_async.core import (
     DEFAULT_TIMEOUT,
     AsyncStatus,
     LazyMock,
     StandardReadable,
 )
-from bluesky.protocols import Movable
 from ophyd_async.epics.core import epics_signal_rw
 
 from dodal.common.signal_utils import create_hardware_backed_soft_signal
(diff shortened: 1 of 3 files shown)


DominicOram (commented):
This works but the error message is obtuse (see https://github.com/bluesky/ophyd-async/issues/750). If there is some reason we can't improve the error in `ophyd_async` then we may want to keep this. I suggest we wait for that issue to have some discussion

DominicOram (approved):
Looks like this will get fixed in `ophyd_async` (https://github.com/bluesky/ophyd-async/pull/752) so happy to merge this

PR #1021, opened 2025-01-27
Add stub description files for I19-1 and i19-2

### LINKED ISSUE #1006: Create I19-1 and I19-2 beamlines definitions
Create i19-1 and i19-2 definitions to be used by [i19-bluesky](https://github.com/DiamondLightSource/i19-bluesky) with what devices are already available.

## Acceptance Criteria
* There are i19_1.py and i19_2.py files with device instantiation

noemifrisina:
Fixes #1006

1. Run dodal connect i19-1
2. Run dodal connect i19-2

### Checks for reviewer


src/dodal/beamlines/i19_2.py
```diff
@@ -0,0 +1,65 @@
...
+    ZebraTTLOutputs,
+)
+from dodal.log import set_beamline as set_log_beamline
+from dodal.utils import BeamlinePrefix
+
+# NOTE all workstations on I19 default to i19-1 as beamline name
+# Unless variable is exported (which is not usually done by scientists)
+# NOTE All PVs for both hutches have the prefix BL19I
+BL = "i19-2"
+PREFIX = BeamlinePrefix(BL)
```
DominicOram, line 22: Should: It's actually a bug that `BeamlinePrefix(i19-2) == BL19I` (https://github.com/DiamondLightSource/dodal/issues/926). So I would change this here and in `i19_1.py` to be explicit i.e. `PREFIX = BeamlinePrefix("i19", "I")`

DominicOram (changes requested):
Great, thanks. One comment in code. Additionally, as a `should`, the default behaviour of `@device_factory` is that the name of the device is the name of the factory, which is always the case for you so I think you can just remove setting the name in all of these

DiamondJoseph:
> ... Additionally, as a `should`, the default behaviour of `@device_factory` is that the name of the device is the name of the factory...

I wasn't sure how opinionated to be about it, I was going to point it out as a `nit`: that and the Synchrotron prefix isn't required (which once device_initialisation goes away we can remove it)

DominicOram (approved):
> > ... Additionally, as a `should`, the default behaviour of `@device_factory` is that the name of the device is the name of the factory...
> 
> I wasn't sure how opinionated to be about it, I was going to point it out as a `nit`: that and the Synchrotron prefix isn't required (which once device_initialisation goes away we can remove it)

You're right, I think it's actually a `could`, makes the code cleaner but not the end of the world. But thanks for doing it!

PR #1032, opened 2025-01-29
Removes/reduces refenece use on i10 IDs

### LINKED ISSUE #1029: Removes/reduces refenece use on i10 IDs
Reduce the use of references and hide some of the less user friendly devices out of i10.py. 

## Acceptance Criteria 
- [ ] User facing devices should be as simple as possible.
- [ ] Users should be able to scan energy and polarisation.
- [ ] Users should be able set energy offset between the pgm and Id, this offset should remain during scan.  
- [ ] User should be able to set id polarisation. 
- [ ] Doc string in i10.py.
- [ ] All test still pass.
- [ ] All devices still connect.

Relm-Arrowny:
Fixes #1029 and it is related to #1002 as it remove the use of references in Apple2.

1. Run test.
2. Confirm connection.

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/apple2_undulator.py
+++ b/src/dodal/devices/apple2_undulator.py
@@ -7,7 +7,6 @@ import numpy as np
 from bluesky.protocols import Movable
 from ophyd_async.core import (
     AsyncStatus,
-    Reference,
     StandardReadable,
     StandardReadableFormat,
     StrictEnum,
@@ -388,8 +387,8 @@ class Apple2(StandardReadable, Movable):
 
         # Attributes are set after super call so they are not renamed to
         # <name>-undulator, etc.
-        self.gap = Reference(id_gap)
-        self.phase = Reference(id_phase)
+        self.gap = id_gap
+        self.phase = id_phase
 
         with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
             # Store the polarisation for readback.
@@ -437,16 +436,16 @@ class Apple2(StandardReadable, Movable):
         """
 
         # Only need to check gap as the phase motors share both fault and gate with gap.
-        await self.gap().check_id_status()
+        await self.gap.check_id_status()
         await asyncio.gather(
-            self.phase().top_outer.user_setpoint.set(value=value.top_outer),
-            self.phase().top_inner.user_setpoint.set(value=value.top_inner),
-            self.phase().btm_inner.user_setpoint.set(value=value.btm_inner),
-            self.phase().btm_outer.user_setpoint.set(value=value.btm_outer),
-            self.gap().user_setpoint.set(value=value.gap),
+            self.phase.top_outer.user_setpoint.set(value=value.top_outer),
+            self.phase.top_inner.user_setpoint.set(value=value.top_inner),
+            self.phase.btm_inner.user_setpoint.set(value=value.btm_inner),
+            self.phase.btm_outer.user_setpoint.set(value=value.btm_outer),
+            self.gap.user_setpoint.set(value=value.gap),
         )
         timeout = np.max(
-            await asyncio.gather(self.gap().get_timeout(), self.phase().get_timeout())
+            await asyncio.gather(self.gap.get_timeout(), self.phase.get_timeout())
         )
         LOGGER.info(
             f"Moving f{self.name} energy and polorisation to {energy}, {self.pol}"
@@ -454,12 +453,10 @@ class Apple2(StandardReadable, Movable):
         )
 
         await asyncio.gather(
-            self.gap().set_move.set(value=1, timeout=timeout),
-            self.phase().set_move.set(value=1, timeout=timeout),
+            self.gap.set_move.set(value=1, timeout=timeout),
+            self.phase.set_move.set(value=1, timeout=timeout),
         )
-        await wait_for_value(
-            self.gap().gate, UndulatorGateStatus.CLOSE, timeout=timeout
-        )
+        await wait_for_value(self.gap.gate, UndulatorGateStatus.CLOSE, timeout=timeout)
         self._energy_set(energy)  # Update energy for after move for readback.
 
     def _get_id_gap_phase(self, energy: float) -> tuple[float, float]:
@@ -524,11 +521,11 @@ class Apple2(StandardReadable, Movable):
         (May be for future one can use the inverse poly to work out the energy and try to match it with the current energy
         to workout the polarisation but during my test the inverse poly is too unstable for general use.)
         """
-        top_outer = await self.phase().top_outer.user_setpoint_readback.get_value()
-        top_inner = await self.phase().top_inner.user_setpoint_readback.get_value()
-        btm_inner = await self.phase().btm_inner.user_setpoint_readback.get_value()
-        btm_outer = await self.phase().btm_outer.user_setpoint_readback.get_value()
-        gap = await self.gap().user_readback.get_value()
+        top_outer = await self.phase.top_outer.user_setpoint_readback.get_value()
+        top_inner = await self.phase.top_inner.user_setpoint_readback.get_value()
+        btm_inner = await self.phase.btm_inner.user_setpoint_readback.get_value()
+        btm_outer = await self.phase.btm_outer.user_setpoint_readback.get_value()
+        gap = await self.gap.user_readback.get_value()
         if gap > MAXIMUM_GAP_MOTOR_POSITION:
             raise RuntimeError(
                 f"{self.name} is not in use, close gap or set polarisation to use this ID"
--- a/src/dodal/devices/i10/i10_apple2.py
+++ b/src/dodal/devices/i10/i10_apple2.py
@@ -8,6 +8,7 @@ import numpy as np
 from bluesky.protocols import Movable
 from ophyd_async.core import (
     AsyncStatus,
+    Device,
     Reference,
     StandardReadable,
     StandardReadableFormat,
@@ -15,6 +16,7 @@ from ophyd_async.core import (
     soft_signal_rw,
 )
 
+# from dodal.beamlines.i10 import pgm
 from dodal.devices.apple2_undulator import (
     Apple2,
     Apple2Val,
@@ -111,15 +113,14 @@ class I10Apple2(Apple2):
             max_energy=max_energy,
             poly_deg=poly_deg,
         )
-
-        super().__init__(
-            id_gap=id_gap,
-            id_phase=id_phase,
-            prefix=prefix,
-            name=name,
-        )
         with self.add_children_as_readables():
-            self.id_jaw_phase = Reference(id_jaw_phase)
+            super().__init__(
+                id_gap=id_gap,
+                id_phase=id_phase,
+                prefix=prefix,
+                name=name,
+            )
+            self.id_jaw_phase = id_jaw_phase
 
     @AsyncStatus.wrap
     async def set(self, value: SupportsFloat) -> None:
@@ -148,8 +149,8 @@ class I10Apple2(Apple2):
         LOGGER.info(f"Setting polarisation to {self.pol}, with {id_set_val}")
         await self._set(value=id_set_val, energy=value)
         if self.pol != "la":
-            await self.id_jaw_phase().set(0)
-            await self.id_jaw_phase().set_move.set(1)
+            await self.id_jaw_phase.set(0)
+            await self.id_jaw_phase.set_move.set(1)
 
     def update_lookuptable(self):
         """
@@ -197,7 +198,7 @@ class I10Apple2PGM(StandardReadable, Movable):
             New device name.
         """
         super().__init__(name=name)
-        self.id_ref = Reference(id)
+        self.id = id
         self.pgm_ref = Reference(pgm)
         with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
             self.energy_offset = soft_signal_rw(float, initial_value=0)
@@ -206,7 +207,7 @@ class I10Apple2PGM(StandardReadable, Movable):
     async def set(self, value: float) -> None:
         LOGGER.info(f"Moving f{self.name} energy to {value}.")
         await asyncio.gather(
-            self.id_ref().set(value=value + await self.energy_offset.get_value()),
+            self.id.set(value=value + await self.energy_offset.get_value()),
             self.pgm_ref().energy.set(value),
         )
 
@@ -229,14 +230,14 @@ class I10Apple2Pol(StandardReadable, Movable):
         """
         super().__init__(name=name)
         with self.add_children_as_readables():
-            self.id = id
+            self.id = Reference(id)
 
     @AsyncStatus.wrap
     async def set(self, value: str) -> None:
-        self.id.pol = value  # change polarisation.
+        self.id().pol = value  # change polarisation.
         LOGGER.info(f"Changing f{self.name} polarisation to {value}.")
-        await self.id.set(
-            await self.id.energy.get_value()
+        await self.id().set(
+            await self.id().energy.get_value()
         )  # Move id to new polarisation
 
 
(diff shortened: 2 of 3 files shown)


src/dodal/devices/apple2_undulator.py
```diff
@@ -380,95 +381,88 @@ def __init__(
...
         self.update_lookuptable()
 
-    @property
-    def pol(self):
-        return self._pol
-
-    @pol.setter
-    def pol(self, pol: str):
-        # This set the polarisation but does not actually move hardware.
+    def check_pol(self, pol: str) -> bool:
```
DominicOram, line 390: Should: Can we use an enum for `pol`? Then we don't need to do this check, the type system will do it for us.

src/dodal/devices/i10/i10_apple2.py
```diff
@@ -299,10 +323,77 @@ async def set(self, value: SupportsFloat) -> None:
...
+                energy_phase_table_path=Path(
+                    look_up_table_dir + "IDEnergy2PhaseCalibrations.csv",
+                ),
+                name="idd_energy",
+                source=source,
+                prefix="",
+            ),
+            pgm=pgm,
+            name="energy",
+        )
```
DominicOram, line 384: Should: It's weird to me that the inner objects are instantiated here. I would expect this instead to just be something like:

```suggestion
        self._id = I10Apple2(prefix, look_up_table_dir, source)
        self.energy = I10Apple2PGM(id=I10Apple2(self._id, pgm, name="energy")
```

and the rest to be instantiated in the constructor of `I10Apple2`

src/dodal/devices/i10/i10_apple2.py
```diff
@@ -177,7 +186,7 @@ def update_lookuptable(self):
 
 class I10Apple2PGM(StandardReadable, Movable):
```
DominicOram, line 187: Could: This might be more obvious if it's named something like `EnergySetter`

src/dodal/devices/apple2_undulator.py
```diff
@@ -497,11 +491,11 @@ async def determinePhaseFromHardware(self) -> tuple[str | None, float]:
         (May be for future one can use the inverse poly to work out the energy and try to match it with the current energy
```
DominicOram, line 491: Could: This is maybe a bit out of scope of the issue but could you rename this function to be `snake_case`?

DominicOram (changes requested):
Thanks, I think it's cleaner but have some more comments to make it even more so.

src/dodal/devices/apple2_undulator.py
```diff
@@ -280,20 +279,22 @@ async def get_timeout(self) -> float:
...
+                    axis.user_readback,
                     axis.velocity,
                 )
                 for axis in axes
             ]
         )
-        return np.max(timeouts)
+        # The extra 2.0 is needed as motors timeout too quicky due
+        # to readback speed is amost twice the actual speed.
+        return np.max(timeouts) * 2.0
```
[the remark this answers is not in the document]
Relm-Arrowny: Updated doc-string to explain why there is a factor of 2.

src/dodal/devices/apple2_undulator.py
Relm-Arrowny: Can I leave this one?  As the available polarisation is define by the lookup table on device creation, I will have to store it in the class and make it available for user to set or I will have to convert user imput into enum during set. I also prefer the custom error message as enum does not show all the available value on error, which is useful when the pol that is not available.

DominicOram: I actually think I do feel quite strongly about this.

> As the available polarisation is define by the lookup table on device creation
But presumably the available polarisations in that file don't actually change that often? So just creating a fixed enum won't cause much pain? If they do change often then this is maybe a reason to keep as is

> make it available for user to set
Yes, this is how we usually deal with enums. eg.
```python
# device
class MyOptions(enum):
   option_0 = 0
   option_1 = 1

class MyDevice():
    async def set(value:MyOptions):
        ...

#plan
from device import MyOptions, MyDevice

yield from bps.set(device, MyOptions.option_1)
```

> I also prefer the custom error message
If the person writing the plan is using types properly they will never get the error in the first place as they will see it before they even run the code. If you want to make it obvious to someone not using a type checker (which we should try and discourage but will probably happen) then you can do something like:
```python
async def set(value: MyOptions):
    if not isinstance(value, MyOptions):
        raise ValueException(f"Must set to one of {[e.value for e in MyOptions]")
```

DominicOram (changes requested):
Thanks, couple of comments in code. Additionally, I think polarisation is basically a derived signal. I will do https://github.com/DiamondLightSource/dodal/issues/1077 so that we can use it here as I think it will clean things up a lot.

DominicOram:
I think you should be able to use https://github.com/DiamondLightSource/dodal/pull/1078 for the polarisation signal now. Happy to help if it's not obvious how.

src/dodal/devices/apple2_undulator.py
```diff
@@ -391,19 +406,19 @@ def __init__(
...
-                f"Polarisation {pol} is not available:"
-                + f"/n Polarisations available:  {self._available_pol}"
-            )
-
-    def set_pol(self, pol: str) -> None:
+    # Hardward backed polarisation function,
+    async def read_pol_setpoint(self):
+        pol = await self.polarisation_setpoint.get_value()
+        # LH3 is not hardware readable as it is the same as lh but it is needed for energy.
+        if pol.value != "lh3":
```
DominicOram, line 413: Could: If you have the enum now you can use it:
```suggestion
        if pol != Pol.LH3:
```

src/dodal/devices/i10/i10_apple2.py
```diff
@@ -129,29 +130,21 @@ def __init__(
...
     @AsyncStatus.wrap
     async def set(self, value: float) -> None:
         """
         Check polarisation state and use it together with the energy(value)
         to calculate the required gap and phases before setting it.
         """
-        pol = await self.polarisation.get_value()
-        if pol not in self._available_pol:
+        pol = await self.polarisation_setpoint.get_value()
+        if pol.value == Pol.NONE:
```
DominicOram, line 140: Could:
```suggestion
        if pol == Pol.NONE:
```

src/dodal/devices/i10/i10_apple2.py
```diff
@@ -129,29 +130,21 @@ def __init__(
...
             LOGGER.warning("Polarisation not set attempting to read from hardware")
             pol, phase = await self.determine_phase_from_hardware()
-            if pol is None:
+            if pol.value == "None":
```
DominicOram, line 143: Could:
```suggestion
            if pol == Pol.NONE:
```

tests/devices/i10/test_i10Apple2.py
```diff
@@ -349,10 +349,10 @@ async def test_I10Apple2_pol_set(
     mock_id_pol.id_ref()._energy_set(energy)
     if pol == "dsf":
         with pytest.raises(ValueError):
-            await mock_id_pol.set(pol)
+            await mock_id_pol.set(Pol(pol))
```
DominicOram, line 352: Could: You can just change the `parametrize` to use `Pol.LH` rather than the `str` "lh"?

src/dodal/devices/apple2_undulator.py
```diff
@@ -373,16 +384,20 @@ def __init__(
...
 
-        with self.add_children_as_readables():
-            # Store the polarisation for readback.
-            self.polarisation, self._polarisation_set = soft_signal_r_and_setter(
-                str, initial_value=None
-            )
+        # Store the polarisation for setpoint.
+        self.polarisation_setpoint, self._polarisation_setpoint_set = (
+            soft_signal_r_and_setter(Pol)
+        )
```
DominicOram, line 390: Should: You can just use the setter added in https://github.com/DiamondLightSource/dodal/pull/1078 for this, right?

DominicOram (changes requested):
Thanks! I think there are still quite a few places where you could be making use of the enum but this is a start. I think using the setter from the other PR will tidy things up a bit though

src/dodal/devices/apple2_undulator.py
```diff
@@ -348,129 +419,158 @@ def __init__(
...
+        # This change the pol setpoint and then change polariastion via set energy.
+        self.set_pol_setpoint(value)
+        await self._set_energy(await self.energy.get_value())
+
+    @abc.abstractmethod
+    async def _set_energy(self, value: float) -> None:
+        """This change the position of all the motors for a given energy and
+        polarisation_setpoint"""
+
+    def _read_pol_setpoint(
```
olliesilvester, line 488: nit: name is implying we are only doing a read. I'd prefer something like `def _update_pol_setpoint_using_hardware_read`

olliesilvester: The `raw_to_derived` function is for converting raw values to derived values. It seems a bit unexpected for this function to also be changing a setpoint. I'm not too familiar with this device yet though, so I could just be misunderstanding

src/dodal/devices/apple2_undulator.py
```diff
@@ -348,129 +419,158 @@ def __init__(
...
         self.update_lookuptable()
 
-    @property
-    def pol(self):
-        return self._pol
-
-    @pol.setter
-    def pol(self, pol: str):
+    def set_pol_setpoint(self, pol: Pol) -> None:
         # This set the polarisation but does not actually move hardware.
```
olliesilvester, line 472:
```suggestion
        # This sets the polarisation but does not actually move hardware.
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -348,129 +419,158 @@ def __init__(
...
-            raise ValueError(
-                f"Polarisation {pol} is not available:"
-                + f"/n Polarisations available:  {self._available_pol}"
+        self._polarisation_setpoint_set(pol)
+
+    async def _set_pol(
+        self,
+        value: Pol,
+    ) -> None:
+        # This change the pol setpoint and then change polariastion via set energy.
```
olliesilvester, line 479:
```suggestion
        # This changes the pol setpoint and then changes polariastion via set energy.
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -348,129 +419,158 @@ def __init__(
...
+        btm_outer: float,
+        gap: float,
+    ) -> Pol:
+        LOGGER.info(
+            f"Reading polarisation setpoint from hardware: "
+            f"top_outer={top_outer}, top_inner={top_inner}, "
+            f"btm_inner={btm_inner}, btm_outer={btm_outer}, gap={gap}."
+        )
+
+        if pol != Pol.LH3:
```
olliesilvester, line 503: Would be useful to have a comment here to explain why we handle the `Pol.LH3` case differently. Also, would it be useful to error/warn if `pol==None`?

src/dodal/devices/i10/i10_apple2.py
```diff
@@ -8,23 +8,27 @@
...
     StandardReadable,
     StandardReadableFormat,
     soft_signal_r_and_setter,
     soft_signal_rw,
 )
 
-from dodal.devices.apple2_undulator import (
+from dodal.log import LOGGER
+
+# from dodal.beamlines.i10 import pgm
```
olliesilvester, line 21: Should: Remove this line

src/dodal/devices/apple2_undulator.py
```diff
@@ -513,62 +615,49 @@ async def determinePhaseFromHardware(self) -> tuple[str | None, float]:
             motor_position_equal(x, 0.0)
             for x in [top_outer, top_inner, btm_inner, btm_outer]
         ):
-            # Linear Horizontal
-            polarisation = "lh"
-            phase = 0.0
-            return polarisation, phase
+            LOGGER.info("Determined polarisation: LH (Linear Horizontal).")
+            return Pol("lh"), 0.0
```
olliesilvester, line 619: Should: use `Pol.LH3` instead here

tests/devices/i10/test_i10Apple2.py
```diff
@@ -296,81 +308,137 @@ def capture_emitted(name, doc):
...
+    set_mock_value(mock_id_pol.id_ref().phase.btm_inner.user_readback, btm_inner)
+    set_mock_value(mock_id_pol.id_ref().phase.btm_outer.user_readback, btm_outer)
+    assert (await mock_id_pol.read())["mock_id-polarisation"]["value"] == pol
+
+
 async def test_linear_arbitrary_pol_fail(
     mock_linear_arbitrary_angle: LinearArbitraryAngle,
 ):
-    mock_linear_arbitrary_angle.id_ref().pol = "lh"
+    # set_mock_value(
```
olliesilvester, line 434: Should: remove this comment

src/dodal/devices/apple2_undulator.py
```diff
@@ -321,23 +335,80 @@ async def get_timeout(self) -> float:
...
         """
         return await estimate_motor_timeout(
-            self.jaw_phase.user_setpoint_demand_readback,
             self.jaw_phase.user_setpoint_readback,
+            self.jaw_phase.user_readback,
             self.jaw_phase.velocity,
         )
 
 
 class Apple2(StandardReadable, Movable):
```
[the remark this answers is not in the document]
Relm-Arrowny: Apple2 is an abstract class as it requires beamline lookup table/energy to function, hence the _set here for moving the motors around.  I have added ABC to make this clearer, I also updated docstring to reflect set has to be implemented in subclass.

olliesilvester (changes requested):
Thanks. This is a big change to a very complicated device, in general the code looks nice.

I think some of my comments may be easier to respond to in person, so let me know if you want to chat about any of them.

src/dodal/devices/apple2_undulator.py
```diff
@@ -321,23 +335,74 @@ async def get_timeout(self) -> float:
...
 
     A pair of look up tables are needed to provide the conversion between motor position
-     and energy.
-    This conversion (update_lookuptable) and the way the id move (set) are two abstract
-     methods that are beamline specific and need to be implemented.
+    and energy.
+
+    Attributes
+    ----------
+    gap : UndulatorGap
```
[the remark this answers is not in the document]
Relm-Arrowny: This was raised before, it is motor with a few different in behaviour see [here](https://github.com/DiamondLightSource/dodal/pull/1002#pullrequestreview-2572280357)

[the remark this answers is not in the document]
olliesilvester: Ah okay thanks. Do you think we could subclass motor (within dodal) to give it the behaviour you want? Then you wouldn't need to create all these signals + manually calculate the timeout. Could create an issue for this and put a link to it in the comments under `UndulatorGap` and `UndulatorPhaseMotor`.

If you don't think this is worth doing then we should at least put a comment explaining why `Motor` hasn't been used

src/dodal/devices/apple2_undulator.py
```diff
@@ -348,129 +413,156 @@ def __init__(
...
+        # This changes the pol setpoint and then changes polariastion via set energy.
+        self.set_pol_setpoint(value)
+        await self._set_energy(await self.energy.get_value())
+
+    @abc.abstractmethod
+    async def _set_energy(self, value: float) -> None:
+        """This change the position of all the motors for a given energy and
+        polarisation_setpoint"""
+
+    def _read_pol(
```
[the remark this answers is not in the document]
olliesilvester: What about the case where
- polarisation is physically at `Pol.PC` (or anything which isn't LH3)
- we set the setpoint to LH3 (but don't actually do a set) 
- we read polarisation? 

Won't it return `Pol.LH3` even though it's still at `Pol.PC`, since the pol setpoint is what this function uses as its `pol` parameter?

src/dodal/devices/apple2_undulator.py
```diff
@@ -348,129 +410,191 @@ def __init__(
...
-            self.phase().set_move.set(value=1, timeout=timeout),
-        )
-        await wait_for_value(
-            self.gap().gate, UndulatorGateStatus.CLOSE, timeout=timeout
+            self.gap.set_move.set(value=1, wait=False, timeout=timeout),
+            self.phase.set_move.set(value=1, wait=False, timeout=timeout),
         )
-        self._energy_set(energy)  # Update energy for after move for readback.
+        await wait_for_value(self.gap.gate, UndulatorGateStatus.CLOSE, timeout=timeout)
+        self._energy_setpoint(energy)  # Update energy after move for readback.
```
olliesilvester, line 546: I think I mentioned in a previous comment, but the name `__energy_setpoint` is a bit confusing to me, since this function is actually setting the energy RBV. `_set_energy_rbv` would make more sense to me

src/dodal/devices/apple2_undulator.py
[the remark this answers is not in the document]
olliesilvester: I don't understand the device well enough to comment on the automatic switch, but otherwise that sounds good I think, as long as users don't get confused when their readback says LH instead of LH3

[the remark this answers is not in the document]
Relm-Arrowny: After thinking about it more, I realized that internal switching is not the right great. It could cause the system to jump from lh to lh3 in the middle of an energy scan. The large intensity mismatch between the harmonics would make the data useless.
Instead, I have now isolated the LH3 readback logic. This ensures the read value behaves correctly without the risk of an automatic switch, which is a much safer solution.

olliesilvester (changes requested):
Sorry, just a few more comments. I think it's very almost there now

src/dodal/devices/apple2_undulator.py
```diff
@@ -348,129 +410,195 @@ def __init__(
...
+            f"top_outer={top_outer}, top_inner={top_inner}, "
+            f"btm_inner={btm_inner}, btm_outer={btm_outer}, gap={gap}."
+        )
+
+        read_pol, _ = self.determine_phase_from_hardware(
+            top_outer, top_inner, btm_inner, btm_outer, gap
+        )
+        # LH3 is indistinguishable from LH see determine_phase_from_hardware's docString
+        # so we return LH3 if the setpoint is LH3 and the readback is LH.
+        if pol == Pol.LH3 and read_pol == Pol.LH:
```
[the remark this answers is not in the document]
Relm-Arrowny: In terms of polarisation, it doesn't really matter as they are identical. While this isn't ideal, it's probably unavoidable since we can't distinguish between them in hardware.

The other option would be to always return lh, which is also fine. However, the current approach provides the user with extra information about when lh3 is being used. 

I have also make the function private as it really should not be use externally, this hopeful stop the setpoint being change without moving hardware.

tests/devices/i10/test_i10Apple2.py
```diff
@@ -296,81 +309,136 @@ def capture_emitted(name, doc):
...
+
+    assert (await mock_id_pol.read())["mock_id-polarisation"]["value"] == pol
+
+
+@pytest.mark.parametrize(
+    "pol,energy, top_outer, top_inner, btm_inner,btm_outer",
+    [
+        ("lh3", 500, 0.0, 0.0, 0.0, 0.0),
+    ],
+)
```
olliesilvester, line 415: You've used `parametrize` but only got one set of parameters - did you mean to make a few more?

olliesilvester (approved):
It's possible I've missed something as I don't have fresh eyes on this code anymore, but I think it looks good now, thanks!

I have a nit on the tests, but I suggest you don't do them in this PR so we can get this one merged: IMO it's better to test the features of your base class in the test file of that base class, rather than the test file of an implementation of that class. For example, `Apple2.determine_phase_from_hardware` is tested in `test_i10Apple2.py` rather than `test_apple2_undulator.py`

tests/devices/i10/test_i10Apple2.py
olliesilvester: Sorry, yes you're right, I was accidentally looking at an old commit when I typed that

PR #1035, opened 2025-02-04
Fix getting expected images from a gridscan

DominicOram:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/777

1. Confirm modified test fails on main but passes here

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/fast_grid_scan.py
+++ b/src/dodal/devices/fast_grid_scan.py
@@ -1,3 +1,4 @@
+import asyncio
 from abc import ABC, abstractmethod
 from typing import Generic, TypeVar
 
@@ -230,9 +231,13 @@ class FastGridScanCommon(StandardReadable, Flyable, ABC, Generic[ParamType]):
         super().__init__(name)
 
     async def _calculate_expected_images(self):
-        x = await self.x_steps.get_value()
-        y = await self.y_steps.get_value()
-        z = await self.z_steps.get_value()
+        x, y, z = asyncio.gather(
+            [
+                await self.x_steps.get_value(),
+                await self.y_steps.get_value(),
+                await self.z_steps.get_value(),
+            ]
+        )
         LOGGER.info(f"Reading num of images found {x, y, z} images in each axis")
         first_grid = x * y
         second_grid = x * z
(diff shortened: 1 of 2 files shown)


src/dodal/devices/fast_grid_scan.py
```diff
@@ -243,6 +229,15 @@ def __init__(self, prefix: str, smargon_prefix: str, name: str = "") -> None:
         }
         super().__init__(name)
 
+    async def _calculate_expected_images(self):
+        x = await self.x_steps.get_value()
+        y = await self.y_steps.get_value()
+        z = await self.z_steps.get_value()
```
callumforrester, line 235: Could: Can this be parallelized? If not then may be worth putting a comment explaining why to avoid tripping someone up in future.

```suggestion
        x, y, z = await asyncio.gather(self.x_steps.get_value(), self.y_steps.get_value(), self.z_steps.get_value()
```

callumforrester (approved):
Looks good to me, I was just looking for something like `create_hardware_backed_soft_signal`

DominicOram:
> Looks good to me, I was just looking for something like `create_hardware_backed_soft_signal`

There's some documentation about it [here](https://diamondlightsource.github.io/dodal/main/reference/device-standards.html#holding-state). I think I might add the words derived signal to that though as it's not that searchable. Worth noting some of the caveats with it from https://github.com/bluesky/ophyd-async/issues/525. Namely it doesn't work in the case of `subscribe`. It would be good to get a cleaner solution into `ophyd-async`

src/dodal/devices/fast_grid_scan.py
DominicOram: Yes, it can be parallelised. I have tended not to bother with parallel reads because they're normally very short anyway but I will add it in.

callumforrester: Fair enough, my concern is mainly that someone in the future might wonder why it wasn't done and go down an archaeological rabbit hole.

callumforrester:
Is the fact that's read-only also a consequence of 525?

Issue #1031, opened 2025-01-28
Document when to make a Device class extend Movable


DiamondJoseph:
Increasingly our new devices are becoming `StandardReadable`s with a structure of signals underneath them- this is good, easy to understand and a vote in favour of the Standardness  of StandardReadable.
Sometimes, these devices have a few settable fields, and it is tempting to define a set method on the type to allow setting them in one step.

e.g. [Backlight](https://github.com/DiamondLightSource/dodal/blob/5c82bbd70cca91e2e3086eafb8ec9dd510705be6/src/dodal/devices/backlight.py#L32), [CurrentAmp](https://github.com/DiamondLightSource/dodal/blob/5c82bbd70cca91e2e3086eafb8ec9dd510705be6/src/dodal/devices/current_amplifiers/sr570.py#L164), [BimorphMirror](https://github.com/DiamondLightSource/dodal/blob/5c82bbd70cca91e2e3086eafb8ec9dd510705be6/src/dodal/devices/bimorph_mirror.py#L106)

**I** think it is useful to document when defining a set method is a value add, and when it is added boilerplate that makes using the device or writing plans for the device more complex without benefit.

I think in the cases where the set method is constructing a dict or dataclass mapping directly to the underlying signals, using this set method requires constructing that dict or dataclass in the plan. If this does not lead to a speedup of the set (e.g. the previously believed behaviour of the Bimorph, that all signals can be set, then the move triggered), **I** do not see the value of the set method, but the method must be documented, tested and maintained.

## Acceptance Criteria
* [] A decision has been documented- perhaps in an ADR?

DiamondJoseph:
@DominicOram @Relm-Arrowny, @callumforrester says this is something you're interested in.

callumforrester:
Definitely think we should document it, possibly upstream in ophyd-async, paging @coretl

coretl:
I will gladly accept docs.

My opinion is that we should only add a `set` method if it is obvious what it means to `set` the Device. I can see that it makes sense to `abs_set(Backlight, BacklightPosition.OUT)`, but I'm not so sure about `abs_set(SR570, SR570GainToCurrentTable.SEN_1)`...

Agree with @DiamondJoseph's position on using primitives rather than dataclass for the set method wherever possible.

stan-dot:
for a 4d Movable like this:
```python
class TablePosition(BaseModel):
    x: float
    y: float
    z: float | None = None
    theta: float | None = None

```

[...]

PR #1037, opened 2025-02-04
1034 pressure cell openseq type

### LINKED ISSUE #1034: Pressure jump cell incorrect OPENSEQ type
The pressure jump cell fails to connect because OPENSEQ PVs return int.

```
high_pressure_xray_cell: NotConnected:
    all_valves_control: NotConnected:
        fast_valve_control: NotConnected:
            5: NotConnected:
                open: TypeError: BL38P-EA-HPXC-01:V5:OPENSEQ with inferred datatype int cannot be coerced to ValveOpenSeqRequest
            6: NotConnected:
                open: TypeError: BL38P-EA-HPXC-01:V6:OPENSEQ with inferred datatype int cannot be coerced to ValveOpenSeqRequest
        valve_control: NotConnected:
            1: NotConnected:
                open: TypeError: BL38P-EA-HPXC-01:V1:OPENSEQ with inferred datatype int cannot be coerced to ValveOpenSeqRequest
            3: NotConnected:
                open: TypeError: BL38P-EA-HPXC-01:V3:OPENSEQ with inferred datatype int cannot be coerced to ValveOpenSeqRequest
    pressure_transducers: NotConnected:
        3: NotConnected:
            slow_beckhoff_voltage_readout: NotConnected: ca://BL38P-EA-ADC-03:CH1
ppump: NotConnected:
    direction: TypeError: Differing datatypes: BL38P-EA-PUMP-01:INFO:DIR has 0, BL38P-EA-PUMP-01:SET:DIR has 3
    state: TypeError: Differing choices: BL38P-EA-PUMP-01:INFO:RUN has ('STOPped', 'STARTed'), BL38P-EA-PUMP-01:SET:RUN has ('STOP', 'START')
```

## Acceptance Criteria
* The pressure jump cell in P38 connects

barnettwilliam:
Fixes #1034

1. `dodal connect p38`
2. Confirm the pressure cell is in the list of connected devices

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/pressure_jump_cell.py
+++ b/src/dodal/devices/pressure_jump_cell.py
@@ -219,21 +219,22 @@ class PressureTransducer(StandardReadable):
         self,
         prefix: str,
         cell_prefix: str,
-        number: int,
+        transducer_number: int,
+        ethercat_channel_number: int,
         name: str = "",
         full_different_prefix_adc: str = "",
     ) -> None:
         final_prefix = f"{prefix}{cell_prefix}"
         with self.add_children_as_readables():
             self.omron_pressure = epics_signal_r(
-                float, f"{final_prefix}PP{number}:PRES"
+                float, f"{final_prefix}PP{transducer_number}:PRES"
             )
-            self.omron_voltage = epics_signal_r(float, f"{final_prefix}PP{number}:RAW")
+            self.omron_voltage = epics_signal_r(float, f"{final_prefix}PP{transducer_number}:RAW")
             self.beckhoff_pressure = epics_signal_r(
-                float, f"{final_prefix}STATP{number}:MeanValue_RBV"
+                float, f"{final_prefix}STATP{transducer_number}:MeanValue_RBV"
             )
             self.slow_beckhoff_voltage_readout = epics_signal_r(
-                float, f"{full_different_prefix_adc}CH1"
+                float, f"{full_different_prefix_adc}CH{ethercat_channel_number}"
             )
 
         super().__init__(name)
@@ -284,13 +285,27 @@ class PressureJumpCell(StandardReadable):
         with self.add_children_as_readables():
             self.pressure_transducers: DeviceVector[PressureTransducer] = DeviceVector(
                 {
-                    i: PressureTransducer(
+                    1: PressureTransducer(
                         prefix=prefix,
-                        number=i,
                         cell_prefix=cell_prefix,
-                        full_different_prefix_adc=f"{prefix}{adc_prefix}-0{i}:",
-                    )
-                    for i in [1, 2, 3]
+                        transducer_number=1,
+                        full_different_prefix_adc=f"{prefix}{adc_prefix}-02:",
+                        ethercat_channel_number=1,
+                    ),
+                    2: PressureTransducer(
+                        prefix=prefix,
+                        cell_prefix=cell_prefix,
+                        transducer_number=2,
+                        full_different_prefix_adc=f"{prefix}{adc_prefix}-01:",
+                        ethercat_channel_number=2,
+                    ),
+                    3: PressureTransducer(
+                        prefix=prefix,
+                        cell_prefix=cell_prefix,
+                        transducer_number=3,
+                        full_different_prefix_adc=f"{prefix}{adc_prefix}-01:",
+                        ethercat_channel_number=1,
+                    ),
                 }
             )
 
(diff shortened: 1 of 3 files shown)


src/dodal/devices/pressure_jump_cell.py
```diff
@@ -155,17 +155,17 @@ class ValveControl(StandardReadable):
     def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
             self.close = epics_signal_rw(ValveControlRequest, prefix + ":CON")
-            self.open = epics_signal_rw(ValveOpenSeqRequest, prefix + ":OPENSEQ")
+            self.open = epics_signal_rw(int, prefix + ":OPENSEQ")
 
         super().__init__(name)
 
-    def set(self, value: ValveControlRequest | ValveOpenSeqRequest) -> AsyncStatus:
```
DominicOram, line 162: Must: I don't _think_ you can do this. You need to stick to the function definition of `Movable` in case we're calling this with a keyword arg e.g `device.set(value=10)`. The reason the IDE isn't picking this up is that you also need `ValveControl` to inherit from `Moveable`. Same with `FastValveControl`

DominicOram (changes requested):
Looks good, just one comment

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Thanks Dominic, fixed in e4440fa9684d1bc4590bbd94c79d9645d61da963

DominicOram (approved):
Thank you! I think this is now good but as a side-note...

The interface for setting `ValveControl` and `FastValveControl` are very odd to me as a naive user. From reading the logic of `set` it looks like if you want to open the valve you need to do `yield from bps.abs_set(valve, ValveOpenSeqRequest.OPEN_SEQ)` and if you want to close it you have to do `yield from bps.abs_set(valve, ValveControlRequest.CLOSE)`. This leads to the question of what is the behaviour if you do `yield from bps.abs_set(valve, ValveControlRequest.OPEN)`? Is this something that should be allowed?

My suggestion would be to hide all this from whoever is using the device and give them:
```python
class ValveControl(Enum):
    OPEN = "Open"
    CLOSE = "Close"

...

def set(self, value: ValveControl) -> AsyncStatus:
    if value == ValveControl.CLOSE:
        return self.close.set(ValveControlRequest.CLOSE)
    elif value == ValveControl.OPEN:
        return self.open.set(ValveOpenSeqRequest.OPEN_SEQ)
```

This makes everything much simpler for the person writing the plan that now just needs to do worry about `ValveControl`, which has simple, obvious options rather than think about both `ValveOpenSeqRequest` and `ValveControlRequest`, which have lots of options.

Is my understanding correct? If it is I'll make a new issue to do the above

barnettwilliam:
You are correct about the `OPEN_SEQ` vs `OPEN` valve interface logic.

It's the intention of the `set_valve()` method of a `PressureJumpCell`'s `all_valves_control` attribute (`AllValvesControl`) to handle the opening/closing by setting `OPEN_SEQ` e.g. `cell.all_valves_control.set_valve( valve-number, request)` from a `FastValveControlRequest` or `ValveControlRequest` to simplify things for the user.

The lower level interface to valves is available via the `AllValvesControl`'s  `fast_valve_control` and `valve_control` attributes e.g. 
`cell.all_valves_control.valve_control[valve-number].close(request)` or `cell.all_valves_control.valve_control[valve-number].open(request)` if finer control was wanted for some reason.

DominicOram:
> It's the intention of the set_valve() method of a PressureJumpCell's all_valves_control attribute (AllValvesControl) to handle the opening/closing by setting OPEN_SEQ e.g. cell.all_valves_control.set_valve( valve-number, request) from a FastValveControlRequest or ValveControlRequest to simplify things for the user.

The user does not have access to `set_value` only to the bluesky verb of `set`. 

I also think the idea of being able to do `yield from bps.set(cell.all_valves_control, AllValvesControlState(valve_1=ValveControlRequest.OPEN, valve_5=FastValveControlRequest.CLOSE)` may need changing based on https://github.com/DiamondLightSource/dodal/issues/1031. I will add a note there.

I will make an issue. Are they being used somewhere in https://github.com/DiamondLightSource/i10-bluesky?

barnettwilliam:
They are not being used as far as I'm aware this for the pressure jump test experiment for I22 that I'm testing at P38.

PR #1054, opened 2025-02-10
Create plans to arm the fast-cs eiger

### LINKED ISSUE #1053: Create plans to arm the fast-cs eiger
The next step in completing ophyd-async's fast-cs Eiger is to benchmark the time taken to arm the Eiger in a two step process:
1. The "load" plan. For this issue, this plan can just set the relevant PV's using `bps.abs_set`, rather than using ophyd async load. This plan should set the mx settings (see currently dodal eiger arming), set the ROI mode, and set the energy threshold
- Wait for the above plan to be finished
2. The `prepare` plan. This should call the fast-cs eigers' `prepare` method, which we will also need to modify slightly for this issue: we need to add the datatype setting (see https://github.com/bluesky/ophyd-async/issues/529); we also need to add  `wait_for_stale_params`.

## Acceptance Criteria
1. Ophyd-Async fast-cs eiger prepare modified to include datatype and `wait_for_stale_params`
2. Plan written to: load, wait for load, do the new prepare
3. Plan tested on a real Eiger and benchmark how long each bit takes

shihab-dls:
Resolves #1053 
This PR is not meant to be merged; kept as a draft to monitor changes to the dev branch.

### DIFF
--- a/src/dodal/devices/util/load_eiger_metadata_plan.py
+++ b/src/dodal/devices/util/load_eiger_metadata_plan.py
@@ -1,6 +1,7 @@
 import time
 
 import bluesky.plan_stubs as bps
+from bluesky.run_engine import RunEngine
 from mx_bluesky.common.utils.log import LOGGER
 from ophyd_async.epics.eiger import EigerDetector, EigerTriggerInfo
 
@@ -116,3 +117,7 @@ def set_odin_pvs(
 
     if wait:
         yield from bps.wait(group)
+
+
+if __name__ == "__main__":
+    RE = RunEngine()
--- a/src/dodal/devices/util/load_eiger_metadata_plan.py
+++ b/src/dodal/devices/util/load_eiger_metadata_plan.py
@@ -10,6 +10,7 @@ def load_metadata(
     detector_params: DetectorParams,
 ):
     assert detector_params.expected_energy_ev
+    yield from bps.stage(eiger)
     yield from set_odin_pvs(eiger, detector_params, wait=True)
     yield from change_roi_mode(eiger, enable, detector_params)
     yield from bps.abs_set(eiger.odin.num_frames_chunks, 1)
--- a/src/dodal/devices/util/load_eiger_metadata_plan.py
+++ b/src/dodal/devices/util/load_eiger_metadata_plan.py
@@ -72,3 +72,6 @@ def set_odin_pvs(
     yield from bps.abs_set(
         eiger.odin.file_name, detector_params.full_filename, group=group
     )
+
+    if wait:
+        yield from bps.wait(group)
--- a/src/dodal/devices/util/load_eiger_metadata_plan.py
+++ b/src/dodal/devices/util/load_eiger_metadata_plan.py
@@ -2,10 +2,25 @@ import time
 
 import bluesky.plan_stubs as bps
 from bluesky.run_engine import RunEngine
-from mx_bluesky.common.utils.log import LOGGER
 from ophyd_async.epics.eiger import EigerDetector, EigerTriggerInfo
 
+from dodal.beamlines.i03 import fastcs_eiger
 from dodal.devices.detector import DetectorParams
+from dodal.log import LOGGER, do_default_logging_setup
+
+params = DetectorParams(
+    expected_energy_ev=12700,
+    exposure_time=0.004,
+    directory="/dls/i03/data/2025/cm40607-2/test_new_eiger",
+    prefix="",
+    detector_distance=255,
+    omega_start=0,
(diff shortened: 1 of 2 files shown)


src/dodal/devices/util/load_eiger_metadata_plan.py
```diff
@@ -0,0 +1,44 @@
+import bluesky.plan_stubs as bps
+from ophyd_async.epics.eiger import EigerDetector
+
+from dodal.devices.detector import DetectorParams
+
+
+def load_metadata(
+    eiger: EigerDetector, energy: float, enable: bool, detector_params: DetectorParams
+):
+    def _inner_plan():
```
DominicOram, line 10: Should: Why do you need an inner plan here?

src/dodal/devices/util/load_eiger_metadata_plan.py
```diff
@@ -0,0 +1,44 @@
...
+    beam_x_pixels, beam_y_pixels = detector_params.get_beam_position_pixels(
+        detector_params.detector_distance
+    )
+
+    yield from bps.abs_set(eiger.drv.beam_centre_x, beam_x_pixels)
+    yield from bps.abs_set(eiger.drv.beam_centre_y, beam_y_pixels)
+    yield from bps.abs_set(eiger.drv.det_distance, detector_params.detector_distance)
+
+    yield from bps.abs_set(eiger.drv.omega_start, detector_params.omega_start)
+    yield from bps.abs_set(eiger.drv.omega_increment, detector_params.omega_increment)
```
DominicOram, line 44: Should: We probably want to group these things as it makes it more async (even if Tom/Gary assure us it's instantaneous). Then we will need to wait on them at some point so we should add that optionally too e.g.
```suggestion
def set_mx_settings_pvs(eiger: EigerDetector, detector_params: DetectorParams, group="mx_settings", wait=True):
    beam_x_pixels, beam_y_pixels = detector_params.get_beam_position_pixels(
        detector_params.detector_distance
    )

    yield from bps.abs_set(eiger.drv.beam_centre_x, beam_x_pixels, group)
    yield from bps.abs_set(eiger.drv.beam_centre_y, beam_y_pixels, group)
    yield from bps.abs_set(eiger.drv.det_distance, detector_params.detector_distance, group)

    yield from bps.abs_set(eiger.drv.omega_start, detector_params.omega_start, group)
    yield from bps.abs_set(eiger.drv.omega_increment, detector_params.omega_increment, group)
    if wait:
        yield from bps.wait(group)
```
This is a pattern we've used in a few places e.g. https://github.com/DiamondLightSource/mx-bluesky/blob/11c8a31f05dfc8421b9ebf205cc49598cd8806be/src/mx_bluesky/beamlines/i24/serial/setup_beamline/setup_beamline.py#L58. It allows the caller to have some flexibility in timings.

We should do the same for other plans in this file too

DominicOram (commented):
Thanks! Couple of comments, I think there are some additional things we need to set too:
* `num_frames_chunks`
* The file name/directory e.g. everything in `set_odin_pvs`
* `dev_shm`

src/dodal/devices/util/load_eiger_metadata_plan.py
shihab-dls: I think it was the first thing I wrote for structure and left in for no reason; removing as definitely not needed :0

shihab-dls:
Currently, this plan does the following arming chain:
`change_roi_mode`
`set_odin_number_of_frame_chunks`
`set_mx_settings_pvs`
Then runs Eiger prepare(), which does:
`set_detector_threhold`

The following are missing:
`change_dev_shm`
`odin.stop`
`set_odin_pvs`
`set_num_triggers_and_captures`

`change_dev_shm` is not included due to limited development/dissemination on the fastcs-eiger/fastcs-odin front, but should not significantly affect timings, and thus does not block testing the arming chain.

I am operating under the assumption `set_num_triggers_and_captures` is addressed in `StandardDetector` during `prepare()`, but will look into this more to check if this is ok.

`odin.stop` and `set_odin_pvs` should be looked at, I believe.

shihab-dls:
Following the last commit, the plan now also includes:
`set_odin_pvs`

However, this does not set `odin.meta.file_name` and `odin.file_writer.id`. This is mainly because of the following comments:
`# file_name should not be set. Set the filewriter file_name and this will be updated in EPICS` in `OdinMetaListener`
`# id should not be set. Set the filewriter file_name and this will be updated in EPICS` in `OdinFileWriter`
But, this should be checked

`odin.stop` is also included now, but is done via `bps.stage(eiger)` as this stops odin file writing and clears trigger info

PR #1058, opened 2025-02-17
1055 tidy up allvalvescontrol

### LINKED ISSUE #1055: Tidy up `AllValvesControl`
Currently, there is some complication in `AllValvesControl` around `set_valve` and types. We should simplify the interfaces along the lines of https://github.com/DiamondLightSource/dodal/pull/1037#pullrequestreview-2605832091 and clean the interface up in line with https://github.com/DiamondLightSource/dodal/issues/1031

## Acceptance Criteria
* The interface for valves in the pressure cell is simplified
* Any plans that use this are updated

barnettwilliam:
Fixes #1055 

1. Run the pressure cell tests
2. Confirm the pressure cell tests pass

## Checks for reviewer

DiamondLightSource/dodal#1055


barnettwilliam:
As the :OPENSEQ pv should always be used ValveOpenSeqRequest is now handled by ValveControl so users only have to be aware of ValveControlRequests similar to the suggestion in https://github.com/DiamondLightSource/dodal/pull/1037#pullrequestreview-2605832091 . There still needs to be a distinction between fast and slow valves because fast valves have additional functionality.

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -68,12 +69,53 @@ class FastValveState(StrictEnum):
...
+
+    def set(self, value: FastValveControlRequest) -> AsyncStatus:
+        set_status = None
+
+        if value == FastValveControlRequest.OPEN:
+            set_status = self._set_open_seq()
+        else:
+            set_status = self.close.set(value)
+
+        return set_status
```
DominicOram, line 118: Should: So I think this can be reduced down quite a bit using generics, see something like https://github.com/DiamondLightSource/dodal/commit/63db94a6a9793b32a334d6bacf1183bffcd2cdcf

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -125,70 +167,23 @@ async def set_valve(
...
-                await self.fast_valve_control[valve].set(ValveOpenSeqRequest.OPEN_SEQ)
-                await asyncio.sleep(OPENSEQ_PULSE_LENGTH)
-                await self.fast_valve_control[valve].set(ValveOpenSeqRequest.INACTIVE)
-            else:
-                await self.fast_valve_control[valve].set(value)
+            await self.fast_valve_control[valve].set(value)
 
     @AsyncStatus.wrap
-    async def set(self, value: AllValvesControlState):
+    async def set(
```
DominicOram, line 176: Should: As I said in the issue, I think we decided in https://github.com/DiamondLightSource/dodal/issues/1031 (though it still needs to be documented) that we shouldn't allow setting using collections like this. It doesn't give you much in the end e.g. rather than:
```python
# Set new values
all_valve_request = {
    1: ValveControlRequest.CLOSE,
    3: ValveControlRequest.CLOSE,
    5: FastValveControlRequest.CLOSE,
    6: FastValveControlRequest.CLOSE,
}

await cell.all_valves_control.set(all_valve_request)
```
it's just as easy to do:
```python
# Set new values
await cell.all_valves_control[1].set(ValveControlRequest.CLOSE)
await cell.all_valves_control[3].set(ValveControlRequest.CLOSE)
await cell.all_valves_control[5].set(FastValveControlRequest.CLOSE)
await cell.all_valves_control[6].set(FastValveControlRequest.CLOSE)
```
and you get the benefit of type checking

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -125,70 +167,23 @@ async def set_valve(
         value: ValveControlRequest | FastValveControlRequest,
     ):
         if valve in self.slow_valves and (isinstance(value, ValveControlRequest)):
```
DominicOram, line 169: Should: I'm not sure what `set_value` gives you. All it means is that you can do `await cell.all_valves_control.set_valve(1, ValveControlRequest.CLOSE)` rather than `await cell.all_valves_control[1].set(ValveControlRequest.CLOSE)`? Given that you now have to support both interfaces it seems better to just use the standard one with `[1]`? Especially given that you are not able to do `set_value` in a plan so it will very rarely be used anyway

tests/devices/unit_tests/test_pressure_jump_cell.py
```diff
@@ -358,3 +364,114 @@ async def test_reading_pjumpcell_includes_read_fields(
...
+    )
+    set_mock_value(
+        cell.all_valves_control.valve_control[3].open,
+        ValveOpenSeqRequest.INACTIVE.value,
+    )
+
+    set_mock_value(
+        cell.all_valves_control.fast_valve_control[5].close,
+        FastValveControlRequest.OPEN,
+    )
```
DominicOram, line 391: Should: I still think it's confusing here to be able to set `OPEN` on a signal called `close`. I would suggest renaming `close` to something more generic like `control`

DominicOram (changes requested):
Thanks! Some suggestions in comments.

Additionally I feel like there must be a clean way of allowing the caller of the device to change valves without caring if they're fast or slow (I assume that a user may want to e.g open/close them all without knowing which are fast/slow?). I have tried to do this in https://github.com/DiamondLightSource/dodal/commit/d95ae6bde3c144bb40c2a392755b29adf0aa289a but to get the type system to be happy with the enums and still allow the fast valves to be set with the fine control I had to do a horrible thing with having two signals to one PV. I wonder if another @DiamondLightSource/bluesky-reviewers can think of a cleaner way?

tests/devices/unit_tests/test_pressure_jump_cell.py
barnettwilliam: Renamed in 51fedb11b5c188d811b3281f1be4fbe8997ebf6d

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Removed in 8ddd206f1a52391110f7a37b7cf153e1588e2b1f

src/dodal/devices/pressure_jump_cell.py
barnettwilliam: Revised in 2587e7d1abcbd3ab0f89e5be0a6aeb28473e77f4 d6576f2d07fc7ff240f60ae200970e513d5b2596

Thank you for the suggestion Dominic it is an improvement being able to iterate over all valves with the same request. I also included `slow_valve_control` as part of `AllValvesControl` so there is effectively the choice of iterating over slow, fast, or all the valves.

Rather than have multiple signals on a PV I opted for `FastValveControl.set()` to accept `ValveControlRequest` and convert  it as required 

I didn't include `FastValveControl.fine_control()` as the request can be made through `AllValvesControl.fast_valve_control[n]`

olliesilvester (changes requested):
Nice, this looks a lot better to me than what's currently in main. Most of my comments are optional suggestions.

I haven't read the tests thoroughly yet, but please re-request a review from me when you're happy and I will try to be quick with a re-review

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -60,23 +60,84 @@ class ValveState(StrictEnum):
...
-    valve_1: ValveControlRequest | None = None
-    valve_3: ValveControlRequest | None = None
-    valve_5: FastValveControlRequest | None = None
-    valve_6: FastValveControlRequest | None = None
+class ValveControlBase(StandardReadable, Movable):
+    open: SignalRW[int]
+    control: SignalRW[ValveControlRequest | FastValveControlRequest]
+
+    @AsyncStatus.wrap
+    async def _set_open_seq(self):
```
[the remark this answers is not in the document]
barnettwilliam: Yes this may not be required I've created a separate issue https://github.com/DiamondLightSource/dodal/issues/1349 to check and remove this if it is not

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -60,23 +60,84 @@ class ValveState(StrictEnum):
...
 
-@dataclass
-class AllValvesControlState:
-    valve_1: ValveControlRequest | None = None
-    valve_3: ValveControlRequest | None = None
-    valve_5: FastValveControlRequest | None = None
-    valve_6: FastValveControlRequest | None = None
+class ValveControlBase(StandardReadable, Movable):
+    open: SignalRW[int]
+    control: SignalRW[ValveControlRequest | FastValveControlRequest]
```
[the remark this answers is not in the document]
barnettwilliam: Added in fdc6532b0cacdd0a3eaddb4cb7f9fb8c57ecc0be

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -88,110 +149,43 @@ def __init__(
...
         prefix: str,
         name: str = "",
-        fast_valves: tuple[int, ...] = (5, 6),
-        slow_valves: tuple[int, ...] = (1, 3),
+        fast_valves_numbers: tuple[int, ...] = (5, 6),
+        slow_valves_numbers: tuple[int, ...] = (1, 3),
     ) -> None:
-        self.fast_valves = fast_valves
-        self.slow_valves = slow_valves
+        self._fast_valves_numbers = fast_valves_numbers
```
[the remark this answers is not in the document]
barnettwilliam: Yes a valve is either a slow valve or a fast valve based on its hardware

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -60,23 +60,84 @@ class ValveState(StrictEnum):
...
     NONE = "Unused"
 
 
-@dataclass
-class AllValvesControlState:
-    valve_1: ValveControlRequest | None = None
-    valve_3: ValveControlRequest | None = None
-    valve_5: FastValveControlRequest | None = None
-    valve_6: FastValveControlRequest | None = None
+class ValveControlBase(StandardReadable, Movable):
```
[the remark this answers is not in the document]
olliesilvester: Yeah, you're right. I think I out some outdated linting options during my previous review

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -60,26 +60,91 @@ class ValveState(StrictEnum):
...
+                f"Class {ValveControlBase} requires signals open and control to be defined"
+            )
+
+        super().__init__(name)
+
+    @AsyncStatus.wrap
+    async def _set_open_seq(self):
+        await self.open.set(ValveOpenSeqRequest.OPEN_SEQ.value)
+        await asyncio.sleep(OPENSEQ_PULSE_LENGTH)
+        await self.open.set(ValveOpenSeqRequest.INACTIVE.value)
```
DiamondJoseph, line 87: If you make the base class an ABC, you can then make use of abstract methods that subclasses have to fill in:

```suggestion
    @property
    @abstractmethod
    def _open(self) -> SignalRW[int]:
        pass

    @AsyncStatus.wrap
    async def _set_open_seq(self):
        await self._open.set(ValveOpenSeqRequest.OPEN_SEQ.value)
        await asyncio.sleep(OPENSEQ_PULSE_LENGTH)
        await self._open.set(ValveOpenSeqRequest.INACTIVE.value)
```

Control doesn't appear to be being used?

DiamondJoseph: In fact, looking at how open is being used, I don't think you need to make it this overcomplicated, you can just define a Generic ValveControl that takes either of your enum classes on instantiation and then either opens in this standard way or passes the enum to its control signal.

```python
TValveControlRequest = TypeVar("TValveControlRequest", bound = ValveControlRequest | FastValveControlRequest)

class ValveControl(StandardReadable, Movable[TValveControlRequest], Generic[TValveControlRequest]):

    def __init__(self, prefix: str, valve_control_type: type[TValveControlRequest], name: str = ""):
        with self.add_children_as_readable():
             self.open = epics_signal_rw(int, prefix + ":OPENSEQ")
             self.control = epics_signal_rw(valve_control_type, prefix + ":OPENSEQ")
        super().__init__(name)

    @AsyncStatus.wrap
    async def set(self, value: TValveControlRequest):
        if value == ValveControlRequest.OPEN:
            await self.open.set(ValveOpenSeqRequest.OPEN_SEQ.value)
            await asyncio.sleep(OPENSEQ_PULSE_LENGTH)
            await self.open.set(ValveOpenSeqRequest.INACTIVE.value)
        else:
             await self.control.set(value)


# [...] much further down the page

        self.fast_valve_control = {
            i: ValveControl(f"{prefix}V{i}", FastValveControlRequest) for i in self._fast_valves_numbers
        }

        self.slow_valve_control = {
            i: ValveControl(f"{prefix}V{i}", ValveControlRequest) for i in self._slow_valves_numbers
        }
```

DiamondJoseph: I've made a PR of what I think the device could look like here https://github.com/DiamondLightSource/dodal/pull/1386/files

barnettwilliam: Thanks Joseph your suggestion in PR https://github.com/DiamondLightSource/dodal/pull/1386 is much better so will go with that

tests/devices/unit_tests/test_pressure_jump_cell.py
```diff
@@ -107,56 +109,66 @@ async def test_pjumpcell_set_valve_sets_valve_fields(
...
+            },
+        },
     )
+    await assert_reading(
+        cell.all_valves_control.valve_control[6],
+        {
+            "pjump-all_valves_control-valve_control-6-open": {
+                "value": int(ValveOpenSeqRequest.OPEN_SEQ.value),
+                "timestamp": ANY,
+                "alarm_severity": 0,
```
olliesilvester, line 161: When doing `assert_reading` we don't actually need the 
```
"timestamp": ANY,
"alarm_severity": 0,
```
parts. I have a feeling maybe you used to need it when you first wrote this PR, but you can remove it now!

olliesilvester (approved):
Minor comments, but looks good overall! Sorry this took so long to review

Issue #1066, opened 2025-02-21
Add (real) unit tests to hutch shutter devices


CoePaul:
In recent discussion of an i19 plan for control of the hutch shutter it was determined that unit tests 
belong in dodal and might currently be absent

## Acceptance Criteria
* When tests cover the behaviour of hutch shutters ( including those for i19 ) at the unit level
( existing tests cover behaviour already )

CoePaul:
I think this is now obsolete - so will close it

PR #1068, opened 2025-02-21
Distinguish shutter types in existing tests

CoePaul:
* Renames ambiguously named "shutter tests" according to their specific shutter type Examples (of renaming) covered here: shutter -> zebra_shutter shutter -> hutch_shutter

Fixes ambiguous names ( side work for Issue #1066  )

1. Run all tests in project
2. Confirm tests all pass

### Checks for reviewer


tests/devices/i19/test_unit_hutch_conditional_shutter.py
```diff
@@ -0,0 +1,39 @@
...
+    await unit_test_shutter_created_without_raising_errors(
+        hutch_from_which_shutter_is_controlled
+    )
+
+
+async def test_shutter_created_for_eh_two_without_raising_errors():
+    hutch_from_which_shutter_is_controlled = HutchState.EH2
+    await unit_test_shutter_created_without_raising_errors(
+        hutch_from_which_shutter_is_controlled
+    )
```
DominicOram, line 33: Could: You can use `@pytest.mark.parametrize` to do this for you:
```suggestion
@pytest.mark.parametrize(
    "hutch_from_which_shutter_is_controlled", [HutchState.EH1, HutchState.EH2]
)
async def test_shutter_created_without_raising_errors(
    hutch_from_which_shutter_is_controlled,
):
    await unit_test_shutter_created_without_raising_errors(
        hutch_from_which_shutter_is_controlled
    )
```
This will create two tests called `test_shutter_created_without_raising_errors[EH1]` and `test_shutter_created_without_raising_errors[EH2]`. 

In the case of just having the two options it's neither here nor there but when there's like 3 or 4 it tidies up a lot of the copy/paste.

tests/devices/i19/test_unit_hutch_conditional_shutter.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+
+from dodal.devices.i19.shutter import (
+    HutchConditionalShutter,
+    HutchInvalidError,
+    HutchState,
+)
+
+
+async def unit_test_shutter_created_without_raising_errors(
```
DominicOram, line 12: Could: This actually need not be an `async` function. To a first approximation an `async` function is only needed if you are calling other `async` functions from it. The constructor of `HutchConditionalShutter` is synchronous and so this isn't needed. If you were calling the `set` here it would be needed. Does no harm though so not the end of the world to leave it in.

tests/devices/i19/test_unit_hutch_conditional_shutter.py
```diff
@@ -0,0 +1,39 @@
...
+    hutch_from_which_shutter_is_controlled = HutchState.EH2
+    await unit_test_shutter_created_without_raising_errors(
+        hutch_from_which_shutter_is_controlled
+    )
+
+
+async def test_shutter_not_created_for_invalid_hutch():
+    not_a_hutch = HutchState.INVALID
+    with pytest.raises(HutchInvalidError):
+        await unit_test_shutter_created_without_raising_errors(not_a_hutch)
```
DominicOram, line 39: Nit: It's a little odd that you're calling a function called `unit_test_shutter_created_without_raising_errors` and expecting to actually raise the error. But not the end of the world.

DominicOram (changes requested):
Looks good, thanks! Mostly minor points. As discussed, I think these should also be in `test_hutch_conditional_shutter.py`.

tests/devices/i19/test_hutch_conditional_shutter.py
```diff
@@ -0,0 +1,160 @@
...
+    shutter = HutchConditionalShutter("", HutchState.EH2, name="mock_shutter")
+    await shutter.connect(mock=True)
+
+    def set_status(value: ShutterDemand, *args, **kwargs):
+        value_sta = ShutterState.OPEN if value == "Open" else ShutterState.CLOSED
+        set_mock_value(shutter.shutter.status, value_sta)
+
+    callback_on_mock_put(shutter.shutter.control, set_status)
+
+    return shutter
```
DominicOram, line 44: Could: I would combine the common bits here: 
```python
async def make_mock_shutter(hutch: HutchState) -> HutchConditionalShutter:
    shutter = HutchConditionalShutter("", hutch, name="mock_shutter")
    await shutter.connect(mock=True)

    def set_status(value: ShutterDemand, *args, **kwargs):
        value_sta = ShutterState.OPEN if value == "Open" else ShutterState.CLOSED
        set_mock_value(shutter.shutter.status, value_sta)

    callback_on_mock_put(shutter.shutter.control, set_status)

    return shutter


@pytest.fixture
async def eh1_shutter(RE: RunEngine) -> HutchConditionalShutter:
    return await make_mock_shutter(HutchState.EH1)


@pytest.fixture
async def eh2_shutter(RE: RunEngine) -> HutchConditionalShutter:
    return await make_mock_shutter(HutchState.EH2)
```

tests/devices/i19/test_hutch_conditional_shutter.py
```diff
@@ -0,0 +1,160 @@
...
+    with pytest.raises(ShutterNotSafeToOperateError):
+        await eh2_shutter.set(ShutterDemand.OPEN)
+
+
+# Tests above
+# Auxiliary functions below
+
+
+def assertIsInstance(an_instance, expected_type):
+    assert isinstance(an_instance, expected_type)
```
DominicOram, line 150: Should: This should be snake case as a python function
```suggestion
def assert_is_instance(an_instance, expected_type):
    assert isinstance(an_instance, expected_type)
```

DominicOram: Could: I would also personally argue that `assert_is_instance(x, y)` is no more readable than `assert isinstance(x, y)` and so feel this function is redundant but I don't feel that strongly.

DominicOram (changes requested):
Great, thank you! Snake_case is the only comment I really care about, happy for you to take/leave the others.

tests/devices/i19/test_hutch_conditional_shutter.py
CoePaul: Might well be "no more readable" but it is more ophiomorphic ( Greek ophis = snake )

DominicOram (approved):
Great, thank you!

PR #1072, opened 2025-02-24
Update bluesky and pin pyright

### LINKED ISSUE #1041: Update bluesky version
We are currently on v1.13 of bluesky but there are a couple of pre-released versions ahead of that which we could move to.

We can then update typing, for example https://github.com/DiamondLightSource/dodal/pull/1002#discussion_r1939815998

## Acceptance Criteria
* `dodal` uses the latest pre-release version of `bluesky`
* New typing is made use of
* Tests all still pass

DominicOram:
Fixes #1041 

1. Run unit tests and confirm they still pass
2. Confirm new type hints look correct
3. Confirm no new typing errors

### Checks for reviewer

### DIFF
--- a/src/dodal/plan_stubs/wrapped.py
+++ b/src/dodal/plan_stubs/wrapped.py
@@ -79,8 +79,7 @@ def move(moves: Mapping[Movable, Any], group: Group | None = None) -> MsgGenerat
     """
 
     return (
-        # type ignore until https://github.com/bluesky/bluesky/issues/1809
-        yield from bps.mv(*itertools.chain.from_iterable(moves.items()), group=group)  # type: ignore
+        yield from bps.mv(*itertools.chain.from_iterable(moves.items()), group=group)
     )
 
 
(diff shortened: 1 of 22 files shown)


DominicOram:
Latest `pyright` is also more fussy, I will pin that in this PR too but make a new issue to improve.

src/dodal/plan_stubs/wrapped.py
```diff
@@ -12,17 +12,17 @@
 
 Group = Annotated[str, "String identifier used by 'wait' or stubs that await"]
 
+T = TypeVar("T")
+
 
-# After bluesky 1.14, bounds for stubs that move can be narrowed
-# https://github.com/bluesky/bluesky/issues/1821
 def set_absolute(
```
olliesilvester, line 18: This can be another issue, but maybe we can remove a bunch of these now that typing has improved? @DiamondJoseph are these wrappers still useful?

olliesilvester (approved):
Thanks! Question about the wrapped functions but we can put that in a separate issue

src/dodal/plan_stubs/wrapped.py
DiamondJoseph: Oooh the bluesky type hinting went in

DiamondJoseph: These wrapped stubs are still required for blueapi because the bluesky version plan stubs aren't serializable.

PR #1074, opened 2025-02-24
Beam centre lookup table should support linear extrapolation

rtuck99:
Fixes 
* DiamondLightSource/mx-bluesky#719 

1. It should now be possible to calculate the beam centre when outside the bounds of the beam-centre lookup table
2. The beam centre should perform linear extrapolation by projecting linearly from the first (last) two data points in the lookup table

### Checks for reviewer


DominicOram (approved):
This looks good, thank you!

PR #1078, opened 2025-02-27
Allow setting on hardware backed signals

### LINKED ISSUE #1077: Allow passing a setter to `create_hardware_backed_soft_signal`
There are many cases where we need a derived signal (see https://github.com/bluesky/ophyd-async/issues/525). In the absence of one in `ophyd-async` it is preferable that we try and keep the places where we're using derived signals going through one code path in `dodal`. This is so that the migration to the standard solution is easier.

Currently we have `create_hardware_backed_soft_signal` which creates a read-only signal but there are cases where the derived signal should be read/write. It is very easy to change `create_hardware_backed_soft_signal` to do this so it is better to do this now, and keep everyone using the same style/code path, than to wait on https://github.com/bluesky/ophyd-async/issues/525.

## Acceptance Criteria
* `create_hardware_backed_soft_signal` creates a read/write signal that uses a provided `set` method for the writing.

DominicOram:
Fixes #1077

1. Confirm new tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/common/signal_utils.py
+++ b/src/dodal/common/signal_utils.py
@@ -66,7 +66,7 @@ def create_rw_hardware_backed_soft_signal(
     )
 
 
-def create_hardware_backed_soft_signal(
+def create_r_hardware_backed_soft_signal(
     datatype: type[SignalDatatypeT],
     get_from_hardware_func: Callable[[], Coroutine[Any, Any, SignalDatatypeT]],
     units: str | None = None,
--- a/src/dodal/common/signal_utils.py
+++ b/src/dodal/common/signal_utils.py
@@ -51,8 +51,7 @@ def create_rw_hardware_backed_soft_signal(
     and send something to the hardware.
 
     This will allow you to make soft signals derived from arbitrary hardware signals.
-    However, calling subscribe on this signal does not give you a sensible value and
-    the signal is currently read only. See https://github.com/bluesky/ophyd-async/issues/525
+    However, calling subscribe on this signal does not give you a sensible value. See https://github.com/bluesky/ophyd-async/issues/525
     for a more full solution.
     """
     return SignalRW(
--- a/src/dodal/common/signal_utils.py
+++ b/src/dodal/common/signal_utils.py
@@ -4,12 +4,14 @@ from typing import Any
 from bluesky.protocols import Reading
 from ophyd_async.core import SignalDatatypeT, SignalR, SignalRW, SoftSignalBackend
 
+SetHardwareType = Callable[[SignalDatatypeT], Coroutine[Any, Any, None]]
+
 
 class HardwareBackedSoftSignalBackend(SoftSignalBackend[SignalDatatypeT]):
     def __init__(
         self,
         get_from_hardware_func: Callable[[], Coroutine[Any, Any, SignalDatatypeT]],
-        set_to_hardware_func: Callable[[], Coroutine[SignalDatatypeT, Any, Any]] = None,
+        set_to_hardware_func: SetHardwareType | None = None,
         *args,
         **kwargs,
     ) -> None:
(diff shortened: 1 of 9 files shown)


src/dodal/common/signal_utils.py
```diff
@@ -2,22 +2,24 @@
...
 from bluesky.protocols import Reading
-from ophyd_async.core import SignalDatatypeT, SignalR, SoftSignalBackend
+from ophyd_async.core import SignalDatatypeT, SignalR, SignalRW, SoftSignalBackend
 
 
 class HardwareBackedSoftSignalBackend(SoftSignalBackend[SignalDatatypeT]):
     def __init__(
         self,
         get_from_hardware_func: Callable[[], Coroutine[Any, Any, SignalDatatypeT]],
+        set_to_hardware_func: Callable[[], Coroutine[SignalDatatypeT, Any, Any]] = None,
```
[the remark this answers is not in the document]
DominicOram: Thanks! I missed it 'cos the linter is still broken due to https://github.com/DiamondLightSource/dodal/pull/1072

src/dodal/common/signal_utils.py
```diff
@@ -27,6 +31,40 @@ async def get_value(self) -> SignalDatatypeT:
...
+            return await super().put(value, wait)
+
+
+def create_rw_hardware_backed_soft_signal(
+    datatype: type[SignalDatatypeT],
+    get_from_hardware_func: Callable[[], Coroutine[Any, Any, SignalDatatypeT]],
+    set_to_hardware_func: SetHardwareType,
+    units: str | None = None,
+    precision: int | None = None,
+):
```
Relm-Arrowny, line 48: Should: I think we should pass the initial_value as we are using it in put. same for the create_hardware_backed_soft_signal
```suggestion
def create_rw_hardware_backed_soft_signal(
    datatype: type[SignalDatatypeT],
    get_from_hardware_func: Callable[[], Coroutine[Any, Any, SignalDatatypeT]],
    set_to_hardware_func: Callable[[SignalDatatypeT], Coroutine[Any, Any, None]],
    initial_value: SignalDatatypeT | None = None,
    units: str | None = None,
    precision: int | None = None,
) -> SignalRW:
```

src/dodal/common/signal_utils.py
```diff
@@ -27,6 +31,40 @@ async def get_value(self) -> SignalDatatypeT:
...
+            get_from_hardware_func,
+            set_to_hardware_func,
+            datatype,
+            units=units,
+            precision=precision,
+        )
+    )
+
 
 def create_hardware_backed_soft_signal(
```
Relm-Arrowny, line 69: Could: Make it match soft_r /soft_rw ?
```suggestion
def create_r_hardware_backed_soft_signal(
```

Relm-Arrowny (commented):
As I try to use it, I spotted a couple of things.

src/dodal/common/signal_utils.py
DominicOram: Yh, I think it's an edge case anyway and `initial_value` is probably never going to be `None`

src/dodal/common/signal_utils.py
```diff
@@ -27,8 +31,42 @@ async def get_value(self) -> SignalDatatypeT:
         await self._update_value()
         return await super().get_value()
 
+    async def put(self, value: SignalDatatypeT | None, wait: bool) -> None:
+        if self.set_to_hardware_func:
+            write_value = self.initial_value if value is None else value
+            await self.set_to_hardware_func(write_value)
+        else:
+            return await super().put(value, wait)
```
olliesilvester, line 39: Do we need this else statement? If `set_to_hardware_func` is set to None, then this should be the backend of a `SignalR` and so `set_value` can't ever be called

olliesilvester (changes requested):
Looks good, just a few minor comments and questions

The device standards page in the docs refer to `create_hardware_backed_soft_signal`, so we should change that too

DominicOram:
I've addressed some comments but given https://github.com/bluesky/ophyd-async/pull/661 is close to complete and will replace this I don't want to spend too much more time on it.

olliesilvester (approved):

PR #1085, opened 2025-03-03
Add new path provider to use numtracker with blueapi

### LINKED ISSUE #991: Create `PathProvider` that uses numtracker
We now have an instance of [numtracker](https://github.com/DiamondLightSource/numtracker) deployed at https://numtracker.diamond.ac.uk. We should write a [`PathProvider`](https://github.com/bluesky/ophyd-async/blob/caa0f5ccd3259203f3d724195c108bbde1ac1e0e/src/ophyd_async/core/_providers.py#L33) iterating on [`StaticVisitPathProvider`](https://github.com/DiamondLightSource/dodal/blob/a44bce1c1c3fbf22795eff5e7ab7bfc90941df24/src/dodal/common/visit.py#L100) to make use of it.

## Acceptance Criteria
* TBD

abbiemery:
Fixes #991 

Needs https://github.com/DiamondLightSource/blueapi/pull/845 and https://github.com/DiamondLightSource/blueapi/pull/846 to be used

1. See https://github.com/DiamondLightSource/blueapi/pull/846

### Checks for reviewer

TODO:
- Write unit tests for the new path provider
- add a test for the restructured metadata wrapper

### DIFF
--- a/src/dodal/beamlines/training_rig.py
+++ b/src/dodal/beamlines/training_rig.py
@@ -1,5 +1,3 @@
-from pathlib import Path
-
 from ophyd_async.epics.adaravis import AravisDetector
 from ophyd_async.fastcs.panda import HDFPanda
 
@@ -10,7 +8,7 @@ from dodal.common.beamlines.beamline_utils import (
 )
 from dodal.common.beamlines.beamline_utils import set_beamline as set_utils_beamline
 from dodal.common.beamlines.device_helpers import DET_SUFFIX, HDF5_SUFFIX
-from dodal.common.visit import LocalDirectoryServiceClient, StaticVisitPathProvider
+from dodal.common.visit import StartDocumentBasedPathProvider
 from dodal.devices.training_rig.sample_stage import TrainingRigSampleStage
 from dodal.log import set_beamline as set_log_beamline
 from dodal.utils import BeamlinePrefix, get_beamline_name
@@ -31,13 +29,18 @@ PREFIX = BeamlinePrefix(BL)
 set_log_beamline(BL)
 set_utils_beamline(BL)
 
-set_path_provider(
-    StaticVisitPathProvider(
-        BL,
-        Path("/exports/mybeamline/data/2025"),
-        client=LocalDirectoryServiceClient(),
-    )
-)
+# Leave the old path provider so it can be uncommented and used in the meantime while
+# we transition
+
+# set_path_provider(
+#     StaticVisitPathProvider(
+#         BL,
+#         Path("/exports/mybeamline/data/2025"),
+#         client=LocalDirectoryServiceClient(),
+#     )
+# )
+
+set_path_provider(StartDocumentBasedPathProvider())
 
 
 @device_factory()
--- a/src/dodal/beamlines/training_rig.py
+++ b/src/dodal/beamlines/training_rig.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from ophyd_async.epics.adaravis import AravisDetector
 from ophyd_async.fastcs.panda import HDFPanda
 
@@ -8,7 +10,10 @@ from dodal.common.beamlines.beamline_utils import (
 )
 from dodal.common.beamlines.beamline_utils import set_beamline as set_utils_beamline
 from dodal.common.beamlines.device_helpers import DET_SUFFIX, HDF5_SUFFIX
-from dodal.common.visit import StartDocumentBasedPathProvider
+from dodal.common.visit import (
+    LocalDirectoryServiceClient,
+    StaticVisitPathProvider,
+)
 from dodal.devices.training_rig.sample_stage import TrainingRigSampleStage
 from dodal.log import set_beamline as set_log_beamline
 from dodal.utils import BeamlinePrefix, get_beamline_name
@@ -29,18 +34,14 @@ PREFIX = BeamlinePrefix(BL)
 set_log_beamline(BL)
 set_utils_beamline(BL)
 
-# Leave the old path provider so it can be uncommented and used in the meantime while
-# we transition
 
-# set_path_provider(
-#     StaticVisitPathProvider(
-#         BL,
-#         Path("/exports/mybeamline/data/2025"),
-#         client=LocalDirectoryServiceClient(),
-#     )
-# )
-
-set_path_provider(StartDocumentBasedPathProvider())
+set_path_provider(
+    StaticVisitPathProvider(
+        BL,
+        Path("/exports/mybeamline/data/2025"),
+        client=LocalDirectoryServiceClient(),
+    )
+)
 
 
 @device_factory()
--- a/src/dodal/common/beamlines/beamline_utils.py
+++ b/src/dodal/common/beamlines/beamline_utils.py
@@ -1,11 +1,19 @@
 import inspect
 from collections.abc import Callable
+from pathlib import Path
 from typing import Annotated, Final, TypeVar, cast
 
 from bluesky.run_engine import call_in_bluesky_event_loop
+from event_model import RunStart
 from ophyd import Device as OphydV1Device
 from ophyd.sim import make_fake_device
-from ophyd_async.core import DEFAULT_TIMEOUT
+from ophyd_async.core import (
+    DEFAULT_TIMEOUT,
+    FilenameProvider,
+    PathProvider,
+    StaticPathProvider,
+    UUIDFilenameProvider,
+)
 from ophyd_async.core import Device as OphydV2Device
 from ophyd_async.core import wait_for_connection as v2_device_wait_for_connection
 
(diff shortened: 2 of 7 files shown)


src/dodal/common/visit.py
```diff
@@ -150,3 +151,20 @@ def __call__(self, device_name: str | None = None) -> PathInfo:
         return PathInfo(
             directory_path=self._root, filename=self._filename_provider(device_name)
         )
+
+
+DEFAULT_TEMPLATE = "{device_name}-{instrument}-{scan_id}"
+
+
+class StartDocumentBasedPathProvider(PathProvider):
```
callumforrester, line 159: Should: Docstrings on this class would be good

callumforrester (changes requested):

src/dodal/common/visit.py
abbiemery: I have added some now.

src/dodal/beamlines/training_rig.py
```diff
@@ -31,13 +29,18 @@
...
+
+# set_path_provider(
+#     StaticVisitPathProvider(
+#         BL,
+#         Path("/exports/mybeamline/data/2025"),
+#         client=LocalDirectoryServiceClient(),
+#     )
+# )
+
+set_path_provider(StartDocumentBasedPathProvider())
```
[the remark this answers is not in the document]
abbiemery: How will blueapi know if they want to use one or not? Presence of the metadata is not sufficient because you may have a situation where you do have the metadata, but not numtracker configured because you weren't intending to use it. Or you could have numtracker configured, but want to use a static one for local testing reasons.

src/dodal/plan_stubs/data_session.py
```diff
@@ -10,7 +13,7 @@
 
 
 def attach_data_session_metadata_wrapper(
-    plan: MsgGenerator, provider: UpdatingPathProvider | None = None
+    plan: MsgGenerator, provider: PathProvider | None = None
```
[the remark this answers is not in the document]
abbiemery: The reason I left it was because this line complains about the types:
```
provider = provider or get_path_provider()
```
Other types of `PathProvider`, will get passed in here, even though they don't do anything. Since we know this at least we are being explicit about the code paths.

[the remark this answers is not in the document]
abbiemery: Decided to leave as is here, and can make a future issue to change should we wish.

src/dodal/beamlines/training_rig.py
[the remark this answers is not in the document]
abbiemery: Not backwards compatible so this can be ignored.

src/dodal/beamlines/training_rig.py
```diff
@@ -31,13 +29,18 @@
...
+# Leave the old path provider so it can be uncommented and used in the meantime while
+# we transition
+
+# set_path_provider(
+#     StaticVisitPathProvider(
+#         BL,
+#         Path("/exports/mybeamline/data/2025"),
+#         client=LocalDirectoryServiceClient(),
+#     )
+# )
```
DiamondJoseph, line 41:
```suggestion
```

We can get it back from the git history

src/dodal/beamlines/training_rig.py
DiamondJoseph, line 43: Are we decided on calling this from inside the dodal module or from blueapi?

src/dodal/common/beamlines/beamline_utils.py
```diff
@@ -22,7 +24,6 @@
 
 ACTIVE_DEVICES: dict[str, AnyDevice] = {}
 BL = ""
```
DiamondJoseph, line 26:
```suggestion
BL = ""
PATH_PROVIDER: PathProvider = StaticPathProvider(UUIDFilenameProvider(), "/tmp")
```

Gives us a sensible default when the PathProvider hasn't been explicitly set (by blueapi presumably).

src/dodal/common/visit.py
```diff
@@ -150,3 +151,42 @@ def __call__(self, device_name: str | None = None) -> PathInfo:
...
+
+    """
+
+    def __init__(self) -> None:
+        self._doc = {}
+
+    def update_run(self, name: str, start_doc: RunStart) -> None:
+        """Cache a start document.
+
+        This can be plugged into the run engines subscribe method.
```
DiamondJoseph, line 175: The RunEngine subscribe method can be passed the names of documents you expect to receive, if the name check is just for sanity's sake it's fine, but if you're receiving all of the Document types because you're not making use of it, I would add it in, else the start_doc param may cause issues when passed another Document type.

RE.subscribe(update_run, name="start")

DiamondJoseph (dismissed):

src/dodal/common/beamlines/beamline_utils.py
abbiemery: Currently I _think_ every beamline module requires one and has one set? Although I guess this won't do harm?

abbiemery: Although it changes the behavior of dodal somewhat by having it. I think I prefer to leave it and still demand you set it yourself.

src/dodal/common/visit.py
abbiemery: This is done in blueapi, so we are only getting start documents.

src/dodal/beamlines/training_rig.py
abbiemery: This gets called when blueapi loads the context.

src/dodal/common/beamlines/beamline_utils.py
abbiemery: Opened a discussion her https://github.com/DiamondLightSource/dodal/issues/1129. It is an problem for another issue.

src/dodal/common/visit.py
DominicOram: Maybe nitpicky but this looks weird to me because all MX callbacks inherit from `CallbackBase` and would use `def start(` for this. I appreciate that's not easy given that `PathProvider` and `CallbackBase` both want to make use of `__call__` though. I think it's particularly confusing because the definition and the actual usecase is split across different repos

src/dodal/beamlines/training_rig.py
callumforrester: From discussion, I think the idea is to revert this change and make blueapi call `set_path_provider(StartDocumentBasedPathProvider())` when given the proper config, but correct me if I'm remembering wrongly.

abbiemery: I have put this back to use the old one, so we leave it in a working state. There is another issue for changing it to the new provider which depends on a numtracker issue to configure the training rigs.

https://github.com/DiamondLightSource/dodal/issues/1130

and 

https://github.com/DiamondLightSource/numtracker/issues/134

src/dodal/common/visit.py
abbiemery: We made an issue for this refactor here. https://github.com/DiamondLightSource/dodal/issues/1132

DominicOram: It was a suggestion, happy to leave as is

src/dodal/common/visit.py
```diff
@@ -150,3 +151,42 @@ def __call__(self, device_name: str | None = None) -> PathInfo:
...
```
callumforrester, line 175: Nit (sorry): 
```suggestion
        This can be plugged into the run engine's subscribe method.
```

src/dodal/plan_stubs/data_session.py
```diff
@@ -30,14 +33,19 @@ def attach_data_session_metadata_wrapper(
...
+        ress = yield from bps.wait_for([provider.data_session])
+        data_session = ress[0].result()
+        # https://github.com/DiamondLightSource/dodal/issues/452
+        # As part of 452, write each dataCollection into their own folder, then can use resource_dir directly
+        yield from bpp.inject_md_wrapper(plan, md={DATA_SESSION: data_session})
+    else:
+        logging.warn(
+            f"{provider} is not an UpdatingPathProvider, {attach_data_session_metadata_wrapper.__name__} will have no effect"
+        )
+        yield from plan
```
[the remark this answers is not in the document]
abbiemery: The only providers used currently are `StaticVisitPathProvider` and `PandASubpathProvider` which are both subclass of `UpdatingPathProvider` . If this wrapper is used, it will still work as the is instance check on the parent will pass for both the childern.

src/dodal/common/visit.py
```diff
@@ -150,3 +151,42 @@ def __call__(self, device_name: str | None = None) -> PathInfo:
...
         )
+
+
+DEFAULT_TEMPLATE = "{device_name}-{instrument}-{scan_id}"
+
+
+class StartDocumentPathProvider(PathProvider):
+    """A PathProvider that sources from metadata in a RunStart document.
+
+    This uses metadata a RunStart document to determine file names and data session
```
callumforrester, line 162: Nit:
```suggestion
    This uses metadata from a RunStart document to determine file names and data session
```

callumforrester (approved):
LGTM, couple of small typos

PR #1099, opened 2025-03-10
Add undulator check as preprocessor

olliesilvester:
Fixes [mx-bluesky 891](https://github.com/DiamondLightSource/mx-bluesky/issues/891)

For now, this plan requires the "common" DCM, but there are a few different implementations of this.  https://github.com/DiamondLightSource/dodal/issues/592 will make this preprocessor more useful.

Also requires a change to mx-bluesky (PR in progress)

1. Confirm code meets issue requirements
2. Check tests and confirm they pass

### Checks for reviewer

### DIFF
--- a/src/dodal/plans/preprocessors/verify_undulator_gap.py
+++ b/src/dodal/plans/preprocessors/verify_undulator_gap.py
@@ -1,0 +1,49 @@
+from bluesky.preprocessors import plan_mutator
+from bluesky.utils import Msg, MsgGenerator, make_decorator
+
+from dodal.plans.verify_undulator_gap import CheckUndulatorDevices, verify_undulator_gap
+
+
+def verify_undulator_gap_before_run_wrapper(
+    plan: MsgGenerator,
+    devices: CheckUndulatorDevices,
+    run_key_to_wrap: str | None = None,
+):
+    """
+    Modifies the wrapped plan so that it checks the undulator gap before the specified run is opened and sets it to the correct value if needed.
+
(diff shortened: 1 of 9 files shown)


src/dodal/plans/preprocessors/verify_undulator_gap.py
```diff
@@ -0,0 +1,49 @@
+from bluesky.preprocessors import plan_mutator
+from bluesky.utils import Msg, MsgGenerator, make_decorator
+
+from dodal.plans.verify_undulator_gap import CheckUndulatorDevices, verify_undulator_gap
+
+
+def verify_undulator_gap_before_run_wrapper(
```
DominicOram, line 7: Could: We have the same pattern in the `xbpm_feedback` preprocessor. I wonder if we can make a helper function that's something like `def do_on_run(on_run_open, on_run_close = None, run_key=None)`?

DominicOram: This may even be useful to feedback to `bluesky`

DominicOram (approved):
Looks good, thanks! One optional comment to make things cleaner, feel free to spin into a new issue if you'd rather though

src/dodal/plans/preprocessors/verify_undulator_gap.py
olliesilvester: Good idea, I'll put it into a new issue though

olliesilvester: https://github.com/DiamondLightSource/dodal/issues/1103

PR #1109, opened 2025-03-12
add_electron_analyser_region_dataclasses

oliwenmandiamond:
Add region data classes for electron analysers #1122


DominicOram (changes requested):
Looks like a good start, hard for me to comment on some bits without seeing the bigger context of what you're doing. It would have been useful to first write an issue that explains what the user wants and what you are trying to provide them. Then it means that people that aren't so involved in the beamline can review, they also act as good documentation for a later date. Additionally, I would suggest having some unit tests on `SESSequence`, as I said, I'm happy to help you with that. Finally, `src/dodal/devices/i09/vgscienta_analyser plan.py` should be `snake_case`.

src/dodal/devices/i09/base_region.py
```diff
@@ -0,0 +1,38 @@
...
+    slices: int = 1
+    iterations: int = 1
+    lensMode: str = ""
+    lowEnergy: float = -1
+    highEnergy: float = -1
+    stepTime: float = -1
+    energyStep: float = -1
+
+
+class SPECSRegion(AbstractBaseRegion):
```
Villtord, line 32: Looks like you forgot to delete it?

src/dodal/devices/electron_analyser/base_analyser.py
```diff
@@ -0,0 +1,54 @@
...
+    PV_HIGH_ENERGY = "HIGH_ENERGY"
+    PV_SLICES = "SLICES"
+    PV_ITERATIONS = "NumExposures"
+    PV_ENERGY_STEP = "STEP_SIZE"
+
+    def __init__(self, prefix: str, name: str = "") -> None:
+        self.prefix = prefix
+
+        with self.add_children_as_readables():
+            self.low_energy_signal = epics_signal_w(
```
Relm-Arrowny, line 26: Must: epics_signal_w does not have read, therefore you cannot add it as readables, if you want both read and write use epics_signal_rw

src/dodal/devices/electron_analyser/specs/specs_analyser.py
```diff
@@ -0,0 +1,25 @@
...
+                float, self.prefix + SpecsAnalyser.PV_PSU_MODE
+            )
+            self.values_signal = epics_signal_w(
+                float, self.prefix + SpecsAnalyser.PV_VALUES
+            )
+            self.centre_energy_signal = epics_signal_w(
+                float, self.prefix + SpecsAnalyser.PV_CENTRE_ENERGY
+            )
+
+        super().__init__(prefix, name)
```
Relm-Arrowny, line 25: Must: Because you inherited movable from base_analyser you will need to define the set method here.

PR #1110, opened 2025-03-12
Mx bluesky 908 ci check walrus in asserts

jacob720:
Fixes #[908](https://github.com/DiamondLightSource/mx-bluesky/issues/908) in mx-bluesky.

1. Add in a walrus operator inside an assert
2. Check ruff shouts at you with `ruff check`
3. Add in a walrus operator inside an assert in /tests or /system_tests
4. Check ruff **doesn't** shout at you.

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -178,14 +178,15 @@ lint.select = [
     "I",   # isort - https://docs.astral.sh/ruff/rules/#isort-i
     "UP",  # pyupgrade - https://docs.astral.sh/ruff/rules/#pyupgrade-up
     "SLF", # self - https://docs.astral.sh/ruff/settings/#lintflake8-self
+    "RUF018" # walrus operators in asserts - https://docs.astral.sh/ruff/rules/assignment-in-assert/
 ]
 
 [tool.ruff.lint.per-file-ignores]
 # By default, private member access is allowed in tests
 # See https://github.com/DiamondLightSource/python-copier-template/issues/154
 # Remove this line to forbid private member access in tests
-"tests/**/*" = ["SLF001"]
-"system_tests/**/*" = ["SLF001"]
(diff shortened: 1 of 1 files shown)


DominicOram (changes requested):
Thanks. I think it would be better to split the `RUF018` change from all the other fixes. Happy to merge the `RUF018` check, see https://github.com/DiamondLightSource/mx-bluesky/issues/908 for my thoughts on the others

pyproject.toml
```diff
@@ -178,14 +178,15 @@ lint.select = [
     "I",   # isort - https://docs.astral.sh/ruff/rules/#isort-i
     "UP",  # pyupgrade - https://docs.astral.sh/ruff/rules/#pyupgrade-up
     "SLF", # self - https://docs.astral.sh/ruff/settings/#lintflake8-self
+    "RUF018" # walrus operators in asserts - https://docs.astral.sh/ruff/rules/assignment-in-assert/
 ]
 
 [tool.ruff.lint.per-file-ignores]
 # By default, private member access is allowed in tests
 # See https://github.com/DiamondLightSource/python-copier-template/issues/154
 # Remove this line to forbid private member access in tests
```
DominicOram, line 187: Nit: Would be good to change this comment too in light of the additional ignore e.g. why we have this ignore as well and specify just removing the one thing for allowing private member access

DominicOram (approved):
Great, thanks. One comment take it or leave it. It also sounds like from the discussion on the issue that adding this into the copier template isn't too controversial. Can you make a PR to add it in to https://github.com/DiamondLightSource/python-copier-template/blob/main/template/pyproject.toml.jinja?

pyproject.toml
jacob720: Sounds good, I'll add that in

PR #1112, opened 2025-03-13
Fix all possible lint errors with ruff check --fix

jacob720:
Fixes #[908](https://github.com/DiamondLightSource/mx-bluesky/issues/908) in mx-bluesky.

1. Look at the changes and rules they fix
2. Decide if this seems worth doing

I also added in over 1000 instant fixes Ruff could do when looking at all possible [rules](https://docs.astral.sh/ruff/rules/).
The specific rules it was fixing can be found below, in the order of how many times it was fixed (or see the attached file)

[fixable.txt](https://github.com/user-attachments/files/19225162/fixable.txt)

769	COM812 	 missing-trailing-comma
140	D212   	 multi-line-summary-first-line
 97	PT006  	 pytest-parametrize-names-wrong-type
 85	PT007  	 pytest-parametrize-values-wrong-type
 85	D413   	 missing-blank-line-after-last-section
 53	RET504 	 unnecessary-assign
 49	D202   	 blank-line-after-function
 42	EM102  	 f-string-in-exception
 42	T201   	 print
 41	ERA001 	 commented-out-code
 38	D200   	 unnecessary-multiline-docstring
 38	D209   	 new-line-after-last-paragraph
 21	EM101  	 raw-string-in-exception
 12	D411   	 no-blank-line-before-section
  9	RET505 	 superfluous-else-return
  8	PT022  	 pytest-useless-yield-fixture
  7	PT001  	 pytest-fixture-incorrect-parentheses-style
  6	SIM300 	 yoda-conditions
  5	SIM118 	 in-dict-keys
  4	RSE102 	 unnecessary-paren-on-raise-exception
  4	TC002  	 typing-only-third-party-import
  4	D403   	 first-word-uncapitalized
  4	RUF022 	 unsorted-dunder-all
  3	PT003  	 pytest-extraneous-scope-function
  3	RET506 	 superfluous-else-raise
  3	TC001  	 typing-only-first-party-import
  3	D300   	 triple-single-quotes
  3	D410   	 no-blank-line-after-section
  3	F401   	 unused-import
  3	RUF015 	 unnecessary-iterable-allocation-for-first-element
  2	RET503 	 implicit-return
  2	SIM105 	 suppressible-exception
  2	TID252 	 relative-imports
  2	TC003  	 typing-only-standard-library-import
  2	PTH201 	 path-constructor-current-directory
  2	PLR0402	 manual-from-import
  2	RUF005 	 collection-literal-concatenation
  2	RUF100 	 unused-noqa
  2	TRY201 	 verbose-raise
  1	EM103  	 dot-format-in-exception
  1	ICN001 	 unconventional-import-alias
  1	PIE790 	 unnecessary-placeholder
  1	PIE810 	 multiple-starts-ends-with
  1	PT014  	 pytest-duplicate-parametrize-test-cases
  1	SIM103 	 needless-bool
  1	SIM108 	 if-else-block-instead-of-if-exp
  1	D214   	 overindented-section
  1	D409   	 mismatched-section-underline-length
  1	D412   	 blank-lines-between-header-and-content
  1	PLC0208	 iteration-over-set
  1	PLR1714	 repeated-equality-comparison
  1	PLR1722	 sys-exit-alias
  1	PLR5501	 collapsible-else-if
  1	UP015  	 redundant-open-modes

### Checks for reviewer


DominicOram (changes requested):
Thank you @jacob720, could you redo this with adding the correct config to the `pyproject.toml` to ignore the ones Callum and I have discussed above? I think you can be clever and e.g. ignoring `EM10` will ignore all the `EM10X` ones (which is probably what we want for `TC00` too)

> F401: These should already be caught, @DominicOram to check

I couldn't reproduce this on main, it could be that one of the other fixes removed an import and so triggered this, lets see if it comes back when we do the ignores

> Also also, we should make an ADR at this point

 I can do this

PR #1113, opened 2025-03-13
(mx-bluesky#894) Add goniometer device

JamesDoingStuff:
Fixes [mx-bluesky#894](https://github.com/DiamondLightSource/mx-bluesky/issues/894)

1. On lab PC (ws581) run `dodal connect aithre`
2. Check that the connection is made

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/aithre.py
+++ b/src/dodal/beamlines/aithre.py
@@ -1,9 +1,9 @@
 from dodal.common.beamlines.beamline_utils import device_factory
 from dodal.devices.aithre_lasershaping.goniometer import Goniometer
 
-PREFIX = "LA18L-MO-LSR-01:"
+PREFIX = "LA18L"
 
 
 @device_factory()
 def goniometer() -> Goniometer:
-    return Goniometer(PREFIX, "aithre")
+    return Goniometer(f"{PREFIX}-MO-LSR-01:", "goniometer")
--- a/src/dodal/beamlines/aithre.py
+++ b/src/dodal/beamlines/aithre.py
@@ -1,0 +1,9 @@
+from dodal.common.beamlines.beamline_utils import device_factory
+from dodal.devices.aithre_lasershaping.goniometer import Goniometer
+
+PREFIX = "LA18L-MO-LSR-01:"
+
+
+@device_factory()
+def goniometer() -> Goniometer:
+    return Goniometer(PREFIX, "aithre")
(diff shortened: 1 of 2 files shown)


JamesDoingStuff:
@DominicOram I thought I'd do this as a draft for now since it's lacking tests and is just based on the instantiation files from other beamlines, so I have no idea whether it works or not - please let me know if I'm way off!

src/dodal/beamlines/aithre.py
```diff
@@ -0,0 +1,9 @@
+from dodal.common.beamlines.beamline_utils import device_factory
+from dodal.devices.aithre_lasershaping.goniometer import Goniometer
+
+PREFIX = "LA18L-MO-LSR-01:"
```
DominicOram, line 4: Could: So normally the prefix of a beamline is just the first bit (i.e. `LA18L` in this case) and the rest is different for each device but I'm not sure if that's the case for the lab. We can change it or leave it as is and we'll quickly find out when we do the next device

src/dodal/beamlines/aithre.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+@device_factory()
+def goniometer() -> Goniometer:
+    return Goniometer(PREFIX, "aithre")
```
DominicOram, line 9: Should; The second parameter here is the name of the specific device. `aithre` is the name of the whole project but the device can just be called `goniometer` I think

src/dodal/devices/aithre_lasershaping/goniometer.py
```diff
@@ -0,0 +1,15 @@
+from ophyd_async.core import StandardReadable
+from ophyd_async.epics.motor import Motor
+
+"""Goniometer and the stages it sits on"""
```
DominicOram, line 4: Nit: This might be better as a docstring in the class as then sphinx can pick it up nicer

DominicOram (changes requested):
As you said, it's hard to test w/o working hardware but I can run `dodal connect aithre` and the failed PVs look reasonable so I think that's good enough for me!

Other than that, some minor comments in the code. We should probably tag @co2e14 in all the `aithre` work we do too, just so he knows where we are

src/dodal/beamlines/aithre.py
JamesDoingStuff: I've made the change to the prefix and moved the rest to the argument of the device call, following examples in the other beamlines - is that right?

DominicOram: Yep, this looks more like what I would expect, thanks!

DominicOram (approved):
Great, this all looks good now. Happy to merge and we can address any issues that arise when the hardware is back up separately

Issue #1122, opened 2025-03-17
Create the region dataclasses for electron analysers


oliwenmandiamond:
Surface beamlines define sequence files that hold region settings. These are in by the detector and configure it with the settings used before data collection starts. We need to add the region classes to be able to read and save sequence files.

They should:
- Have a generic base class that is common to all analysers
- Have specialised region classes for specific electron analyser settings.
- Have tests to show they read in the expected data.

PR #1124, opened 2025-03-18
1109 add region data classes

oliwenmandiamond:
#1122 
Add region data classes for electron analysers

Instructions to reviewer on how to test:
- Confirm that region data is loaded successfully and matches expected data from tests


oliwenmandiamond:
I renamed the remote branch which for some reason closed my old PR... So sorry for confusion

src/dodal/devices/electron_analyser/vgscienta/vgscienta_region.py
```diff
@@ -0,0 +1,77 @@
...
+    lensMode: str = "Angular45"
+    passEnergy: int | float = 5
+    acquisitionMode: str = AcquisitionMode.SWEPT
+    lowEnergy: float = 8.0
+    highEnergy: float = 10.0
+    stepTime: float = 1.0
+    energyStep: float = Field(default=200.0)
+    # Specific to this class
+    regionId: str = Field(default=str(uuid.uuid4()))
+    excitationEnergySource: str = "source1"
```
DominicOram, line 39: Could: it would be cleaner if this was a `VGScientaExcitationEnergySource` rather than a `str`, then you don't need the logic in `get_excitation_energy_source_by_region` any more. I appreciate this is hard with the file format that you're currently trying to use though so feel free to pull it into a new issue

src/dodal/devices/electron_analyser/vgscienta/vgscienta_region.py
```diff
@@ -0,0 +1,77 @@
...
+        filtered_excitation_energy_sources = [
+            e
+            for e in self.excitationEnergySources
+            if e.name == region.excitationEnergySource
+        ]
+        return (
+            filtered_excitation_energy_sources[0]
+            if len(filtered_excitation_energy_sources) == 1
+            else None
+        )
```
DominicOram, line 77: Could: 
```suggestion
        return next(
            filter(
                lambda e: e.name == region.excitationEnergySource,
                self.excitationEnergySources,
            ),
            None,
        )
```
maybe slightly cleaner

src/dodal/common/data_util.py
```diff
@@ -0,0 +1,20 @@
...
+) -> TBaseModel:
+    with open(file) as f:
+        json_obj = f.read()
+    cls = t.model_validate_json(json_obj)
+    return cls
+
+
+def save_class_to_json_file(model: BaseModel, file: str) -> None:
+    with open(file, "w") as f:
+        f.write(model.model_dump_json())
```
DominicOram, line 20: As an aside, how are you envisaging this working? I would naively think you'll be sending a `Region` object in when you call BlueAPI, which then sets up the analyser. In which case I'm not sure where the read/writing to disk is needed?

DominicOram (approved):
Great, minor suggestions and a question for my understanding but looks good, thanks for the tests

src/dodal/common/data_util.py
oliwenmandiamond: So these are generic helper functions that anyone can use to read in json files to a data class and save a data class to a json file. Currently, load_json_file_to_class is only used in the tests I've created. You're also correct that the idea is to hopefully pass on the region json to the analyser from GDA when we use BlueAPI. However, once we implement data collection natively within Bluesky, were also going to want the detector to read the file directly and also have users be able to manipulate the files within Bluesky. These small functions are thinking ahead to that goal.

src/dodal/devices/electron_analyser/vgscienta/vgscienta_region.py
oliwenmandiamond: Unfortunately, no. The idea of this file format is that regions simply point to a single excitation energy source. The ExcitationEnergySource.value must be in sync with all regions (currently for client and server purposes on GDA end). This is simplest and also means if you need to update anything about the excitation energy source e.g value, you only need to update the excitation energy source value rather than each individual region that points to it.

DominicOram:
I think the linting issues are fixed on main if you merge it in

src/dodal/devices/electron_analyser/base_region.py
```diff
@@ -0,0 +1,64 @@
...
+    acquisitionMode: str
+    lowEnergy: float
+    highEnergy: float
+    stepTime: float
+    energyStep: float
+    energyMode: EnergyMode = EnergyMode.KINETIC
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_energy_mode(cls, data: Any) -> Any:
```
Villtord, line 34: Perhaps it would be better to rename to "validate_energy_mode" or "correct_energy_mode"? This function is for load of legacy sequences/regions and there is no real PV value behind this field.

DominicOram: How `legacy` are they? Is there a way we can just say that we're not dealing with them in Bluesky going forward? In MX we're trying as much as possible to drop the legacy stuff as we switch over

Villtord: Legacy in terms that they exist in all current saved sequences, but we can do completely without them in future new saved sequences

src/dodal/common/data_util.py
DominicOram: Ok, I think this is fine for now but we should chat to @DiamondLightSource/developers-daq-core about this, the idea of the new architecture is to try and remove our reliance on files so discussing why we need the files at all would be good in the future

src/dodal/devices/electron_analyser/base_region.py
DominicOram: How would the scientists feel about not being able to load old sequences in bluesky?

DominicOram: Alternatively you could do this conversion in the GDA side before you trigger BlueAPI. Ultimately up to you guys but I think it's good to view the migration to bluesky as an opportunity to clean this stuff up as much as you can as you go

PR #1125, opened 2025-03-18
Add analyser classes

oliwenmandiamond:
Add initial analyser region controller devices 

1. Check analyser classes setup region data correctly
2. Confirm tests are adequate and pass

### Checks for reviewer


src/dodal/devices/electron_analyser/base_analyser.py
```diff
@@ -0,0 +1,107 @@
...
+    async def set(
+        self,
+        value: TBaseRegion,
+        excitation_energy_eV: float | None = None,
+        *args,
+        **kwargs,
+    ):
+        """
+        This is intended to be used with bluesky plan
+        bps.abs_set(analyser, region, excitation_energy)
```
DominicOram, line 72: Must: I didn't realise that `set` let you provide multiple arguments. We don't do this anywhere else and I'm not sure we want to do it @DiamondLightSource/developers-daq-core, what do you think? I would personally make this a plan:
```python
def configure_analyser(analyser: BaseAnalyser, region: TBaseRegion, excitation_energy_ev: float):
        LOGGER.info(f"Configuring analyser with region {region.name}")

        is_binding_energy = region.energyMode == EnergyMode.BINDING
        low_energy = (
            excitation_energy_eV - region.highEnergy
            if is_binding_energy
            else region.lowEnergy
        )
        high_energy = (
            excitation_energy_eV - region.lowEnergy
            if is_binding_energy
            else region.highEnergy
        )
        # Cast pass energy to correct type for PV to take
        pass_energy_type = self.get_pass_energy_type()
        pass_energy = analyser.pass_energy_type(region.passEnergy)
        # Set detector settings, wait for them all to have completed
        yield from bps.mv(
            analyser.low_energy_signal, low_energy,
            analyser.high_energy_signal, high_energy,
            analyser.slices_signal, region.slices,
            analyser.lens_mode_signal, region.lensMode,
            analyser.pass_energy_signal, pass_energy,
            analyser.iterations_signal, region.iterations,
            analyser.acquisition_mode_signal, region.acquisitionMode,
        )
```

src/dodal/devices/electron_analyser/base_analyser.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+            self.energy_step_signal = epics_signal_rw(
+                float, self.prefix + BaseAnalyser.PV_ENERGY_STEP
+            )
+            self.iterations_signal = epics_signal_rw(
+                int, self.prefix + BaseAnalyser.PV_ITERATIONS
+            )
+            self.acquisition_mode_signal = epics_signal_rw(
+                str, self.prefix + BaseAnalyser.PV_ACQISITION_MODE
+            )
```
DominicOram, line 55: Should: I think the word `_signal` is redundant in these definitions. `abs_set(analyser.high_energy, 1000)` is just as readable as `abs_set(analyser.high_energy_signal, 1000)` and shorter. Other devices in dodal do not specify `signal` everywhere

src/dodal/devices/electron_analyser/base_analyser.py
```diff
@@ -0,0 +1,107 @@
...
+        **kwargs,
+    ):
+        """
+        This is intended to be used with bluesky plan
+        bps.abs_set(analyser, region, excitation_energy)
+        """
+
+        region = value
+        if excitation_energy_eV is None:
+            raise ValueError("excitation_energy_eV must be specified.")
```
DominicOram, line 77: Should: If it is an error for `excitation_energy_eV ` to be `None` then the function shouldn't type it so that it could be `None`. The general best practice is to lean on the type system here e.g. make the function only accept a `float` then assume inside the function it is a `float`. If someone calls it with something that isn't a float then the type system will shout at them

src/dodal/devices/electron_analyser/base_region.py
```diff
@@ -26,9 +26,15 @@ class BaseRegion(BaseModel):
...
     stepTime: float
-    energyStep: float
+    energyStep: float  # in eV
     energyMode: EnergyMode = EnergyMode.KINETIC
 
+    def get_energy_step_eV(self) -> float:
+        """
+        Each sub class can override this if they are specified in different units
+        """
+        return self.energyStep
```
DominicOram, line 36: Should: All subclasses use the same implementation of this currently, right? I assume there are future region definitions that don't? I would suggest removing this now and when you have to implement the future ones look at better standardisation. Otherwise this is just patching over poor standards from GDA

src/dodal/devices/electron_analyser/base_analyser.py
```diff
@@ -0,0 +1,107 @@
...
+                int, self.prefix + BaseAnalyser.PV_ITERATIONS
+            )
+            self.acquisition_mode_signal = epics_signal_rw(
+                str, self.prefix + BaseAnalyser.PV_ACQISITION_MODE
+            )
+
+        super().__init__(name)
+
+    def get_pass_energy_type(self) -> type:
+        return float
```
DominicOram, line 60: Should: This typing should be standardised in the EPICS layer. When we have come across similar differences in PVs in MX we have pushed for controls to standardise and they are generally happy to. Have you discussed this with controls?

src/dodal/devices/electron_analyser/specs/specs_analyser.py
```diff
@@ -0,0 +1,49 @@
+import asyncio
+
+from ophyd_async.core import AsyncStatus
+from ophyd_async.epics.core import epics_signal_rw
+
+from dodal.devices.electron_analyser.base_analyser import BaseAnalyser
+from dodal.devices.electron_analyser.specs.specs_region import SpecsRegion
+
+
+class SpecsAnalyser(BaseAnalyser):
```
DominicOram, line 10: Should: As in base class re constants, `_signal`, `self.prefix` and use of `set`

src/dodal/devices/electron_analyser/vgscienta/vgscienta_analyser.py
```diff
@@ -0,0 +1,83 @@
...
+
+from dodal.devices.electron_analyser.base_analyser import BaseAnalyser
+from dodal.devices.electron_analyser.base_region import EnergyMode
+from dodal.devices.electron_analyser.vgscienta.vgscienta_region import (
+    DetectorMode,
+    VGScientaRegion,
+)
+
+
+class VGScientaAnalyser(BaseAnalyser):
```
DominicOram, line 14: Should: As in base class re constants, `_signal`, `self.prefix` and use of `set`

src/dodal/devices/electron_analyser/specs/specs_analyser.py
```diff
@@ -0,0 +1,49 @@
...
+        region = value
+        # These units need to be converted depending on the region
+        energy_step_eV = region.get_energy_step_eV()
+
+        await asyncio.gather(
+            super().set(region, excitation_energy_eV),
+            self.values_signal.set(region.values),
+            self.psu_mode_signal.set(region.psuMode),
+            self.centre_energy_signal.set(region.centreEnergy)
+            if region.acquisitionMode == "Fixed Transmission"
```
DominicOram, line 44: Could: if `acquisitionMode ` only has a set number of modes it should be an enum not a string

tests/devices/unit_tests/electron_analyser/test_specs_analyser.py
```diff
@@ -0,0 +1,30 @@
...
+@pytest.fixture
+def sequence_class() -> type[SpecsSequence]:
+    return SpecsSequence
+
+
+def test_analyser_abs_set(sim_analyser: SpecsAnalyser, sequence: SpecsSequence) -> None:
+    RE = RunEngine()
+    excitation_energy = 100
+    for r in sequence.get_enabled_regions():
+        RE(abs_set(sim_analyser, r, excitation_energy))
```
DominicOram, line 30: Should: This isn't a great test for a few reasons:
* It relies on reading the sequence file and on `get_enabled_regions` both of which are not part of the `SpecsAnalyser`. In general we want each of our test files to just test the one thing as much as possible so that if we make a change elsewhere we don't have to fix lots of places
* Because we're relying on the loaded file it's complicated to see what we actually expect the analyser to be called with
* There are no `asserts`, the analyser could be doing nothing at all and this would still pass
* You're not testing different examples, i.e. in this case the different behavior for `"Fixed Transmission"` vs `Fixed Energy`

Here is an example of a test I would write for this:
```python
def test_given_region_using_fixed_transmission_then_centre_energy_set_to_region_centre_energy(
    sim_analyser: SpecsAnalyser,
):
    RE = RunEngine()
    my_region = SpecsRegion()
    my_region.acquisitionMode = "Fixed Transmission"
    my_region.centreEnergy = 1234

    RE(abs_set(sim_analyser, my_region, 1000))

    get_mock_put(sim_analyser.centre_energy_signal).assert_called_once_with(
        1234, wait=True
    )
```
I would probably write 4/5 of these kind of tests for each of your analysers covering all the little bits of logic in them

tests/devices/unit_tests/electron_analyser/test_vgscienta_analyser.py
```diff
@@ -0,0 +1,35 @@
...
+def sequence_file() -> str:
+    return "vgscienta_sequence.seq"
+
+
+@pytest.fixture
+def sequence_class() -> type[VGScientaSequence]:
+    return VGScientaSequence
+
+
+def test_analyser_abs_set(
```
DominicOram, line 28: As above re splitting into smaller tests

src/dodal/devices/electron_analyser/base_analyser.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @AsyncStatus.wrap
+    async def set(
```
DominicOram, line 63: Should: I would have some specific unit tests on just the `BaseAnalyser` too

src/dodal/devices/electron_analyser/base_analyser.py
```diff
@@ -0,0 +1,107 @@
...
+
+from bluesky.protocols import Movable
+from ophyd_async.core import AsyncStatus, StandardReadable
+from ophyd_async.epics.core import epics_signal_rw
+
+from dodal.devices.electron_analyser.base_region import EnergyMode, TBaseRegion
+from dodal.log import LOGGER
+
+
+class BaseAnalyser(StandardReadable, Movable, Generic[TBaseRegion]):
```
DominicOram, line 12: Do you expect a BaseAnalyser to ever be instantiated itself? If not you could consider making it an [ABC](https://docs.python.org/3/library/abc.html)

DominicOram (changes requested):
Apologies, a few comments. I would also suggest making a beamline file for the beamlines that have these analysers. This would mean that you (or a reviewer) can do `dodal connect i05` and confirm that the PVs are all connecting properly. Feel free to do this in a new issue/PR though

src/dodal/devices/electron_analyser/base_analyser.py
oliwenmandiamond: Making the excitation energy a signal could also work and then just use the custom plan to give both arguments might be easiest solution

src/dodal/devices/electron_analyser/specs/specs_analyser.py
oliwenmandiamond: I agree, but this is a little more complicated as each beamline has it's own set of valid acquisition modes, including ones that are using the same type of analyser. It is something I will address in the future

src/dodal/devices/electron_analyser/base_analyser.py
oliwenmandiamond: I've changed them to be abstract

src/dodal/devices/electron_analyser/base_analyser.py
oliwenmandiamond: No, have changed to be abstract instead.

src/dodal/devices/electron_analyser/base_analyser.py
DominicOram: Yes, I think you could either do it with just composition of plans:
```python
def set_generic_analyser(analyser: BaseAnalyser, ...):
    ...

def set_specs_analyser(analyser: SpecsAnalyser, ...):
    yield from set_generic_analyser(...)
    ...
```
Or having `excitation_energy_eV` as a soft signal you set first would work too. I'm not sure which is cleaner tbh

DominicOram: @callumforrester, thoughts on multiple arguments to `set`? I think we should not allow it

DominicOram (commented):
I'm also not sure why the tests are failing but I think it is to do with the `conftest.py` that you have added

src/dodal/devices/electron_analyser/base_region.py
oliwenmandiamond: I have assumed all will be in eV. Am fixing GDA end to fix this standardisation.

src/dodal/devices/electron_analyser/base_analyser.py
oliwenmandiamond: Will hopefully be fixed in controls to be a standard. Assuming will be str.

.gitignore
```diff
@@ -78,3 +78,4 @@ lockfiles/
 
 # custom CI setup 
 report.json
+Pipfile
```
callumforrester, line 81: Must: This shouldn't be needed, we stopped using `Pipfile`s a long time ago. Please use the devcontainer to develop with dodal if at all possible, happy to help you set up your dev environment.

src/dodal/devices/electron_analyser/base_analyser.py
callumforrester: I agree we shouldn't allow multiple arguments, I suspect that will break downstream. It could be a plan, or we could use a model for this, something like:

```python
class MyAnalyserParameters(BaseModel):
        value: TBaseRegion
        excitation_energy_eV: float | None = None

class MyAnalyser(AnalyserBase[MyAnalyserParameters]):
    @AsyncStatus.wrap
    async def set(self, params: MyAnalyserParameters) -> None:
        ...
```

I'm not sure which is best without understanding the use case better, I might suggest going with a plan for now, but we may decide to bake into the device later once we've learned more. It's easier to do it that way around than start with the logic baked into the device.

callumforrester (changes requested):
@oliwenmandiamond This looks promising, it may have been discussed already but I started writing comments about sticking to [PEP8](https://peps.python.org/pep-0008/#function-and-variable-names) in various places, but I've turned them into a catch-all here:

Should: Please use `snake_case` to name variables unless you have a _very_ good reason not to (sorry, I appreciate that it's hard to switch between Python and Java), for example `pass_energy` instead of `passEnergy`.

Also I'm less concerned about this one so nit: In general PEP8 recommends that you follow the case convention over the way something is normally spelt, so `excitation_energy_eV` may be the correct scientific notation but it should be `excitation_energy_ev` (lowercase `v`) in code.

callumforrester:
@DominicOram and @oliwenmandiamond re: tests I think the same can be achieved in a more Pythonic fashion with parameterized fixtures rather than classes, but without looking closer it's hard to tell. Happy to put aside some time to go through it.

src/dodal/devices/electron_analyser/specs_analyser_controller.py
```diff
@@ -0,0 +1,14 @@
+from ophyd_async.epics.core import epics_signal_rw
+
+from dodal.devices.electron_analyser.abstract_analyser_controller import (
+    AbstractAnalyserController,
+)
+
+
+class SpecsAnalyserController(AbstractAnalyserController):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        super().__init__(prefix, name)
```
DominicOram, line 10: Must: The `super()` call needs to be at the end otherwise the children do not get properly named and so when you read the analyser you do not get the `psu_mode` etc.

```python
async def test_device_has_expected_reading(
    sim_analyser: SpecsAnalyserController,
) -> None:
    reading = await sim_analyser.read()
    assert "analyser-psu_mode" in reading.keys()
```
Or, even more completely, using https://github.com/bluesky/ophyd-async/blob/2984ea49b307b661f8e9c32b66208c0d8da1940e/src/ophyd_async/testing/_assert.py#L45

src/dodal/devices/electron_analyser/vgscienta_analyser_controller.py
```diff
@@ -0,0 +1,22 @@
...
+    AbstractAnalyserController,
+)
+from dodal.devices.electron_analyser.vgscienta_region import (
+    DetectorMode,
+)
+
+
+class VGScientaAnalyserController(AbstractAnalyserController):
+    def __init__(self, prefix: str, name: str = "") -> None:
+        super().__init__(prefix, name)
```
DominicOram, line 13: Must: As above RE order of `super`

src/dodal/plan_stubs/electron_analyser/configure_controller.py
```diff
@@ -0,0 +1,78 @@
...
+    yield from bps.mv(
+        analyser.values, region.values,
+        analyser.psu_mode, region.psu_mode,
+    )
+    # fmt: on
+    if region.acquisition_mode == "Fixed Transmission":
+        yield from bps.mv(analyser.centre_energy, region.centre_energy)
+
+    if region.acquisition_mode == "Fixed Energy":
+        yield from bps.mv(analyser.energy_step, region.energy_step)
```
DominicOram, line 58: Could: It would be good to test this codepath too

tests/devices/unit_tests/electron_analyser/test_util.py
```diff
@@ -0,0 +1,29 @@
...
+    assert r.is_kinetic_energy() == is_kinetic_energy
+    assert r.is_kinetic_energy() != is_binding_energy
+
+
+def assert_region_has_expected_values(
+    r: AbstractBaseRegion, expected_region_values: dict[str, Any]
+) -> None:
+    for key in r.__dict__:
+        if key in expected_region_values:
+            assert r.__dict__[key] == expected_region_values[key]
```
DominicOram, line 27: Could: We use `DeepDiff` in other places to do comparisons like this, but up to you

DominicOram (changes requested):
Great, thanks! I think the moving of the `super` is the only blocker, otherwise all looks good.

src/dodal/devices/electron_analyser/abstract_region.py
```diff
@@ -0,0 +1,116 @@
...
+        new_data[f(key)] = values
+    return new_data
+
+
+class JavaToPythonModel(BaseModel):
+    @model_validator(mode="before")
+    @classmethod
+    def before_validation(cls, data: dict) -> dict:
+        data = switch_case_validation(data, java_to_python_case)
+        return data
```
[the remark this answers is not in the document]
oliwenmandiamond: switch_case_validation is used in AbstractBaseRegion and JavaToPythonModel. This is because if you add a model_validator in parent class and then use one in sub class, it overrides. So will keep this code.

DominicOram (approved):
Great, thank you. Just need to fix the linting. Do you have the `pre-commit` hooks installed? That would have caught the linting error I think. See https://diamondlightsource.github.io/dodal/main/how-to/lint.html#running-pre-commit

callumforrester (approved):
LGTM! Thanks for being patient :)

PR #1126, opened 2025-03-18
make turbo slit movable

stan-dot:
Work towards SWIFT testing at i20-1

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i20_1.py
+++ b/src/dodal/beamlines/i20_1.py
@@ -1,7 +1,5 @@
 from pathlib import Path
 
-from networkx import triad_type
-
 from dodal.common.beamlines.beamline_utils import (
     device_factory,
     set_path_provider,
@@ -34,6 +32,7 @@ set_path_provider(
 )
 
 
+# NOTE this is mock as we cannot move items on the beamline until we get sign-off to do so
 @device_factory(mock=True)
 def turbo_slit() -> TurboSlit:
     """
--- a/src/dodal/beamlines/i20_1.py
+++ b/src/dodal/beamlines/i20_1.py
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+from networkx import triad_type
+
 from dodal.common.beamlines.beamline_utils import (
     device_factory,
     set_path_provider,
@@ -32,7 +34,7 @@ set_path_provider(
 )
 
 
-@device_factory()
+@device_factory(mock=True)
 def turbo_slit() -> TurboSlit:
     """
     turboslit for selecting energy from the polychromator
--- a/src/dodal/devices/turbo_slit.py
+++ b/src/dodal/devices/turbo_slit.py
@@ -1,9 +1,9 @@
-from bluesky.protocols import Movable, Readable
-from ophyd_async.core import AsyncStatus
+from bluesky.protocols import Movable
+from ophyd_async.core import AsyncStatus, StandardReadable
 from ophyd_async.epics.motor import Motor
 
 
-class TurboSlit(Movable, Readable):
+class TurboSlit(StandardReadable, Movable):
     """
     This collection of motors coordinates time resolved XAS experiments.
     It selects a beam out of the polychromatic fan.
--- a/src/dodal/devices/turbo_slit.py
+++ b/src/dodal/devices/turbo_slit.py
@@ -3,7 +3,7 @@ from ophyd_async.core import AsyncStatus, StandardReadable
 from ophyd_async.epics.motor import Motor
 
 
-class TurboSlit(StandardReadable, Movable):
+class TurboSlit(StandardReadable, Movable[float]):
     """
     This collection of motors coordinates time resolved XAS experiments.
     It selects a beam out of the polychromatic fan.
--- a/src/dodal/devices/turbo_slit.py
+++ b/src/dodal/devices/turbo_slit.py
@@ -1,8 +1,9 @@
-from ophyd_async.core import StandardReadable
+from bluesky.protocols import Movable, Readable
+from ophyd_async.core import AsyncStatus
 from ophyd_async.epics.motor import Motor
 
 
-class TurboSlit(StandardReadable):
+class TurboSlit(Movable, Readable):
     """
     This collection of motors coordinates time resolved XAS experiments.
     It selects a beam out of the polychromatic fan.
@@ -23,3 +24,15 @@ class TurboSlit(StandardReadable):
             self.arc = Motor(prefix=prefix + "ARC")
             self.xfine = Motor(prefix=prefix + "XFINE")
         super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def set(self, value: float):
+        """this will move the default XFINE"""
+        await self.xfine.set(value)
+
+    @AsyncStatus.wrap
+    async def set_full(self, gap: float, arc: float, xfine: float):
+        """TODO double check the operation order here"""
+        await self.gap.set(gap)
+        await self.arc.set(arc)
+        await self.xfine.set(xfine)
--- a/src/dodal/devices/turbo_slit.py
+++ b/src/dodal/devices/turbo_slit.py
@@ -29,10 +29,3 @@ class TurboSlit(StandardReadable, Movable):
     async def set(self, value: float):
         """this will move the default XFINE"""
         await self.xfine.set(value)
-
-    @AsyncStatus.wrap
-    async def set_full(self, gap: float, arc: float, xfine: float):
-        """TODO double check the operation order here"""
-        await self.gap.set(gap)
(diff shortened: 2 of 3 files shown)


src/dodal/beamlines/i20_1.py
```diff
@@ -1,5 +1,7 @@
 from pathlib import Path
 
+from networkx import triad_type
```
DominicOram, line 3: Should: Was this added accidentally? Not sure where is it is used?

src/dodal/devices/turbo_slit.py
```diff
@@ -1,8 +1,9 @@
-from ophyd_async.core import StandardReadable
+from bluesky.protocols import Movable, Readable
+from ophyd_async.core import AsyncStatus
 from ophyd_async.epics.motor import Motor
 
 
-class TurboSlit(StandardReadable):
+class TurboSlit(Movable, Readable):
```
DominicOram, line 6: Must: This is throwing type errors in `i20_1` as it needs to be a `HasName` and implement `read`, additionally it means that . I think it probably needs to be:
```suggestion
class TurboSlit(StandardReadable, Movable):
```

src/dodal/devices/turbo_slit.py
```diff
@@ -23,3 +24,15 @@ def __init__(self, prefix: str, name: str = ""):
...
             self.xfine = Motor(prefix=prefix + "XFINE")
         super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def set(self, value: float):
+        """this will move the default XFINE"""
+        await self.xfine.set(value)
+
+    @AsyncStatus.wrap
+    async def set_full(self, gap: float, arc: float, xfine: float):
```
DominicOram, line 34: Should: Where are you using `set_full`? You will be unable to actually call this function from a plan

src/dodal/beamlines/i20_1.py
```diff
@@ -32,7 +34,7 @@
 )
 
 
-@device_factory()
+@device_factory(mock=True)
```
DominicOram, line 37: Could: Why is this mocked? Might be useful to add a comment

DominicOram (changes requested):
A few comments :)

src/dodal/beamlines/i20_1.py
stan-dot: no idea, maybe fallacious automatic import

src/dodal/devices/turbo_slit.py
stan-dot: why would I be unable? that is if we'd like to use the full range of PVs for the device. it's used sometimes even though the default set (xfine) is the one used most often

src/dodal/beamlines/i20_1.py
stan-dot: because we are banned from moving things at the beamline atm

stan-dot:
```python
+ 
+     @AsyncStatus.wrap
+     async def set_full(self, gap: float, arc: float, xfine: float):
+         """TODO double check the operation order here"""
+         await self.gap.set(gap)
+         await self.arc.set(arc)
+         await self.xfine.set(xfine)
```


codecov fails due to no tests for the set_full method - @iain-hall what would be a realistic scenario when we use this?

src/dodal/beamlines/i20_1.py
DominicOram: Ok, my recommendation would be to just not put it on the beamline then. The issue with marking this as mocked is that you can't do `dodal connect i20-1` to test it. But it's up to you

src/dodal/devices/turbo_slit.py
DominicOram: So I can call the `set` method in a plan by doing `yield from bps.abs_set(turbo_slits, 1.234)` or ` yield from bps.mv(turbo_slits, 1.234)` but there is no `bluesky` verb to allow you to call `set_full`. To me this looks like logic that should be in the plan e.g.
```python
def move_turbo_slit(turbo_slit: TurboSlit, gap: float, arc: floar, xfine:float):
    yield from bps.mv(turbo_slit.gap, gap)
    yield from bps.mv(turbo_slit.arc, arc)
    yield from bps.mv(turbo_slit.xfine, xfine)
```
note that this will move them all sequentially and wait for each to move before doing the next, I assume you want to do this as it's how you've written `set_full` but I would recommend moving them in parallel if there's not hardware reason you can't. The only time I would move this logic into the device is if the axes must be moved in a particular order for some hardware reason, particularly if it's a safety issue, is that the case here?

src/dodal/beamlines/i20_1.py
stan-dot: but we *want* to run it in mock-mode only on a beamline while the other devices are real

stan-dot:
okay, removing the lines; we'll do this in an alignment plan

src/dodal/devices/turbo_slit.py
```diff
@@ -1,8 +1,9 @@
-from ophyd_async.core import StandardReadable
+from bluesky.protocols import Movable
+from ophyd_async.core import AsyncStatus, StandardReadable
 from ophyd_async.epics.motor import Motor
 
 
-class TurboSlit(StandardReadable):
+class TurboSlit(StandardReadable, Movable):
```
DiamondJoseph, line 6:
```suggestion
class TurboSlit(StandardReadable, Movable[float]):
```

Now that bluesky 1.13.1 is in again, should reflect that

DiamondJoseph: But also, the set method just defers to a child device- blueapi can get the child devices from a string (e.g. "turbo.xfine"), so is this necessary?

DiamondJoseph: How are they using the device currently? If the plans are written to take the `turbo_slit` device directly and do the mapping to appropriate Movables does that achieve what they want?

I really think that if you have a device with 3 children, and calling set on that device calls set on exactly 1 of the children with the same value, that's not the correct design for the device.

stan-dot:
just need to add a simple test

src/dodal/devices/turbo_slit.py
stan-dot: deleted the `set_full` function

src/dodal/devices/turbo_slit.py
stan-dot: > I really think that if you have a device with 3 children, and calling set on that device calls set on exactly 1 of the children with the same value, that's not the correct design for the device.

please justify the statement. you put forward and abstract argument while the design I implemented reflects the word usage of spectroscopy scientists. they say to set slit to something, not slit xfine during the scan. the other two children are only set during alignment, not scans.

I feel you're saying opinionated things knowing less context than the people who you say are wrong.

stan-dot: out of date, applied the fix for float

src/dodal/devices/turbo_slit.py
```diff
@@ -23,3 +24,8 @@ def __init__(self, prefix: str, name: str = ""):
             self.arc = Motor(prefix=prefix + "ARC")
             self.xfine = Motor(prefix=prefix + "XFINE")
         super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def set(self, value: float):
+        """this will move the default XFINE"""
```
DominicOram, line 30: Nit: 
```suggestion
        """This will move the default XFINE"""
```

src/dodal/devices/turbo_slit.py
DominicOram: >I really think that if you have a device with 3 children, and calling set on that device calls set on exactly 1 of the children with the same value, that's not the correct design for the device.

I think this is probably one of those things that comes down to taste but wouldn't mind if we had a rule that said we couldn't do it. We do it in MX for a few things where it's very clear from the scientist's perspective that a device only has one thing it can do e.g. we have `Backlight` where we let the user do `backlight.set(IN)`. I think this particular case is less clear but I have no strong opinion.

However, I do think that some of the discussion here was verging on the rude. The purpose of reviews is to try and make the codebase _that we all have some responsibility for_ to be the best it can be and we all ultimately want the same thing. There's no need to get adversarial or defensive.

src/dodal/devices/turbo_slit.py
DominicOram: Could: It would also be useful to know the units that value should be in.

DominicOram (approved):
Thanks. Couple of minor comments. Otherwise looks good to me.

src/dodal/devices/turbo_slit.py
stan-dot: referenced in a new issue

Issue #1141, opened 2025-03-25
Fix tests from new ophyd-async version


DominicOram:
Using the new ophyd-async version has caused the errors seen in https://github.com/DiamondLightSource/dodal/actions/runs/14060107542/job/39368539114?pr=1068

We should fix these

## Acceptance Criteria
* Tests pass on latest `ophyd-async`

PR #1142, opened 2025-03-25
Pin ophyd-async to 0.9.0 until compatibility issues are fixed

rtuck99:
Temporarily pin ophyd-async until issues with new ophyd-async in #1141 are fixed


DominicOram (approved):
Looks good, assuming tests pass

PR #1143, opened 2025-03-25
Update to latest ophyd async

### LINKED ISSUE #1141: Fix tests from new ophyd-async version

DominicOram:
Fixes #1141

As part of this I noticed that a number of inner functions of Devices were `AsyncStatus.wrap` decorated. This is only required on external interfaces like `set` and so I have tidied it up.

1. Confirm tests pass again

### Checks for reviewer


DominicOram:
FYI @coretl

DominicOram:
Type issues that are left are, thoughts @coretl and @callumforrester?:
* The merlin is trying to use it's own `ImageMode` (https://github.com/DiamondLightSource/dodal/blob/f135ed47e8d9da4b7a1293d1c1c0380506b51c92/src/dodal/devices/i13_1/merlin_io.py#L6) - What is the best way to do this in the `StandardDetector`?
* `Motor` is no longer `Readable` but instead `AsyncReadable` this seems potentially bad as if I want to write a plan that just works against an arbitrary `Readable` I can't pass it in a `Motor`, is this expected?

coretl:
>     * The merlin is trying to use it's own `ImageMode` (https://github.com/DiamondLightSource/dodal/blob/f135ed47e8d9da4b7a1293d1c1c0380506b51c92/src/dodal/devices/i13_1/merlin_io.py#L6
>       ) - What is the best way to do this in the `StandardDetector`?

`ADImageMode` has changed to a `SubsetEnum`, so there is no need to reimplement this in `ADMerlin`, it can just use the baseclass and those extra enums will be supported. They aren't used in logic so that's fine.

>     * `Motor` is no longer `Readable` but instead `AsyncReadable` this seems potentially bad as if I want to write a plan that just works against an arbitrary `Readable` I can't pass it in a `Motor`, is this expected?

I didn't think that had changed? The only change I can see is the addition of protocols, but `StandardReadable` is `AsyncReadable`, and always has been. 

https://github.com/bluesky/ophyd-async/commit/befee4271304352b815ed6730d143c13f2919d50#diff-e9eaed6f80e09c47971dcf09f4aefec5a6373de640d80c35e02a579b7a837878

The `AsyncReadable` is really a narrowing of `Readable` (which accepts both sync and async functions), but I'm not sure how to communicate that to the type system. What was the error you get?

PR #1144, opened 2025-03-26
Add baton device

DominicOram:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/358

1. See `mx-bluesky` PR

### Checks for reviewer


src/dodal/devices/baton.py
```diff
@@ -0,0 +1,17 @@
+from typing import Annotated as A
+
```
rtuck99, line 2: Is it a deliberate decision not to include this in the beamline `i03.py`?

DominicOram: It doesn't yet exist on the beamline but I will add it in with an appropriate `skip` I think

jacob720 (approved):
Looks good

noemifrisina (approved):

PR #1149, opened 2025-03-28
Mx bluesky 136 get murko results

jacob720:
Fixes #[136](https://github.com/DiamondLightSource/mx-bluesky/issues/136) from mx-bluesky

1. Run tests
2. Check they pass
3. If possible, run on the beamline to get actual results from murko.

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -353,7 +353,7 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
         name="oav_to_redis_forwarder",
         redis_host=REDIS_HOST,
         redis_password=REDIS_PASSWORD,
-        redis_db=7,
+        db=7,
     )
 
 
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -17,6 +17,7 @@ from dodal.devices.diamond_filter import DiamondFilter, I04Filters
 from dodal.devices.eiger import EigerDetector
 from dodal.devices.fast_grid_scan import ZebraFastGridScan
 from dodal.devices.flux import Flux
+from dodal.devices.i04.murko_results import MurkoResultsDevice
 from dodal.devices.i04.transfocator import Transfocator
 from dodal.devices.ipin import IPin
 from dodal.devices.motors import XYZPositioner
@@ -349,6 +350,20 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
     """
     return OAVToRedisForwarder(
         f"{PREFIX.beamline_prefix}-DI-OAV-01:",
+        name="oav_to_redis_forwarder",
+        redis_host=REDIS_HOST,
+        redis_password=REDIS_PASSWORD,
+        redis_db=7,
+    )
+
+
+@device_factory()
+def murko_results() -> MurkoResultsDevice:
+    """Get the i04 OAV to redis forwarder, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return MurkoResultsDevice(
+        "",
         name="oav_to_redis_forwarder",
         redis_host=REDIS_HOST,
         redis_password=REDIS_PASSWORD,
--- a/src/dodal/devices/i04/murko_results.py
+++ b/src/dodal/devices/i04/murko_results.py
@@ -103,8 +103,6 @@ class MurkoResultsDevice(StandardReadable, Triggerable, Stageable):
         omega: float,
         search_angle: float,
     ) -> np.ndarray | None:
-        print(omega)
-        print(search_angle)
         LOGGER.info(f"Compare {omega}, {search_angle}, {self._last_omega}")
         if (  # if last omega is closest
             abs(omega - search_angle) >= abs(self._last_omega - search_angle)
--- a/src/dodal/devices/i04/murko_results.py
+++ b/src/dodal/devices/i04/murko_results.py
@@ -35,10 +35,11 @@ class MurkoResultsDevice(StandardReadable, Triggerable, Stageable):
 
     def __init__(
         self,
+        prefix="",
         redis_host=REDIS_HOST,
         redis_password=REDIS_PASSWORD,
         db=MURKO_REDIS_DB,
-        redis_channel="murko",
+        name="",
     ):
         self.redis_client = StrictRedis(
             host=redis_host,
@@ -58,6 +59,7 @@ class MurkoResultsDevice(StandardReadable, Triggerable, Stageable):
             self.x_um, self._x_um_setter = soft_signal_r_and_setter(float | None)
             self.y_um, self._y_um_setter = soft_signal_r_and_setter(float | None)
             self.z_um, self._z_um_setter = soft_signal_r_and_setter(float | None)
+        super().__init__(name=name)
 
     @AsyncStatus.wrap
     async def stage(self):
@@ -66,7 +68,6 @@ class MurkoResultsDevice(StandardReadable, Triggerable, Stageable):
         self._x_um_setter(None)
         self._y_um_setter(None)
         self._z_um_setter(None)
-        self.found_90 = False
 
     @AsyncStatus.wrap
     async def unstage(self):
--- a/src/dodal/devices/i04/murko_results.py
+++ b/src/dodal/devices/i04/murko_results.py
@@ -1,0 +1,131 @@
+import json
+import pickle
+from typing import TypedDict
+
+from bluesky.protocols import Stageable, Triggerable
+from ophyd_async.core import (
+    AsyncStatus,
+    StandardReadable,
+    soft_signal_r_and_setter,
+    soft_signal_rw,
+)
+from redis import StrictRedis
+
+from dodal.beamlines.i04 import MURKO_REDIS_DB, REDIS_HOST, REDIS_PASSWORD
+from dodal.devices.oav.oav_calculations import camera_coordinates_to_xyz
+from dodal.log import LOGGER
+
+MurkoResult = dict
+FullMurkoResults = dict[str, list[MurkoResult]]
+
+
+class MurkoMetadata(TypedDict):
+    zoom_percentage: float
+    microns_per_x_pixel: float
+    microns_per_y_pixel: float
+    beam_centre_i: float
+    beam_centre_j: float
+    sample_id: str
+    omega_angle: float
+    uuid: str
+
+
+class MurkoResultsDevice(StandardReadable, Triggerable, Stageable):
+    TIMEOUT_S = 2
+
+    def __init__(
+        self,
+        redis_host=REDIS_HOST,
+        redis_password=REDIS_PASSWORD,
+        db=MURKO_REDIS_DB,
+        redis_channel="murko",
+    ):
+        self.redis_client = StrictRedis(
+            host=redis_host,
+            password=redis_password,
+            db=db,
+        )
+        self.pubsub = self.redis_client.pubsub()
+        self._last_omega = 0
+        self.sample_id = soft_signal_rw(str)  # Should get from redis
+        self.found_90 = False
+        self.current_x_y_z_um = soft_signal_rw(
+            tuple[float, float, float]
+        )  # Should get from redis
+
+        with self.add_children_as_readables():
+            # Diffs from current x/y/z
+            self.x_um, self._x_um_setter = soft_signal_r_and_setter(float | None)
+            self.y_um, self._y_um_setter = soft_signal_r_and_setter(float | None)
+            self.z_um, self._z_um_setter = soft_signal_r_and_setter(float | None)
+
+    @AsyncStatus.wrap
+    async def stage(self):
+        # Subscribe to redis pub/
+        self.pubsub.subscribe("murko-results")
+        self._x_um_setter(None)
+        self._y_um_setter(None)
+        self._z_um_setter(None)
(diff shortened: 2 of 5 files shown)


tests/devices/i04/test_murko_results.py
```diff
@@ -0,0 +1,497 @@
...
+        patch.object(
+            fake_murko_results,
+            "get_beam_centre",
+            side_effect=lambda *args, **kwargs: (50, 50),
+        ) as mock_get_beam_centre,
+    ):
+        fake_murko_results.pubsub.get_message = mock_get_message
+        fake_murko_results.redis_client.hget = mock_hget
+        fake_murko_results.get_beam_centre = mock_get_beam_centre
+        await fake_murko_results.trigger()
```
DominicOram, line 121: Could: This is very similar in the other tests, I would pull into a helper function.

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,211 @@
...
+    uuid: str
+
+
+class Coord(Enum):
+    x = 0
+    y = 1
+    z = 2
+
+
+class MurkoResultsDevice(StandardReadable, Triggerable, Stageable):
```
DominicOram, line 50: Should: Can we get a docstring describing the class? Then somewhere some detail on the logic of what we're doing, maybe here or in some of the calc methods

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,211 @@
...
+            password=redis_password,
+            db=db,
+        )
+        self.pubsub = self.redis_client.pubsub()
+        self._last_omega = 0
+        self._last_result = None
+        self.sample_id = soft_signal_rw(str)  # Should get from redis
+        self.coords = {"x": {}, "y": {}, "z": {}}
+        self.search_angles = OrderedDict(
+            [(90, ("x", "z")), (180, ("x", "y")), (270, ())]
```
DominicOram, line 72: What's the 270 one? Where we stop?

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,211 @@
...
+        return next_angle
+
+    def process_result(
+        self, result: dict, uuid: int, metadata_str: str, search_angle: float | None
+    ) -> float:
+        if search_angle is None:
+            return None
+        metadata = MurkoMetadata(
+            json.loads(metadata_str)
+        )  # Metadata is JSON but data is pickle?
```
DominicOram, line 191: Yes, the metadata is small amounts of simple data, the `data` itself is large numpy arrays so doesn't json as well

DominicOram (commented):
This is definitely looking nicer than what I had originally. I think it could do with some more docs, both in the code and if you can update the general murko docs too that would be good

tests/devices/i04/test_murko_results.py
jacob720: Have done this

src/dodal/devices/i04/murko_results.py
jacob720: Have added this, let me know if it's not clear enough

src/dodal/devices/i04/murko_results.py
jacob720: Yeah where we stop, I remember you mentioning we want to continue past 180°, but maybe 270° is too far. We could also collect another x and potentially z value at 270° (or the closest omega to 270°) and get a better average?

jacob720:
I've added to the docs in mx-bluesky https://github.com/DiamondLightSource/mx-bluesky/tree/136_use_murko_results

src/dodal/beamlines/i04.py
```diff
@@ -352,6 +353,20 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
         name="oav_to_redis_forwarder",
         redis_host=REDIS_HOST,
         redis_password=REDIS_PASSWORD,
+        db=7,
```
DominicOram, line 356: Must: Running `dodal connect i04` gives an error on `oav_to_redis_forwarder` and on `murko_results` with `db` and `redis_db` not being recognised respectively. I think you've mixed the keyword names up. Additionally, these are the same thing anyway so we should have the same keyword name for both, don't mind which it is

src/dodal/beamlines/i04.py
```diff
@@ -352,6 +353,20 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
...
+    )
+
+
+@device_factory()
+def murko_results() -> MurkoResultsDevice:
+    """Get the i04 OAV to redis forwarder, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return MurkoResultsDevice(
+        "",
```
DominicOram, line 366: Could: `prefix` is usually for the PV prefix, `murko_results` doesn't connect to PVs so the prefix is meaningless. I think you can remove it here and in the device.

src/dodal/beamlines/i04.py
```diff
@@ -352,6 +353,20 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
...
+        name="oav_to_redis_forwarder",
```
DominicOram, line 367: Must: This needs a unique name and it's not the forwarder.

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+from bluesky.protocols import Stageable, Triggerable
+from ophyd_async.core import (
+    AsyncStatus,
+    StandardReadable,
+    soft_signal_r_and_setter,
+    soft_signal_rw,
+)
+from redis import StrictRedis
+
+# from dodal.beamlines.i04 import MURKO_REDIS_DB, REDIS_HOST, REDIS_PASSWORD
```
DominicOram, line 18: Must: This is commented out because of the circular import, right? I think we still need to get the constants from a common place. Can we put them in a `constants` file somewhere.

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+        self._y_mm_setter(None)
+        self._z_mm_setter(None)
+
+    @AsyncStatus.wrap
+    async def unstage(self):
+        # Unsubscribe
+        await self.pubsub.unsubscribe()
+
+    def get_beam_centre(self) -> tuple[int, int]:
+        return (442, 363)
```
DominicOram, line 109: Should: We can get this out of the `MurkoMetadata` instead.

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+        with self.add_children_as_readables():
+            # Diffs from current x/y/z
+            self.x_mm, self._x_mm_setter = soft_signal_r_and_setter(float)
+            self.y_mm, self._y_mm_setter = soft_signal_r_and_setter(float)
+            self.z_mm, self._z_mm_setter = soft_signal_r_and_setter(float)
+        super().__init__(name=name)
+
+    @AsyncStatus.wrap
+    async def stage(self):
+        # Subscribe to redis pub/
```
DominicOram, line 97: Nit: I would say this comment doesn't add much

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+            beam_dist_px[0],
+            beam_dist_px[1],
+            closest_omega,
+            metadata["microns_per_x_pixel"],
+            metadata["microns_per_y_pixel"],
+        )
+
+    @AsyncStatus.wrap
+    async def trigger(self):
+        # May be better as kickoff/complete?
```
DominicOram, line 156: Should: Leaving comments like this just confuses future devs. We should either:
* Make it a `kickoff/complete` if we think it's better
* Decide that trigger is best and remove the comment
* If we want more discussion make an issue to discuss it

In all cases the comment should be removed

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+        # Wait for results
+        sample_id = await self.sample_id.get_value()
+        next_angle = self.angles_to_search.pop(0)
+        final_message = None
+        while next_angle:  # Goes round in a loop until 270 degrees is found
+            # waits here for next batch to be recieved (10 scans?). pickle deserialises data
```
DominicOram, line 162: Nit: Again, number of scans per batch is variable so saying it's 10 here may just lead to confusion later. Also not sure what the `pickle deserialises data` bit gives us

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+        assert y_values, "No y values"
+        self._x_mm_setter(np.mean(x_values))
+        self._y_mm_setter(np.mean(y_values))
+        self._z_mm_setter(np.mean(z_values))
+
+    async def process_batch(self, message, sample_id, next_angle):
+        if message and message["type"] == "message":
+            batch_results = pickle.loads(message["data"])
+            for results in batch_results:
+                print(f"Got {results} from redis")
```
DominicOram, line 185: Should: Use the logger rather than printing

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+                        next_angle = self.process_result(
+                            result, uuid, metadata_str, next_angle
+                        )
+        return next_angle
+
+    def process_result(
+        self, result: dict, uuid: int, metadata_str: str, search_angle: float | None
+    ) -> float:
+        if search_angle is None:
+            return None
```
DominicOram, line 200: Should: The return type on here should be `float | None`, I'm surprised the type system is happy how it is

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+        y_values = list(self.coords["y"].values())
+        z_values = list(self.coords["z"].values())
+        assert x_values, "No x values"
+        assert z_values, "No z values"
+        assert y_values, "No y values"
+        self._x_mm_setter(np.mean(x_values))
+        self._y_mm_setter(np.mean(y_values))
+        self._z_mm_setter(np.mean(z_values))
+
+    async def process_batch(self, message, sample_id, next_angle):
```
DominicOram, line 181: Could: type hints on this.

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+                for uuid, result in results.items():
+                    if metadata_str := await self.redis_client.hget(
+                        f"murko:{sample_id}:metadata", uuid
+                    ):
+                        next_angle = self.process_result(
+                            result, uuid, metadata_str, next_angle
+                        )
+        return next_angle
+
+    def process_result(
```
DominicOram, line 196: Should: It seems a bit odd that `process_results` and `process_batch` return the next angle to be searched and that we `pop` off the different angles in two different places. Is it possible to get an architecture where we do that once and don't need to pass it back and forth so much?

tests/devices/i04/test_murko_results.py
```diff
@@ -0,0 +1,360 @@
...
+                batches=3,
+                messages_per_batch=2,
+                images_per_message=2,
+                start_x=0.5,
+                step_x=0.01,
+                start_y=0.3,
+                step_y=0.02,
+            )
+        )
+    )  # 2 messages, both (0.5, 0.5)
```
DominicOram, line 286: Must: This comment is wrong?

DominicOram (changes requested):
Sorry, a few more comments. On your tests, I feel a bit like they're over commented, my concern when you have lots of comments with complicated maths like that is that:
* It's not that readable
* Often code can change without the comment also changing, in a case like this you end up with lots of potential to confuse

I would normally prefer doing the maths as actual code in the tests too or making variables/function names self documenting but I appreciate that these may be too complex to do that with. It's also just tests so less of a big deal than prod code so I will leave it with you if you think you can tidy it but no big deal

src/dodal/devices/i04/murko_results.py
jacob720: Have put them in `constants.py` in `dodal/devices/i04/` but let me know if there's a more appropriate place.

src/dodal/beamlines/i04.py
jacob720: This is fixed now

src/dodal/devices/i04/murko_results.py
jacob720: I've tweaked this to it's not passed back and forth anymore. Any function can find the search angle by looking at index 0 of `self.angles_to_search`. It's only popped once now, once the closest omega to that search angle has been found.

src/dodal/devices/i04/murko_results.py
```diff
@@ -0,0 +1,221 @@
...
+            )
+        ) is not None:
+            LOGGER.info(f"Using result {uuid}, {metadata_str}, {result}")
+            for coord in self.search_angles[search_angle]:
+                self.coords[coord][omega_angle] = movement[Coord[coord].value]
+            if self.angles_to_search:
+                search_angle = self.angles_to_search.pop(0)
+            else:
+                search_angle = None
+            LOGGER.info(f"Setting {search_angle} to {movement}")
```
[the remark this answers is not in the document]
jacob720: I've deleted this log

DominicOram (approved):
Great, thanks! I will make a new PR addressing some of my comments and we can get this merged :)

PR #1155, opened 2025-03-31
Allow interpolation in undulator gap

### LINKED ISSUE #1154: Interpolate between undulator gaps
When an energy is set the undulator gap is calculated from a look up table. Currently this uses the value in the table closest to the selected energy. However, it should interpolate between the closest two values in the lookup table. The table is a little odd in it sometimes has multiple values for the same energy eg:
```
energy, gap
10000, 5
15000, 8
15000, 3
16000, 5
```

This signifies that between 10000 and 15000 a gap should be interpolated from between 5 and 8 but between 15000 and 16000 a gap should be selected between 3 and 5.

Additionally, there should be more  logging on the device as it currently just says "selecting gap for energy of x" without saying what the gap it's selecting is.

## Acceptance Criteria
* Undulator gap is interpolated when an energy is selected between two values in the lookup table.

DominicOram:
Fixes #1154

1. Confirm tests match the issue description
2. Confirm tests now pass

### Checks for reviewer


stan-dot (approved):
lgtm

RJCD-Diamond (commented):
What is the desired gap at exactly 15000 ev? 3 or 8? This code will return 3 at 15000ev. If that's expected and desired then lgtm

DominicOram:
> What is the desired gap at exactly 15000 ev? 3 or 8? This code will return 3 at 15000ev. If that's expected and desired then lgtm

Discussed w/ @npat113. The answer was that at this transition energy the flux should be basically the same regardless of which harmonic is chosen so it doesn't matter

npat113:
Hold up - gap value of 3? Can't do that.

npat113:
15000eV has a single entry of 5.987

npat113:
Yes, at exact transition either value is appropriate. The double entry is there to allow interpolation above and below the transition

DominicOram:
Oh, sorry, that was just my nonsense made up numbers to prove the concept. In prod we use the numbers in `/dls_sw/i03/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt` and so cannot go to 3. Apologies for the confusion.

RJCD-Diamond (approved):

PR #1161, opened 2025-04-03
Create a common base device that sends requests to blueapi on I19 optics 

### LINKED ISSUE #1153: Create a common `BlueAPIBackedDevice` for I19
Create a `BlueAPIBackedDevice` that optics hutch devices for I19 can inherit from. This device should implement the logic to make a REST call to the optics blueapi running on the beamline on set. 
See https://github.com/DiamondLightSource/dodal/blob/f9830fc1eb468ef45e3c62153d1507eb78bdd8cc/src/dodal/devices/i19/shutter.py#L60

## Acceptance Criteria
* There is a separate device implementing ClientSession and the shutter (and any future optics hutch device) inherits from it

noemifrisina:
Closes #1153 

1. Run tests

### Checks for reviewer


src/dodal/devices/i19/blueapi_device.py
```diff
@@ -0,0 +1,102 @@
...
+    hutch running on the I19 cluster, which will evaluate the current hutch in use vs \
+    the hutch sending the request and decide if the plan will be run or not.
+
+    For details see the architecture described in \
+    https://github.com/DiamondLightSource/i19-bluesky/issues/30.
+    """
+
+    def __init__(self, name: str = "") -> None:
+        self.url = OPTICS_BLUEAPI_URL
+        self.headers = HEADERS
```
callumforrester, line 34: Could: Dependency injection, might make tests easier, also means the actual URL can be defined in `dodal.beamlines.i19_1/2.py`, which I think is probably better.

```suggestion
    def __init__(self, name: str = "", url: str | None = None, headers: dict[str, str] | None = None) -> None:
        self.url = url or OPTICS_BLUEAPI_URL
        self.headers = headers or HEADERS
```

src/dodal/devices/i19/blueapi_device.py
```diff
@@ -0,0 +1,102 @@
...
+            while not plan_complete:
+                async with session.get(f"/tasks/{task_id}") as res:
+                    plan_result = await res.json()
+                    plan_complete = plan_result["is_complete"]
+                    errors = plan_result["errors"]
+                    if len(errors) > 0:
+                        message = "\n".join(errors)
+                        LOGGER.error(f"Plan {value['name']} failed: {message}")  # type:ignore
+                        raise RuntimeError(f"Plan failed with error: {message}")
+                    time.sleep(interval)  # noqa
```
callumforrester, line 101: Must: This is an async function so asyncio sleep is fine

```suggestion
                    await asyncio.sleep(interval)
```

Also could: Move outside of the context block, otherwise you're keeping the HTTP connection alive while you're not doing anything (if my read of the [aiohttp client docs](https://docs.aiohttp.org/en/stable/client_advanced.html) is correct).

```suggestion
                await asyncio.sleep(interval)
```

src/dodal/devices/i19/shutter.py
```diff
@@ -39,7 +26,7 @@ def __init__(self, prefix: str, hutch: HutchState, name: str = "") -> None:
         with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
             self.shutter_status = epics_signal_r(ShutterState, f"{prefix}STA")
         self.hutch_request = hutch
-        self.url = OPTICS_BLUEAPI_URL
+        # self.url = OPTICS_BLUEAPI_URL
```
callumforrester, line 29: Should: Remove

callumforrester (changes requested):

src/dodal/devices/i19/blueapi_device.py
noemifrisina: I'm not entirely sure about this tbh, it would mean adding the option for eg. the url -which is pretty much a constant and has to be the same for both hutches anyway - to every device that inherits from this one to make it work. It's okay for the shutter - but thinking of future devices, it might end up increasing the probability of making a mistake.

callumforrester: I suppose if you ever need different URLs it's a refactor anyway, was just going to parity with PVs. In a dodal beamline module you define the PV prefixes, I would expect the URL prefixes for HTTP devices in the same place. Not fussed though.

callumforrester (approved):

PR #918, opened 2024-11-26
MX Bluesky 651 fix energy changes

rtuck99:
Dodal changes for DiamondLightSource/mx-bluesky#651

See MX-bluesky PR DiamondLightSource/mx-bluesky#676

### Instructions to reviewer on how to test:
1. Energy changes work on beamline
2. Unit tests pass

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://diamondlightsource.github.io/dodal/main/reference/device-standards.html)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

PR #1162, opened 2025-04-04
Add required grid-detect devices to i04

jacob720:
Part of [#918](https://github.com/DiamondLightSource/mx-bluesky/issues/918) in `mx-bluesky`

The i03 beamstop device has been moved to a new directory, `dodal/devices/mx_phase1` so that the code can be shared with i04. This is so the `grid_detect` plan can run on i04.

1. Confirm all relevant files have been moved
2. Confirm functionality is maintained

### Checks for reviewer


src/dodal/beamlines/i04.py
```diff
@@ -107,13 +116,14 @@ def ipin() -> IPin:
...
+def beamstop() -> Beamstop:
     """Get the i04 beamstop device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i04, it will return the existing object.
     """
-    return XYZPositioner(
+    return Beamstop(
         f"{PREFIX.beamline_prefix}-MO-BS-01:",
-        "beamstop",
+        beamline_parameters=get_beamline_parameters(),
+        name="beamstop",
```
DominicOram, line 126: Could: `device_factory` by default will fill in the `name` for us so I think we can just not provide it.

src/dodal/beamlines/i04.py
```diff
@@ -359,10 +369,56 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
...
+@device_factory()
+def panda() -> HDFPanda:
+    """Get the i04 panda device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return HDFPanda(
+        f"{PREFIX.beamline_prefix}-EA-PANDA-01:",
+        path_provider=get_path_provider(),
+        name="panda",
+    )
```
DominicOram, line 402: Should: I don't think we have these on i04 at the moment and they seem unrelated to the beamstop issue anyway, can we remove for now?

src/dodal/beamlines/i04.py
```diff
@@ -359,10 +369,56 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
...
+
+@device_factory()
+def zocalo() -> ZocaloResults:
+    """Get the i04 ZocaloResults device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return ZocaloResults(
+        name="zocalo",
+        prefix=PREFIX.beamline_prefix,
+    )
```
DominicOram, line 413: Should: `ZocaloResults` doesn't need a prefix and as above, name will get filled in for us, so we could just do:
```suggestion
    return ZocaloResults()
```

src/dodal/beamlines/i04.py
```diff
@@ -359,10 +369,56 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
...
+
+
+@device_factory(skip=BL == "s03")
+def pin_tip_detection() -> PinTipDetection:
+    """Get the i04 pin tip detection device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return PinTipDetection(
+        f"{PREFIX.beamline_prefix}-DI-OAV-01:",
+        "pin_tip_detection",
```
DominicOram, line 423: Could: As above, we don't need to provide a name

src/dodal/beamlines/i04.py
```diff
@@ -359,10 +369,56 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
...
+    """Get the i04 ZocaloResults device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return ZocaloResults(
+        name="zocalo",
+        prefix=PREFIX.beamline_prefix,
+    )
+
+
+@device_factory(skip=BL == "s03")
```
DominicOram, line 416: Should: s03 isn't a thing for i04, it isn't really a thing any more at all so we can remove this skip condition

DominicOram (changes requested):
Thanks, some comments in code but otherwise good.

src/dodal/beamlines/i04.py
jacob720: Have removed these

src/dodal/beamlines/i04.py
jacob720: I've fixed this now

src/dodal/beamlines/i04.py
jacob720: Have done this

src/dodal/beamlines/i04.py
jacob720: Have removed this

src/dodal/beamlines/i04.py
```diff
@@ -359,10 +365,28 @@ def oav_to_redis_forwarder() -> OAVToRedisForwarder:
...
         data_type=I04Filters,
     )
+
+
+@device_factory()
+def zocalo() -> ZocaloResults:
+    """Get the i04 ZocaloResults device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return ZocaloResults()
```
olliesilvester, line 382: After testing we noticed that this device has a channel which defaults to `"xrc.i03"`. We should instantiate it with 

```suggestion
    return ZocaloResults("xrc.i04")
```

src/dodal/beamlines/i04.py
olliesilvester: This is still here - maybe you forgot to push?

olliesilvester (changes requested):
A few small comments, Also, I think the PR title would be better as something like "Add required grid-detect devices to i04"

src/dodal/beamlines/i04.py
jacob720: Not sure what happened there, have fixed it now

src/dodal/beamlines/i04.py
jacob720: Have added this

olliesilvester: This needs to be a named argument, `return ZocaloResults(channel="xrc.i04")`

olliesilvester (approved):
Thanks, looks good

PR #1165, opened 2025-04-07
Get detector data from analyser drivers

oliwenmandiamond:
Added additional data that is captured for electron analyser detectors e.g

- image
- spectrum
- energy_axis
- binding_energy_axis
- angle_axis
- total_intentisty

1. Check implementation makes sense with abstract and sub classes
2. Check tests are relevant and pass

### Checks for reviewer


src/dodal/beamlines/b07.py
```diff
@@ -1,10 +1,6 @@
-from dodal.common.beamlines.beamline_utils import (
-    device_factory,
-)
+from dodal.common.beamlines.beamline_utils import device_factory
 from dodal.common.beamlines.beamline_utils import set_beamline as set_utils_beamline
-from dodal.devices.electron_analyser.specs_analyser_io import (
-    SpecsAnalyserDriverIO,
-)
+from dodal.devices.electron_analyser import SpecsAnalyserDriverIO
```
DiamondJoseph, line 3: Must: these unrelated formatting changes to other files shouldn't be included- are you using the devcontainer or have you got a python formatter installed that's fighting with others?

DiamondJoseph: e: the SpecsAnalyser one is valid because the import package changed, but the device factory and the device constructor call shouldn't be changing here if possible

oliwenmandiamond: I made the changes manually to use the package rather than the direct import

src/dodal/devices/electron_analyser/specs_analyser_io.py
```diff
@@ -8,12 +11,60 @@
...
+        axis = np.array([min_angle + offset + i * width for i in range(slices)])
+        return axis
+
+    def _get_energy_axis_signal(self, prefix: str = "") -> SignalR:
+        """
+        Override abstract and return soft signal that calculates energy axis
+        """
+        if hasattr(self, "energy_axis"):
+            return self.energy_axis
+        energy_axis = create_r_hardware_backed_soft_signal(
```
Relm-Arrowny, line 54: Head up: This will be move to derived signals in ophyd_async #1148 soonish.

src/dodal/devices/electron_analyser/specs_analyser_io.py
```diff
@@ -8,12 +11,60 @@
...
+            Array1D[np.float64], self._calculate_energy_axis, units="eV"
+        )
+        return energy_axis
+
+    async def _calculate_energy_axis(self) -> Array1D[np.float64]:
+        min_energy = await self.low_energy.get_value()
+        max_energy = await self.high_energy.get_value()
+        total_points_iterations = await self.slices.get_value()
```
Relm-Arrowny, line 62: could: parallel read here.
```suggestion
        min_energy, max_energy, total_points_iterations =asyncio.gather( 
        self.low_energy.get_value(),
        self.high_energy.get_value(), 
        self.slices.get_value()
        )
```

src/dodal/devices/electron_analyser/abstract_analyser_io.py
```diff
@@ -26,12 +41,63 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
 
-    def to_kinetic_energy(
-        self, value: float, excitation_energy: float, mode: EnergyMode
-    ) -> float:
-        return excitation_energy - value if mode == EnergyMode.BINDING else value
+    async def _calculate_total_time(self) -> float:
+        return (
+            await self.total_steps.get_value()
+            * await self.step_time.get_value()
+            * await self.iterations.get_value()
```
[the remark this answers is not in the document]
oliwenmandiamond: I think this is just longer where as the original code is simpler

DiamondJoseph (changes requested):

src/dodal/devices/electron_analyser/util.py
```diff
@@ -0,0 +1,13 @@
...
+def to_kinetic_energy(
+    value: float, value_mode: EnergyMode, excitation_energy: float
+) -> float:
+    return excitation_energy - value if value_mode == EnergyMode.BINDING else value
+
+
+def to_binding_energy(
+    value: float, value_mode: EnergyMode, excitation_energy: float
+) -> float:
+    return excitation_energy - value if value_mode == EnergyMode.KINETIC else value
```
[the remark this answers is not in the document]
Relm-Arrowny: I know this one as I was looking at this and thinking this look identical. Total(excitation) energy is energy to get the electrons out plus the momentum in energy they have.

src/dodal/devices/electron_analyser/specs_analyser_io.py
```diff
@@ -8,12 +14,58 @@
...
+        # SPECS returns the extreme edges of the range, not the centre of the pixels
+        width = (max_angle - min_angle) / slices
+        offset = width / 2
+
+        axis = np.array([min_angle + offset + i * width for i in range(slices)])
+        return axis
+
+    def _get_energy_axis_signal(self, prefix: str = "") -> SignalR:
+        if hasattr(self, "energy_axis"):
+            return self.energy_axis
```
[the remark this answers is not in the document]
oliwenmandiamond: So every analyser has an energy axis and angle axis. However, we don't know how it is implemented. Some are directly from a PV, others we need to calculate. 

So I created an abstract method which implements these as a signal and returns it. This is called via the abstract class and calls the abstract method. However, if for whatever reason this method is called again I wanted it to return the same signal object and not create the same, this was my way around it. If you know a cleaner way please let me know

src/dodal/devices/electron_analyser/util.py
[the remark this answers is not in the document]
oliwenmandiamond: Yes, excitation energy = kinetic + binding.

They are identical except for the energy mode part. The energy mode supplied tells if it is currently in binding or kinetic. If we want kinetic and it is already kinetic, it will return kinetic. If it is binding, it will convert to binding. 

Now I could implement the check to see if something is in the correct energy type or not, but I thought was much cleaner to add functions that state the type of energy were looking to get easier.

src/dodal/devices/electron_analyser/abstract_analyser_io.py
```diff
@@ -1,20 +1,35 @@
...
+class AbstractAnalyserDriverIO(ABC, StandardReadable, ADBaseIO):
     """
     Generic device to configure electron analyser with new region settings.
     Electron analysers should inherit from this class for further specialisation.
     """
 
     def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
+            # Used for setting up region data acquisition. Read per scan
+            self.region_name = soft_signal_rw(str, initial_value="null")
```
DiamondJoseph, line 28: should: If these are all read per-scan, they should be in a block marked as Config signals, which are only read once per scan:
```
with self.add_children_as_readables(CONFIG_SIGNAL):
    self.region_name = ...
    self.total_time = ...

with self._add_children_as_readables():
    self.image = ...
```

You can see the [types of readable children here](https://github.com/bluesky/ophyd-async/blob/80e70fc31c56c4b198768717644bf6dce0bb152f/src/ophyd_async/core/_readable.py#L17)- if there's a field/some fields you consider the primary important ones that people want to access 90% of the time (like the position of a motor), then you can also make use of HINTED (

src/dodal/devices/electron_analyser/specs_analyser_io.py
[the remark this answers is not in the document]
DiamondJoseph: Ah ok, so it's for caching it?

src/dodal/devices/electron_analyser/util.py
```diff
@@ -0,0 +1,13 @@
+from dodal.devices.electron_analyser import EnergyMode
+
+
+def to_kinetic_energy(
+    value: float, value_mode: EnergyMode, excitation_energy: float
+) -> float:
+    return excitation_energy - value if value_mode == EnergyMode.BINDING else value
```
DiamondJoseph, line 7: nit: I'd flip these clauses around to be a little clearer on purpose

`return value if value_mode == EnergyMode.KINETIC else excitation - value`

DiamondJoseph (changes requested):

src/dodal/devices/electron_analyser/abstract_analyser_io.py
```diff
@@ -26,12 +49,63 @@ def __init__(self, prefix: str, name: str = "") -> None:
             self.iterations = epics_signal_rw(int, prefix + "NumExposures")
             self.acquisition_mode = epics_signal_rw(str, prefix + "ACQ_MODE")
 
-        super().__init__(name)
+            # Read once per scan after data acquired
```
DiamondJoseph, line 52: @coretl config signals that get read *after* the read: is this possible? I assume not as it's read to construct the Descriptor document

src/dodal/devices/electron_analyser/specs_analyser_io.py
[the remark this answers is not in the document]
DiamondJoseph: I think since this is only run once (on instantiation of the device) you can and should get rid of the hasattr check and branching behaviour: it's harder to maintain and allows unexpected behaviour if it is set on the concrete impl of the class.

src/dodal/devices/electron_analyser/vgscienta_analyser_io.py
```diff
@@ -10,17 +12,37 @@
...
+            self.external_io = epics_signal_r(Array1D[np.float64], prefix + "EXTIO")
 
         super().__init__(prefix, name)
 
+    def _get_energy_axis_signal(self, prefix: str = "") -> SignalR[Array1D[np.float64]]:
+        """
+        Override abstract and return epics signal
+        """
+        if hasattr(self, "energy_axis"):
+            return self.energy_axis
```
DiamondJoseph, line 35: As above, I'd remove this to prevent confusion
```suggestion
```

src/dodal/devices/electron_analyser/vgscienta_analyser_io.py
```diff
@@ -10,17 +12,37 @@
...
+        return epics_signal_r(Array1D[np.float64], prefix + "X_SCALE_RBV")
+
+    def _get_angle_axis_signal(self, prefix: str = "") -> SignalR[Array1D[np.float64]]:
+        """
+        Override abstract and return epics signal
+        """
+        if hasattr(self, "angle_axis"):
+            return self.angle_axis
```
DiamondJoseph, line 43:
```suggestion
```

src/dodal/devices/electron_analyser/specs_analyser_io.py
```diff
@@ -7,13 +9,60 @@
...
+        # SPECS returns the extreme edges of the range, not the centre of the pixels
+        width = (max_angle - min_angle) / slices
+        offset = width / 2
+
+        axis = np.array([min_angle + offset + i * width for i in range(slices)])
+        return axis
+
+    def _get_energy_axis_signal(self, prefix: str = "") -> SignalR[Array1D[np.float64]]:
+        if hasattr(self, "energy_axis"):
+            return self.energy_axis
```
DiamondJoseph, line 47:
```suggestion
```

DiamondJoseph (changes requested):

src/dodal/devices/electron_analyser/abstract_analyser_io.py
oliwenmandiamond: From my testing, it was possible using  `bps.trigger_and_read(...)`

The values were read once and after the trigger with a new stream. However, I might be mistake and this was previous detector data

oliwenmandiamond: As discussed, this was actually previous detector data. We can collect per point for now as not an issue as same data every time. Something we can refine later.

src/dodal/devices/electron_analyser/abstract_analyser_io.py
```diff
@@ -25,13 +40,81 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
+    ) -> Array1D[np.float64]:
+        is_binding = energy_mode == EnergyMode.BINDING
+        return np.array(
+            [
+                to_binding_energy(i_energy_axis, EnergyMode.KINETIC, excitation_energy)
+                if is_binding
+                else i_energy_axis
+                for i_energy_axis in energy_axis
+            ]
+        )
```
DiamondJoseph, line 102:
```suggestion
        is_binding = energy_mode == EnergyMode.BINDING
        return np.array(
            [to_binding_energy(i_energy_axis, energy_mod, excitation_energy) for i_energy_axis in energy_axis]
        )
```

Isn't this the same?

DiamondJoseph (approved):
From what I understand of the device, looks sensible: just one comment about whether the list comprehension could be simplified?

src/dodal/devices/electron_analyser/abstract_analyser_io.py
oliwenmandiamond: No, tests fail if you do it this way. The reason being is we actually need to give it the opposite of the energy_mode supplied. For example, if it is already in binding energy, the pv returns the energy axis to us as kinetic energy and therefore we need to convert it to binding energy.

PR #1174, opened 2025-04-14
(mxb#896) Base OAV for I23 laser shaping

JamesDoingStuff:
Fixes [mxb#896](https://github.com/DiamondLightSource/mx-bluesky/issues/896)

To create an OAV device that doesn't have zoom, I have moved the core functionality of the `OAV` device into a new `OAVBase` device that acts as a parent for the existing `OAV` i.e. the `ZoomController` device is added on top of the `OAVBase` to produce an `OAV` device. 

Rather than redefining every attribute and function to have a version with zoom and another without, I have left them as they are and simply set `OAVBase` to default to a `zoom_level` of `1.0x`, unless a different zoom level is passed as an argument by the child `OAV`. This also ensures full compatibility with existing OAV config files.

Lastly, all `hardware_backed_soft_signal`s are replaced with `derived_signal`s as in #1148
1. Run `dodal connect aithre`
2. Check `oav` connects successfully
3. Repeat for other beamlines to ensure existing OAV devices remain functional

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/aithre.py
+++ b/src/dodal/beamlines/aithre.py
@@ -4,8 +4,8 @@ from dodal.devices.oav.oav_detector_base import OAVBase
 from dodal.devices.oav.oav_parameters import OAVConfig
 from dodal.devices.robot import BartRobot
 
-ZOOM_PARAMS_FILE = "/dls_sw/i23/software/aithre_oav.xml"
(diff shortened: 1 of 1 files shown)


DominicOram:
> Lastly, all hardware_backed_soft_signals are replaced with derived_signals as in https://github.com/DiamondLightSource/dodal/pull/1148

To avoid confusion this means we should probably merge that PR first

PR #1175, opened 2025-04-14
Use all murko results

jacob720:
See [#136](https://github.com/DiamondLightSource/mx-bluesky/issues/136) from mx-bluesky. This is a branch off of https://github.com/DiamondLightSource/dodal/pull/1149

An explanation of some of the maths can be found in the issue linked above. Would be good to add in some outlier rejection as well as incorporating confidence values from Murko if they are given.

1. Test with real Murko results
2. Confirm the sample is centred correctly.

### Checks for reviewer


src/dodal/devices/i04/murko_results.py
```diff
@@ -42,12 +41,19 @@ class Coord(Enum):
...
-    (x, z) coords can be read at 90°, and (x, y) at 180° (or the closest omega angle to
-    90° and 180°). The average of the x values at these angles is taken, and sin(omega)z
-    and cosine(omega)y are taken to account for the rotation. This value is used to
-    calculate a number of mm the sample needs to move to be in line with the beam centre.
+    The most_likely_click[1] value from Murko corresponds with the x coordinate of the
+    sample. The most_likely_click[0] value from Murko corresponds with a component of
+    the y and z coordinates of the sample, depending on the omega angle, as the sample
+    is rotated around the x axis.
+
+    Given a most_liekly_click value at a certain omega angle θ:
```
DominicOram, line 51: Nit: 
```suggestion
    Given a most_likely_click value at a certain omega angle θ:
```
Do you have spell check on vscode? I find it very useful

src/dodal/devices/i04/murko_results.py
```diff
@@ -58,6 +64,7 @@ def __init__(
         redis_password=RedisConstants.REDIS_PASSWORD,
         redis_db=RedisConstants.MURKO_REDIS_DB,
         name="",
+        stop_angle=270,
```
DominicOram, line 67: Hmm... We should probably be able to stop based on a message from redis saying that we've reached the end of the rotation. Can you make an issue for that?

src/dodal/devices/i04/murko_results.py
```diff
@@ -98,94 +101,76 @@ async def unstage(self):
...
+            -beam_dist_px[0] * metadata["microns_per_x_pixel"] / 1000
         )
+        self.y_dists_mm.append(
+            -beam_dist_px[1] * metadata["microns_per_y_pixel"] / 1000
+        )
+        self.omegas.append(omega)
+        self._last_omega = omega
+
+    @staticmethod
+    def get_yz_least_squares(v_values: list, omegas: list) -> tuple[float, float]:
```
DominicOram, line 161: Could: IMHO, `@staticmethod` isn't very pythonic, you can just use a standalone function outside the class

tests/devices/i04/test_murko_results.py
```diff
@@ -127,253 +154,265 @@ async def test_no_movement_given_sample_centre_matches_beam_centre(
     mock_strict_redis, mock_murko_results, mock_setters
 ):
     mock_x_setter, mock_y_setter, mock_z_setter = mock_setters
-    messages = iter(pickled_messages(get_messages()))  # 2 messages, both (0.5, 0.5)
-    metadata = json_metadata(
-        2, start=90, step=90, beam_centre_i=50, beam_centre_j=50
-    )  # at 90, 180 degrees
+    messages, metadata = get_messages(
+        images_per_message=10, omega_start=50, omega_step=5
+    )  # Crystal alligned with beam centre
```
DominicOram, line 159: Nit: 
```suggestion
    )  # Crystal aligned with beam centre
```

DominicOram (changes requested):
This is great, thanks! Hopefully it gives good answers when tested. Some minor comments in code.

For your tests, they're quite complicated because they're testing the whole of the device at once. I would instead build them up a bit e.g.
* Start with tests on `get_yz_least_squares`, you should be able to have a parameterised set of inputs/outputs we expect that we can reason about
* Send single messages into `process_results` and confirm that `x_dists_mm`, `y_dists_mm` and `omegas` gets what you expect, paramaterise a variety of different message contents
* Patch out `process_results`, send multiple messages into `process_batch` and confirm that `process_results` is called when expected, test all the edge cases of the message type being wrong or no metadata existing
* Now you can do the full test of everything but hopefully more of the edge cases have been covered from before so you might need fewer of these

There are some mixed opinions on this. I like splitting them up like above as:
* It's more readable (and tests are docs)
* It makes it more obvious when something breaks exactly where the issue was
* It lends itself more to TDD

But the counter point (I know @rtuck99 and I have had this discussion before) is that you're now testing the innards of the class and making assumptions about how it behaves so your tests are brittle. e.g. if someone decides to change the way the class works inside but it still gives the answer same answer to `z_mm` then this won't break any production code and so shouldn't break tests.

In practice I think having both types of tests is useful and I suggest at least adding a few of the finer grained ones.

src/dodal/devices/i04/murko_results.py
jacob720: https://github.com/DiamondLightSource/dodal/issues/1239

src/dodal/devices/i04/murko_results.py
```diff
@@ -102,75 +102,97 @@ async def trigger(self):
...
-            v_values (list): List of vertical distances from beam centre in mm.
-            thetas_deg (list): List of omega angles in degrees.
-
-        Returns:
-            tuple[float, float]: y, z coordinates
-        """
-        thetas = np.radians(omegas)
-        matrix = np.column_stack([np.cos(thetas), -np.sin(thetas)])
+        LOGGER.info(f"Got most_likely_click: {coords}")
+        if tuple(coords) == (-1, -1):
```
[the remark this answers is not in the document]
jacob720: https://github.com/MartinSavko/murko/issues/9

PR #1178, opened 2025-04-15
Use a generic 1D positioner everywhere

### LINKED ISSUE #1176: Make positioner more generic
The `Positioner` made in https://github.com/DiamondLightSource/dodal/pull/960 is good but it could be more generic. Specifically:
* It should be moved into a util area
* It could be a `Generic` that takes the enum to select
* `Positioner_pv_suffix` should be lower case
* It should replace other uses of positioner
* The logic in `set` should be covered by the type system

DominicOram:
Fixes #1176

Changes the `Y` motor signal of `DiamondFilter` to be `diamond_filter._stage_motion` rather than `diamond_filter.y`

1. Confirm tests still pass
2. Confirm you can `dodal connect` to `i03`, `i04` and `i10`

### Checks for reviewer


src/dodal/devices/positioner.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def __init__(
+        self,
+        prefix: str,
+        datatype: type[StrictEnum],
+        positioner_pv_suffix: str = ":MP:SELECT",
+        name: str = "",
+    ) -> None:
+        self._stage_motion = Motor(prefix=prefix)
+        with self.add_children_as_readables(Format.CONFIG_SIGNAL):
```
Relm-Arrowny, line 29: I think we want to do read to get the postion rather than the config?
```suggestion
        with self.add_children_as_readables():
```

src/dodal/devices/positioner.py
```diff
@@ -0,0 +1,47 @@
...
+        await self.stage_position.set(value=value)
+
+
+def create_positioner(
+    datatype: type[T],
+    prefix,
+    positioner_pv_suffix: str = ":MP:SELECT",
+    name: str = "",
+) -> Positioner1D[T]:
+    return Positioner1D[datatype](prefix, datatype, positioner_pv_suffix, name)
```
stan-dot, line 47: could: this is a constructor of the Positioner1D class, making it a `@classmethod` would feel natural

src/dodal/devices/positioner.py
```diff
@@ -0,0 +1,47 @@
...
+from ophyd_async.epics.core import (
+    epics_signal_rw,
+)
+from ophyd_async.epics.motor import Motor
+
+T = TypeVar("T", bound=StrictEnum)
+
+
+class Positioner1D(StandardReadable, Movable[T]):
+    """1D stage with a enum table to select positions."""
```
stan-dot, line 19: should: document more when to use this

stan-dot (changes requested):
small DX changes suggested

tests/devices/unit_tests/test_positioner.py
```diff
@@ -1,27 +1,36 @@
...
+    TEST_2 = "Test_2"
+
+
+class BadPositions(StrictEnum):
+    BAD_POSITION = "BAD"
 
 
 @pytest.fixture
-async def mock_positioner(prefix: str = "BLXX-EA-007:") -> Positioner:
+async def mock_positioner(prefix: str = "BLXX-EA-007:") -> Positioner1D:
```
DiamondJoseph, line 20:
```suggestion
async def mock_positioner() -> Positioner1D:
```

nit: the default prefix is never overridden

src/dodal/devices/diamond_filter.py
```diff
@@ -25,22 +25,10 @@ class I04Filters(_Filters):
...
     """
-
-    def __init__(
-        self,
-        prefix: str,
-        data_type: type[T],
-        name: str = "",
-    ) -> None:
-        with self.add_children_as_readables():
-            self.y_motor = Motor(prefix + "Y")
```
DiamondJoseph, line 43: Is the y_motor position no longer required on i03/i04?
Must: This should at least be called out in the ticket body so it's clear that it's an intentional change if it is.

DiamondJoseph (changes requested):

src/dodal/devices/positioner.py
DominicOram: @JamesOHeaDLS is there any docs on the EPICS positioner template? I think it's a DLS only thing, right?

src/dodal/devices/positioner.py
DominicOram: I prefer functions over static methods, makes calling it more verbose. I think it's personal preference so going to leave for now unless others feel strongly.

DiamondJoseph (approved):

stan-dot (approved):
lgtm

PR #1186, opened 2025-04-23
write tutorial mentioning the external IO devices

### LINKED ISSUE #1185: Document how to make devices for non-EPICS IO

stan-dot:
Fixes #1185

1. Read the file `external-io-devices.md`
2. Add suggestions for changes and additions

### Checks for reviewer

### DIFF
--- a/docs/how-to/external-io-devices.md
+++ b/docs/how-to/external-io-devices.md
@@ -7,7 +7,7 @@ There are instances when input-output operations are desired that do not work wi
 For example, many GDA-compatible features need disk IO as GDA relies heavily on XML files to define various variables.
 One may need a bluesky calibration plan to start with cached values - those could be in the filesystem or a key value store like Redis.
 
-At the moment direct IO inside plans is not allowed when inside the RunEngine context.
+Direct IO inside plans is not allowed when inside the RunEngine context.
 That means that all IO operations must happen through ophyd-async devices.
 
 ## Extant examples
--- a/docs/how-to/external-io-devices.md
+++ b/docs/how-to/external-io-devices.md
@@ -8,7 +8,9 @@ For example, many GDA-compatible features need disk IO as GDA relies heavily on 
 One may need a bluesky calibration plan to start with cached values - those could be in the filesystem or a key value store like Redis.
 
 Direct IO inside plans is not allowed when inside the RunEngine context.
-That means that all IO operations must happen through ophyd-async devices.
+
+As far as is possible, we want our devices to only talk to EPICS PVs. The [config server](https://github.com/DiamondLightSource/daq-config-server) should fulfil the majority of use cases where we can't do that, it is possible to make a temporary ophyd-async devices, but heavily discouraged.
+It's not recommended to read from the filesystem going forward and instead development effort will be put into the config server.
 
 ## Extant examples
 
--- a/docs/how-to/external-io-devices.md
+++ b/docs/how-to/external-io-devices.md
@@ -1,0 +1,25 @@
+# How to Write External IO Devices
+
+## Motivation and Overview
+
+At Diamond Dodal devices primarily refer to python classes referencing EPICS channels, with some internal logic.
+There are instances when input-output operations are desired that do not work with EPICS.
+For example, many GDA-compatible features need disk IO as GDA relies heavily on XML files to define various variables.
+One may need a bluesky calibration plan to start with cached values - those could be in the filesystem or a key value store like Redis.
+
+At the moment direct IO inside plans is not allowed when inside the RunEngine context.
+That means that all IO operations must happen through ophyd-async devices.
+
+## Extant examples
+
+- [aperturescatterguard](../../src/dodal/devices/aperturescatterguard.py) - reads a set of valid positions from a file.
+- [oav_to_redis_forwarder](../../src/dodal/devices/oav/oav_to_redis_forwarder.py) - pushes data into redis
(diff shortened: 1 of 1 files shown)


docs/how-to/external-io-devices.md
```diff
@@ -0,0 +1,25 @@
+# How to Write External IO Devices
+
+## Motivation and Overview
+
+At Diamond Dodal devices primarily refer to python classes referencing EPICS channels, with some internal logic.
+There are instances when input-output operations are desired that do not work with EPICS.
+For example, many GDA-compatible features need disk IO as GDA relies heavily on XML files to define various variables.
+One may need a bluesky calibration plan to start with cached values - those could be in the filesystem or a key value store like Redis.
+
+At the moment direct IO inside plans is not allowed when inside the RunEngine context.
```
DominicOram, line 10: Should: Remove `At the moment`, I don't this will ever be the case

docs/how-to/external-io-devices.md
```diff
@@ -0,0 +1,25 @@
...
+- [oav_to_redis_forwarder](../../src/dodal/devices/oav/oav_to_redis_forwarder.py) - pushes data into redis
+- [OAV_detector](../../src/dodal/devices/oav/oav_detector.py) - detector configuration is based on a file on disk
+
+## Existing patterns
+
+Those three devices have been mostly at the MX beamlines. To see a more up to date list, narrow down the search to the [beamlines folder](../../src/dodal/beamlines/) and search for a `dls` string. You will see `Path` calls that are about data writing, but also some ALL_CAPS constants such as `DISPLAY_CONFIG`.
+There are two cases, either the IO operation we're looking into is just on device start, or it's an ongoing thing. If it's just in the start as a config, the established pattern is to provide a Class for that object, and make a standlone function in the device file to load it from the filesystem, taking path as a parameter.
+Then inside the specific file in the beamlines folder, the device takes the product of calling the function  with a beamline-specific path. The fact the device just takes an object makes it easier to write mocks for testing.
```
DominicOram, line 23:
```suggestion
Then inside the specific file in the beamlines folder, the device takes the product of calling the function with a beamline-specific path. The fact the device just takes an object makes it easier to write mocks for testing.
```

DominicOram (changes requested):
Minor comments, I also think we need to mention the configserver here as most of these usecases will be replaced with that in the near future. FYI @olliesilvester

docs/how-to/external-io-devices.md
callumforrester: Agreed, it goes against the design of bluesky

callumforrester:
I'm glad this is documented, but I feel it should be discouraged unless absolutely necessary, and maybe the opening paragraph should make reference to that. But I welcome others' opinions.

DominicOram:
> I'm glad this is documented, but I feel it should be discouraged unless absolutely necessary, and maybe the opening paragraph should make reference to that. But I welcome others' opinions.

I agree, this is why I said we should mention the config server as the preferred option

docs/how-to/external-io-devices.md
stan-dot: removed the time specification three words

DominicOram (changes requested):
> does config server support during-plan talking to say, redis or filesystem?

Yes, the config server will need support getting config data into a plan and into a device (though it may be that you get the data into the plan via some kind of dummy config reading device).

Once you update these docs to say that it's not recommended to read from the filesystem going forward and instead effort will be put into the config server I'm happy to approve.

callumforrester:
@stan-dot As far as is possible (and I do appreciate that it's not always possible), we want our devices to only talk to EPICS PVs. The config server should fulfil the majority of use cases where we can't do that, I/O outside of those regiemes will still be supported but heavily discouraged. Projects like https://github.com/DiamondLightSource/fastcs are being developed to make going via the EPICS easy/a path of low resistance.

docs/how-to/external-io-devices.md
```diff
@@ -0,0 +1,27 @@
...
+## Motivation and Overview
+
+At Diamond Dodal devices primarily refer to python classes referencing EPICS channels, with some internal logic.
+There are instances when input-output operations are desired that do not work with EPICS.
+For example, many GDA-compatible features need disk IO as GDA relies heavily on XML files to define various variables.
+One may need a bluesky calibration plan to start with cached values - those could be in the filesystem or a key value store like Redis.
+
+Direct IO inside plans is not allowed when inside the RunEngine context.
+
+As far as is possible, we want our devices to only talk to EPICS PVs. The [config server](https://github.com/DiamondLightSource/daq-config-server) should fulfil the majority of use cases where we can't do that, it is possible to make a temporary ophyd-async devices, but heavily discouraged.
```
callumforrester, line 12: Unsure what you mean by temporary devices? Also typo

```suggestion
As far as is possible, we want our devices to only talk to EPICS PVs. The [config server](https://github.com/DiamondLightSource/daq-config-server) should fulfil the majority of use cases where we can't do that, it is possible to make temporary ophyd-async devices, but heavily discouraged.
```

stan-dot: temporary as in `there until the config server supports their feature`

stan-dot: added

DominicOram (approved):
> Thanks, I'm happy to approve this once @DominicOram and @Relm-Arrowny are too

I'm always happy to defer to your approval @callumforrester, especially for docs. Velocity of getting them in is more important to me than my approving them all.

Relm-Arrowny (approved):

PR #1193, opened 2025-04-28
Remove s03 system tests

rtuck99:
Fixes DiamondLightSource/mx-bluesky#631

Removes:
   * the `s03` pytest test marker, and tests depending on s03 devices.
   * s03 beamline and associated dictionary entries

Introduces the `system_test` pytest marker for those tests not related to s03 but previously marked as such

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://diamondlightsource.github.io/dodal/main/reference/device-standards.html)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

PR #1196, opened 2025-04-28
Exclude system test marker from tox tests command

olliesilvester:
After https://github.com/DiamondLightSource/dodal/pull/1193, `tox -e tests` runs the tests marked with @pytest.mark.system_test. We want to exclude these since the marker means the tests require external dependencies to pass

1. Confirm `tox -e tests` now passes

### Checks for reviewer


callumforrester (approved):
Happy with this, could also just do it by directory and exclude the `system_tests` directory altogether, but I'm not fussed

DiamondJoseph:
@callumforrester re-requesting a review because I have changed the name of the pytest mark to something that I think is more obvious what is intended (e.g. there are tests in `system_tests/` that were not marked `@system_test`, and the mark is actually for diamond internal stuff, but it is a change to the test running proceedure

callumforrester:
@DiamondJoseph renaming is a good shout, but I'm not sure if internal is particularly clear either. I'd maybe go with `requires_live_environment`?

DominicOram:
I think this is complicated enough that we should add something in the docs e.g.
* What goes in `/system_tests/`
* What goes in `adsim`? I would in-fact suggest we called this something more generic e.g `requires_other_services`
* What goes in `internal`? I would also suggest we call this something like `requires_internal_services` or `requires_live_environment` as callum says

DiamondJoseph:
>     * What goes in `adsim`? I would in-fact suggest we called this something more generic e.g `requires_other_services`

This is the only point I disagree on: the mark should to be specific enough to enable/disable sets of tests when you have the required infrastructure. Renaming `adsim` to `requires_adsim_containerised_iocs` (leaving the future open for `requires_s04_containerised_iocs`, `requires_lab_iocs` etc.) would make sense to me. `requires("adsim")` looks promising, and I think we can then run multiple overlapping sets (e.g. `pytest -m adsim -m s04` is possible as would be `pytest -m requires(adsim) -m requires(s04)`

DominicOram:
Sure, I don't care too much

olliesilvester:
I can re-review the extra changes here. I can also add a PR onto this one to add the docs suggested by @DominicOram

PR #1205, opened 2025-04-30
Create initial analyserscan plan

oliwenmandiamond:
Create basic analyserscan plan, allowing for reading of a sequence file, configuring analyser with region settings, and saving the data for each region into separate streams. 

1. Check that plan logic makes sense
2. Make sure tests pass

### Checks for reviewer


src/dodal/devices/electron_analyser/specs_analyser.py
```diff
@@ -16,7 +18,7 @@ def __init__(self, prefix: str, name: str = "") -> None:
         with self.add_children_as_readables():
             # Used for setting up region data acquisition. Per scan.
             self.psu_mode = epics_signal_rw(str, prefix + "SCAN_RANGE")
-            self.values = epics_signal_rw(int, prefix + "VALUES")
+            self.sanpeshot_values = epics_signal_rw(int, prefix + "VALUES")
```
Villtord, line 21: I think this is a typo

src/dodal/devices/electron_analyser/abstract_analyser.py
```diff
@@ -0,0 +1,287 @@
...
+        for the underlying analyser software and cannot be changed on epics side.
+        """
+
+
+TAbstractAnalyserDriverIO = TypeVar(
+    "TAbstractAnalyserDriverIO", bound=AbstractAnalyserDriverIO
+)
+
+
+class AnalyserController(Generic[TAbstractAnalyserDriverIO]):
```
Villtord, line 142: Does this class need to be a child of TAbstractAnalyserDriverIO if it gets driver of the same class TAbstractAnalyserDriverIO as a parameter?

PR #1215, opened 2025-05-07
Add generic calculation for translating from sample space to lab space and use for i23 OAV

### LINKED ISSUE #1168: I23 serial: Make derived signals for moving in the OAV plane
To do the I23 serial experiment (https://github.com/DiamondLightSource/mx-bluesky/issues/962) we would like to move in the plane of the OAV. The plane of the camera is such that when we are at omega of 0 we should move x, y and when we are at omega -90 we should move x, z. This should be generalised to any omega.

## Acceptance Criteria
* There are derived signals for the above

### LINKED ISSUE #1232: Reconcile I23 goniometer with Aithre goniometer
Code should be reused between the goniometer on the I23 beamline and the goniometer in the Aithre laser shaping setup - especially for the OAV-plane signals proposed in #1164 and #1215.

The I23 goniometer is constructed as as `XYZPositioner` upon which `omega`, `kappa` and `phi` are added (and now these signals). The Aithre goniometer, on the other hand, does not have a sample X stage, nor `kappa` and `phi` rotation axes. It does, however, have additional X Y and Z translation axes that the goniometer sits on. These PVs clash in naming convention with those of the I23 goniometer. For a summary of PVs of the two:

| PV in I23 Goniometer | Equivalent PV in Aithre Goniometer |
|---|---|
| X | - |
| Y | SAMPY |
| Z | SAMPZ |
| OMEGA | OMEGA |
| KAPPA |  - |
| PHI |-|
| - | X |
| - | Y |
| - | Z

## Acceptance Criteria
* Duplicated code is minimised for the two goniometers
* Each connects and tests pass as expected

JamesDoingStuff:
Fixes #1168 and fixes #1232
 
1. Confirm new tests pass
2. Confirm `dodal connect i23` connects the goniometer
3. Confirm docs are good on the utility function for someone not that familiar with the problem

### Checks for reviewer


src/dodal/devices/aithre_lasershaping/goniometer.py
```diff
@@ -1,9 +1,8 @@
-import asyncio
```
rtuck99, line 1: Missing `__init__.py` - without this the API docs aren't generated for the package

src/dodal/devices/motors.py
```diff
@@ -41,6 +91,31 @@ def __init__(
 
 
 class SixAxisGonio(XYZPositioner):
+    """
+
+    Six-axis goniometer with a standard xyz stage and three axes of rotation: kappa, phi
+    and omega.
+
+        Parameters
```
rtuck99, line 99: Can we please use standard google pydoc notation here: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings? It helps the generated API docs format nicely.

You can also use doctest style examples: https://docs.python.org/3/library/doctest.html
The advantage is that the examples then get checked for correctness which helps make sure that the docstrings stay up to date. However although the formatting is nice if you don't want the tests to actually run, then you can use` # doctest: +SKIP`

src/dodal/devices/motors.py
```diff
@@ -1,7 +1,57 @@
...
+
+    The usual use case for this is translating from sample space to lab space. For
+    example, if you have a sample that is mounted on a goniometer to the right hand side
+    of an OAV view this can provide an axis that will move the sample up/down in that
+    view regardless of the omega orientation of the sample.
+
+    Args:
+        parallel_to_0 (Motor): this is the axis that, when the sample is at 0 deg rotation,
+                               a move here is entirely parallel with the derived axis.
+        parallel_to_minus_90 (Motor): this is the axis that, when the sample is at 90 deg
```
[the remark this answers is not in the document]
rtuck99: So I think this doc comment can be made clearer. For me, I find the naming of the args a bit confusing, because the frame of reference of the angle in the name isn't clear and is contradicted by the arg description (although I think I understand your intent with the sign flip, it is I think unnecessarily confusing).

I think it's better to describe the geometric assumptions in the doc comment, and label the axes neutrally, because this function is applicable to a variety of different geometries and labelling the axes  things like whether any given axis is right- or left- handed, up or down etc are essentially not relevant, as long as the inputs conform to the documented requirements it should give the right answer.

While I was reviewing this, to make it easier for myself to follow the implementation, I renamed the args `motor_i`, `motor_j`  and `motor_theta`. I tried to come up with some wording for the comment that would summarise less ambiguously what the function does, what I came up with was this:

>    Given a signal that controls a motor in a rotation axis and two other
>    signals controlling motors on a pair of orthogonal axes, these axes being in the
>    rotating frame of reference created by the first axis,
>    create a derived signal that is a projection of the two axes in the non-rotating
>    frame of reference.
>
>    The projection is onto the axis defined by i when the rotation angle is 0.

For me I think this is clearer, although it is quite technical and I think you should retain the illustrative text in the comment because not everyone prefers this kind of explanation and it is generally helpful. Also I think my sentence structure is still a bit clunky and overly long. Perhaps breaking it up into bullet points would help, I'm open to improvements in that regard. For me, the important facts are that:

* There are two frames of reference, one is stationary, one is rotating
* One input is a motor controls the angle between the two frames of reference
* The other inputs are on orthogonal axes in the rotating frame, with the given orientations at specified angles of theta
* The output is a motor which is a projection onto the stationary frame parallel to i when theta == 0

Also, I think it might be better if the rotation motor was moved to the first argument of the function, because it's the important thing that controls how the other two axes are interpreted.

@olliesilvester do you think my text above is helpful?

tests/devices/unit_tests/test_motors.py
```diff
@@ -6,12 +7,29 @@
...
     with init_devices(mock=True):
         gonio = SixAxisGonio("")
+    patch_motor(gonio.omega)
+    patch_motor(gonio.z)
+    patch_motor(gonio.y)
+    patch_motor(gonio.x)
+
+    return gonio
+
+
```
rtuck99, line 24: These fixtures are the same?

src/dodal/devices/motors.py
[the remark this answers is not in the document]
DominicOram: Completely agree, sounds a lot better. Thanks!

Issue #900, opened 2024-11-13
 I24 serial: have the PMAC program runner check the counter PV


noemifrisina:
The collection time passed as timeout to `complete` on the PMAC `ProgramRunner` is currently calculated from the input parameters (exposure time, laser exposure and delay, number of images) plus the time needed for the PMAC moves plus a buffer. For most short experiments this works well, but for some of the longer collections, it has proven to be difficult to estimate as the buffer needed seems to increase exponentially with the number of blocks. While we should still check that the status PV (`BL24I-MO-STEP-14:signal:P2401`) goes to zero at the end of the collection - so tha the PMAC is not left in a weird state - we can also check when a collection is over by monitoring the counter PV (`BL24I-MO-STEP-14:signal:P2402`).

ProgramRunner should check that this PV has already reached the number of frames requested by timeout and if not, continue waiting.

## Acceptance Criteria
* ProgramRunner doesn't complete until the counter PV reaches the expected number of frames.

DominicOram:
As discussed, we should actually monitor the counter and timeout if it freezes, see https://github.com/DiamondLightSource/mx-bluesky/issues/859

PR #1216, opened 2025-05-07
1105 add monitor for frame counter

### LINKED ISSUE #1105: I24 serial: add a monitor for the frame counter to PMAC ProgramRunner
Once #1104 is done, we need to add a monitor for the  frame counter PV on the PMAC. This monitor should determine whether the collection is done or if the pmac has stopped prematurely.

eg.
If the expected number of frames is 25600, the ProgramRunner should complete once this number is reached or timeout if the frame counter value has not reached the value but hasn't changed in 10-30s.

## Acceptance Criteria
* The PMAC ProgramRunner has a monitor that checks the fram counter for either collection done or timeout.

adaudon:
Fixes #1105 and completes #900

1. Confirm tests still pass

### Checks for reviewer


src/dodal/devices/i24/pmac.py
```diff
@@ -1,4 +1,5 @@
 from asyncio import sleep
+from asyncio.exceptions import TimeoutError as TE
```
DominicOram, line 2: Should: Why change this to `TE`? I think it makes the code less readable

src/dodal/devices/i24/pmac.py
```diff
@@ -150,17 +154,19 @@ async def kickoff(self):
         )
 
     @AsyncStatus.wrap
-    async def complete(self):
-        """Stop collecting when the scan status PV goes to 0.
-
-        Args:
-            complete_time (float): total time required by the collection to \
-            finish correctly.
+    async def complete(self, counter_timeout: float = 30):
```
DominicOram, line 157: Should: `complete` should implement the function as defined in `Flyable` and so not take any parameters. A better way to do this would be a soft signal that holds the timeout time

src/dodal/devices/i24/pmac.py
```diff
@@ -120,12 +122,14 @@ def __init__(
...
         prog_num_sig: SignalRW,
         collection_time_sig: SignalRW,
         name: str = "",
     ) -> None:
         self._signal_ref = Reference(pmac_str_sig)
         self._status_ref = Reference(status_sig)
+        self._counter_ref = Reference(counter_sig)
         self._prog_num_ref = Reference(prog_num_sig)
 
         self._collection_time_ref = Reference(collection_time_sig)
```
DominicOram, line 135: Should: This collection time is now no longer used, I think we should probably remove it here and where it is set in `mx-bluesky`. Instead setting the timeout for each frame.

src/dodal/devices/i24/pmac.py
```diff
@@ -150,17 +154,19 @@ async def kickoff(self):
...
-        )
+        try:
+            async for signal, value in observe_signals_value(
+                self._status_ref(), self._counter_ref(), timeout=counter_timeout
+            ):
+                if signal is self._status_ref():
+                    if value == ScanState.DONE:
+                        break
+        except TE:
+            pass
```
DominicOram, line 169: Must: Why are we swallowing this error? The issue was to make sure it does time out

tests/devices/unit_tests/i24/test_pmac.py
```diff
@@ -88,6 +88,30 @@ async def go_high_then_low():
     assert await fake_pmac.pmac_string.get_value() == "&2b11r"
 
 
+@pytest.mark.parametrize("sleepTime, expected", [(0.05, 3), (0.2, 2)])
+async def test_counter_refresh_timeout(sleepTime, expected, fake_pmac: PMAC, RE):
```
DominicOram, line 92: Should: We're in python. Please use `snake_case`

DominicOram: Could: `expected` is a bit vague, maybe `expected_counter_value`?

tests/devices/unit_tests/i24/test_pmac.py
```diff
@@ -88,6 +88,30 @@ async def go_high_then_low():
...
+        set_mock_value(fake_pmac.counter, 3)
+        set_mock_value(fake_pmac.scanstatus, 0)
+
+    callback_on_mock_put(
+        fake_pmac.pmac_string,
+        lambda *args, **kwargs: asyncio.create_task(update_counter()),  # type: ignore
+    )
+
+    await fake_pmac.run_program.kickoff()
+    await fake_pmac.run_program.complete(counter_timeout=0.1)
```
DominicOram, line 110: Must: I would expect this to test that if we hit the timeout it raised the expected exception

DominicOram (changes requested):
Thank you! Some comments in code

src/dodal/devices/i24/pmac.py
adaudon: Added an alias because python kept referring to a different timeout. But TE isn't a very good alias. I've changed the code so it imports exceptions and the code says exceptions.TimeoutError so it's clear.

src/dodal/devices/i24/pmac.py
DominicOram: Moved to https://github.com/DiamondLightSource/mx-bluesky/issues/1050 instead

tests/devices/unit_tests/i24/test_pmac.py
```diff
@@ -106,10 +105,36 @@ async def update_counter():
...
+    async def update_counter():
+        set_mock_value(fake_pmac.scanstatus, 1)
+        set_mock_value(fake_pmac.counter, 0)
+        await asyncio.sleep(0.05)
+        set_mock_value(fake_pmac.counter, 1)
+        await asyncio.sleep(0.05)
+        set_mock_value(fake_pmac.counter, 2)
+        await asyncio.sleep(0.2)
+        set_mock_value(fake_pmac.counter, 3)
+        set_mock_value(fake_pmac.scanstatus, 0)
```
DominicOram, line 125: Could: This can just be in a separate function to avoid the code duplication e.g.
```python
    async def update_counter(sleep_time: float):
        set_mock_value(fake_pmac.scanstatus, 1)
        set_mock_value(fake_pmac.counter, 0)
        await asyncio.sleep(0.05)
        set_mock_value(fake_pmac.counter, 1)
        await asyncio.sleep(0.05)
        set_mock_value(fake_pmac.counter, 2)
        await asyncio.sleep(sleep_time)
        set_mock_value(fake_pmac.counter, 3)
        set_mock_value(fake_pmac.scanstatus, 0)
```
Then in each test you can use:
```python
from functools import partial
    callback_on_mock_put(
        fake_pmac.pmac_string,
        lambda *args, **kwargs: asyncio.create_task(partial(update_counter, 0.2)),  # type: ignore
    )

DominicOram (approved):
Thanks, one optional comment, otherwise good.

PR #1218, opened 2025-05-08
Update P38 bluesky data directory

barnettwilliam:
Updated P38 bluesky data directory for 2025


DiamondJoseph (approved):
I've set the CI running again now that the docs build should be passing

PR #1220, opened 2025-05-08
I23 Detector Device

JamesDoingStuff:
For [mxb#1013](https://github.com/DiamondLightSource/mx-bluesky/issues/1013)

1. Run `dodal connect i23`
2. Confirm connection of `detector`

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i23.py
+++ b/src/dodal/beamlines/i23.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+from ophyd_async.core import StrictEnum
 from ophyd_async.epics.adpilatus import PilatusDetector
 
 from dodal.common.beamlines.beamline_utils import (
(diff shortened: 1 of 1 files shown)


src/dodal/beamlines/i23.py
```diff
@@ -43,6 +45,12 @@
 )
 
 
+class I23DetectorPositions(StrictEnum):
+    IN = "In"
+    OUT = "Out"
+    CHANGE = "sample change"
```
DominicOram, line 51: Nit: `SAMPLE_CHANGE` is probably a better name

src/dodal/beamlines/i23.py
```diff
@@ -93,3 +101,12 @@ def pilatus() -> PilatusDetector:
         drv_suffix="cam1:",
         fileio_suffix=HDF5_SUFFIX,
     )
+
+
+@device_factory()
+def detector() -> Positioner1D[I23DetectorPositions]:
```
DominicOram, line 107: Should: I think calling it `detector` is a bit misleading as it doesn't have stuff to control the detector itself, just move it. In `i03.py` we call it `detector_motion`. When we have the rest of the detector in too we can think about combining them

DominicOram (changes requested):
Thanks, some suggestions in the code

PR #1222, opened 2025-05-09
Fix Tetramm timeout during prepare

barnettwilliam:
Fixes https://github.com/bluesky/ophyd-async/issues/841#issuecomment-2865558193

1. Create a plan that includes a Tetramm and waits on prepare with `error_on_timeout=True`
2. Run the plan
3. Confirm that no error occurs

### Checks for reviewer


RJCD-Diamond (approved):
lgtm

barnettwilliam:
The initial fix was fragile so updated to follow the pattern used by AreaDetector, originally tetramm was missing the guard.  Additionally` wait_for_set_completion` needed to be set to false confirmed by testing on i1 in p38.

https://github.com/bluesky/ophyd-async/blob/9300a2c7a68425b0680c35cce5527c27fd613c21/src/ophyd_async/epics/adcore/_core_logic.py#L59

PR #1227, opened 2025-05-12
Make beamstop selected_pos attribute settable

jacob720:
Fixes #[1060](https://github.com/DiamondLightSource/mx-bluesky/issues/1060) in mx-bluesky. Makes `beamstop.selected_pos` settable so that a plan can move the beamstop into position if needed.

1. Confirm this is the best way to do this
2. Confirm tests are thorough enough

### Checks for reviewer


DominicOram (changes requested):
Thanks, I think a better way to do this would be to do more in the device. We already have `selected_pos`, we should make that writable. So that we can do:
```python
yield from bps.abs_set(beamstop._selected_pos, BeamstopPositions.DATA_COLLECTION)
```
and that moves it in.

src/dodal/devices/mx_phase1/beamstop.py
```diff
@@ -77,3 +86,11 @@ def _get_selected_position(self, x: float, y: float, z: float) -> BeamstopPositi
             return BeamstopPositions.DATA_COLLECTION
         else:
             return BeamstopPositions.UNKNOWN
+
+    async def _set_selected_position(self, position: BeamstopPositions) -> None:
+        if position == BeamstopPositions.DATA_COLLECTION:
+            await self.x_mm.set(self._in_beam_xyz_mm[0])
+            await self.y_mm.set(self._in_beam_xyz_mm[1])
+            await self.z_mm.set(self._in_beam_xyz_mm[2])
```
DominicOram, line 94: Should: I think all these motors can be set in parallel, e.g. we can do something like https://github.com/DiamondLightSource/dodal/blob/e07801a81ef458e1ea64b5db4dbd12731bb0e39f/src/dodal/devices/aperturescatterguard.py#L286

DominicOram (changes requested):
Great, thanks. Couple of comments. Could you also add into the `mx-bluesky` plan to actually move it in?

src/dodal/devices/mx_phase1/beamstop.py
jacob720: That's good to know, I was thinking there must be a way to stop them moving 1 by 1

src/dodal/devices/mx_phase1/beamstop.py
```diff
@@ -39,6 +43,8 @@ class Beamstop(StandardReadable):
         z: beamstop z position in mm
         selected_pos: Get the current position of the beamstop as an enum. Currently this
             is read-only.
+        in_beam_position_mm: Dictionary containing the x, y, z positions in mm for when
+            the beamstop is in beam.
```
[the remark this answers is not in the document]
jacob720: Have updated this docstring

DominicOram (approved):
Great, thanks! I don't think this actually fixes the issue until you're calling it in `mx-bluesky` though. Which PR is that in?

PR #1234, opened 2025-05-16
Allow nested lists in beamline parameters

### LINKED ISSUE #1233: Fix beamline parameters parser to allow nested lists
Currently the beamline parameters parser can't deal with nested lists. When trying to parse a beamline parameters file with nested lists, a warning is thrown and the line in question is parsed as a string. 

## Acceptance Criteria
* The `parse_value` function in `beamline_parameters.py` can deal with a string containing nested lists
* There are tests to check this works

jacob720:
Fixes #1233

Allows nested lists in beamline parameters file, which the parser couldn't previously deal with.

1. Pass beamline parameters file containing nested lists
2. Confirm all lines are parsed correctly with no warnings or errors

### Checks for reviewer

### DIFF
--- a/src/dodal/common/beamlines/beamline_parameters.py
+++ b/src/dodal/common/beamlines/beamline_parameters.py
@@ -73,8 +73,23 @@ class GDABeamlineParameters:
             return float(value)
 
     @classmethod
-    def find_first_index(cls, s: str, chars: Iterable):
-        return min((s.find(c) for c in chars if c in s), default=-1)
+    def find_first_index(cls, value: str, chars: Iterable):
+        return min((value.find(c) for c in chars if c in value), default=-1)
+
+    @classmethod
+    def find_close_square_bracket(cls, value: str):
+        index = 0
+        open = 1
+        close = 0
+        if not value[0] == "[":
+            raise ValueError(f"List must start with '[': {value}")
+        while open > close:
+            index += 1
+            if value[index] == "[":
+                open += 1
+            elif value[index] == "]":
+                close += 1
+        return index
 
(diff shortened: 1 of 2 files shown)


src/dodal/common/beamlines/beamline_parameters.py
```diff
@@ -71,17 +72,45 @@ def parse_list_element(cls, value: str):
         else:
             return float(value)
 
+    @classmethod
+    def find_first_index(cls, value: str, chars: Iterable):
+        return min((value.find(c) for c in chars if c in value), default=-1)
+
+    @classmethod
+    def find_close_square_bracket(cls, value: str):
```
DiamondJoseph, line 80: What filetype are the beamline parameter files? Is it possible to convert to JSON or something that can be parsed by existing libraries?

DominicOram (changes requested):
Great, thanks. Given that the use of `Yes` to mean `True` is one of the oddities of this format can we add that to the tests?

Despite what I said this morning about not spending too much effort on this I agree with you that using `ast.literal_eval` will clean it up a lot. If the only thing we need to do to be able to use that is change `Yes` to `True` and `No` to `False` I would suggest running a string replace for that then running `ast.literal_eval` against it. If that passes all tests then it should be fine.

jacob720:
Thanks, I've implemented this. The only weird case I can think of is if somehow there was a string like `Nowhere` which will now get parsed as `Falsewhere`, but this seems very unlikely to ever come up.

DominicOram (approved):
Great, a lot cleaner, thanks!

PR #1249, opened 2025-05-29
Removed references to CPU and GPU comparisons

adaudon:
Partially fixes mx-bluesky#449

1. Confirm tests pass

### Checks for reviewer


DominicOram (changes requested):
Great, thank you! I think we can also remove a bit more:
* `get_dict_differences` is no longer called
* This means that `deepdiff` is no longer needed in `dodal` so we can remove it from `pyproject.toml`
* However, we do use it in `mx-bluesky`, where we expect it to have come in from `dodal` so we will need to add it in to the `pyproject.toml` of `mx-bluesky`

rtuck99:
We can also remove all references to `FeatureFlags.compare_cpu_and_gpu_zocalo`

DominicOram (approved):
Looks good, thanks!

> We can also remove all references to FeatureFlags.compare_cpu_and_gpu_zocalo

I think this has already been addressed in https://github.com/DiamondLightSource/mx-bluesky/pull/1099

PR #1250, opened 2025-05-29
Add docs for where to put dodal logic

### LINKED ISSUE #617: Document guidance for deciding at which layer a piece of logic should go
As discussed with @DominicOram , it would be nice to have a flowchart-style set of rules to decide whether a piece of logic should go into `EPICS`, `ophyd-async`, or `Bluesky`

Something like this:
- If the logic only affects one device, is simple, won't change often, and/or is useful for scientists to be able to run outside of bluesky, it should go in EPICS

- If it only affects one device, and the logic is complicated, and it may need occasional changes and enhancements, it should go in `ophyd-async`. For this statement, where we have composite `ophyd-async` devices, we can treat the top level device and everything beneath it as 'one device'

- If it affects multiple devices, and/or if the logic is tied to the experiment where changes are required often, it should go in `bluesky`. 

Tagging a few people who may have thoughts:
@callumforrester @abbiemery @coretl 

## Acceptance Criteria
* A set of rules to determine where logic should go is agreed upon and documented

olliesilvester:
Fixes #617 

1. run `tox -e docs autobuild` and read the docs added
2. Furiously debate the wording

After writing this I noticed a very similar section in ophyd-async https://blueskyproject.io/ophyd-async/main/explanations/where-device-logic.html . This dodal page is more diamond specific, but I'm perhaps it's better to delete the overlap?

### Checks for reviewer


olliesilvester:
I think linking to the ophyd async docs covers a lot of what I had written, but hopefully the new flowchart helps. There is still a bit of repetition in the how to guide for creating a device, but I don't think it's too bad - it goes into more detail

PR #1251, opened 2025-05-29
Interpolate omega when no metadata

jacob720:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/1040

Interpolates omega values from previous values and timestamps, to avoid throwing out Murko results. This is done using `time.time()`, but if we could get timestamps directly from EPICS this may prove more accurate. 

Also, currently the omega velocity is calculated using just the previous two omegas and their timestamps, however it may be more accurate to use an average velocity calculated from all available omegas and timestamps. It depends if measured differences in speed are down to actual differences (in which case using just the last two makes sense), or if the velocity is very stable and differences are just down to measurement.

1. Confirm omega is interpolated correctly
2. Test on the beamline if possible

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/i04/murko_results.py
+++ b/src/dodal/devices/i04/murko_results.py
@@ -1,5 +1,6 @@
 import json
 import pickle
+import time
 from enum import Enum
 from typing import TypedDict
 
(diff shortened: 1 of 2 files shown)


DominicOram (changes requested):
I think this is better done in `MurkoCallback` as we already have access to the metadata and EPICS timestamps there. I will re-implement it.

PR #1254, opened 2025-06-02
B16 initial config

douglaswinter:
Initial devices: fds2 detector, Attocube ECC100 piezos (attol1, 2, 3, and attorot1) and sim X, Y, Z stages.


src/dodal/devices/b16/detector.py
```diff
@@ -0,0 +1,34 @@
...
+    ADBaseIO,
+    ADTIFFWriter,
+    AreaDetector,
+)
+
+from dodal.common.beamlines.beamline_utils import get_path_provider
+from dodal.common.beamlines.device_helpers import CAM_SUFFIX, TIFF_SUFFIX
+
+
+class ConstantDeadTimeController(ADBaseController):
```
DominicOram, line 12: Could: @coretl, is this something that should be in `ophyd-async` already? Feels like it might be generically useful

tests/beamlines/unit_tests/test_b16.py
```diff
@@ -0,0 +1,9 @@
+from unittest.mock import Mock
+
+from dodal.devices.b16.detector import ConstantDeadTimeController
+
```
DominicOram, line 4: Could: Would be good to have a test here that `software_triggered_tiff_area_detector` gives you what you expect too

src/dodal/beamlines/b16.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@device_factory()
+def attol3() -> Motor:
+    return Motor("BL16B-EA-ECC-03:ACT2")
+
+
+@device_factory()
+def attorot1() -> Motor:
+    return Motor("BL16B-EA-ECC-02:ACT2")
```
DominicOram, line 52: Could: I would pull these all into an `Attocube` device for neatness and then reference them in plans as e.g. `attocube.rot1`. I would also make the names more descriptive as someone who doesn't know the beamline it's not clear what the attocube actually moves or what axes `l1`, `l2` etc. are. My personal view is that in a world of autocomplete it's better to have longer descriptive names but even if you want short names a few docstrings on here that describe what the attocube moves and what each axis is would be great.

src/dodal/beamlines/b16.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@device_factory()
+def fds2() -> AreaDetector:
```
DominicOram, line 56: Could: Similarly as above with naming or docstrings

DominicOram (approved):
Reviewed based on what I would do in MX but all comments are optional

src/dodal/beamlines/b16.py
douglaswinter: Will have a think about how to organise these as b16 config grows. Although I would love to rename them (not really sure myself what the names mean) I'm hesitant to do that because staff have referred to these by these names for years. Similarly unsure about grouping them into devices - these four motors belong to two separate attocube controllers, and they have at least 6 controllers, and I don't think it means much to them which controller each axis belongs to. Maybe a generic MotorGroup device would be good for this, not necessarily mapping to a single physical device.

tests/beamlines/unit_tests/test_b16.py
douglaswinter: Sure, done.

src/dodal/beamlines/b16.py
DominicOram: > not really sure myself what the names mean

I think this definitely means that we should, if not rename them, then at least have good docstrings describing them.

PR #1255, opened 2025-06-03
Add K11 beamline module

douglaswinter:
Initial config consist of path provider and two motors.

### DIFF
--- a/src/dodal/devices/electron_analyser/__init__.py
+++ b/src/dodal/devices/electron_analyser/__init__.py
@@ -1,10 +1,15 @@
-from .abstract.base_detector import (
+from .detector import (
     ElectronAnalyserDetector,
     ElectronAnalyserRegionDetector,
     TElectronAnalyserDetector,
     TElectronAnalyserRegionDetector,
 )
-from .types import EnergyMode
+from .enums import EnergyMode
+from .types import (
+    ElectronAnalyserDetectorImpl,
+    GenericElectronAnalyserDetector,
+    GenericElectronAnalyserRegionDetector,
+)
 from .util import to_binding_energy, to_kinetic_energy
 
 __all__ = [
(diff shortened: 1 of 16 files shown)


src/dodal/beamlines/k11.py
```diff
@@ -0,0 +1,35 @@
...
+)
+
+
+@device_factory()
+def kb_cs_x() -> Motor:
+    return Motor("BL11K-OP-KBM-01:CS:X")
+
+
+@device_factory()
+def kb_cs_y() -> Motor:
```
DominicOram, line 34: Could: As with the other PR, I think this could do with some more descriptive names and/or docstrings. Also, if they're moving the same thing, just in different axes, I would combine them into a single device

src/dodal/devices/motors.py
```diff
@@ -59,3 +61,24 @@ def __init__(
...
+
+class MotorGroup(StandardReadable):
+    """
+    Generic group of motors, dynamically created from a name -> PV dictionary
+    """
+
+    def __init__(self, motor_name_to_pv: dict[str, str], name=""):
+        with self.add_children_as_readables():
+            for raw_motor_name, pv in motor_name_to_pv.items():
+                motor_name = safe_identifier(raw_motor_name)
```
DiamondJoseph, line 74: You don't need to set the motor name- if a device's name is `""`, it will be named after it's parent-

```python
m = MotorGroup(name="foo")
m.x.name = "foo-x"
m.y.name = "foo-y"
```

src/dodal/devices/motors.py
```diff
@@ -59,3 +61,24 @@ def __init__(
...
+                setattr(self, motor_name, Motor(pv, motor_name))
```
DiamondJoseph, line 75:
```suggestion
        with self.add_children_as_readables():
            self.x = Motor(f"{prefix}X")
            self.y = Motor(f"{prefix}T")
```

Then in your beamline module you just need:


```python
@device_factory()
def kb() -> MotorGroup:
    return MotorGroup(f"{PREFIX.beamline_prefix}-OP-KBM-01:CS:")
```

src/dodal/devices/motors.py
DiamondJoseph: And devices that are created with `@device_factory()` will take the name of the function if their names are `""`

DiamondJoseph (changes requested):

src/dodal/devices/motors.py
douglaswinter: Reverted the scientifically-insignificant class

PR #1257, opened 2025-06-03
Waiting for xbpm feedback

adaudon:
Fixes DiamondLightSource/mx-bluesky#1003

1. Check tests pass

### Checks for reviewer


src/dodal/devices/xbpm_feedback.py
```diff
@@ -22,6 +26,29 @@ def __init__(self, prefix: str = "", name: str = "xbpm_feedback") -> None:
...
+                    LOGGER.info("Waiting for XBPM")
+                    await asyncio.sleep(delay)
+                    n += 1
+
+        log_task = asyncio.create_task(periodic_reminder())
+
+        try:
+            async for value in observe_value(self.pos_stable):
+                if value:
+                    log_task.cancel()
```
DominicOram, line 51: Nit: The `finally` block also gets called after the return statement so the extra `cancel` here is not necessary.

src/dodal/devices/xbpm_feedback.py
```diff
@@ -22,6 +26,29 @@ def __init__(self, prefix: str = "", name: str = "xbpm_feedback") -> None:
 
     @AsyncStatus.wrap
     async def trigger(self):
-        async for value in observe_value(self.pos_stable):
-            if value:
-                return
+        async def periodic_reminder():
```
DominicOram, line 29: Could: It would be nice to put this in a separate function in case others find it useful. If you do pull it out it should:
* Go in somewhere like `common/device_utils`
* Have a good docstring describing what it is doing
* Optionally have a nice interface for the caller so that they do not need to worry about how to clean it up e.g.
```python
with periodic_logger("My log message"):
    ...
```
Where the `with` will always clean up the task (see https://docs.python.org/3/library/contextlib.html)

I recommend you give this a go as a learning opportunity, even if we don't end up reusing the logger much.

src/dodal/devices/xbpm_feedback.py
```diff
@@ -22,6 +26,29 @@ def __init__(self, prefix: str = "", name: str = "xbpm_feedback") -> None:
...
+            schedule = [  # seconds, frequency
+                (0.1, 3),
```
DominicOram, line 31: Should: I think the gap here is a bit small:
```suggestion
                (1, 3),
```

tests/devices/unit_tests/test_xbpm_feedback.py
```diff
@@ -41,3 +41,26 @@ def plan():
         yield from bps.wait(0.1)
 
     RE(plan())
+
+
+def test_logging_while_waiting_for_XBPM(
```
DominicOram, line 46: Should: Ah, I see why you went for the `0.1` length now. How about instead doing something like:
```python
@patch("dodal.devices.xbpm_feedback.asyncio.sleep")
def test_logging_while_waiting_for_XBPM(
    asyncio_sleep: MagicMock, RE: RunEngine, fake_xbpm_feedback: XBPMFeedback, caplog
):
    set_mock_value(fake_xbpm_feedback.pos_stable, False)

    current_number_of_sleep_calls = 0
    number_of_sleep_calls_before_stable = 10

    def go_stable_after_a_number_of_sleep_calls(*args, **kwargs):
        nonlocal current_number_of_sleep_calls
        if current_number_of_sleep_calls > number_of_sleep_calls_before_stable:
            set_mock_value(fake_xbpm_feedback.pos_stable, True)
        current_number_of_sleep_calls += 1

    asyncio_sleep.side_effect = go_stable_after_a_number_of_sleep_calls

    with caplog.at_level("INFO"):
        RE(bps.trigger(fake_xbpm_feedback))

    # asyncio_sleep now contains all the calls to sleep, which you can test against
    # caplog contains all the messages, which you can test against
```
This shouldn't take any time and will allow you to test for various different numbers of `number_of_sleep_calls_before_stable`

DominicOram (changes requested):
Thanks! Some comments in code, additionally can you make the PR title a bit more readable as it will go in the auto-generated release notes

src/dodal/devices/xbpm_feedback.py
adaudon: That's fair, removed. It was suggested to me to ensure it's cancelled but if it's redundant then best to remove it.

PR #1271, opened 2025-06-06
Enable discovery of the baton device now that the PVs are present

rtuck99:
Remove skip=True from the i03 baton device, and update with PV prefix
now that the device is available.

This is implemented as part of 
* DiamondLightSource/mx-bluesky#987

See related PRs:
* DiamondLightSource/blueapi#1048
* mx-bluesky TBD

1. `dodal connect i03` connects to the baton device

### Checks for reviewer

### DIFF
--- a/tests/devices/unit_tests/test_bart_robot.py
+++ b/tests/devices/unit_tests/test_bart_robot.py
@@ -1,5 +1,3 @@
-"""PYTEST_DONT_REWRITE"""
-
 import traceback
 from asyncio import create_task, sleep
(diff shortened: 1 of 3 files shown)


tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -1,3 +1,6 @@
+"""PYTEST_DONT_REWRITE"""
```
DominicOram, line 1: Should: Don't think this is needed?

DominicOram (approved):
Great, thanks. Just the one comment.

PR #1277, opened 2025-06-12
chore: Unify Motor group implementations

DiamondJoseph:
Unifies basic devices made of groups of motors to allow plans to be re-used.

1. Check that beamlines that have been modified still connect to all devices
2. Check that devices still emit documents of the same kind, and fields are still the same
nb. P45 is currently out of order, so cannot be checked

### Checks for reviewer


src/dodal/devices/motors.py
```diff
@@ -2,60 +2,147 @@
...
+
+    """
+
+    def __init__(
+        self, prefix: str, name: str = "", x_infix: str = "X", y_infix: str = "Y"
+    ):
+        with self.add_children_as_readables():
+            self.x = Motor(prefix + x_infix)
+            self.y = Motor(prefix + y_infix)
+        super().__init__(name=name)
```
DiamondJoseph, line 53: Or should these be unified as TwoAxisStage, with self.fast_axis and self.slow_axis and configuration to decide what it is called on read? I just would like to avoid the construction of device classes where you do not know what attrs each instance has, to aid in plan development.

olliesilvester (approved):
A nice improvement. Motor devices connected, except for b01-1 and training rig, but they also wasn't connecting on main.

The only wrinkle is that in MX we try to put the units in signal variable names, eg in our [Beamstop](https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/mx_phase1/beamstop.py). This is a separate issue really though (https://github.com/bluesky/ophyd-async/issues/657)

src/dodal/devices/motors.py
```diff
@@ -2,60 +2,173 @@
...
+        with self.add_children_as_readables():
+            self.x = Motor(prefix + x_infix)
+            self.theta = Motor(prefix + theta_infix)
+        super().__init__(name=name)
+
+
+class XYStage(StandardReadable):
+    """
+
+    ophyd_async x y motor stage, combining 2 Motors, mutually perpendicular to the beam.
```
DominicOram, line 34: Nit: Coordinate systems are horrible beasts. Some issues with defining it here:
* You haven't defined it in any of the other generic devices, we should probably do it in all of them or none
* There are very poor standards on coordinate systems in general, I am sure someone is using y to be parallel to the beam (and z to be vertical) somewhere
* Some devices have moving frames of reference compared to the beam. For example, our gonio has x,y as perpendicular to the beam when it is at an omega of 0 and y parallel to the beam when at omega 90

I would suggest maybe changing to something like "Usually defined as x and y perpendicular to the beam and y vertical in lab space".

src/dodal/devices/motors.py
```diff
@@ -2,60 +2,173 @@
...
-            self.omega = Motor(prefix + infix[5])
-        super().__init__(name=name, prefix=prefix, infix=infix[0:3])
+            self.kappa = Motor(prefix + kappa_infix)
+            self.phi = Motor(prefix + phi_infix)
+            self.omega = Motor(prefix + omega_infix)
+        super().__init__(prefix, name, x_infix, y_infix, z_infix)
+
+
+class YZStage(StandardReadable):
+    """Physical motion for detector travel
```
DominicOram, line 152: Should: There's no reason this has to have anything to do with detectors

src/dodal/devices/motors.py
```diff
@@ -2,60 +2,173 @@
...
+
+    ophyd_async y z motor stage, combining 2 Motors: a horizontal stage parallel to the
+    beam and a vertical stage perpendicular to it.
```
DominicOram, line 155: Nit: It's not immediately clear which of the description refers to y and which to z. Again, I would also say something like "usually"

src/dodal/devices/smargon.py
```diff
@@ -116,7 +116,7 @@ class CombinedMove(TypedDict):
     chi: NotRequired[float | None]
 
 
-class Smargon(StandardReadable, Movable):
+class Smargon(XYZStage, Movable):
```
DominicOram, line 119: Not for self really, this should be a `SixAxisGonio`, we can fix this up at a later date though

src/dodal/devices/motors.py
```diff
@@ -2,60 +2,173 @@
...
+            self.z = Motor(prefix + z_infix)
+        super().__init__(prefix, name, x_infix, y_infix)
 
 
-class SixAxisGonio(XYZPositioner):
+class XYZThetaStage(XYZStage):
+    """
+
+    ophyd_async x y z theta motor stage, combining 3  mutually perpendicular Motors and
+    a rotational stage in theta.
```
DominicOram, line 89: Nit: What do you mean by "in theta" here? Is theta a specifically defined axis or is it just one of any rotation axes about x, y or z? Given you have defined a co-ordinate system elsewhere this is now ill-defined.

src/dodal/devices/motors.py
```diff
@@ -2,60 +2,173 @@
...
-        super().__init__(name=name)
+            self.z = Motor(prefix + z_infix)
+        super().__init__(prefix, name, x_infix, y_infix)
 
 
-class SixAxisGonio(XYZPositioner):
+class XYZThetaStage(XYZStage):
+    """
+
+    ophyd_async x y z theta motor stage, combining 3  mutually perpendicular Motors and
```
DominicOram, line 88: Nit: Remove additional space, same in all other docstrings
```suggestion
    ophyd_async x y z theta motor stage, combining 3 mutually perpendicular Motors and
```

DominicOram (changes requested):
Thanks, I had been meaning to do this for a while. Comments are mainly around co-ordinate systems, which have tripped us up multiple times. Additionally, could you add a line in https://diamondlightsource.github.io/dodal/main/how-to/create-device.html#reusing-an-existing-class about reusing these rather than making your own?

DiamondJoseph:
I'm going to re-base this on top of #1305 because it will be a pain otherwise and that change will make upping the test coverage in this one easier.

src/dodal/devices/motors.py
```diff
@@ -1,61 +1,118 @@
...
-    """
 
-    Standard ophyd_async xyz motor stage, by combining 3 Motors,
-    with added infix for extra flexibliy to allow different axes other than x,y,z.
+class Stage(StandardReadable, ABC):
+    """
+    For these devices, the following co-ordinates are typical but not enforced:
+    - z is horizontal & parallel to the direction of beam travel
+    - y is vertical and antiparallel to the force of gravity
+    - x is the cross product of y🞬z
```
rtuck99, line 14: Perhaps should change this to plain text as this code point doesn't render here or in my editor

src/dodal/devices/motors.py
```diff
@@ -1,61 +1,118 @@
...
+            self.phi = Motor(prefix + phi_infix)
+            self.omega = Motor(prefix + omega_infix)
+        super().__init__(prefix, name, x_infix, y_infix, z_infix)
+
+
+class YZStage(StandardReadable):
+    def __init__(
+        self, prefix: str, name: str = "", y_infix: str = _Y, z_infix: str = _Z
+    ) -> None:
+        self.y = Motor(prefix + y_infix)
```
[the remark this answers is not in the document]
DiamondJoseph: Good catch

Issue #1278, opened 2025-06-13
Linting failing due to incompatibility with event-model 1.23


DiamondJoseph:
Event-model 1.23 modifies the shape field from list[int] to list[int | None]- until bluesky picks up this change in its next release (1.14.3 or 1.15.0?) we must pin to specific versions of event-model for compatibility

## Acceptance Criteria
* event-model is unpinned and we support whatever version comes with bluesky

PR #1279, opened 2025-06-13
fix: Temporarily pin event-model to fix CI

DiamondJoseph:
Linting in CI is currently failing as it fetches the latest version of event model (as bluesky only pins to >=1.19.8). 1.23 which released earlier in the week changes shape in a DataKey from list[int] to list[int | None]. This temporarily pins the version of event-model until a release of bluesky has been made which picks up this new change.

1. Confirm that CI passes

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
     "redis",
     "deepdiff",
     "scanspec>=0.7.3",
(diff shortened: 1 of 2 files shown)


callumforrester (approved):
Linking for posterity: https://github.com/DiamondLightSource/dodal/issues/1278

PR #1281, opened 2025-06-13
Add LensMode type to electron analyser drivers

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/1284. Add different LensMode values and added it as a type passed to the analyser. Updated tests and added more to check valid lens modes.

2. Check tests pass
3. Check new tests are valid

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/electron_analyser/types.py
+++ b/src/dodal/devices/electron_analyser/types.py
@@ -20,23 +20,29 @@ from dodal.devices.electron_analyser.vgscienta.detector import (
     VGScientaDetector,
 )
 
-# Alias names so can read the type placements easier
-AcquisitionMode = StrictEnum
-LensMode = SupersetEnum
-ElectronAnalyserDetectorImpl = VGScientaDetector[LensMode] | SpecsDetector[LensMode]
+AnyAcqMode = StrictEnum
+AnyLensMode = SupersetEnum | StrictEnum
+
+# Electron analyser types that encompasses all implementations, useful for tests and
+# plans
+ElectronAnalyserDetectorImpl = (
+    VGScientaDetector[AnyLensMode] | SpecsDetector[AnyLensMode]
+)
 ElectronAnalyserDriverImpl = (
-    VGScientaAnalyserDriverIO[LensMode] | SpecsAnalyserDriverIO[LensMode]
+    VGScientaAnalyserDriverIO[AnyLensMode] | SpecsAnalyserDriverIO[AnyLensMode]
 )
 
-AbstractBaseRegion = AbstractBaseRegion[AcquisitionMode, LensMode]
+# Short hand the type so less verbose
+AbstractBaseRegion = AbstractBaseRegion[AnyAcqMode, AnyLensMode]
 
+# Generic electron analyser types that supports full typing with the abstract classes.
 GenericElectronAnalyserDetector = ElectronAnalyserDetector[
-    AbstractAnalyserDriverIO[AbstractBaseRegion, AcquisitionMode, LensMode],
-    AbstractBaseSequence[AbstractBaseRegion, LensMode],
+    AbstractAnalyserDriverIO[AbstractBaseRegion, AnyAcqMode, AnyLensMode],
+    AbstractBaseSequence[AbstractBaseRegion, AnyLensMode],
     AbstractBaseRegion,
 ]
 
 GenericElectronAnalyserRegionDetector = ElectronAnalyserRegionDetector[
-    AbstractAnalyserDriverIO[AbstractBaseRegion, AcquisitionMode, LensMode],
+    AbstractAnalyserDriverIO[AbstractBaseRegion, AnyAcqMode, AnyLensMode],
     AbstractBaseRegion,
 ]


src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -125,10 +131,12 @@ async def set(self, region: TAbstractBaseRegion):
             self.low_energy.set(low_energy),
             self.high_energy.set(high_energy),
             self.slices.set(region.slices),
-            self.lens_mode.set(region.lens_mode),
+            self.lens_mode.set(self.lens_mode_type(region.lens_mode)),
             self.pass_energy.set(pass_energy),
             self.iterations.set(region.iterations),
-            self.acquisition_mode.set(region.acquisition_mode),
+            self.acquisition_mode.set(
```
olliesilvester, line 137: draft: check if typing can be improved on this

src/dodal/beamlines/i09.py
```diff
@@ -1,14 +1,22 @@
...
+from dodal.devices.i09 import DCM, Grating, LensMode
 from dodal.devices.pgm import PGM
 from dodal.devices.synchrotron import Synchrotron
 from dodal.log import set_beamline as set_log_beamline
 from dodal.utils import BeamlinePrefix, get_beamline_name
 
+
+class EnergyMode(StrictEnum):
+    KINETIC = "Kinetic"
+    BINDING = "Binding"
```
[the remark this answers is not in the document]
olliesilvester: I also can't see this `Enum` being used anywhere yet, probably better to remove it until there's a PR which uses it

olliesilvester (changes requested):
A nice change, there's just a few places where the Generic type wasn't passed through properly, so we aren't actually getting the static typing benefit yet

src/dodal/beamlines/b07_1.py
```diff
@@ -22,12 +22,13 @@ def synchrotron() -> Synchrotron:
...
 
 
 @device_factory()
-def analyser_driver() -> SpecsAnalyserDriverIO:
-    return SpecsAnalyserDriverIO(
+def analyser_driver() -> SpecsAnalyserDriverIO[LensMode]:
+    return SpecsAnalyserDriverIO[LensMode](
         f"{PREFIX.beamline_prefix}-EA-DET-01:CAM:",
+        LensMode,
         {"source1": pgm().energy.user_readback},
```
Villtord, line 33: This seems to be excessive from first view - LensMode is used 2 times

Villtord (changes requested):
My suggestion is to remove TLensMode from base abstract classes and just pass it as a parameter (as it does both right now, not sure why). This brakes quite some tests on analyser, but I still think it's worth doing as makes it a bit simpler to instantiate/use that classes.

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -24,13 +24,16 @@
...
+TLensMode = TypeVar("TLensMode", bound=StrictEnum)
+
 
 class AbstractAnalyserDriverIO(
     ABC,
     StandardReadable,
     ADBaseIO,
     Movable[TAbstractBaseRegion],
-    Generic[TAbstractBaseRegion],
+    Generic[TAbstractBaseRegion, TAcquisitionMode, TLensMode],
```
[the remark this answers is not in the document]
oliwenmandiamond: The reason both are needed is because unfortunately python doesn't behave the same as Java. In Java, you could just do 

`AbstractAnalyserDriverIO<LensMode> driver = new AbstractAnalyserDriverIO<>();`, you can use the generic type to be passed as a type.

Python doesn't retain generic parameters at runtime by default and Python is dynamically typed — the runtime often has no idea about T. So we have to pass it as a generic and as an argument to achieve the same effect

src/dodal/beamlines/b07_1.py
oliwenmandiamond: The static type checking is optional and doesn't need to referenced later in plans if not required. As Tom mentioned, the benefit it marginal but as we have already added it, we should leave it in.

src/dodal/devices/electron_analyser/vgscienta/region.py
```diff
@@ -14,11 +14,11 @@
...
 
 
-class VGScientaRegion(AbstractBaseRegion[AcquisitionMode]):
+class VGScientaRegion(AbstractBaseRegion):
     # Override defaults of base region class
-    lens_mode: str = "Angular45"
+    lens_mode: str
     pass_energy: int = 5
-    acquisition_mode: AcquisitionMode = AcquisitionMode.SWEPT
+    acquisition_mode: str = AcquisitionMode.SWEPT
```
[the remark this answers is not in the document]
oliwenmandiamond: So to be consistent, I've opted to make the region data model hold these values only as strings for simplicity. When trying to load pydantic models using multiple layers of inheritance with Generics, it fails because it is losing the type prorogation with lens mode as it requires it to be a StrictEnum. This is something I will try to explore in a different change though as this is getting too large

src/dodal/devices/electron_analyser/abstract/base_region.py
```diff
@@ -43,7 +44,18 @@ def energy_mode_validation(data: dict) -> dict:
     return data
 
 
-class AbstractBaseRegion(ABC, JavaToPythonModel):
+TAcquisitionMode = TypeVar("TAcquisitionMode", bound=StrictEnum)
+# Allow SupersetEnum. Specs analysers can connect to Lens mode separately to the
+# analyser which leaves the enum to either be "Not connected" OR the available enums
+# when connected.
+TLensMode = TypeVar("TLensMode", SupersetEnum, StrictEnum)
```
olliesilvester, line 51: Must: Need to use `bound` in the `TypeVar`, otherwise the static typing will let you use any `StrictEnum` or `SupersetEnum` for the `lens_mode` signals. For example, the type checker doesn't complain if I do something incorrect like

```
class MYEnum(StrictEnum):
    VAL = "val"

test_det = SpecsDetector("thing", AcquisitionMode, {"thing": epics_signal_r(int, "rd")})
test_det.driver.lens_mode.set(MYEnum.VAL)
```

However, my type-checker will complain after I do
`TLensMode = TypeVar("TLensMode", bound = SupersetEnum | StrictEnum)`

olliesilvester (changes requested):
Haven't done a full review yet, but spotted a few things

src/dodal/devices/i09/enums.py
```diff
@@ -0,0 +1,15 @@
+from ophyd_async.core import StrictEnum
+
+
+class Grating(StrictEnum):
+    G_300 = "300 lines/mm"
+    G_400 = "400 lines/mm"
+    G_800 = "800 lines/mm"
+
+
+class LensMode(StrictEnum):
```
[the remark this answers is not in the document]
oliwenmandiamond: VGScienta `LensMode` do not behave the same, they never get in a state of "Not connected", therefore they can always use a `StrictEnum` as epics always reflects what the devices enum returns.

src/dodal/devices/electron_analyser/abstract/base_region.py
```diff
@@ -46,7 +44,18 @@ def energy_mode_validation(data: dict) -> dict:
     return data
 
 
-class AbstractBaseRegion(ABC, JavaToPythonModel, Generic[TStrictEnum]):
+TAcquisitionMode = TypeVar("TAcquisitionMode", bound=StrictEnum)
+# Allow SupersetEnum. Specs analysers can connect to Lens mode separately to the
+# analyser which leaves the enum to either be "Not connected" OR the available enums
+# when connected.
+TLensMode = TypeVar("TLensMode", bound=SupersetEnum | StrictEnum)
```
Villtord, line 51: minor: should it be defined in `base_driver_io.py` rather than here?

Villtord (approved):

olliesilvester (commented):
This looks good now! I'm a bit confused about the typing aliases though, will discuss in person before approving

src/dodal/devices/electron_analyser/types.py
```diff
@@ -18,15 +20,23 @@
...
-ElectronAnalyserDriverImpl = VGScientaAnalyserDriverIO | SpecsAnalyserDriverIO
+# Alias names so can read the type placements easier
+AcquisitionMode = StrictEnum
+LensMode = SupersetEnum
+ElectronAnalyserDetectorImpl = VGScientaDetector[LensMode] | SpecsDetector[LensMode]
+ElectronAnalyserDriverImpl = (
+    VGScientaAnalyserDriverIO[LensMode] | SpecsAnalyserDriverIO[LensMode]
+)
+
+AbstractBaseRegion = AbstractBaseRegion[AcquisitionMode, LensMode]
```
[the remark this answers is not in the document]
oliwenmandiamond: As discussed, will rename to AnyLensMode and AnyAcqMode with some comments to make less confusing. These are use in plans and tests. I found I needed a generic way to define the electron analysers and I kept redefining it in sm-bluesky and tests in dodal. Therefore I defined them in a single place to import from

src/dodal/devices/electron_analyser/abstract/base_region.py
oliwenmandiamond: No, because it is now used at the region level too

E.g `VGScientaRegion[MyLensMode] `and `SpecsRegion[MyLensMode]`

However, they could belong in their own file inside abstract called types.py possibly

PR #1285, opened 2025-06-13
Hutch shutter raises an error on Open if not safe

noemifrisina:
This changes the hutch shutter to only raise an error on an `Open` demand if the hutch is not locked and searched. 
This should allow to test methods trying to close the hutch shutter during commissioning - when the hutch door might be open - as the shutter would always be closed anyway.

1. Run test

### Checks for reviewer


src/dodal/devices/hutch_shutter.py
```diff
@@ -81,14 +81,14 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
-                "The hutch has not been locked, not operating shutter."
-            )
         if not TEST_MODE:
             if value == ShutterDemand.OPEN:
+                interlock_state = await self.interlock.shutter_safe_to_operate()
+                if not interlock_state:
+                    # If not in test mode, fail. If in test mode, the optics hutch may be open.
+                    raise ShutterNotSafeToOperateError(
+                        "The hutch has not been locked, not operating shutter."
+                    )
```
DiamondJoseph, line 91: Is this something that has to be checked in plans or is there a check also done at the EPICS level?

PR #1290, opened 2025-06-17
Use correct redis murko host env variable name

olliesilvester:
Corrects an environment variable name (see https://github.com/DiamondLightSource/mx-bluesky/issues/1110#issuecomment-2980641827)

To test:
- Check that i04's BlueAPI pod can see this environment variable, and not the environment variable that existed before this commit

### Checks for reviewer


DominicOram (approved):
Great, thanks!

PR #1294, opened 2025-06-18
Use timeout built into observe_value for pin detect

### LINKED ISSUE #1293: Use timeout in `observe_value` when waiting on pin detect
As discussed (and solved) in https://github.com/bluesky/ophyd-async/pull/650 there is an edge case on observing PVs that update fast and using `asyncio.wait_for`. We have seen this come up again on i03 today (https://ispyb.diamond.ac.uk/samples/sid/6590773) where hyperion got stuck in an infinite loop when no pin was loaded. This is likely because in the `trigger` of `PinTipDetection` we are still using `asyncio.wait_for`. We should instead use the one built into `observe_value`.

## Acceptance Criteria
* The timeout internal to `observe_value` is used rather than relying on `asyncio.wait_for`

DominicOram:
Fixes #1293 

1. Confirm logic on linked issues makes sense
2. Confirm tests still pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/oav/pin_image_recognition/__init__.py
+++ b/src/dodal/devices/oav/pin_image_recognition/__init__.py
@@ -141,14 +141,16 @@ class PinTipDetection(StandardReadable):
 
     @AsyncStatus.wrap
     async def trigger(self):
-        async def _set_triggered_tip():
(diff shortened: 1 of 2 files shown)


rtuck99 (approved):
Approved

PR #1295, opened 2025-06-19
Add ccmc b07

### LINKED ISSUE #1291: Add channel cut monochromator device (ccmc) to b07-1
B07-1 in addition to plane grating monochromator PGM has another type of monochromator - channel cut - with a set of positions (energies) to be used.

## Acceptance Criteria
* dodal connect b07-1 connect ccmc

Villtord:
Fixes #1291 

1. Run dodal connect b07_1 and make sure ccmc connected
2. Run tests
3. Confirm all tests pass

### Checks for reviewer


tests/devices/unit_tests/b07_1/test_ccmc.py
```diff
@@ -0,0 +1,58 @@
...
+from bluesky.plan_stubs import mv
+from bluesky.run_engine import RunEngine
+from ophyd_async.core import init_devices
+from ophyd_async.testing import assert_configuration, assert_reading, assert_value
+
+from dodal.devices.b07_1.ccmc import CCMC, CCMCPositions
+
+
+@pytest.fixture
+async def mock_CCMC(RE: RunEngine) -> CCMC:
```
oliwenmandiamond, line 11: Change to `mock_ccmc`

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,67 @@
...
+        ----------
+        prefix:
+            Beamline specific part of the PV
+        positions:
+            The Enum for the positions names.
+        name:
+            Name of the device
+        """
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            # normal motors in epics
```
oliwenmandiamond, line 44: If normal motors, why not use from `ophyd_async.epics.motor import Motor` ?

Villtord: Fair enough, but there are 2 thoughts about it:
1. Do we really want to allow full control of those motors in GDA? The primary and main use of CCMC is via pos_selest.

2. If still yes then I think I can not put Motor object as child with CONFIG_SIGNAL flag (not sure if it's a bug or feature), so I need to put it in CHILD and then I will get motor readings with every `read()` call which I am not sure I want to.

tests/devices/unit_tests/b07_1/test_ccmc.py
```diff
@@ -0,0 +1,58 @@
...
+                "value": CCMCPositions.OUT,
+            },
+            f"{mock_CCMC.name}-energy_in_ev": {
+                "value": 0.0,
+            },
+        },
+    )
+
+
+async def test_move_crystal(
```
oliwenmandiamond, line 51: Can you also add a test to cover returning zero?

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,67 @@
...
+            self.energy_in_ev = derived_signal_r(
+                self._convert_pos_to_eV, pos_signal=self.pos_select
+            )
+
+        super().__init__(name)
+
+    def _convert_pos_to_eV(self, pos_signal: CCMCPositions) -> float:
+        if pos_signal != CCMCPositions.OUT:
+            return float(str(pos_signal.value).split("Xtal_")[1])
+        return 0.0
```
oliwenmandiamond, line 67: I think instead of returning zero, we throw an error saying there is no beam, the device isn't moved to the correct position

tests/devices/unit_tests/b07_1/test_ccmc.py
Villtord: done

tests/devices/unit_tests/b07_1/test_ccmc.py
Villtord: done

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,67 @@
...
+            self.y_rotation = epics_signal_rw(
+                float,
+                read_pv=prefix + "ROTY:POS:RD",
+                write_pv=prefix + "ROTY:MOV:WR",
+            )
+
+        with self.add_children_as_readables():
+            # Must be a CHILD as read() must return this signal
+            self.pos_select = epics_signal_rw(positions, prefix + "CRYSTAL:MP:SELECT")
+            self.energy_in_ev = derived_signal_r(
```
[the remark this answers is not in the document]
oliwenmandiamond: but pos_select/crystal already gives us the energy value, just as string though?

tests/devices/unit_tests/b07_1/test_ccmc.py
```diff
@@ -0,0 +1,82 @@
...
+import pytest
+from bluesky.plan_stubs import mv
+from bluesky.run_engine import RunEngine
+from ophyd_async.core import StrictEnum, init_devices
+from ophyd_async.testing import assert_configuration, assert_reading, assert_value
+
+from dodal.devices.b07_1.ccmc import CCMC, CCMCPositions
+
+
+class TestEnum(StrictEnum):
```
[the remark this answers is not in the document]
Villtord: oopsie

src/dodal/devices/b07_1/ccmc.py
Villtord: done

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,85 @@
...
+
+ccmc_lower_limit = 1500.0
+ccmc_upper_limit = 3000.0
+error_message = "Can not get energy value in ev from ccmc position: "
+
+
+T = TypeVar("T", bound=StrictEnum)
+
+
+class CCMC(StandardReadable, Movable[T]):
```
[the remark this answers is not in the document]
Villtord: Yes, sure

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,85 @@
...
+            self.crystal = epics_signal_rw(positions, prefix + "CRYSTAL:MP:SELECT")
+
+        # energy derived signal as property
+        self.energy_in_ev = derived_signal_r(
+            self._convert_pos_to_ev, pos_signal=self.crystal
+        )
+
+        super().__init__(name=name)
+
+    def _convert_pos_to_ev(self, pos_signal: ChannelCutMonochromatorPositions) -> float:
```
DiamondJoseph, line 76: must: you're not using the generic bounds, so any other CCMC would have to be `Movable[ChannelCutMonochromatorPositions]` 

```suggestion
    def _convert_pos_to_ev(self, pos_signal: T) -> float:
```

Either this or remove the Generic-ness of it.

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,72 @@
...
+
+class CCMCPositions(StrictEnum):
+    OUT = "Out of Beam"
+    XTAL_2000 = "Xtal_2000"
+    XTAL_2250 = "Xtal_2250"
+    XTAL_2500 = "Xtal_2500"
+
+
+ccmc_lower_limit = 1500.0
+ccmc_upper_limit = 3000.0
```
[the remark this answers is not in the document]
DiamondJoseph: Is this the only CCMC that we have or will have though? If you're being generic with the Movable type, you should consider that other devices may have different handling.

I would maybe even let each device define their own validation/conversion function?

```python
class ChannelCutMonochromator(StandardReadable):

    def __init__(
        self,
        prefix: str,
        positions: type[T],
        pos_to_ev: Callable[[T], float],
        name: str = "",
    ) -> None:
        ...

@device_factory()
def ccmc -> ChannelCutMonochromator():
  def energy_conversion(position: ChannelCutMonochromatorPositions) -> float:
      return match position:
         case XTAL_2000:
             return 2000
      ...
  return ChannelCutMonochromator(prefix, ChannelCutMonochromatorPositions, energy_conversion)

File: src/dodal/devices/b07_1/ccmc.py [hunk 10]
```diff
@@ -0,0 +1,85 @@
...
+error_message = "Can not get energy value in ev from ccmc position: "
+
+
+T = TypeVar("T", bound=StrictEnum)
+
+
+class ChannelCutMonochromator(StandardReadable, Movable[T]):
```
DiamondJoseph, line 30: We're generally opposing defining Movable when it's just being used as a shortcut to one or several fields under the device- when you're writing your plan, you can easily write

```python
def move_out(ccmc: ChannelCutMonochromator[ChannelCutMonochromatorPositions]) -> MsgGenerator:
  yield from abs_set(ccmc.crystal, ChannelCutMonochromatorPositions.OUT) 
```

or

```python
def move_out(ccmc_crystal: Movable[ChannelCutMonochromatorPositions]) -> MsgGenerator:
    yield from abs_set(ccmc_crystal, ChannelCutMonochromatorPositions.OUT) 
```

And call it from blueapi with {"ccmc": "ccmc.crystal")

src/dodal/devices/b07/__init__.py
```diff
@@ -1,3 +1,4 @@
 from dodal.devices.b07.grating import B07BGrating
 
+#
```
DiamondJoseph, line 3:
```suggestion
```

src/dodal/devices/b07_1/ccmc.py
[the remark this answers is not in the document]
DiamondJoseph: As is, these values can't be returned by your current `_convert_pos_to_ev` method either way, so they're dead code.

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,85 @@
...
+    multi-positioner. The positions are named after the nominal energies of the
+    crystals. To change energy select one of the crystals from the list.
+    This causes the Y motor to move that crystal into the beam and other
+    motors have to align the angles correctly.
+    """
+
+    def __init__(
+        self,
+        prefix: str,
+        positions: type[StrictEnum],
```
DiamondJoseph, line 43:
```suggestion
        positions: type[T],
```

DiamondJoseph (changes requested):

src/dodal/devices/b07_1/ccmc.py
Villtord: it's not a shortcut - I genuinely want to define moving of this device as only moving of crystal to enum position, any other moves must be done in epics.

src/dodal/devices/b07_1/ccmc.py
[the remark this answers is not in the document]
DiamondJoseph: > Those are used internally in _convert_pos_to_ev to validate result, why do they need to be returned at all?

```python
            energy = float(str(pos_signal.value).split("Xtal_")[1])
            if ccmc_lower_limit < energy < ccmc_upper_limit:
                return energy
```

energy can only be 2000, 2250, 2500 with the current enum - always within the limits. I'm not sure why that limit checking is required?

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,82 @@
...
+        # piezo motor in epics
+        self._y_rotation = epics_signal_rw(
+            float,
+            read_pv=prefix + "ROTY:POS:RD",
+            write_pv=prefix + "ROTY:MOV:WR",
+        )
+
+        with self.add_children_as_readables():
+            # Must be a CHILD as read() must return this signal
+            self.crystal = epics_signal_rw(positions, prefix + "CRYSTAL:MP:SELECT")
```
DiamondJoseph, line 64:
```suggestion
        name: str = "",
    ) -> None:
        """
        Parameters
        ----------
        prefix:
            Beamline specific part of the PV
        positions:
            The Enum for the positions names.
        name:
            Name of the device
        """
        # crystal motors internal
        self._xyz = XYZStage(prefix)
        # piezo motor in epics
        self._y_rotation = epics_signal_rw(
            float,
            read_pv=prefix + "ROTY:POS:RD",
            write_pv=prefix + "ROTY:MOV:WR",
        )

        with self.add_children_as_readables():
            # Must be a CHILD as read() must return this signal
            self.crystal = epics_signal_rw(ChannelCutMonochromatorPositions, prefix + "CRYSTAL:MP:SELECT")
```

Now that it's not generic passing the type isn't required anymore

tests/devices/unit_tests/b07_1/test_ccmc.py
```diff
@@ -0,0 +1,88 @@
...
+from ophyd_async.core import StrictEnum, init_devices
+from ophyd_async.testing import assert_configuration, assert_reading, assert_value
+
+from dodal.devices.b07_1.ccmc import (
+    ChannelCutMonochromator,
+    ChannelCutMonochromatorPositions,
+)
+
+
+class WrongEnum(StrictEnum):
```
oliwenmandiamond, line 13: This isn't used anywhere? Do you need to add a test to check if moving to position using enum that is wrong will throw an error?

Villtord: Thanks, this test got lost in multiple iterations commits

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,81 @@
...
+            Name of the device
+        """
+        # crystal motors internal
+        self._xyz = XYZStage(prefix)
+        # piezo motor in epics
+        self._y_rotation = epics_signal_rw(
+            float,
+            read_pv=prefix + "ROTY:POS:RD",
+            write_pv=prefix + "ROTY:MOV:WR",
+        )
```
DiamondJoseph, line 57: Are these signals effectively read-only signals that are set when the Crystal is set, or will they require manual setting? Do you ever want to manually set or adjust them?

DiamondJoseph (changes requested):
Lots of nits and one thing that absolutely must be changes if I can't persuade you on the nits

src/dodal/devices/pgm.py
```diff
@@ -37,4 +37,4 @@ def __init__(
             self.grating = epics_signal_rw(grating, prefix + gratingPv)
             self.cff = epics_signal_rw(float, prefix + "CFF")
 
-        super().__init__(name=name)
+        super().__init__(name)
```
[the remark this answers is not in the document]
Villtord: have no idea why I changed that - probably unintentionally while resolving merge

src/dodal/devices/b07_1/ccmc.py
Villtord: Current use case is that users never set them and they could be essentially read-only. Not sure about future beamline plans in case they need a tweak - might be very helpful to have them writeable and include that micro-adjustment in plan. Still in that case long-term solution is on the Epics side I assume.
In GDA they are just a group of scannable motors (all write-read)

tests/devices/unit_tests/b07_1/test_ccmc.py
```diff
@@ -0,0 +1,112 @@
...
+    mock_ccmc: ChannelCutMonochromator,
+):
+    await assert_reading(
+        mock_ccmc._xyz,
+        {
+            f"{mock_ccmc.name}-_xyz-x": {"value": 0.0},
+            f"{mock_ccmc.name}-_xyz-y": {"value": 0.0},
+            f"{mock_ccmc.name}-_xyz-z": {"value": 0.0},
+        },
+    )
```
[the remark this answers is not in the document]
Villtord: yes, probably I do need to include them in every read as it is now in GDA. But that won't change their ugly names.

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,81 @@
...
+    def _convert_pos_to_ev(self, pos_signal: ChannelCutMonochromatorPositions) -> float:
+        if pos_signal != ChannelCutMonochromatorPositions.OUT:
+            energy = float(str(pos_signal.value).split("Xtal_")[1])
+            if ccmc_lower_limit < energy < ccmc_upper_limit:
+                return energy
+        raise ValueError(error_message)
+
+    @AsyncStatus.wrap
+    async def set(self, demand: ChannelCutMonochromatorPositions) -> None:
+        await self.crystal.set(demand, wait=True)
```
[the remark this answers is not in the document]
DiamondJoseph: Right, sorry it's AsyncMovable https://github.com/bluesky/ophyd-async/blob/ecff1d1a7d9711dfd18d2dd33df532bc3be7b907/src/ophyd_async/core/_protocol.py#L102 that requires `value`, and since you're return an AsyncStatus you'd be compliant with that protocol if it was named the same but otherwise are not.

If a device isn't a fully finished design then defining a `set` method prematurely- and potentially having to refactor plans that have started using it if you decide there's a more useful `set` method that e.g. takes both the crystal and some fields of the xyz/rotational stages that you'd prefer to have access to- is more work than writing your initial plans with `abs_set(ccmc.crystal)` being a redundant call vs. `abs_set(ccmc)`

src/dodal/devices/b07_1/ccmc.py
```diff
@@ -0,0 +1,81 @@
...
+class ChannelCutMonochromatorPositions(StrictEnum):
+    OUT = "Out of Beam"
+    XTAL_2000 = "Xtal_2000"
+    XTAL_2250 = "Xtal_2250"
+    XTAL_2500 = "Xtal_2500"
+
+
+ccmc_lower_limit = 1500.0
+ccmc_upper_limit = 3000.0
+error_message = "Can not get energy value in ev from ccmc position: "
```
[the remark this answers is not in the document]
Villtord: done

PR #1296, opened 2025-06-19
add i20 module and gas_injector

### LINKED ISSUE #1289: Add a gas injection rig device
Based on the i18 spectroscopy work - it seems better to put this logic into a device

https://github.com/DiamondLightSource/spectroscopy-bluesky/pull/22

need to use the monitoring feature

```python
from ophyd_async.core import observe_value, observe_signals_value
 line_pressure_iterator = observe_value(
 gas_injector.pressure_controller_1.readout
 )
 value = await anext(line_pressure_iterator)

```

update: will also be used at b18 at some point https://github.com/DiamondLightSource/dodal/issues/1309

stan-dot:
Fixes #1289

1. use bluesky locally on i20
2. Run `dodal connect i20`
3. confirm that the gas_injector device connects

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/i20/gas_injector.py
+++ b/src/dodal/devices/i20/gas_injector.py
@@ -55,20 +55,16 @@ class GasInjector(StandardReadable):
         with self.add_children_as_readables():
             self.vacuum_pump = epics_signal_rw(int, prefix + "VACP1:CON")
             self.line_valve = epics_signal_rw(int, prefix + "V5:CON")
-            # Gas valves as a dict
-            self.gas_valves = {
-                GasToInject.KRYPTON: epics_signal_rw(int, prefix + "V1:CON"),
-                GasToInject.NITROGEN: epics_signal_rw(int, prefix + "V2:CON"),
-                GasToInject.ARGON: epics_signal_rw(int, prefix + "V3:CON"),
-                GasToInject.HELIUM: epics_signal_rw(int, prefix + "V4:CON"),
-            }
-            # Chamber valves as a dict
-            self.chambers = {
-                IonChamberToFill.i0: epics_signal_rw(int, prefix + "V6:CON"),
-                IonChamberToFill.iT: epics_signal_rw(int, prefix + "V7:CON"),
-                IonChamberToFill.iRef: epics_signal_rw(int, prefix + "V8:CON"),
-                IonChamberToFill.i1: epics_signal_rw(int, prefix + "V9:CON"),
-            }
+            # Gas valves as flat signals
+            self.gas_valve_krypton = epics_signal_rw(int, prefix + "V1:CON")
+            self.gas_valve_nitrogen = epics_signal_rw(int, prefix + "V2:CON")
+            self.gas_valve_argon = epics_signal_rw(int, prefix + "V3:CON")
+            self.gas_valve_helium = epics_signal_rw(int, prefix + "V4:CON")
+            # Chamber valves as flat signals
+            self.chamber_i0 = epics_signal_rw(int, prefix + "V6:CON")
+            self.chamber_it = epics_signal_rw(int, prefix + "V7:CON")
+            self.chamber_iref = epics_signal_rw(int, prefix + "V8:CON")
+            self.chamber_i1 = epics_signal_rw(int, prefix + "V9:CON")
             self.pressure_controller_1 = PressureController(
                 prefix + "PCTRL1:", name="pressure_controller_1"
             )
@@ -77,10 +73,30 @@ class GasInjector(StandardReadable):
             )
 
     def get_gas_valve(self, gas: GasToInject):
-        return self.gas_valves[gas]
+        match gas:
+            case GasToInject.KRYPTON:
+                return self.gas_valve_krypton
+            case GasToInject.NITROGEN:
+                return self.gas_valve_nitrogen
+            case GasToInject.ARGON:
+                return self.gas_valve_argon
+            case GasToInject.HELIUM:
+                return self.gas_valve_helium
+            case _:
+                raise ValueError(f"Unknown gas: {gas}")
 
     def get_chamber_valve(self, chamber: IonChamberToFill):
-        return self.chambers[chamber]
+        match chamber:
+            case IonChamberToFill.i0:
+                return self.chamber_i0
+            case IonChamberToFill.iT:
+                return self.chamber_it
+            case IonChamberToFill.iRef:
+                return self.chamber_iref
+            case IonChamberToFill.i1:
+                return self.chamber_i1
+            case _:
+                raise ValueError(f"Unknown chamber: {chamber}")
 
     async def wait_for_equilibration(
         self,


tests/devices/i20/test_gas_injector.py
```diff
@@ -0,0 +1,221 @@
...
+        patch(
+            "dodal.devices.i20.gas_injector.epics_signal_rw",
+            side_effect=lambda *a, **k: AsyncMock(),
+        ) as _,
+        patch(
+            "dodal.devices.i20.gas_injector.epics_signal_r",
+            side_effect=lambda *a, **k: AsyncMock(),
+        ),
+    ):
+        yield GasInjector("BLXX:")
```
DiamondJoseph, line 28:
```suggestion
    # Patch epics_signal_rw and epics_signal_r to return AsyncMocks
    with init_devices(mock=True):
        yield GasInjector("BLXX:")
```

Why do you need the signals to be AsyncMocks? Don't you just need them to have Mock backends?

.vscode/tasks.json
```diff
@@ -11,14 +11,6 @@
...
             },
             "problemMatcher": [],
-        },
-        {
-            "label": "load_dials_env",
-            "command": "module",
-            "args": [
-                "load dials/latest"
-            ],
-            "type": "shell"
```
DiamondJoseph, line 21: While I appreciate it is annoying that it pops up when trying to debug in dodal, this should be removed as it's own ticket- if there isn's a ticket already pelase make one. I'm sure MX have some insight and can make this less disruptive

src/dodal/devices/i20/gas_injector.py
```diff
@@ -0,0 +1,201 @@
...
+    HELIUM = "helium"
+    KRYPTON = "krypton"
+    NITROGEN = "nitrogen"
+
+
+class IonChamberToFill(enum.Enum):
+    i0 = "I0"
+    i1 = "I1"
+    iT = "iT"
+    iRef = "iRef"
```
DiamondJoseph, line 20: ~~These should extend ophyd_async's StrictEnum~~ 
only if they're being used as Signal types. But they're not

src/dodal/devices/i20/gas_injector.py
```diff
@@ -0,0 +1,201 @@
...
+                GasToInject.ARGON: epics_signal_rw(int, prefix + "V3:CON"),
+                GasToInject.HELIUM: epics_signal_rw(int, prefix + "V4:CON"),
+            }
+            # Chamber valves as a dict
+            self.chambers = {
+                IonChamberToFill.i0: epics_signal_rw(int, prefix + "V6:CON"),
+                IonChamberToFill.iT: epics_signal_rw(int, prefix + "V7:CON"),
+                IonChamberToFill.iRef: epics_signal_rw(int, prefix + "V8:CON"),
+                IonChamberToFill.i1: epics_signal_rw(int, prefix + "V9:CON"),
+            }
```
DiamondJoseph, line 71: I'm really opposed to having arbitrary dicts in device definitions- these could be modified on the instance, plans can't do type checking on what is returned or reliably know which attrs exist. I might make Chambers and Valves their own definitions, with normal field definitions, or else just make these normal field definitions on the type- you can still have the `get_` methods if it makes writing the plan simpler?

src/dodal/devices/i20/gas_injector.py
```diff
@@ -0,0 +1,201 @@
...
+            self.pressure_controller_1 = PressureController(
+                prefix + "PCTRL1:", name="pressure_controller_1"
+            )
+            self.pressure_controller_2 = PressureController(
+                prefix + "PCTRL2:", name="pressure_controller_2"
+            )
+
```
DiamondJoseph, line 78:
```suggestion
            self.pressure_controller_1 = PressureController(
                prefix + "PCTRL1:"
            )
            self.pressure_controller_2 = PressureController(
                prefix + "PCTRL2:"
            )

```

Should: child devices get automatically named by the parent on connect, you shouldn't overwrite that without good reason

DiamondJoseph (changes requested):

src/dodal/devices/i20/gas_injector.py
```diff
@@ -0,0 +1,201 @@
...
+                current=current_pressure,
+                initial=line_pressure,
+                target=LIMIT_PRESSURE,
+                time_elapsed=time.monotonic() - start,
+            )
+            if abs(current_pressure - LIMIT_PRESSURE) < tolerance:
+                break
+
+        await self.line_valve.set(ValveCommands.CLOSE.value)
+        await self.vacuum_pump.set(VacuumPumpCommands.OFF.value)
```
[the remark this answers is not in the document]
stan-dot: maybe, I never heard of that

src/dodal/devices/i20/gas_injector.py
stan-dot: how could I know?

src/dodal/devices/i20/gas_injector.py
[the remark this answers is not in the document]
stan-dot: it's literallly used only in 2 other places in the whole of dodal. 

but it's fine, I applied this into the new commit

src/dodal/devices/i20/gas_injector.py
DiamondJoseph: These signals also won't be connected as they're not child attrs of a device that is being connected.

stan-dot: wait, they are inside the `self.add_children_as_readables` ?? are they not?

DiamondJoseph: `self.chambers` is a child of `self`, but `self.chambers` isn't a Device and `chambers[IonChamberToFill.i0]` isn't a child of `self.chambers` if it were. 

Compare [DeviceVector from ophyd-async](https://github.com/bluesky/ophyd-async/blob/9f724d25e39b1be30f6441566a8153acd3388e77/src/ophyd_async/core/_device.py#L229) which is a Device and its elements are its children because of how it overwrites `children(self)`. See [discussion here about support DeviceVector mappings](https://github.com/bluesky/ophyd-async/issues/903) which as I've said I don't think should be done.

DiamondJoseph:
> how to get the float value then? I'm trying different things and the IDE and CI are both complaining

try `check_pressure = await chamber_pressure.readout.get_value()`

stan-dot:
from conversation with @iain-hall  -this requires local testing with RunEngine - even before we get the cluster at i20.

therefore we suspend this PR for now.

src/dodal/devices/i20/gas_injector.py
DiamondJoseph: If they're Devices (which Signals are) and are members of the class, they will be found by the `children(self)` method and connected.

stan-dot: I did ask in `athena_componetns` what is the best way to do this. I got no answers so far. In the first place I had that logic inside the plans  - @DominicOram suggested to bring this into a device, but apparently I am not able to do that right

src/dodal/devices/i20/gas_injector.py
```diff
@@ -0,0 +1,201 @@
+import enum
```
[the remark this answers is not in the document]
DominicOram: See conversation at https://github.com/DiamondLightSource/spectroscopy-bluesky/pull/22#discussion_r2139698113. Happy to discuss further if you want @DiamondJoseph?

DiamondJoseph:
I think you may have force pushed and overwritten some of your own changes, I've unresolved the relevant comments.

DominicOram:
@iain-hall, is this still relevant? Could we either assign it to you or close it?

PR #1297, opened 2025-06-19
fix: Allow automated updates of the copier template and update to latest version

DiamondJoseph:
Renovate is able to update copier templates, but not if the copier template is using the gh: short link, nor if the template makes use of scripts that require `--trust`. The python-copier-template since 4.0.0 no longer requires --trust, so will be able to be automated by Renovate, allowing picking up any changes from upstream. It additionally includes some minor changes to remove workarounds for issues that have been fixed upstream, and deduplicates workflows in a more efficient and cleaner manner

### Checks for reviewer


DiamondJoseph:
Compare the readability of the "checks" tab of this PR to a pre-existing PR:
image
image

The de-duplicated duplicate has always been sorted to the top, now it is simply not created and the run workflows are prioritised.

.github/actions/install_requirements/action.yml
```diff
@@ -15,7 +15,8 @@ runs:
       run: |
         PYTHON_VERSION="${{ inputs.python-version }}"
         if [ $PYTHON_VERSION == "dev" ]; then
-          PYTHON_VERSION=$(sed -n "s/ARG PYTHON_VERSION=//p" Dockerfile)
+          # python version from Dockerfile, removing potential pinned sha
```
stan-dot, line 18: ` removing potential pinned sha` 

probably you mean `pinned sha hash` as 'sha' means just secure hashing algorithm and we are not pinning the algorithm - I'd 
```suggestion
          # python version from Dockerfile, removing potential pinned commit hash
```

stan-dot: apparently I'm wrong and `sha` is used in this context - however [there ](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/about-commits#about-commits) as well as in one other place in the codebase we use uppercase SHA

stan-dot: still 'hash' might be clearer

.github/workflows/ci.yml
```diff
@@ -2,26 +2,24 @@ name: CI
...
       tox: pre-commit,type-checking
 
   test:
-    needs: check
-    if: needs.check.outputs.branch-pr == ''
     strategy:
       matrix:
         runs-on: ["ubuntu-latest"] # can add windows-latest, macos-latest
-        python-version: ["3.10", "3.11", "3.12"]
+        python-version: ["3.11", "3.12"]
```
stan-dot, line 22: are we dropping 3.10 in the same PR?

pyproject.toml
```diff
@@ -39,7 +38,7 @@ dependencies = [
 dynamic = ["version"]
 license.file = "LICENSE"
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
stan-dot, line 41: as above - comment on dropping 3.10

stan-dot (approved):
minor requests for clarification

rtuck99 (approved):
Approved

PR #1301, opened 2025-06-20
Added base table class

### LINKED ISSUE #1299: Create base table for i22
I22 uses a base table for sample alignment etc, it has x, y and pitch

## Acceptance Criteria

RJCD-Diamond:
Fixes #1299

Added base_tables.py and two classes for sample base tables, these are intended for i22 and b21 but should be generic enough to be used on any beamline

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/base_table.py
+++ b/src/dodal/devices/base_table.py
@@ -1,0 +1,35 @@
+from ophyd_async.core import StandardReadable
+from ophyd_async.epics.motor import Motor
+
+
+class BaseTable(StandardReadable):
+    """Base class for a sample table with two motors representing X and Y directions."""
+
+    def __init__(
+        self,
+        prefix: str,
+        x: str = "X",
+        y: str = "Y",
+        name: str = "",
+    ) -> None:
+        with self.add_children_as_readables():
+            self.x = Motor(prefix + x)
+            self.y = Motor(prefix + y)
+        super().__init__(name=name)
+
+
+class PitchedBaseTable(BaseTable):
+    """Base class for a sample table with two motors representing gaps
+    in X and Y directions, with additional motor for pitch."""
+
+    def __init__(
+        self,
+        prefix: str,
+        x: str = "X",
+        y: str = "Y",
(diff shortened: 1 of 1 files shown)


DiamondJoseph:
Just bringing to your attention #1277 that I am happy for someone to take over and address Dom's comments if they have capacity, since I've been flooded for a few days

DiamondJoseph:
> > Just bringing to your attention #1277 that I am happy for someone to take over and address Dom's comments if they have capacity, since I've been flooded for a few days
> 
> So although #1277 covers most examples of the types of Motor group implementation it doesn't cover a this use case on i22 and b21, which means I'd have to create another one. If we wanted to make something super general, we could do something like this, either using a tuple or a dict. [...]

Adding another one is fine, we all still get to benefit from common things. I am really opposed to the `dict` device creation- the Stages I added have docstrings defining the relationship between the motors. If the relationship between the motors isn't important, they should be individual Motor devices- a group of Motors is only useful if it means something for the type of experiments it can run (e.g. X and Y are both mutually perpendicular to the beam) or cannot run (e.g. XFine and XCoarse are parallel to each other) or behaviour (e.g. does setting my YAW axis reverse the relationship between X and Y and the beam?).

DiamondJoseph:
I've added an XYPitchStage to #1277, together with the existing XYStage?

DiamondJoseph:
As I said above and have added to the docs in #1277, 

> I am really opposed to the dict device creation- the Stages I added have docstrings defining the relationship between the motors. If the relationship between the motors isn't important, they should be individual Motor devices- a group of Motors is only useful if it means something for the type of experiments it can run (e.g. X and Y are both mutually perpendicular to the beam) or cannot run (e.g. XFine and XCoarse are parallel to each other) or behaviour (e.g. does setting my YAW axis reverse the relationship between X and Y and the beam?).

RJCD-Diamond:
> @RJCD-Diamond I believe this is no longer required with #1277 ?

Correct

PR #1306, opened 2025-06-20
chore: Remove redundant names from device_factory managed devices

DiamondJoseph:
@device_factory names a device after it has finished connecting the same name as the device factory function that it wraps. This change re-orders or provides defaults where name was a required argument, and removes all names that were explicitly passed into the device factories. This is best practices, prevents duplication and sets an example for how the codebase should be managed.

It also updates some out of date docs

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/beamlines/b21.py
```diff
@@ -116,8 +116,6 @@ def slits_7() -> Slits:
         prefix=f"{PREFIX.beamline_prefix}-AL-SLITS-07:",
         x_gap="X:GAP",
         y_gap="Y:GAP",
-        x_centre="X:CENTRE",
-        y_centre="Y:CENTRE",
```
DiamondJoseph, line 120: TODO: this may have been overzealous

docs/reference/device-standards.rst
```diff
@@ -27,10 +27,32 @@ Device Best Practices
...
+Device should provide the ability to override their name while maintaining a default name of ""-
+this allows the device to be named on connection from the name of its factory function when using
+the device_factory decorator.
+
+When a device is named in this way, all of its child devices are named appropriately.
+
+.. code-block:: python
+    class MyDevice(Device):
+        def __init__(self, prefix: str, name: str = "")
+            x = Motor(prefix + "X")
```
rtuck99, line 47: this example doesn't actually call the superclass init

src/dodal/beamlines/i04.py
```diff
@@ -83,7 +82,6 @@ def gonio_positioner() -> XYZStage:
     """
     return XYZStage(
         f"{PREFIX.beamline_prefix}-MO-GONIO-01:",
-        "lower_gonio_stages",
```
rtuck99, line 86: This doesn't retain the original name?

src/dodal/beamlines/i24.py
```diff
@@ -197,7 +185,6 @@ def focus_mirrors() -> FocusMirrorsMode:
     """Get the i24 focus mirror devise to find the beam size."""
     return FocusMirrorsMode(
         f"{PREFIX.beamline_prefix}-OP-MFM-01:",
-        "focus_mirrors",
     )
```
rtuck99, line 190: ditto

src/dodal/beamlines/i24.py
```diff
@@ -206,7 +193,6 @@ def eiger_beam_center() -> DetectorBeamCenter:
     """A device for setting/reading the beamcenter from the eiger on i24."""
     return DetectorBeamCenter(
         f"{PREFIX.beamline_prefix}-EA-EIGER-01:CAM:",
-        "eiger_bc",
     )
```
rtuck99, line 198: ditto

rtuck99 (commented):
I would prefer not to change the names of devices unless there's a particular need, as the names of the devices may be in various beamline plans and scripts in callback handlers etc

src/dodal/beamlines/i24.py
DiamondJoseph: Think this one was a false positive, function name and device name are both focus_mirrors

src/dodal/beamlines/i24.py
```diff
@@ -206,7 +194,7 @@ def eiger_beam_center() -> DetectorBeamCenter:
     """A device for setting/reading the beamcenter from the eiger on i24."""
     return DetectorBeamCenter(
         f"{PREFIX.beamline_prefix}-EA-EIGER-01:CAM:",
-        "eiger_bc",
+        "pilatus_bc",
```
DiamondJoseph, line 197:
```suggestion
        "eiger_bc",
```

rtuck99 (approved):
Approved

PR #1311, opened 2025-06-25
Uncommitted beamline changes as of 2025-06-25

rtuck99:
Fixes #ISSUE

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,8 @@ dependencies = [
     "redis",
     "deepdiff",
     "scanspec>=0.7.3",
-    "event-model>=1.23",  # Until bluesky pins it https://github.com/DiamondLightSource/dodal/issues/1278
+    "event-model>=1.23",      # Until bluesky pins it https://github.com/DiamondLightSource/dodal/issues/1278
+    "pyzmq==26.3.0",
 ]
 
 dynamic = ["version"]
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -371,6 +371,7 @@ def zocalo() -> ZocaloResults:
     return ZocaloResults(
         name="zocalo",
         prefix=PREFIX.beamline_prefix,
+        use_gpu=True
     )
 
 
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -33,7 +33,8 @@ dependencies = [
     "redis",
     "deepdiff",
     "scanspec>=0.7.3",
-    "event-model>=1.23",  # Until bluesky pins it https://github.com/DiamondLightSource/dodal/issues/1278
+    "event-model>=1.23",      # Until bluesky pins it https://github.com/DiamondLightSource/dodal/issues/1278
+    "pyzmq==26.3.0",
```
DominicOram, line 37: Should: Can we have a comment on why we need to pin this please?

src/dodal/beamlines/i03.py
```diff
@@ -368,10 +368,7 @@ def zocalo() -> ZocaloResults:
     """Get the i03 ZocaloResults device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i03, it will return the existing object.
     """
-    return ZocaloResults(
-        name="zocalo",
-        prefix=PREFIX.beamline_prefix,
-    )
+    return ZocaloResults(name="zocalo", prefix=PREFIX.beamline_prefix, use_gpu=True)
```
DominicOram, line 371: I'm not really sure what's going on here. When do we change this parameter? I don't think we do in `mx_bluesky` at all, even when told not to use the GPU. Maybe this will change with https://github.com/DiamondLightSource/dodal/pull/1249 so not sure if we need to address now but it needs looking at. Could you make a new issue if there isn't one already?

rtuck99:
> As discussed, can you re-instate the deffered move thing? Maybe new issue for the race condition and we just keep an eye on it

When I reinstate the code, the test that I wrote specifically for this breaks. Thinking about it, if the calls to `user_set_point` don't serialize in the correct order, I can't see how the on-the-wire command order is going to be correct. So I would question whether the CombinedMove is actually working or whether we are just doing 6 separate motor moves. I will put the code back as it doesn't work in reality the way it is currently, but I think we really do need to talk with controls to see if the Defer Moves is actually even usable with the current behaviour.

src/dodal/beamlines/i03.py
rtuck99: https://github.com/DiamondLightSource/mx-bluesky/issues/1140

PR #1318, opened 2025-07-02
b01-1: Fix PandA PV prefix

GDYendell:
1. Do `dodal connect b01_1`
2. Confirm it succeeds instead of `panda: NotConnected: pva://BL01C-EA-PANDA-01:PVI`

### Checks for reviewer


olliesilvester (approved):
Since this is a PVA device, `dodal connect b01_1` will fail for the panda unless you're on the beamline network - I will take your word that the prefix is correct though!

PR #1320, opened 2025-07-03
Add ViSR docstrings

VictoriaBeilsten-Edmands:
### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/b01_1.py
+++ b/src/dodal/beamlines/b01_1.py
@@ -42,6 +42,11 @@ https://argocd.diamond.ac.uk/applications?showFavorites=false&proj=&sync=&autoSy
 
 @device_factory()
 def panda() -> HDFPanda:
+    """Provides triggering of the detectors.
+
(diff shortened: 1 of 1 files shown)


src/dodal/beamlines/b01_1.py
```diff
@@ -50,11 +55,25 @@ def panda() -> HDFPanda:
 
 @device_factory()
 def synchrotron() -> Synchrotron:
+    """The synchrotron light source.
+
+    Returns:
+        Synchrotron: The synchrotron light source device.
+    """
```
callumforrester, line 62: Could: Arguably not worth bothering with this one :)

callumforrester (approved):
Looks good to me, thanks!

src/dodal/beamlines/b01_1.py
VictoriaBeilsten-Edmands: Yes, it's there for completeness rather than being in anyway helpful!

Issue #1328, opened 2025-07-07
Create enum for electron analyser psu_mode


oliwenmandiamond:
Currently, for `SpecsRegion`, we have the field `psu_mode`
```Python
class SpecsRegion(AbstractBaseRegion):
    ...
    psu_mode: str = "1.5kV"
```

This should be updated to use enums . Below is an example for B07 but it is beamline dependent.

[xol73553@i05-ws004:~] $ caget -d 31 BL07B-EA-DET-01:CAM:SCAN_RANGE
BL07B-EA-DET-01:CAM:SCAN_RANGE
    Native data type: DBF_ENUM
    Request type:     DBR_CTRL_ENUM
    Element count:    1
    Value:            3.5kV
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            ( 5)
                      [ 0] 3.5kV
                      [ 1] 1.5kV
                      [ 2] 400V
                      [ 3] 100V
                      [ 4] 10V

Therefore, we need to implement a unique enum for each beamline for `psu_mode`.

This also applies to other analysers like VGScienta, except it is called `element_set` instead and is stored on the `VGScientaSequence` rather than the `VGScientaRegion`

PR #1329, opened 2025-07-07
Add PsuMode enum for analysers, update classes and tests to use

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/1328

1. Check enum makes sense and used correctly
2. Check tests pass

### Checks for reviewer


oliwenmandiamond:
This is ready for review again. However, the only issue is that dodal connect will break with this change as we are assuming electron analysers are all using `PSU_MODE` as the PV for the power supply. Specs currently uses `SCAN_RANGE` and VGScienta uses `ELEMENT_SET`. We need to speak to controls about aliasing this PV for the simulators and on the beamline so we have a common PV.

oliwenmandiamond:
We have already created Jiras tracking the issue, with some of it already completed

https://jira.diamond.ac.uk/browse/CGP-422
https://jira.diamond.ac.uk/browse/CGP-418

oliwenmandiamond:
Done

olliesilvester (approved):
Couple more comments but I won't block a merge.

There are some tests relevant to this PR in https://github.com/DiamondLightSource/dodal/pull/1340/files , but I'm reviewing that next anyway

PR #1335, opened 2025-07-09
Fix energy axis calculation for specs driver

Villtord:
Fixes incorrect energy axis size for Specs analysers.
This is a small fix to use total points PV instead of slices for calculating energy axis size of data.

1. Do scan with analyser
2. Confirm the energy axis is correct

### Checks for reviewer


oliwenmandiamond (approved):
Looks good to me, need someone from @DiamondLightSource/bluesky-reviewers  to approve.

PR #1340, opened 2025-07-11
Create pass energy enums for analysers

oliwenmandiamond:
Fixes #ISSUE

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/electron_analyser/specs/driver_io.py
+++ b/src/dodal/devices/electron_analyser/specs/driver_io.py
@@ -19,10 +19,16 @@ from dodal.devices.electron_analyser.abstract.types import TLensMode, TPsuMode
 from dodal.devices.electron_analyser.specs.enums import AcquisitionMode
 from dodal.devices.electron_analyser.specs.region import SpecsRegion
 
+PassEnergy = float
+
 
 class SpecsAnalyserDriverIO(
     AbstractAnalyserDriverIO[
-        SpecsRegion[TLensMode, TPsuMode], AcquisitionMode, TLensMode, TPsuMode
+        SpecsRegion[TLensMode, TPsuMode],
+        AcquisitionMode,
+        TLensMode,
+        TPsuMode,
+        PassEnergy,
     ],
     Generic[TLensMode, TPsuMode],
 ):
@@ -47,7 +53,13 @@ class SpecsAnalyserDriverIO(
             )
 
         super().__init__(
-            prefix, AcquisitionMode, lens_mode_type, psu_mode_type, energy_sources, name
+            prefix=prefix,
+            acquisition_mode_type=AcquisitionMode,
+            lens_mode_type=lens_mode_type,
+            psu_mode_type=psu_mode_type,
+            pass_energy_type=PassEnergy,
+            energy_sources=energy_sources,
+            name=name,
         )
 
     @AsyncStatus.wrap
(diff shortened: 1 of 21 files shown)


src/dodal/devices/electron_analyser/specs/driver_io.py
```diff
@@ -17,12 +17,16 @@
...
 
 
 class SpecsAnalyserDriverIO(
     AbstractAnalyserDriverIO[
-        SpecsRegion[TLensMode, TPsuMode], AcquisitionMode, TLensMode, TPsuMode
+        SpecsRegion[TLensMode, TPsuMode],
+        AcquisitionMode,
+        TLensMode,
+        TPsuMode,
+        PassEnergy,
```
olliesilvester, line 29: nit: this would be less confusing to me if we just used `float`

```suggestion
        float,
```

src/dodal/devices/electron_analyser/specs/region.py
```diff
@@ -9,13 +9,16 @@
 from dodal.devices.electron_analyser.abstract.types import TLensMode, TPsuMode
 from dodal.devices.electron_analyser.specs.enums import AcquisitionMode
 
+PassEnergy = float
```
olliesilvester, line 12: See other comment, I'd prefer we just use `float`

src/dodal/devices/electron_analyser/types.py
```diff
@@ -23,28 +23,35 @@
 AnyAcqMode = StrictEnum
 AnyLensMode = SupersetEnum | StrictEnum
 AnyPsuMode = SupersetEnum | StrictEnum
+AnyPassEnergy = StrictEnum | int
```
olliesilvester, line 26: Should this be(?)
```suggestion
AnyPassEnergy = StrictEnum | float
```

src/dodal/devices/electron_analyser/types.py
```diff
@@ -23,28 +23,35 @@
...
+PassEnergyEnum = StrictEnum
```
olliesilvester, line 27:
```suggestion
AnyPassEnergyEnum = StrictEnum
```

tests/devices/unit_tests/electron_analyser/conftest.py
```diff
@@ -54,10 +56,25 @@ async def sim_detector(
 ) -> ElectronAnalyserDetectorImpl:
     lens_mode_class = get_args(detector_class)[0]
     psu_mode_class = get_args(detector_class)[1]
-    async with init_devices(mock=True, connect=True):
-        sim_detector = detector_class(
-            "TEST:", lens_mode_class, psu_mode_class, energy_sources
-        )
+
+    if get_origin(detector_class) == VGScientaDetector:
```
olliesilvester, line 60: Am I right in thinking this fixture only works if you have a 
```
@pytest.fixture
def detector_class()
```
defined somewhere more locally? If so, it might be better to make this a function rather than a fixture, so then it's clear you actually need to provide the `detector_class` yourself

olliesilvester: Same comment to any other fixtures which do this kind of thing

olliesilvester (changes requested):
Mostly looks good, just a few comments for readability

src/dodal/devices/electron_analyser/types.py
oliwenmandiamond: Yep, thank you

olliesilvester (approved):
Looks good!

PR #1345, opened 2025-07-14
Fix ViSR detector names

douglaswinter:
Renamed detectors to correctly reflect their associated experiments in line with ViSR terminology. Previous names were inadvertently swapped and inconsistent with stakeholder conventions.


callumforrester:
@douglaswinter Tests should work now if you rebase

GDYendell (approved):
Looks good once CI passing

PR #1355, opened 2025-07-17
Add devices to i20_1 for running trajectory scans

iain-hall:
Updated i20_1.py to include turbo slit motor and Panda needed for running trajectory scans.

### DIFF
--- a/src/dodal/beamlines/i20_1.py
+++ b/src/dodal/beamlines/i20_1.py
@@ -40,6 +40,7 @@ and you need to be on the beamline network to access them.
 The simplest way to do this is to `ssh i20-1-ws001` and run dodal connect i20_1 from there.
 """
 
+
 @device_factory()
 def turbo_slit() -> TurboSlit:
     """
@@ -47,6 +48,7 @@ def turbo_slit() -> TurboSlit:
     """
 
     return TurboSlit(f"{PREFIX.beamline_prefix}-OP-PCHRO-01:TS:")
+
 
 @device_factory()
 def turbo_slit_x() -> Motor:
@@ -59,7 +61,8 @@ def turbo_slit_x() -> Motor:
 @device_factory()
 def panda() -> HDFPanda:
     return HDFPanda(
-        f"{PREFIX.beamline_prefix}-EA-PANDA-02:", path_provider=get_path_provider())
+        f"{PREFIX.beamline_prefix}-EA-PANDA-02:", path_provider=get_path_provider()
+    )
 
 
 # Use mock device until motors are reconnected on the beamline


src/dodal/beamlines/i20_1.py
```diff
@@ -12,6 +16,12 @@
 from dodal.log import set_beamline as set_log_beamline
 from dodal.utils import BeamlinePrefix, get_beamline_name
 
+# introduces dependency on 'velocityprofile' package 
+from ophyd_async.epics.pmac import (
+    Pmac,
+    PmacMotor,
```
DominicOram, line 22: Must: These don't exist in the latest `ophyd-async` (hence the test errors). Is there a specific `ophyd-async` branch that has these that you want to use? Or did they used to exist and been renamed?

src/dodal/beamlines/i20_1.py
```diff
@@ -31,29 +41,62 @@
...
     """
-
     return Xspress3(
         f"{PREFIX.beamline_prefix}-EA-DET-03:",
         num_channels=16,
     )
 
 
-@device_factory()
+# Skip device - synchrotron PVs are not available (CA gateway machine has been removed)
```
DominicOram, line 99: Nit: Is there an issue/jira ticket describing this in more detail? E.g. is it permanent? If it is going to be fixed what needs to be done, where is that fix being tracked?

DominicOram (changes requested):
Thanks, mostly the comment re PMAC. I tried to run `dodal connect i20-1` and nothing connected but looking at the synoptic it looks like the whole beamline is disconnected anyway. I assume this is a temporary thing?

src/dodal/beamlines/i20_1.py
iain-hall: I think this ia permanent thing - the gateway machine warranty expired, and IT decided to remove it. This is also why the PVs aren't visible outside of the beamline network and dodal connect fails unless run from i20-1 network. I think there are plans to replace CA gateway machines with some kubernetes service, but not any time soon. So maybe best to just remove Synchrotron completely.

src/dodal/beamlines/i20_1.py
coretl: Yes, there are a bunch of changes in the parent https://github.com/bluesky/ophyd-async/issues/427 that we'll be making shortly that will feed into thos

src/dodal/beamlines/i20_1.py
```diff
@@ -31,29 +35,52 @@
...
     """
 
     return TurboSlit(f"{PREFIX.beamline_prefix}-OP-PCHRO-01:TS:")
 
+@device_factory()
+def turbo_slit_x() -> Motor:
+    """
+    turbo slit x motor
+    """
+    return Motor(f"{PREFIX.beamline_prefix}-OP-PCHRO-01:TS:XFINE")
```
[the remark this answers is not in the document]
iain-hall: https://github.com/DiamondLightSource/dodal/issues/1417

DominicOram (approved):
Thanks

PR #1363, opened 2025-07-21
use adcore acquire_time in analyser driver

### LINKED ISSUE #1368: Use parent ADBaseIO field acquire_time in analyser driver class
Currently analyser driver has a step_time parameter wired to the base AD camera PV: AcquireTime and populated each time from the region json file. Yet there is already existing acquire_time field in parent ADBaseIO class.

## Acceptance Criteria
* All tests pass

Villtord:
Fixes #1368 

1. Run tests
2. Confirm all tests pass

### Checks for reviewer


src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -106,6 +109,7 @@ def __init__(
 
         with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
             # Read once per scan after data acquired
+            self.acquire_time = self.acquire_time
```
DominicOram, line 112: Should: What is this line doing? It seems like a noop?

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -75,6 +75,9 @@ def __init__(
         self.psu_mode_type = psu_mode_type
         self.pass_energy_type = pass_energy_type
 
+        # must call first to initiate parent pydantic fields
```
DominicOram, line 78: Should: I'm a bit confused by this comment, it doesn't look like this class inherits from a pydantic model?

src/dodal/devices/electron_analyser/specs/region.py
```diff
@@ -18,10 +18,18 @@ class SpecsRegion(
...
+    )
+    low_energy: float = Field(
+        default=500, validation_alias=AliasChoices("start_energy", "low_energy")
+    )
+    high_energy: float = Field(
+        default=510, validation_alias=AliasChoices("end_energy", "high_energy")
+    )
+    energy_step: float = Field(
+        default=0.15, validation_alias=AliasChoices("step_energy", "energy_step")
+    )
```
DominicOram, line 32: Could: Why were these changes made? It would be good to have a unit test that covers the need for them

DominicOram (dismissed):
Thanks, a few minor comments

src/dodal/devices/electron_analyser/specs/region.py
Villtord: correct, there are 2 types of region files - old and new - and I was going to make a separate PR adding tests for a new type as there is currenlty only a test for an old region file. I assumed initially that  `low_energy: float = Field(default=500, alias="start_energy")` will also work parsing  "low_energy" from "new" region file but it doesn't in fact - it works fine for alias field "start_energy" which is in old type of region file.

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
Villtord: Still if I don't call `super.__init__()` first I will get an error later in tests  AttributeError: 'VGScientaAnalyserDriverIO' object has no attribute 'acquire_time'

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
DominicOram: Hmm... I'm 50/50 whether this would actually work or not and I don't think you have a test that checks what the config signals of the device are. Could you confirm it with a test? If it does work that's fine but I think it needs a comment next to it probably as at first glance it looks very strange

Villtord: yes, will do that later, I noticed as well there is no test but I can confirm it running against simulated device

src/dodal/devices/electron_analyser/specs/region.py
Villtord: I will make a separate PR on that.

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
DominicOram: Ok, I see. I would put the comment above the section and be more explicit:
```python
# The following config signals are used for setting up region data acquisition.
...
# These config signals are read once per scan after data acquired
...
```
But not that big a deal

src/dodal/devices/b07/enums.py
```diff
@@ -32,6 +32,7 @@ class PsuMode(SupersetEnum):
     V1500 = "1.5kV"
     V400 = "400V"
     V100 = "100V"
+    V40 = "40V"
```
oliwenmandiamond, line 35: I don't think this should be committed, be a separate issue with the simulators. Other then this, change looks good!

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
Villtord: so I tested this and indeed that line didn't add acquire_time as readable config, I had to add explicit line adding that in configs `self.add_readables()`

Villtord: did that

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -75,6 +75,9 @@ def __init__(
         self.psu_mode_type = psu_mode_type
         self.pass_energy_type = pass_energy_type
 
+        # must call first to initiate parent variables
+        super().__init__(prefix=prefix, name=name)
```
DiamondJoseph, line 79: I believe making this call here means all of your signals defined afterwards won't be named (or possibly even connected?)

Villtord: Right, this is not an issue anymore as Olli mentioned.

Villtord:
thanks, I just need to fix tests now after merge with main

PR #1364, opened 2025-07-21
Modify devices for i03's default UDC state

DominicOram:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/396

1. Confirm tests pass

### Checks for reviewer


src/dodal/devices/fluorescence_detector_motion.py
```diff
@@ -1,16 +1,14 @@
...
     IN = "In"
 
 
 class FluorescenceDetector(StandardReadable):
     def __init__(self, prefix: str, name: str = ""):
         with self.add_children_as_readables():
-            self.pos = epics_signal_r(
-                FluorescenceDetectorControlState, prefix + "-EA-FLU-01:CTRL"
-            )
+            self.pos = epics_signal_rw(InOut, prefix + "-EA-FLU-01:CTRL")
```
DiamondJoseph, line 13: I thought we had in our docs that the prefix should be split to allow some degree of disparate PV prefixes, but I can't find it right now.

Any objection while you're changing this to make the prefix that is passed in `f"{PREFIX.beamline_prefix}-EA-FLU-01:"` for i03 and the signal `prefix + "CTRL"`?

src/dodal/beamlines/i03.py
```diff
@@ -457,3 +458,31 @@ def baton() -> Baton:
...
+        get_beamline_parameters(),
+    )
+
+
+@device_factory()
+def collimation_table() -> CollimationTable:
+    """Get the i03 device for moving the collimation table, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i03, it will return the existing object.
+    """
+    return CollimationTable(PREFIX.beamline_prefix)
```
DiamondJoseph, line 488:
```suggestion
    return CollimationTable(motion_prefix=f"{PREFIX.beamline_prefix}-MO-TABLE-01", power_supply_prefix=f"{PREFIX.beamline_prefix}-MO-PMAC-02:CRATE2_HEALTHY")
```

Now that we don't have to support device_instantiation and its assumptions of a single prefix and name, these devices could be adjusted to be more reuseable

Other than the PMAC power supply

src/dodal/devices/scintillator.py
```diff
@@ -1,10 +1,81 @@
...
+        prefix: str,
+        aperture_scatterguard: Reference[ApertureScatterguard],
+        beamline_parameters: GDABeamlineParameters,
+        name: str = "",
+    ):
         with self.add_children_as_readables():
-            self.y = Motor(prefix + "-MO-SCIN-01:Y")
-            self.z = Motor(prefix + "-MO-SCIN-01:Z")
+            self.y_mm = Motor(prefix + "-MO-SCIN-01:Y")
+            self.z_mm = Motor(prefix + "-MO-SCIN-01:Z")
```
DiamondJoseph, line 36: As above, move these onto the prefix of the device instance rather than on the class

DiamondJoseph (commented):
I'm trying to find the docs where this is written down

DominicOram:
> I'm trying to find the docs where this is written down

This was the speediest review I've ever had, thank you! You're right, I was being lazy, fixed now.

tests/devices/unit_tests/test_aperture_scatterguard.py
```diff
@@ -521,6 +587,68 @@ async def test_given_out_and_aperture_selected_when_move_in_then_correct_y_selec
...
+        ApertureValue.LARGE,
+        ApertureValue.OUT_OF_BEAM,
+    ],
+)
+async def test_given_parked_and_aperture_selected_when_move_in_then_z_moved_out_first(
+    selected_aperture: ApertureValue,
+    ap_sg: ApertureScatterguard,
+    aperture_positions: dict[ApertureValue, AperturePosition],
+):
+    # Ophyd-async should make this parent out the box?
```
olliesilvester, line 604: Is this worth an issue?

src/dodal/devices/cryostream.py
```diff
@@ -8,6 +8,9 @@ class InOut(StrictEnum):
 
 
 class CryoStream(StandardReadable):
+    MAX_TEMP_K = 110
```
olliesilvester, line 11: nit: If possible, it would be good to add a link to where these numbers were from

src/dodal/devices/scintillator.py
```diff
@@ -1,10 +1,81 @@
...
+
 
 class Scintillator(StandardReadable):
-    def __init__(self, prefix: str, name: str = ""):
+    """Moves a scintillator into and out of the beam.
+
+    The scintillator will light up when hit with xrays, this allows scientists to use it
+    in conjunction with the optical OAV camera to commission the beamline.
+
+    When moved out of the beam it is parked under the table. This parking has a potential
```
olliesilvester, line 23: Is there anything in the IOC level to protect against this as well? I always feel like collision protection should be at the control layer

src/dodal/devices/mx_phase1/beamstop.py
```diff
@@ -87,10 +84,11 @@ def _get_selected_position(self, x: float, y: float, z: float) -> BeamstopPositi
...
     async def _set_selected_position(self, position: BeamstopPositions) -> None:
         if position == BeamstopPositions.DATA_COLLECTION:
+            # Move z first as it could be under the table
+            await self.z_mm.set(self._in_beam_xyz_mm[2])
             await asyncio.gather(
                 self.x_mm.set(self._in_beam_xyz_mm[0]),
                 self.y_mm.set(self._in_beam_xyz_mm[1]),
-                self.z_mm.set(self._in_beam_xyz_mm[2]),
             )
         elif position == BeamstopPositions.UNKNOWN:
```
olliesilvester, line 93: Should: Better to use else, as it's possible for someone to ignore typing and do `_set_selected_position(self, "anything")`

src/dodal/devices/scintillator.py
```diff
@@ -1,10 +1,81 @@
...
+            if (
+                self._aperture_scatterguard().selected_aperture.get_value()
+                != ApertureValue.PARKED
+            ):
+                raise ValueError(
+                    "Cannot move scintillator out if aperture/scatterguard is not parked"
+                )
+            await self.y_mm.set(self._scintillator_out_yz_mm[0])
+            await self.z_mm.set(self._scintillator_out_yz_mm[1])
+        elif position == InOut.UNKNOWN:
```
olliesilvester, line 80: See other comment about using else

tests/devices/unit_tests/test_scintillator.py
```diff
@@ -0,0 +1,99 @@
...
+        scintillator = Scintillator(
+            prefix="",
+            name="test_scin",
+            aperture_scatterguard=mock_ap_sg,
+            beamline_parameters=mock_beamline_parameters,
+        )
+    with ExitStack() as motor_patch_stack:
+        for motor in [scintillator.y_mm, scintillator.z_mm]:
+            motor_patch_stack.enter_context(patch_motor(motor))
+        yield scintillator, mock_ap_sg
```
olliesilvester, line 45: I've not seen this before - is there something specific which doesn't cleanup properly here without it?

tests/devices/unit_tests/test_scintillator.py
```diff
@@ -0,0 +1,99 @@
...
+
+async def test_when_set_to_unknown_position_then_error_raised(
+    scintillator_and_ap_sg: tuple[Scintillator, ApertureScatterguard],
+):
+    scintillator, _ = scintillator_and_ap_sg
+    with pytest.raises(ValueError):
+        await scintillator.selected_pos.set(InOut.UNKNOWN)
+
+
+async def test_given_aperture_scatterguard_parked_when_set_to_out_position_then_(
```
olliesilvester, line 78:
```suggestion
async def test_given_aperture_scatterguard_parked_when_set_to_out_position_then_returns_expected(
```

olliesilvester (approved):
Looks good! Left a few comments, but mostly just suggestions, so I will approve and let you decide

src/dodal/devices/cryostream.py
DominicOram: I think the answer is basically that they come out of Neil's head

src/dodal/devices/aperturescatterguard.py
```diff
@@ -202,21 +202,20 @@ def __init__(
...
-            return True
-        return False
+    async def _unpark(self, position_to_move_to: ApertureValue):
+        """When the aperture is parked it is under the collimation table. It needs to be
+        moved out from under the table before it is moved up to beam height.
+        """
+        position = self._loaded_positions[position_to_move_to]
+        await self.aperture.z.set(position.aperture_z)
 
     async def _set_current_aperture_position(self, value: ApertureValue) -> None:
```
olliesilvester, line 212: Should: Don't unpark if value == ApertureValue.PARKED

Also, could add a check at the start of this function to "do nothing" if already at requested position - unless this logic is already somewhere in `derived_signal_rw`?

olliesilvester (approved):
Looks good, just one comment

PR #1367, opened 2025-07-21
Only move the DCM perp if it's outside tolerance

### LINKED ISSUE #1316: Energy change plan shouldn't move the mirror + dcm motors unless necessary
Since the optics motors operate in a vacuum even small changes can result in excess heat generation.
It has been found that even though the motors are set to essentially the same value this results in a rise in motor activity and temperatures
https://diamondlightsource.slack.com/archives/C048RD0KFFF/p1751299650243839

We should really only be running the change energy plan if there's a change in energy or it's the first run, alternatively this could be a controls fix.

## Acceptance Criteria
* DCM, mirror, energy changes are minimised.

DominicOram:
Fixes #1316

See also https://github.com/DiamondLightSource/mx-bluesky/pull/1166

1. Confirm new test passes

### Checks for reviewer


rtuck99 (approved):
Approved

PR #1370, opened 2025-07-22
Remove references to 3.10 and to asyncio.TimeoutError

### LINKED ISSUE #1237: Drop support for python 3.10
As per https://github.com/DiamondLightSource/dodal/blob/main/docs/reference/standards.rst#supported-python-versions we are following https://numpy.org/neps/nep-0029-deprecation_policy.html, which dropped python 3.10 support in April. We should do the same

## Acceptance Criteria
* Python 3.10 support dropped

DominicOram:
Fixes #1237

1. Search 3.10 and confirm there is no reference to it
2. Confirm there are no more references to `asyncio.TimeoutError`

### Checks for reviewer

### DIFF
--- a/docs/how-to/dev-install.md
+++ b/docs/how-to/dev-install.md
@@ -19,7 +19,7 @@ in a container under [VSCode](https://code.visualstudio.com/)
 If on a DLS machine make sure you have python >3.11 running by doing `module load python/3.11`
 
 ```
-python -m venv venv
+python3 -m venv venv
 source venv/bin/activate
 pip install --upgrade pip
 pip install -e '.[dev]'
--- a/docs/how-to/dev-install.md
+++ b/docs/how-to/dev-install.md
@@ -8,16 +8,20 @@ First clone the repository locally using [Git](https://git-scm.com/downloads). T
 
 ## Install dependencies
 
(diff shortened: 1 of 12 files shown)


docs/how-to/dev-install.md
```diff
@@ -8,16 +8,20 @@ First clone the repository locally using [Git](https://git-scm.com/downloads). T
...
 <!-- https://sphinx-design.readthedocs.io/en/latest/tabs.html# -->
 ::::{tab-set}
 
 :::{tab-item} Local virtualenv
+
+If on a DLS machine make sure you have python >3.11 running by doing `module load python/3.11`
+
 ```
-python3 -m venv venv
+python -m venv venv
```
DiamondJoseph, line 22:
```suggestion
python3 -m venv venv
```

What's the behaviour on a Diamond workstation when doing `module load python/3.11; which python3`? Having just had to reinstall my laptop, `python3` is the more correct  executable to use outside of Diamond's systems: `python` doesn't exist and all programs are meant to explicitly ask for `python2 | python3`. The user is allowed to e.g. symlink `python -> python3` but there's no guarantee.

DiamondJoseph (approved):
Just one rfc, and I've merged main into the change to pick up pinning ophyd-async which was causing the only test failures.

PR #1373, opened 2025-07-22
chore(deps): Update ophyd-async to 0.12.3

ZohebShaikh:
Fixes [dodal main pytest issue](https://github.com/DiamondLightSource/dodal/actions/runs/16417747325/job/46475552855#step:7:1676)

The ophyd-async issue related to this is https://github.com/bluesky/ophyd-async/issues/976

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,8 +14,7 @@ description = "Ophyd devices and other utils that could be used across DLS beaml
 dependencies = [
     "click",
     "ophyd",
-    "ophyd-async>=0.12.3",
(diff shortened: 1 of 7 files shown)


DominicOram:
Looks like you're still using `aioca==1.8.1`. I suspect the issue is that we're using the `aioca` as specified in the `pyproject.toml` for `dodal`, not the one specified in `ophyd-async`. I think if you use `ophyd-async[ca]>=0.12.3` in the `pyproject.toml` this will fix it. I think that might also mean we can drop the explicit dependency on `aioca` in `dodal`.

Issue #1380, opened 2025-07-24
Change AbstractAnalyserDriverIO set method to be abstract


oliwenmandiamond:
To help simplify some of the electron analyser design, we should make the `set` method on `AbstractAnalyserDriverIO`   abstract and in turn make each electron analyser implementation be in charge of moving the signals from region data with the `set` method. 

The current design is that the abstract class moves some core signals which they have in common with  `AbstractBaseRegion`, however this is not as simple because some models such as `Specs` only move some of these core signals when certain conditions are met. This has led to moving some "core" signals that they have in common with the `AbstractBaseRegion` class but not others, which makes the design confusing as we then have to the other "core" signals on the implementation layer.

This will simplify and clarify the design of the abstract driver class better. The `AbstractAnalyserDriverIO` will define the core signals and readables which electron analysers should inherit from to reduce duplicate code, but it's up to each implementation on how to move the signals when given region data.

## Acceptance Criteria
* Specific criteria that will be used to judge if the issue is fixed

PR #1385, opened 2025-07-25
Update AbstractAnalyserDriverIO set method to be abstract

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/1380

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -127,39 +127,17 @@ def __init__(
 
         super().__init__(prefix=prefix, name=name)
 
+    @abstractmethod
     @AsyncStatus.wrap
     async def set(self, region: TAbstractBaseRegion):
```
DiamondJoseph, line 132:
```suggestion
    @abstractmethod
    async def set(self, region: TAbstractBaseRegion) -> AsyncStatus:
```

I think right now you're wrapping the method that returns an AsyncStatus in an AsyncStatus and the move would appear to be done immediately

DiamondJoseph (changes requested):
Just one problem but otherwise I like this change, much clearer what the intent and behaviour is

PR #1388, opened 2025-07-25
chore: Ensure devices match expectations of documentation

DiamondJoseph:
Updates some older devices that e.g. assumed the behaviour of device_instantiation with a single required prefix argument. Generally makes devices pass in more of a prefix (making them more generically usable) or pass in multiple prefixes, rather than requiring that infixes match a particular pattern. Removes prefix where it is a legacy argument, and makes prefix mandatory which serves as a hint as to how a device class should be instantiated.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/i24/dcm.py
```diff
@@ -12,22 +12,22 @@ class DCM(BaseDCM[RollCrystal, PitchAndRollCrystal]):
     A double crystal monocromator device, used to select the beam energy.
     """
 
-    def __init__(self, prefix: str, name: str = "") -> None:
+    def __init__(self, prefix: str, base_prefix: str, name: str = "") -> None:
```
DiamondJoseph, line 15: Or perhaps motion_prefix?

src/dodal/devices/xbpm_feedback.py
```diff
@@ -14,7 +14,7 @@ class XBPMFeedback(Device, Triggerable):
     """The XBPM feedback device is an IOC that moves the DCM, HFM and VFM to automatically
     hold the beam into place, as measured by the XBPM sensor."""
 
-    def __init__(self, prefix: str = "", name: str = "xbpm_feedback") -> None:
+    def __init__(self, prefix: str, name: str = "xbpm_feedback") -> None:
         self.pos_ok = epics_signal_r(float, prefix + "-EA-FDBK-01:XBPM2POSITION_OK")
```
DiamondJoseph, line 18: #1391 for the XBPM signal refactor, as it also changes from using a single signal to a Motor record (which is making a presumption about how the device works on my part)

src/dodal/devices/i24/dcm.py
```diff
@@ -12,22 +12,22 @@ class DCM(BaseDCM[RollCrystal, PitchAndRollCrystal]):
...
         with self.add_children_as_readables():
             # Temperatures
-            self.xtal1_temp = epics_signal_r(float, prefix + "-DI-DCM-01:PT100-1")
+            self.xtal1_temp = epics_signal_r(float, prefix + "PT100-1")
             self.xtal1_heater_temp = epics_signal_r(
                 float, prefix + "-DI-DCM-01:PT100-2"
             )
```
DiamondJoseph, line 21:
```suggestion
            self.xtal1_temp = epics_signal_r(float, prefix + "PT100-1")
            self.xtal1_heater_temp = epics_signal_r(float, prefix + "PT100-2")
```

src/dodal/devices/i24/dcm.py
```diff
@@ -12,22 +12,22 @@ class DCM(BaseDCM[RollCrystal, PitchAndRollCrystal]):
...
-            self.xtal2_temp = epics_signal_r(float, prefix + "-DI-DCM-01:PT100-4")
+            self.xtal2_temp = epics_signal_r(float, prefix + "PT100-4")
             self.xtal2_heater_temp = epics_signal_r(
                 float, prefix + "-DI-DCM-01:PT100-5"
             )
```
DiamondJoseph, line 25:
```suggestion
            self.xtal2_temp = epics_signal_r(float, prefix + "PT100-4")
            self.xtal2_heater_temp = epics_signal_r(float, prefix + "PT100-5")
```

src/dodal/devices/aperturescatterguard.py
```diff
@@ -158,13 +158,14 @@ class ApertureScatterguard(StandardReadable, Preparable):
 
     def __init__(
         self,
+        aperture_prefix: str,
+        scatterguard_prefix: str,
```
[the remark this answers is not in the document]
DominicOram: I don't want to instantiate them in the beamline file as they should always be used in conjunction due to the collision risk. In fact they should probably be private in here too

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -126,19 +126,17 @@ class ZocaloResults(StandardReadable, Triggerable):
 
     def __init__(
         self,
-        name: str = "zocalo",
+        name: str = "",
         zocalo_environment: str = ZOCALO_ENV,
         channel: str = "xrc.i03",
         sort_key: str = DEFAULT_SORT_KEY.value,
         timeout_s: float = DEFAULT_TIMEOUT,
-        prefix: str = "",
```
DominicOram, line 134: Should: I agree with removing this but we should also remove from the docstring

src/dodal/devices/i24/dcm.py
DominicOram: Yh, I think motion is better

src/dodal/devices/aperturescatterguard.py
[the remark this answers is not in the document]
DiamondJoseph: I'll leave that change for another PR as it's not an expectation of the docs.

src/dodal/devices/zocalo/zocalo_results.py
```diff
@@ -100,25 +100,31 @@ def source_from_results(results):
...
-        use_gpu (bool): When True, ZocaloResults will take the first set of
-        results that it receives (which are likely the GPU results)
+    <<<<<<< HEAD
+            use_gpu (bool): When True, ZocaloResults will take the first set of
+            results that it receives (which are likely the GPU results)
+    =======
+            prefix (str): EPICS PV prefix for the device
+
+            results_source (ZocaloSource): Where to get results from, GPU or CPU analysis
+    >>>>>>> main
```
DominicOram, line 127: Oops

PR #1389, opened 2025-07-25
Add a starting diffractometer device for i19-2

### LINKED ISSUE #1073: Add Diffractometer device of i19-2
Write a device for the 4 circle diffractometer on i19-2, which includes both the goniometer and the detector axes.

Note. May need to keep in mind the collision model (see IS4CCD)

noemifrisina:
Fixes #1073 

In the first instance, just add a diffractometer device for I19-2 which has all 6 gonio axes plus the detector stage. For now, this does not include the collision model, so should only be used for reading the axes positions.

1. Run dodal connect on i19-2

### Checks for reviewer

### DIFF
--- a/src/dodal/cli.py
+++ b/src/dodal/cli.py
@@ -1,12 +1,14 @@
 import os
 from collections.abc import Mapping
+from pathlib import Path
 
 import click
 from bluesky.run_engine import RunEngine
-from ophyd_async.core import NotConnected
+from ophyd_async.core import NotConnected, StaticPathProvider, UUIDFilenameProvider
 from ophyd_async.plan_stubs import ensure_connected
 
 from dodal.beamlines import all_beamline_names, module_name_for_beamline
+from dodal.common.beamlines.beamline_utils import set_path_provider
 from dodal.utils import AnyDevice, filter_ophyd_devices, make_all_devices
 
 from . import __version__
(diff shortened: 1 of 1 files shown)


tests/devices/i19/test_diffractometer.py
```diff
@@ -0,0 +1,7 @@
+from dodal.devices.i19.diffractometer import DetectorMotion, FourCircleDiffractometer
+
+
+def test_diffractometer_created_without_errors():
```
DominicOram, line 4: Could: Might be good to test that the readables are as expected. You can do this with `assert_reading`.

src/dodal/devices/i19/diffractometer.py
```diff
@@ -0,0 +1,28 @@
+from ophyd_async.epics.motor import Motor
+
+from dodal.devices.motors import Stage, XYZStage
+
+CIRC = "-MO-CIRC-02:SAM:"  # for phi, kappa, omega, 2theta and det_z
+SAMP = "-MO-SAMP-02:SAM:"  # for x,y,z
```
DominicOram, line 6: Should: @DiamondJoseph has gone through and pulled these out into parameters to pass to the device (https://github.com/DiamondLightSource/dodal/pull/1388/files). I think that is a good standard and we should do the same here

DominicOram (changes requested):
Great, thanks. I confirmed `dodal connect i19-2` works. Couple of minor comments in code, otherwise good.

PR #1392, opened 2025-07-28
Remove sleeps from bart robot tests

### LINKED ISSUE #943: Fix flakey bart robot unit test

olliesilvester:
Fixes #943 (hopefully)

These tests relied on sleeps to order when to set signals. I've changed it so that the signals are set based on specific logging messages. Not sure if it's the _best_ fix but it should work for now

1. Confirm tests are still sensible
2. Confirm flakiness is resolved

### Checks for reviewer

### DIFF
--- a/tests/devices/unit_tests/test_bart_robot.py
+++ b/tests/devices/unit_tests/test_bart_robot.py
@@ -1,5 +1,6 @@
 import traceback
-from asyncio import create_task, sleep
+from asyncio import create_task
+from functools import partial
 from unittest.mock import ANY, AsyncMock, MagicMock, call, patch
 
 import pytest
@@ -16,8 +17,8 @@ from dodal.devices.robot import BartRobot, PinMounted, RobotLoadFailed, SampleLo
 
 async def _get_bart_robot() -> BartRobot:
     device = BartRobot("robot", "-MO-ROBOT-01:")
-    device.LOAD_TIMEOUT = 0.05  # type: ignore
-    device.NOT_BUSY_TIMEOUT = 0.05  # type: ignore
+    device.LOAD_TIMEOUT = 1  # type: ignore
+    device.NOT_BUSY_TIMEOUT = 1  # type: ignore
     await device.connect(mock=True)
     return device
 
@@ -29,11 +30,10 @@ async def test_bart_robot_can_be_connected_in_sim_mode():
 
 async def test_given_robot_load_times_out_when_load_called_then_exception_contains_error_info():
     device = await _get_bart_robot()
+    device.LOAD_TIMEOUT = 0.01  # type: ignore
+    device.NOT_BUSY_TIMEOUT = 0.01  # type: ignore
 
-    async def _sleep(*args):
-        await sleep(1)
-
-    device._load_pin_and_puck = AsyncMock(side_effect=_sleep)
+    device._load_pin_and_puck = AsyncMock(side_effect=TimeoutError)
 
     set_mock_value(device.prog_error.code, (expected_error_code := 10))
     set_mock_value(device.prog_error.str, (expected_error_string := "BAD"))
@@ -50,6 +50,8 @@ async def test_given_program_running_when_load_pin_then_logs_the_program_name_an
     patch_logger: MagicMock,
 ):
     device = await _get_bart_robot()
+    device.LOAD_TIMEOUT = 0.01  # type: ignore
+    device.NOT_BUSY_TIMEOUT = 0.01  # type: ignore
     program_name = "BAD_PROGRAM"
     set_mock_value(device.program_running, True)
     set_mock_value(device.program_name, program_name)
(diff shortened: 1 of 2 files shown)


tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -94,24 +100,34 @@ async def test_given_program_not_running_and_pin_unmounting_but_new_pin_not_moun
...
+# Use log info messages to determine when to set the gonio_pin_sensor, so we don't have to use any sleeps during testing
+async def set_with_happy_path(
+    device: BartRobot, mock_log_info: MagicMock
+) -> AsyncStatus:
     """Mocks the logic that the robot would do on a successful load"""
-    device.LOAD_TIMEOUT = 0.05  # type: ignore
+
+    mock_log_info.side_effect = partial(_set_pin_sensor_on_log_messages, device)
+    device.LOAD_TIMEOUT = 1  # type: ignore
+    device.NOT_BUSY_TIMEOUT = 1  # type: ignore
```
DominicOram, line 118: Could: The device is already set to this timeout, we can remove here?

tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -121,33 +137,33 @@ async def test_given_program_not_running_and_pin_unmounts_then_mounts_when_load_
 
 async def test_given_waiting_for_pin_to_mount_when_no_pin_mounted_then_error_raised():
     device = await _get_bart_robot()
-    status = create_task(device.pin_mounted_or_no_pin_found())
-    await sleep(0.2)
     set_mock_value(device.prog_error.code, 25)
-    await sleep(0.01)
+    status = create_task(device.pin_mounted_or_no_pin_found())
```
DominicOram, line 141: Could: If we're just relying on the `prog_error` already being set then we needn't spawn this in a new task.

tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -121,33 +137,33 @@ async def test_given_program_not_running_and_pin_unmounts_then_mounts_when_load_
...
+    device._load_pin_and_puck = MagicMock()  # type: ignore
     await device.set(SampleLocation(1, 2))
-    mock_wait_for.assert_awaited_once_with(ANY, timeout=0.1)
+    mock_wait_for.assert_awaited_once_with(ANY, timeout=0.02)
 
 
 async def test_moving_the_robot_will_reset_error_if_light_curtain_is_tripped_and_still_throw_if_error_not_cleared():
     device = await _get_bart_robot()
+    device.LOAD_TIMEOUT = 0.01  # type: ignore
+    device.NOT_BUSY_TIMEOUT = 0.01  # type: ignore
```
DominicOram, line 166: Should: This test doesn't care about timeouts, we need not do this

tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -94,24 +100,34 @@ async def test_given_program_not_running_and_pin_unmounting_but_new_pin_not_moun
         raise
 
 
-async def set_with_happy_path(device: BartRobot) -> AsyncStatus:
+def _set_pin_sensor_on_log_messages(device: BartRobot, msg: str):
```
DominicOram, line 103: Should: If we have to do this then lets at least pull them into constants so that the relationship is more obvious

tests/devices/unit_tests/test_bart_robot.py
```diff
@@ -16,8 +17,8 @@
 
 async def _get_bart_robot() -> BartRobot:
     device = BartRobot("robot", "-MO-ROBOT-01:")
-    device.LOAD_TIMEOUT = 0.05  # type: ignore
-    device.NOT_BUSY_TIMEOUT = 0.05  # type: ignore
+    device.LOAD_TIMEOUT = 1  # type: ignore
+    device.NOT_BUSY_TIMEOUT = 1  # type: ignore
```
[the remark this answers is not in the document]
DominicOram: Could: We could make a `_get_fats_timeout_bart_robot` and use that in those tests?

DominicOram (changes requested):
Ok, I think we have to hook in via the logger then :( Few minor comments in code

tests/devices/unit_tests/test_bart_robot.py
olliesilvester: It seems like the test still waits for the timeout time, even though it raises a `RobotLoadFailed` before it times out. I didn't update the timeout on a previous commit and ci [here](https://github.com/DiamondLightSource/dodal/actions/runs/16568993795/job/46855975054) failed as the test was taking a long time

DominicOram (approved):
Looks good, lets see if it's much more stable

Issue #1403, opened 2025-07-30
Create the Lakeshore devices.


Relm-Arrowny:
Part of #1339 

This create the lakeshore device.

PR #1408, opened 2025-07-30
1404 add base lakeshore temperature controllers

### LINKED ISSUE #1411: Add lakeshore base device
Add lakeshore base device

## Acceptance Criteria

Relm-Arrowny:
Fixes #1411 
Part 2 of #1403 

Added Lakeshore controller to set temperature.
Added two derived signal to keep track of control channel and hint channel.
Corrected heater_output pv. 

1.run test,
2. Confirm it passes

 ### Checks for reviewer


src/dodal/devices/temperture_controller/lakeshore/lakeshore_io.py
```diff
@@ -0,0 +1,86 @@
...
+        else:
+            self.control_channels = DeviceVector(
+                {
+                    i: LakeshoreControlChannel(
+                        prefix=prefix, suffix=str(i), heater_type=heater_setting
+                    )
+                    for i in range(1, num_readback_channel + 1)
+                }
+            )
+
```
oliwenmandiamond, line 74: The if else isn't needed for the DeviceVector, only for the parameters to give it, Maybe something like this?

```Python
suffixes = [""] if single_control_channel else map(str, range(1, num_readback_channel + 1))
self.control_channels = DeviceVector(
    {
        i: LakeshoreControlChannel(prefix=prefix, suffix=suffix, heater_type=heater_setting)
        for i, suffix in enumerate(suffixes, start=1)
    }
)
```
However, I still think `single_control_channel` isn't needed as something you pass in. Is there ever a time where `num_readback_channel == 1 and single_control_channel == False`? If not, then if passing `num_readback_channel = 1` should make `single_control_channel = True`.

src/dodal/devices/temperture_controller/lakeshore/lakeshore.py
```diff
@@ -0,0 +1,157 @@
...
+from bluesky.protocols import Movable
+from ophyd_async.core import (
+    AsyncStatus,
+    SignalDatatypeT,
+    StandardReadable,
+    StandardReadableFormat,
+    derived_signal_rw,
+    soft_signal_rw,
+)
+from ophyd_async.core._readable import _HintsFromName
```
oliwenmandiamond, line 12: I am not sure about this as you are importing a private class from a private module from ophyd-async. I don't think this should be allowed and just because you can do it doesn't mean we should. If you must use this, I would ask ophyd-async to be make this public so you can do

`from ophyd_async.core import HintsFromName`

Issue #1402, opened 2025-07-29
Unify ON/OFF ENABLED/DISABLE Enums


RJCD-Diamond:
There are alot of device that use the same nomenclature for enabling and disabling, such as:

class DeviceEnable(StrictEnum):
    ON = "Enabled"
    OFF = "Disabled"

class DeviceOnOff(StrictEnum):
    ON = "ON"
    OFF = "OFF"

class DeviceONOFF(StrictEnum):
    ON = "On"
    OFF = "Off"

etc..

Some examples include:

BimorphMirrorOnOff
BacklightPower
CryoStreamEnable
TurboEnum
AutotuneControl
ThawerStates
ForceBit
LidHeatEnable
LEDStatus
UndulatorGapAccess
WatsonMarlow323PumpEnable
UpdateRBV

I could probably go on....

We should put a class in dodal.common that can be used for all devices that have similar enums instead of re-writing one for each device

DominicOram:
There's a bunch of these in MX code too, we might as well do them all at the same time.

PR #1415, opened 2025-07-30
Created common enums and updated devices to use

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/1402

1. Check enum names are okay
2. Check enums have been updated correctly for classes
3. Tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i23.py
+++ b/src/dodal/beamlines/i23.py
@@ -45,8 +45,8 @@ I23_ZEBRA_MAPPING = ZebraMapping(
 
 
 class I23DetectorPositions(StrictEnum):
-    IN = InOut.IN
-    OUT = InOut.OUT
+    IN = InOut.IN.value
+    OUT = InOut.OUT.value
     SAMPLE_CHANGE = "sample change"
 
 
--- a/src/dodal/devices/aithre_lasershaping/laser_robot.py
+++ b/src/dodal/devices/aithre_lasershaping/laser_robot.py
@@ -5,9 +5,9 @@ from dodal.devices.robot import BartRobot
 
 
 class ForceBit(StrictEnum):
-    ON = OnOff.ON
+    ON = OnOff.ON.value
     NO = "No"
-    OFF = OnOff.OFF
+    OFF = OnOff.OFF.value
 
 
 class LaserRobot(BartRobot):
--- a/src/dodal/devices/i24/aperture.py
+++ b/src/dodal/devices/i24/aperture.py
@@ -5,8 +5,8 @@ from dodal.devices.motors import XYStage
 
 
 class AperturePositions(StrictEnum):
-    IN = InOut.IN
-    OUT = InOut.OUT
+    IN = InOut.IN.value
+    OUT = InOut.OUT.value
     ROBOT = "Robot"
     MANUAL = "Manual Mounting"
 
--- a/src/dodal/devices/i24/dual_backlight.py
+++ b/src/dodal/devices/i24/dual_backlight.py
@@ -5,8 +5,8 @@ from dodal.common.enums import OnOffUpper
 
 
 class BacklightPositions(StrictEnum):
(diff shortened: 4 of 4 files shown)


src/dodal/common/enums.py
```diff
@@ -0,0 +1,31 @@
+from ophyd_async.core import StrictEnum
+
+
+class OnState(StrictEnum):
+    ON = "On"
+    OFF = "OFF"
```
DiamondJoseph, line 6:
```suggestion
    OFF = "Off"
```

Looks like O**n** O**FF** wasn't used before, so just a slip here.

I'm checking with TC if an epics str pv is necessarily always case sensitive?

src/dodal/common/enums.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class OnStateCaptilised(StrictEnum):
```
DiamondJoseph:
```suggestion
class OnStateCapitalised(StrictEnum):
```

src/dodal/common/enums.py
```diff
@@ -0,0 +1,31 @@
...
+    ON = "ON"
+    OFF = "OFF"
+
+
+class EnabledState(StrictEnum):
+    ENABLED = "Enabled"
+    DISABLED = "Disabled"
+
+
+class EnabledStateCaptilised(StrictEnum):
```
DiamondJoseph:
```suggestion
class EnabledStateCaptalised(StrictEnum):
```

src/dodal/common/enums.py
```diff
@@ -0,0 +1,31 @@
...
+    ENABLED = "ENABLED"
+    DISABLED = "DISABLED"
+
+
+class InOut(StrictEnum):
+    IN = "In"
+    OUT = "Out"
+
+
+class InOutCapitlised(StrictEnum):
```
DiamondJoseph:
```suggestion
class InState(StrictEnum):
    IN = "In"
    OUT = "Out"


class InStateCapitalised(StrictEnum):
```

For consitency

DiamondJoseph (changes requested):

src/dodal/common/enums.py
oliwenmandiamond: Good spot

src/dodal/common/enums.py
```diff
@@ -0,0 +1,31 @@
+from ophyd_async.core import StrictEnum
+
+
+class OnState(StrictEnum):
```
callumforrester, line 4: Could: I'm not a huge fan of the `<First Member>State` naming pattern and I'm wondering about suggesting something like:

```suggestion
class OnOff(StrictEnum):
```

Ditto for the rest of them. I'm not sure about this though, opinions welcome.

callumforrester (changes requested):
Thanks for doing this

src/dodal/common/enums.py
callumforrester: Sorry, I didn't see @DiamondJoseph's review, we seem to be thinking exact opposites so ignore me

callumforrester: Okay, have discussed with @DiamondJoseph offline, I'll comment there

oliwenmandiamond:
Now changes are in ophyd-async, can we have another look at this please? Thanks!

.talismanrc
callumforrester, line 1: What's going on here? Does it just need a rebase?

oliwenmandiamond: Have a read on this thread, when rebasing it was causing issues https://diamondlightsource.slack.com/archives/C051NDJH6BZ/p1754404590964279

callumforrester: Change makes sense, I was asking why the entire file has also been reformatted

callumforrester (approved):

DiamondJoseph (approved):

PR #1418, opened 2025-07-31
Create common mirror classes and populate I05s mirrors

### LINKED ISSUE #1424: Create some common optics mirrors classes and populate I05 optics with them
There are quite a few common focussing/collimating/switching mirrors in GDA code for many beamlines. As always there are subtle differences and all flavours of PV suffices and combinations of various epics records. I05 has 3 different mirrors and some of them can be found on other beamlines.

## Acceptance Criteria

Villtord:
Fixes #1424 

1. dodal connect i05, i05-1
2. Confirm no errors in tests

### Checks for reviewer


Villtord:
Blocked by https://jira.diamond.ac.uk/browse/I05-731

src/dodal/devices/i05/common_mirror.py
```diff
@@ -0,0 +1,120 @@
...
+                read_pv=prefix + fpitch_read_suffix,
+                write_pv=prefix + fpitch_write_suffix,
+            )
+            self.mirror = epics_signal_rw(
+                mirrors,
+                read_pv=prefix + mirror_read_suffix,
+                write_pv=prefix + mirror_write_suffix,
+            )
+
+        self.mirror_change = epics_signal_x(write_pv=prefix + mirror_change_suffix)
```
DominicOram, line 115: Could: For devices in MX we try and hide PVs like this by making a `set` method on the device, something like:
```python
async def set(value:Mirror):
    await self._mirror.set(mirror)
    await self._mirror_change.set()
```
and so the user can just do `bps.set(mirror, MJ6)` and it works. Our reasoning is:
* Most other devices will move when you call set on them e.g. motors etc. and you won't need to call another thing to make it go. This means it's easy for an end user to forget to call the `bps.trigger(mirror.change)`
* In 99% of cases we want to set it and make it go straight away. In the case where you don't the underlying signals are still there for you if you need that, or we have sometimes implemented a `prepare` for it. This is specifically point 4 on https://blueskyproject.io/ophyd-async/main/explanations/when-to-extend-movable.html#when-should-a-device-extend-movable
* In our experience IOCs that do this do it because setting the value causes some processing so you need to make sure that processing is finished before you call the change PV. This is all fine if the user does:
```python
yield from bps.abs_set(mirror.mirror, MJ6, wait=True)
yield from bps.trigger(mirror.change)
```
but will cause issues if you miss out the `wait=True`.
* Additionally, if that processing does take some time then by putting it in the `ophyd` device you can more easily do other things in parallel:
```python
yield from bps.abs_set(mirror.mirror, MJ6, group="mirror_change")
do_something_else()
yield from bps.wait("mirror_change")
```
which you can sort of do as you have currently written it but less efficiently:
```python
yield from bps.abs_set(mirror.mirror, MJ6, group="mirror_set")
do_something_else()
yield from bps.wait("mirror_set")
yield from bps.trigger(mirror.change, group="mirror_change")
do_more_other_things()
yield from bps.wait("mirror_change")
```

src/dodal/devices/i05/common_mirror.py
```diff
@@ -0,0 +1,120 @@
...
+        self.mirror_abort = epics_signal_x(write_pv=prefix + mirror_abort_suffix)
```
DominicOram, line 116: Could: I would add calling this into a `stop` so that you can do `yield from bps.stop(mirror)` to stop it moving, like you would for a motor. I _think_ this will also mean that if you use it in a plan and there's an exception Bluesky will abort the move automatically. This may or may not be behavior you want

src/dodal/devices/i05/common_mirror.py
```diff
@@ -0,0 +1,120 @@
...
+        z_infix: str = _Z,
+        pitch_infix: str = "PITCH",
+        yaw_infix: str = "YAW",
+        roll_infix: str = "ROLL",
+        fpitch_read_suffix: str = "FPITCH:RBV",
+        fpitch_write_suffix: str = "FPITCH:DMD",
+        mirror_read_suffix: str = "MIRCTRL:RBV:MIRROR",
+        mirror_write_suffix: str = "MIRCTRL:DMD:MIRROR",
+        mirror_change_suffix: str = "MIRCTRL:SEQ:CHNG:MIRROR.PROC",
+        mirror_abort_suffix: str = "MIRCTRL:DMD:ABORT.PROC",
```
DominicOram, line 101: Nit: This amount of configurability seems like overkill to me. I would only add an `_infix` option when I have a known example of a device that is exactly the same but with that PV being different. But maybe you already do?

src/dodal/beamlines/i05.py
```diff
@@ -17,6 +25,39 @@ def synchrotron() -> Synchrotron:
...
+
+
+# beamline specific devices
+
+
+# will connect after https://jira.diamond.ac.uk/browse/I05-731
+class M4M5Mirror(StrictEnum):
+    UNKNOWN = "Unknown"
+    MJ6 = "M4"
+    M3 = "M5"
```
DominicOram, line 54: Should: Are these meant to be `MJ6` and `M3` or are they meant to be `M4` and `M5`?

src/dodal/beamlines/i05_1.py
```diff
@@ -13,10 +21,43 @@
...
+
+
+# beamline specific devices
+
+
+# will connect after https://jira.diamond.ac.uk/browse/I05-731
+class Mj7j8Mirror(StrictEnum):
+    UNKNOWN = "Unknown"
+    MJ6 = "MJ8"
+    M3 = "MJ7"
```
DominicOram, line 53: Should: Are these meant to be MJ6 and M3 or are they meant to be MJ8 and MJ7?

src/dodal/beamlines/i05.py
```diff
@@ -17,6 +25,39 @@ def synchrotron() -> Synchrotron:
...
+# will connect after https://jira.diamond.ac.uk/browse/I05-731
+@device_factory()
+def m3mj6() -> XYZPiezoSwitchingMirror:
+    return m3mj6_switching_mirror()
+
+
+# beamline specific devices
+
+
+# will connect after https://jira.diamond.ac.uk/browse/I05-731
```
DominicOram, line 50: ❤️ love the push to standardise the EPICS  here!

Villtord:
@oliwenmandiamond thanks I'm on your side with that, but atm I think we just stick to what we have.

jacob720 (approved):
Looks good, one tiny comment, take it or leave it

oliwenmandiamond (commented):
This looks good. Thank you for making it comply with the new structure! Just minor formatting change and reuse the defaults and then happy to approve

PR #1420, opened 2025-07-31
Mx bluesky 1007 commissioning mode

rtuck99:
Part of fix required for
* DiamondLightSource/mx-bluesky#1007

See related mx-bluesky PR: TBC

* This introduces a PV on the `Baton` device `COMMISSIONING`
* Global functions `is_commissioning_mode_enabled()` and `set_commissioning_signal()` to enable other devices to be able to determine whether the commissioning mode is enabled without needing a child reference to the baton.

1. Test pass

### Checks for reviewer


DominicOram:
Sorry, there is a `mx-bluesky` equivalent, it's link was just missing. Will review both now

rtuck99:
> Thanks. Couple of minor comments. Additionally:
> 
>     * Must: when does the `set_commissioning_signal` get called in production?
> 
>     * Could: Given that we're only (currently) using `is_commissioning_mode_enabled` in the undulator and I can't see it changing we could also just provide it as a reference when creating the undulator. It seems a bit odd to me to have a global signal hanging around but I can see the advantage of something global too.

Good catch, I prevaricated deciding where to set it, because of the whole baton_handler rebuilding devices thing, then forgot to add it.

I thought about adding it as a reference, unfortunately that means the undulator then has to have responsibility for instantiating the baton (or vice-versa), which seems wrong and also is an issue if more than one device wants to use commissioning mode. Everywhere else we use a device reference, they are ultimately "owned" by a top-level device. Unless we want to make `device_factory` support `inject`?

DominicOram:
>I thought about adding it as a reference, unfortunately that means the undulator then has to have responsibility for instantiating the baton (or vice-versa), which seems wrong and also is an issue if more than one device wants to use commissioning mode. Everywhere else we use a device reference, they are ultimately "owned" by a top-level device. Unless we want to make device_factory support inject?

If you instantiate devices through the device factories they all end up as singletons anyway and only actually get instantiated if they haven't already been. e.g. the `undulator_dcm` contains a reference to the `undulator` and `dcm` https://github.com/DiamondLightSource/dodal/blob/d10550b29dfdd943c62422f83bbf860c52746d27/src/dodal/beamlines/i03.py#L276:
```python
undulator = i03.undulator() # This does the instantiation
undulator_dcm = i03.undulator_dcm() # This is just instantiated with the above
```
I guess in this case they are "owned" by the cache inside `device_factory`, which seems fine to me?

DominicOram (approved):
Thanks. One optional comment, take it or leave it.

PR #1431, opened 2025-08-05
Remove test after change in ophyd-async behaviour

GDYendell:
This is now not expected to raise a ValueError. See https://github.com/bluesky/ophyd-async/pull/988

This fixes the tests on main against the latest ophyd-async.

1. Run dodal tests against ophyd-async main

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ description = "Ophyd devices and other utils that could be used across DLS beaml
 dependencies = [
     "click",
     "ophyd",
-    "ophyd-async[ca,pva]>=0.12.3",
+    "ophyd-async[ca,pva]>=0.13.0",
     "bluesky",
     "pyepics",
     "dataclasses-json",


tests/devices/unit_tests/test_tetramm.py
```diff
@@ -171,25 +171,6 @@ async def test_hints_self_by_default(tetramm: TetrammDetector):
     assert tetramm.hints == {"fields": [tetramm.name]}
 
 
-async def test_prepare_with_too_low_a_deadtime_raises_error(
```
DiamondJoseph, line 174: This test is ensuring that we don't try to run the TetrAMM too fast during flyscans and get weird datasets or timeouts at the end. I would really like to see the test restored and the TetrAMM prepare method (the device is defined in this repository) modified to ensure the existing behaviour.

DiamondJoseph (changes requested):

callumforrester:
@DiamondJoseph Alternative PR that special cases Tetramm: https://github.com/DiamondLightSource/dodal/pull/1440

PR #1433, opened 2025-08-05
Add patch_all_motors function and use it

### LINKED ISSUE #1432: Add a patch_all_motors helper function
The `patch_motor` function patches enough of the logic on a motor so that you can do `bps.mv(my_motor, 10)` out of the box without an error.

Often we want to build mock devices where all the motors are patched like this so that we can use them in plans without worrying about exceptions being thrown. Therefore it would be useful to have a function that recursively did patched all motors.

## Acceptance Criteria
* There is a `patch_all_motors` that will recursively patch all motors in a device

DominicOram:
Fixes #1432

1. Confirm tests still pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/util/test_utils.py
+++ b/src/dodal/devices/util/test_utils.py
@@ -25,8 +25,7 @@ def patch_all_motors(parent_device: Device):
     motors = []
 
     def recursively_find_motors(device: Device):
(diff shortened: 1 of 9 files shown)


DiamondJoseph (approved):
Looks like a useful test util function

PR #1443, opened 2025-08-11
Add default image mode setting for analysers and clean up imports

fajinyuan:
Fixes #ISSUE

1. Refector image mode setting to apply to all electron analysers
2. Confirm imports are all correctly done

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/electron_analyser/abstract/base_driver_io.py
+++ b/src/dodal/devices/electron_analyser/abstract/base_driver_io.py
@@ -13,7 +13,7 @@ from ophyd_async.core import (
     derived_signal_r,
     soft_signal_rw,
 )
-from ophyd_async.epics.adcore import ADBaseIO
+from ophyd_async.epics.adcore import ADBaseIO, ADImageMode
 from ophyd_async.epics.core import epics_signal_r, epics_signal_rw
 
 from dodal.devices.electron_analyser.abstract.base_region import (
@@ -138,6 +138,8 @@ class AbstractAnalyserDriverIO(
         Args:
             region: Contains the parameters to setup the driver for a scan.
         """
+        # Set image_mode to SINGLE by default for all analysers
+        await self.image_mode.set(ADImageMode.SINGLE)
 
     def _get_energy_source(self, alias_name: str) -> SignalR[float]:
         energy_source = self.energy_sources.get(alias_name)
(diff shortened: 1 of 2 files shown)


src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -138,6 +138,8 @@ async def set(self, region: TAbstractBaseRegion):
         Args:
             region: Contains the parameters to setup the driver for a scan.
         """
+        # Set image_mode to SINGLE by default for all analysers
+        await self.image_mode.set(ADImageMode.SINGLE)
```
oliwenmandiamond, line 142: This needs to be moved to `AbstractElectronAnalyserDetector.trigger` method. You will also need to adjust some tests.

oliwenmandiamond (changes requested):
Almost there, just need to correct ordering of function calls and update tests to check for new call

oliwenmandiamond (changes requested):
A few comments. You must also add tests to the abstract driver to make sure it is staging correctly and the detector to make sure it is staging the driver correctly.

tests/devices/unit_tests/electron_analyser/abstract/test_base_detector.py
```diff
@@ -128,5 +128,69 @@ async def test_analyser_region_detector_trigger_sets_driver_with_region(
...
+    )
+
+    for reg_det in region_detectors:
+        reg_det.driver.set = AsyncMock()
+
+        assert_stage_calls_driver_stage_and_controller_disarm(
+            reg_det, sim_detector.controller, RE
+        )
+
+        reg_det.driver.set.assert_awaited_once_with(reg_det.region)
```
oliwenmandiamond, line 170: Okay, I understand why this test is failing now. So this relates to how the region detectors are designed. 
So we have our driver and it is wrapped by a detector. When we want to run multiple regions, we create a new instance of the detector which is responsible for setting up the driver with a region. It is passed in a reference to the same driver as it is the same detector, we just have to process the data one by one. 

So if we have 4 regions for example, we get 4 detectors but they all share the same reference to the same driver. This is so we don't have to do any additional connecting to the PV signals and only one thing is responsible for controlling the PV's for the same detector. The reason for this design is located here:

https://github.com/bluesky/bluesky/issues/1899

Essentially, a single detector wont save configuration for each new region setting. So we have to create new detectors for each region to force bluesky to record new configuration as it matches a new reference object as detectors are only designed to record configuration once. I will commit a fix in a moment. 

So the reason this test is failing is because we are checking stage is only called once on the driver. However, were calling it x number of regions.

The reason this works for trigger is because we are checking it was awaited with a different region object each time where as here, were just checking it was awaited once.

oliwenmandiamond (approved):
Looks good to me, but need someone else to approve with permissions

PR #1444, opened 2025-08-11
Fix pressure cell controller PVs

barnettwilliam:
Fixes errors when setting the pressure cell `control` PVs. Uses the wrapped style declaration style used elsewhere by pressure the cell code.  

### Instructions to reviewer on how to test:
1. Confirm the unit tests pass

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://diamondlightsource.github.io/dodal/main/reference/device-standards.html)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

PR #1449, opened 2025-08-11
Feature pressure cell controller moveable

barnettwilliam:
This feature makes `PressureJumpCellController` moveable so that the devices top-level control can be used to set the pressure to a requested value or perform an un-triggered jump with the given parameters. The two types accepted are a float or a `PressureJumpParameters` instance, for example:

1. `bps.move(pcell.control, 100)` sets the pressure to 100 bar
2. `bps.move(pcell.control, PressureJumpParameters(250, 1000) )` performs a pressure jump from 250 bar to 1000 bar

The move completes when the the pressure is reached or the pressure jump has completed indicated by the device setting the busy signal set to false.

1. Confirm the unit tests pass

### Checks for reviewer


src/dodal/devices/pressure_jump_cell.py
```diff
@@ -200,28 +203,59 @@ def __init__(
...
-            self.to_pressure = epics_signal_rw(float, f"{prefix}JUMPT")
-            self.jump_ready = epics_signal_rw(bool, f"{prefix}SETJUMP")
+            # Pressure jump
+            self.from_pressure = epics_signal_rw(int, f"{prefix}JUMPF")
+            self.to_pressure = epics_signal_rw(int, f"{prefix}JUMPT")
+            self.set_jump = epics_signal_rw(bool, f"{prefix}SETJUMP")
 
+            # Common
+            self.busy = epics_signal_r(bool, f"{prefix}GOTOBUSY")
+            self.stop = epics_signal_rw(bool, f"{prefix}STOP")
```
DominicOram, line 224: See my comment at https://github.com/DiamondLightSource/dodal/pull/1444/files#r2269206143

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -240,9 +280,7 @@ def __init__(
         self.all_valves_control = AllValvesControl(f"{prefix}{cell_prefix}", name)
         self.pump = Pump(f"{prefix}{cell_prefix}", name)
 
-        self.controller = PressureJumpCellController(
-            f"{prefix}{cell_prefix}CTRL:", name
-        )
+        self.control = PressureJumpCellController(f"{prefix}{cell_prefix}CTRL:", name)
```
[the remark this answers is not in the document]
barnettwilliam: Currently only used in i22-bluesky following PR to update https://github.com/DiamondLightSource/i22-bluesky/pull/157

src/dodal/devices/pressure_jump_cell.py
```diff
@@ -200,28 +203,65 @@ def __init__(
...
 
+    @AsyncStatus.wrap
+    async def set(self, value: int | PressureJumpParameters):
+        """
+        Sets the desired pressure waiting for the device to complete the operation.
+
+        If value is of type int, the pressure is set to the given fixed value.
+        If value is is of type PressureJumpParameters, a pressure jump is performed
+        given by its pressure_from and pressure_to.
+        """
```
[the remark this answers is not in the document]
DominicOram: Sorry, my comment here wasn't very clear. I meant that you could do:
```python
class DoJump(Triggerable):
    def __init__(self, prefix: str, busy_signal: Reference, timeout_signal: Reference):
        self._busy_signal = busy_signal
        self._timeout_signal = timeout_signal
        self.set_jump = epics_signal_rw(bool, f"{prefix}SETJUMP")

    @AsyncStatus.wrap
    async def trigger(self):
        await set_and_wait_for_other_value(
            self.set_jump, True, self._busy_signal(), True
        )

        await wait_for_value(self._busy_signal(), False, self._timeout_signal())

class PressureJumpCellController(StandardReadable, Movable, Stoppable):
            # Pressure jump
            self.from_pressure = epics_signal_rw(int, f"{prefix}JUMPF")
            self.to_pressure = epics_signal_rw(int, f"{prefix}JUMPT")
            self.do_jump = DoJump(prefix, Reference(self.timeout), Reference(self.busy))

def do_pressure_jump_plan(cell: PressureJumpCell, from_pressure: int, to_pressure: int):
    yield from bps.mv(
        cell.control.from_pressure, from_pressure, cell.control.to_pressure, to_pressure
    )
    yield from bps.trigger(cell.control.do_jump, wait=True)
```

Then the `set` on the `PressureJumpCellController` can just take an `int`. This obeys the advice in `ophyd-async` of set methods only being primitives https://github.com/bluesky/ophyd-async/blob/a221add534630462eb111174a89bd85a9317d3db/docs/explanations/when-to-extend-movable.md?plain=1#L13. I still think this is generally a cleaner but happy to chat further

tests/devices/test_pressure_jump_cell.py
```diff
@@ -414,3 +407,176 @@ async def test_setting_all_pressure_cell_valves(
...
+            "pjump-control-timeout": partial_reading(0.0),
+            "pjump-control-_stop": partial_reading(False),
+            "pjump-control-from_pressure": partial_reading(0.0),
+            "pjump-control-to_pressure": partial_reading(0.0),
+            "pjump-control-do_jump-set_jump": partial_reading(False),
+        },
+    )
+
+
+@patch("dodal.devices.pressure_jump_cell.wait_for_value")
```
DominicOram: Could: Because you have mocked out the busy behaviour you don't also need to patch this now. I you just remove the patch the test works fine

barnettwilliam: Removed in e06fd30b29f4dc94c92885feb519e44fd9668006

PR #1451, opened 2025-08-12
Add util function to log based on watcher events

### LINKED ISSUE #1450: Create utility functions using ophys-async's watchable statuses
I noticed we currently aren't using ophyd-async `WatchableAsyncStatus` in dodal, or in MX bluesky plans, but they provide us with useful information about the completion of a status. It also isn't massively obvious as to how the watchers should be use, so a dodal utility function to log messages based on watcher events will be quite useful.

## Acceptance Criteria
* Watcher utility function added

olliesilvester:
Fixes #1450 

1. Is the function useful and does it work?
2. Check new test

### Checks for reviewer


src/dodal/common/watcher_utils.py
```diff
@@ -0,0 +1,71 @@
...
+        target: Number | None = None,
+        name: str | None = None,
+        unit: str | None = None,
+        precision: int | None = None,
+        fraction: float | None = None,
+        time_elapsed: float | None = None,
+        time_remaining: float | None = None,
+    ):
+        if isinstance(current, Number) and isinstance(target, Number) and target:
+            current_percent = int((current / target) * 100)
```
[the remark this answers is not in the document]
jsouter: oh whoops. Even still, shouldn't the % through the move be looking based of `current-initial`?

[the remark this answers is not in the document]
olliesilvester: Ah yeah good spot, thanks

src/dodal/common/watcher_utils.py
```diff
@@ -0,0 +1,80 @@
...
+class _LogOnPercentageProgressWatcher(Watcher[Number]):
+    def __init__(
+        self,
+        status: WatchableAsyncStatus[Number],
+        message_prefix: str,
+        percent_interval: Number = 25,
+    ):
+        status.watch(self)
+        self.percent_interval = percent_interval
+        self._current_percent_interval = 0
```
DominicOram, line 17: Should: This doesn't need to be an instance variable

src/dodal/common/watcher_utils.py
```diff
@@ -0,0 +1,80 @@
...
+
+Number = int | float
+
+
+class _LogOnPercentageProgressWatcher(Watcher[Number]):
+    def __init__(
+        self,
+        status: WatchableAsyncStatus[Number],
+        message_prefix: str,
+        percent_interval: Number = 25,
```
DominicOram, line 13: Should: Can you use an annotated type here then I don't think you need the check on it being >0, something like `percent_interval: Annotated[Number, Gt(0)]= 25`

src/dodal/common/watcher_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        precision: int | None = None,
+        fraction: float | None = None,
+        time_elapsed: float | None = None,
+        time_remaining: float | None = None,
+    ):
+        if (
+            isinstance(current, Number)
+            and isinstance(target, Number)
+            and isinstance(initial, Number)
+            and target - initial
```
DominicOram, line 40: Could: I guess this is to check that `target != initial`? It feels a bit unintuitive, feels like just having `target != initial` here is clearer?

src/dodal/common/watcher_utils.py
```diff
@@ -0,0 +1,80 @@
...
+    Args:
+        status: A WatchableAsyncStatus. For example, Ophyd-async produces this status from a Motor.set method
+
+        message_prefix: The string at the start of each of the produced logging messages
+
+        percent_interval: How often to produce logging message, in terms of percentage completion
+        of the status.
+
+    Note that when using with Bluesky plan stubs you will need to cast the status (as of
+    Bluesky v1.14.2), since a Bluesky status doesn't use generics
```
DominicOram, line 70: Could: If we're going to do this properly we should make an issue in bluesky about this. It will require a bit of back and forth with ophyd and ophyd-async but it looks like these should be equivalent and so share a protocol:
* https://github.com/bluesky/ophyd/blob/578a24ce16f7de853bed5b84507785a6582cb63b/ophyd/status.py#L719
* https://github.com/bluesky/ophyd-async/blob/9d2016b893c6df0662b57273bf69f3e267a54137/src/ophyd_async/core/_status.py#L164

Though I guess there may be some status' that do not have it so you might still end up with `complete` not returning it

src/dodal/common/watcher_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        message_prefix: str,
+        percent_interval: Number = 25,
+    ):
+        status.watch(self)
+        self.percent_interval = percent_interval
+        self._current_percent_interval = 0
+        self.message_prefix = message_prefix
+        if self.percent_interval <= 0:
+            raise ValueError(
+                f"Percent interval on class _LogOnPercentageProgressWatcher must be a positive number, but recieved {self.percent_interval}"
```
DominicOram, line 21: Should: i before e, except after c

src/dodal/common/watcher_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        of the status.
+
+    Note that when using with Bluesky plan stubs you will need to cast the status (as of
+    Bluesky v1.14.2), since a Bluesky status doesn't use generics
+
+    Example usage within a Bluesky plan:
+    yield from bps.kickoff(my_detector)
+    status = yield from bps.complete(my_detector, group="collection complete")
+    status = cast(WatchableAsyncStatus, status)
+    log_on_percentage_complete(status, "Data collection triggers recieved", 10)
```
DominicOram, line 76: Should: i before e, except after c

DominicOram (changes requested):
I'm a bit torn on this, it sort of feels like we're just recreating https://blueskyproject.io/bluesky/main/progress-bar.html#progress-bar, which lets you see the progress of things and I would rather people use the out of the box bluesky tools. However, for when you don't have a user sat in front of the terminal, I do see the advantage of putting this in logs instead of displaying it in the command line so maybe we just add a comment on this class that points to the progress bar docs and suggests that might be better in some cases?

src/dodal/common/watcher_utils.py
olliesilvester: Hmm, I don't think the `Annotated` type will allow me to remove the explicit check - mypy seems to ignore the info given in `Annotated`. I will try pydantic

olliesilvester: Pydantic doesn't particularly help here either sadly

DominicOram: Hmm... What does `pyright` think?

src/dodal/common/watcher_utils.py
olliesilvester: It needs to be 0 on the first call though, and it needs to be persistent across calls of `__call__`

olliesilvester: All good :laughing:

src/dodal/common/watcher_utils.py
DominicOram: For future reference, issue is https://github.com/bluesky/bluesky/issues/1948

PR #1456, opened 2025-08-13
Convert devices to context-dependant manager

callumforrester:
Fixes #ISSUE

1. Confirm unit tests pass
2. Confirm `dodal connect b01-1` still works
3. Confirm `dodal connect <other beamline>` still works where the other beamline does not use the new factory manager

### Checks for reviewer

### DIFF
--- a/src/dodal/utils.py
+++ b/src/dodal/utils.py
@@ -6,13 +6,16 @@ import re
 import socket
 import string
 from collections.abc import Callable, Iterable, Mapping
-from dataclasses import dataclass
+from dataclasses import dataclass, field
+from enum import Enum
 from functools import update_wrapper, wraps
 from importlib import import_module
 from inspect import signature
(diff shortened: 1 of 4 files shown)


src/dodal/utils.py
```diff
@@ -237,6 +237,93 @@ def make_device(
...
+                skip,
+            )
+            self._factories[factory.__name__] = controller
+            return controller
+
+        return decorator
+
+
 def make_all_devices(
     module: str | ModuleType | None = None, include_skipped: bool = False, **kwargs
```
DiamondJoseph, line 328: So is blueapi then intended to call with kwargs of path_provider=our_path_provider?

callumforrester:
Going to park this for now, it's an idea for a device factory that doesn't depend on global state, if the idea proves popular I can carry on with it

PR #1470, opened 2025-08-20
Make undulator gap flyable

Relm-Arrowny:
Fixes #ISSUE

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/apple2_undulator.py
+++ b/src/dodal/devices/apple2_undulator.py
@@ -5,19 +5,22 @@ from math import isclose
 from typing import Any, Generic, TypeVar
 
 import numpy as np
-from bluesky.protocols import Movable
+from bluesky.protocols import Movable, Preparable
 from ophyd_async.core import (
     AsyncStatus,
+    FlyMotorInfo,
     SignalR,
     SignalW,
     StandardReadable,
     StandardReadableFormat,
     StrictEnum,
     derived_signal_rw,
+    error_if_none,
     soft_signal_r_and_setter,
     wait_for_value,
 )
 from ophyd_async.epics.core import epics_signal_r, epics_signal_rw, epics_signal_w
+from ophyd_async.epics.motor import MotorLimitsException
 from pydantic import BaseModel, ConfigDict, RootModel
 
 from dodal.log import LOGGER
@@ -153,7 +156,7 @@ class SafeUndulatorMover(StandardReadable, Movable[T], Generic[T]):
             raise RuntimeError(f"{self.name} is already in motion.")
 
 
-class UndulatorGap(SafeUndulatorMover[float]):
+class UndulatorGap(SafeUndulatorMover[float], Preparable):
     """A device with a collection of epics signals to set Apple 2 undulator gap motion.
     Only PV used by beamline are added the full list is here:
     /dls_sw/work/R3.14.12.7/support/insertionDevice/db/IDGapVelocityControl.template
@@ -186,8 +189,7 @@ class UndulatorGap(SafeUndulatorMover[float]):
         self.high_limit_travel = epics_signal_r(float, prefix + "BLGAPMTR.HLM")
         self.low_limit_travel = epics_signal_r(float, prefix + "BLGAPMTR.LLM")
 
-        # This is calculated acceleration from speed
-        self.acceleration_time = epics_signal_r(float, prefix + "IDGSETACC")
+        self.acceleration_time = epics_signal_r(float, prefix + "BLGSETVEL.ACCL")
 
         with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
             # Unit
@@ -197,7 +199,73 @@ class UndulatorGap(SafeUndulatorMover[float]):
         with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
             # Gap readback value
             self.user_readback = epics_signal_r(float, prefix + "CURRGAPD")
+
+        # Currently requested fly info, stored in prepare
+        self._fly_info: FlyMotorInfo | None = None
+
+        # Set on kickoff(), complete when motor reaches self._fly_completed_position
+        self._fly_status: AsyncStatus | None = None
+
         super().__init__(self.set_move, prefix, name)
+
+    @AsyncStatus.wrap
+    async def prepare(self, value: FlyMotorInfo):
+        """Move to the beginning of a suitable run-up distance ready for a fly scan."""
+        # Velocity, at which motor travels from start_position to end_position, in motor
+        # egu/s.
+        max_velocity, min_velocity, egu = await asyncio.gather(
+            self.max_velocity.get_value(),
+            self.min_velocity.get_value(),
+            self.motor_egu.get_value(),
+        )
+        if min_velocity > abs(value.velocity) or abs(value.velocity) > max_velocity:
+            raise ValueError(
+                f"Velocity {abs(value.velocity)} {egu}/s was requested for a id gap motor "
+                f" speed must be between {max_velocity} and {min_velocity} {egu}/s"
+            )
+
+        acceleration_time = await self.acceleration_time.get_value()
+        ramp_up_start_pos = value.ramp_up_start_pos(acceleration_time)
+        ramp_down_end_pos = value.ramp_down_end_pos(acceleration_time)
+
+        motor_lower_limit, motor_upper_limit, egu = await asyncio.gather(
+            self.low_limit_travel.get_value(),
+            self.high_limit_travel.get_value(),
+            self.motor_egu.get_value(),
+        )
+
+        if (
+            not motor_upper_limit >= ramp_up_start_pos >= motor_lower_limit
+            or not motor_upper_limit >= ramp_down_end_pos >= motor_lower_limit
+        ):
+            raise MotorLimitsException(
+                f"Motor trajectory for requested fly is from "
+                f"{ramp_up_start_pos}{egu} to "
+                f"{ramp_down_end_pos}{egu} but motor limits are "
+                f"{motor_lower_limit}{egu} <= x <= {motor_upper_limit}{egu} "
+            )
+
+        # move to prepare position at maximum velocity
+        await self.velocity.set(abs(max_velocity))
+        await self.set(ramp_up_start_pos)
+
+        # Set velocity we will be using for the fly scan
+        await self.velocity.set(abs(value.velocity))
+
+    @AsyncStatus.wrap
+    async def kickoff(self):
+        """Begin moving motor from prepared position to final position."""
+        fly_info = error_if_none(
+            self._fly_info, "Motor must be prepared before attempting to kickoff"
+        )
+
+        acceleration_time = await self.acceleration_time.get_value()
+        self._fly_status = self.set(fly_info.ramp_down_end_pos(acceleration_time))
+
+    def complete(self) -> AsyncStatus:
+        """Mark as complete once motor reaches completed position."""
+        fly_status = error_if_none(self._fly_status, "kickoff not called")
+        return fly_status
 
     async def _set_demand_positions(self, value: float) -> None:
         await self.user_setpoint.set(str(value))
--- a/src/dodal/devices/apple2_undulator.py
+++ b/src/dodal/devices/apple2_undulator.py
@@ -5,7 +5,7 @@ from math import isclose
 from typing import Any, Generic, TypeVar
 
 import numpy as np
-from bluesky.protocols import Movable, Preparable
+from bluesky.protocols import Flyable, Movable, Preparable
 from ophyd_async.core import (
     AsyncStatus,
     FlyMotorInfo,
@@ -156,7 +156,7 @@ class SafeUndulatorMover(StandardReadable, Movable[T], Generic[T]):
             raise RuntimeError(f"{self.name} is already in motion.")
 
 
-class UndulatorGap(SafeUndulatorMover[float], Preparable):
+class UndulatorGap(SafeUndulatorMover[float], Preparable, Flyable):
     """A device with a collection of epics signals to set Apple 2 undulator gap motion.
     Only PV used by beamline are added the full list is here:
     /dls_sw/work/R3.14.12.7/support/insertionDevice/db/IDGapVelocityControl.template
--- a/tests/devices/unit_tests/test_apple2_undulator.py
+++ b/tests/devices/unit_tests/test_apple2_undulator.py
@@ -1,11 +1,12 @@
 from collections import defaultdict
-from unittest.mock import AsyncMock
+from unittest.mock import AsyncMock, MagicMock
 
 import bluesky.plan_stubs as bps
 import pytest
 from bluesky.plans import scan
 from bluesky.run_engine import RunEngine
-from ophyd_async.core import init_devices
+from ophyd_async.core import FlyMotorInfo, init_devices
+from ophyd_async.epics.motor import MotorLimitsException
 from ophyd_async.testing import (
     assert_emitted,
     assert_reading,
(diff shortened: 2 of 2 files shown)


src/dodal/devices/apple2_undulator.py
```diff
@@ -113,7 +119,36 @@ async def estimate_motor_timeout(
     return abs((target_pos - cur_pos) * 2.0 / vel) + DEFAULT_MOTOR_MIN_TIMEOUT
 
 
-class SafeUndulatorMover(StandardReadable, Movable[T], Generic[T]):
+class UndulartorBase(abc.ABC, Device, Generic[T]):
```
olliesilvester, line 122:
```suggestion
class UndulatorBase(abc.ABC, Device, Generic[T]):
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -138,22 +173,70 @@ async def set(self, value: T) -> None:
...
-        if await self.fault.get_value() != 0:
-            raise RuntimeError(f"{self.name} is in fault state")
-        if await self.gate.get_value() == UndulatorGateStatus.OPEN:
-            raise RuntimeError(f"{self.name} is already in motion.")
+    def __init__(self, prefix: str, name: str = ""):
+        super().__init__(prefix=prefix, name=name)
+        del self.motor_stop  # Remove motor_stop from the public interface
+
+    async def stop(self, success=False):
+        LOGGER.info(f"Stopping {self.name} is not supported.")
```
DiamondJoseph, line 185:
```suggestion
    async def stop(self, success: bool = False):
        LOGGER.info(f"Stopping {self.name} is not supported.")
```

nit: Type hints

From discussion IRL, we think that this doesn't need to raise an exception although JW would prefer if it did. The plan should be allowed to continue running when the device has been asked to stop, because stopping the InsertionDevice before the move is finished causes the device to lose home and need to recalibrate the whole beamline.

tests/devices/unit_tests/test_apple2_undulator.py
```diff
@@ -18,13 +19,28 @@
...
     UndulatorGateStatus,
     UndulatorJawPhase,
     UndulatorPhaseAxes,
 )
 
 
+@pytest.fixture(scope="function")
+def logger(caplog: pytest.LogCaptureFixture):
+    logger = getLogger()
+    _ = [logger.removeHandler(h) for h in logger.handlers if h != caplog.handler]  # type: ignore
```
olliesilvester, line 33:
```suggestion
    [logger.removeHandler(h) for h in logger.handlers if h != caplog.handler]
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -138,22 +173,70 @@ async def set(self, value: T) -> None:
         await self.set_move.set(value=1, timeout=timeout)
         await wait_for_value(self.gate, UndulatorGateStatus.CLOSE, timeout=timeout)
```
DiamondJoseph, line 174: ~~there is an ophyd-async method called [set_and_wait_for_other_value](https://github.com/bluesky/ophyd-async/blob/9d2016b893c6df0662b57273bf69f3e267a54137/src/ophyd_async/core/_signal.py#L582) that might be useful here- rather than doubling up on the timeout?~~

https://github.com/bluesky/ophyd-async/issues/1029

I think the set should remove the `timeout` param and use the default timeout.

src/dodal/devices/apple2_undulator.py
```diff
@@ -138,22 +173,70 @@ async def set(self, value: T) -> None:
...
+
+        split_pv = prefix.split("-")
+        fault_pv = f"{split_pv[0]}-{split_pv[1]}-STAT-{split_pv[3]}ANYFAULT"
+        self.fault = epics_signal_r(float, fault_pv)
+        self.set_move = set_move
+        super().__init__(prefix=prefix + "BLGAPMTR", name=name)
+
+    @WatchableAsyncStatus.wrap
+    async def set(self, new_position: float, timeout=DEFAULT_MOTOR_MIN_TIMEOUT):
+        self._set_success = True
```
DiamondJoseph, line 205:
```suggestion
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -138,22 +173,70 @@ async def set(self, value: T) -> None:
...
+        LOGGER.info(f"Stopping {self.name} is not supported.")
 
 
-class UndulatorGap(SafeUndulatorMover[float]):
+class GapSafeUndulatorMotor(MotorWithoutStop, UndulartorBase):
+    """A device that will check it's safe to move the undulator before moving it and
+    wait for the undulator to be safe again before calling the move complete.
+    """
+
+    def __init__(self, set_move: SignalW, prefix: str, name: str = ""):
```
DiamondJoseph, line 193:
```suggestion
    def __init__(self, set_move: SignalW[float], prefix: str, name: str = ""):
```

And also in the `Base`

Because it's 0 or 1, try [bool] and if that doesn't work [int]

src/dodal/devices/apple2_undulator.py
```diff
@@ -226,25 +309,11 @@ def __init__(self, prefix: str, infix: str, name: str = ""):
...
-        self.acceleration_time = epics_signal_rw(float, fullPV + ".ACCL")
-        self.precision = epics_signal_r(int, fullPV + ".PREC")
-        self.deadband = epics_signal_r(float, fullPV + ".RDBD")
-        self.motor_done_move = epics_signal_r(int, fullPV + ".DMOV")
-        self.low_limit_travel = epics_signal_rw(float, fullPV + ".LLM")
-        self.high_limit_travel = epics_signal_rw(float, fullPV + ".HLM")
-        super().__init__(name=name)
+        pv = f"{prefix}BL{infix}"
+        fullPV = pv + "MTR"
+        super().__init__(prefix=fullPV, name=name)
```
DiamondJoseph, line 314:
```suggestion
        super().__init__(prefix=prefix, name=name)
```
Pass down the prefix where the constructor is used, to meet dodal device standards

src/dodal/devices/apple2_undulator.py
```diff
@@ -226,25 +309,11 @@ def __init__(self, prefix: str, infix: str, name: str = ""):
...
+        self.user_setpoint = epics_signal_w(str, pv + "SET")
```
DiamondJoseph, line 315: Check if the default signal on the Motor works, and if this is an alias or if that signal is incorrect.

tests/devices/unit_tests/test_apple2_undulator.py
```diff
@@ -97,6 +117,13 @@ async def test_in_motion_error(
         await mock_jaw_phase.set(2)
 
 
+async def test_unstoppable_motor_stop_not_implemented(
+    unstoppable_motor: MotorWithoutStop, caplog
```
DiamondJoseph, line 121:
```suggestion
    unstoppable_motor: MotorWithoutStop, caplog: pytest.LogCaptureFixture
```

nit: types

tests/devices/unit_tests/test_apple2_undulator.py
```diff
@@ -18,13 +19,28 @@
...
     UndulatorJawPhase,
     UndulatorPhaseAxes,
 )
 
 
+@pytest.fixture(scope="function")
+def logger(caplog: pytest.LogCaptureFixture):
+    logger = getLogger()
+    _ = [logger.removeHandler(h) for h in logger.handlers if h != caplog.handler]  # type: ignore
+    return logger
```
DiamondJoseph, line 34: I don't think this is necessary, we only instantiate the local logging Handler in tests/dodal connect. Don't need to worry about forwarding to graylog unexpectedly.
```suggestion
```

DiamondJoseph (changes requested):
Looks good other than these nits and questions.

src/dodal/devices/apple2_undulator.py
Relm-Arrowny: I believe it's this way because set_move.set doesn't return as completed until the motion is done, and UndulatorGateStatus closes shortly after the motion is done.

src/dodal/devices/apple2_undulator.py
```diff
@@ -69,10 +74,39 @@ async def estimate_motor_timeout(
...
+
+    def __init__(self, name=""):
+        # Gate keeper open when move is requested, closed when move is completed
+        self.gate: SignalR[UndulatorGateStatus]
+        self.fault: SignalR[float]
+        super().__init__(name=name)
+
+    @abc.abstractmethod
+    async def _set_demand_positions(self, value: T) -> None:
+        """Set the demand positions on the device without actually hitting move."""
```
DiamondJoseph, line 96:
```suggestion
class UndulatorBase(abc.ABC, Device, Movable[T], Generic[T]):
    """Abstract base class for Apple2 undulator devices.

    This class provides common functionality for undulator devices, including
    gate and fault signal management, safety checks before motion, and abstract
    methods for setting demand positions and estimating move timeouts.
    """

    def __init__(self, name: str = ""):
        # Gate keeper open when move is requested, closed when move is completed
        self.gate: SignalR[UndulatorGateStatus]
        self.fault: SignalR[float]
        super().__init__(name=name)

    @abc.abstractmethod
    async def _set_demand_positions(self, value: T) -> None:
        """Set the demand positions on the device without actually hitting move."""
```

nit: Can't this base class also be a Movable, if it's defining `_set_demand_positions`? It's an ABC, so it's not required to actually implement `set` yet.

src/dodal/devices/apple2_undulator.py
```diff
@@ -69,10 +74,39 @@ async def estimate_motor_timeout(
...
+
+    async def raise_if_cannot_move(self) -> None:
+        if await self.fault.get_value() != 0:
+            raise RuntimeError(f"{self.name} is in fault state")
+        if await self.gate.get_value() == UndulatorGateStatus.OPEN:
+            raise RuntimeError(f"{self.name} is already in motion.")
 
 
-class SafeUndulatorMover(StandardReadable, Movable[T], Generic[T]):
+class SafeUndulatorMover(StandardReadable, UndulatorBase, Movable[T]):
```
DiamondJoseph, line 109:
```suggestion
class SafeUndulatorMover(StandardReadable, UndulatorBase[T], Generic[T]):
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -94,25 +128,72 @@ async def set(self, value: T) -> None:
...
-    async def raise_if_cannot_move(self) -> None:
-        if await self.fault.get_value() != 0:
-            raise RuntimeError(f"{self.name} is in fault state")
-        if await self.gate.get_value() == UndulatorGateStatus.OPEN:
-            raise RuntimeError(f"{self.name} is already in motion.")
+    def __init__(self, prefix: str, name: str = ""):
+        super().__init__(prefix=prefix, name=name)
+        del self.motor_stop  # Remove motor_stop from the public interface
 
+    async def stop(self, success=False):
```
DiamondJoseph, line 142:
```suggestion
    async def stop(self, success: bool = False):
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -94,25 +128,72 @@ async def set(self, value: T) -> None:
...
+        self.gate = epics_signal_r(UndulatorGateStatus, prefix + "BLGATE")
+
+        split_pv = prefix.split("-")
+        fault_pv = f"{split_pv[0]}-{split_pv[1]}-STAT-{split_pv[3]}ANYFAULT"
+        self.fault = epics_signal_r(float, fault_pv)
+        self.set_move = set_move
+        super().__init__(prefix=prefix + "BLGAPMTR", name=name)
+
+    @WatchableAsyncStatus.wrap
+    async def set(self, new_position: float, timeout=DEFAULT_TIMEOUT):
```
DiamondJoseph, line 162:
```suggestion
    async def set(self, new_position: float, timeout: CalculatableTimeout = DEFAULT_TIMEOUT):
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -130,50 +211,56 @@ def __init__(self, prefix: str, name: str = ""):
...
-            # Gap velocity
-            self.velocity = epics_signal_rw(float, prefix + "BLGSETVEL")
-        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
-            # Gap readback value
-            self.user_readback = epics_signal_r(float, prefix + "CURRGAPD")
-        super().__init__(self.set_move, prefix, name)
+        """ Clear the motor config_signal as we need new PV for velocity."""
+        self._describe_config_funcs = ()
+        self._read_config_funcs = ()
+        self.velocity = epics_signal_rw(float, prefix + "BLGSETVEL")
```
DiamondJoseph, line 223: Is it possible to get this signal aliased to `prefix + ".VELO"` by the ID group? It would mean that this implementation of Motor only extends and adds signals, rather than modifying the `config_signals`, which if the base `Motor` implementation changes will miss out on those other changes.

DiamondJoseph: But the current signal is always returning 0 you said- is it viable to replace that signal with the actual calculation by the ID group?

tests/devices/i09_2_shared/test_i09_apple2.py
```diff
@@ -336,16 +336,16 @@ async def test_j09_apple2_controller_set_pol_does_not_go_via_lh_if_already_at_lh
...
-    ).assert_called_once_with(f"{top_inner_phase:.6f}", wait=True)
+    ).assert_called_once_with(f"{float(top_inner_phase)}", wait=True)
     get_mock_put(
         mock_id_controller.apple2().phase().btm_inner.user_setpoint
-    ).assert_called_once_with(f"{btm_inner_phase:.6f}", wait=True)
+    ).assert_called_once_with(f"{float(btm_inner_phase)}", wait=True)
     get_mock_put(
         mock_id_controller.apple2().phase().btm_outer.user_setpoint
-    ).assert_called_once_with(f"{btm_outer_phase:.6f}", wait=True)
+    ).assert_called_once_with(f"{float(btm_outer_phase)}", wait=True)
```
oliwenmandiamond: Same again

tests/devices/insertion_device/test_apple2_undulator.py
```diff
@@ -163,12 +178,53 @@ async def test_given_gate_never_closes_then_setting_phases_times_out(
...
+        {
+            "mock_id_gap-velocity": {
+                "value": 2.0,
+            },
+            "mock_id_gap-motor_egu": {
+                "value": "c",
+            },
+            "mock_id_gap-offset": {
+                "value": 0.0,
+            },
```
oliwenmandiamond: Update to use `partial_reading` for `{"value" : ...}`

oliwenmandiamond (changes requested):

oliwenmandiamond (commented):
The apple2_undulator.py file is growing to over 800 lines of code which is making it hard to read and get lost what I am looking at when reviewing. Now that this inside it's own insertion_device module, could we try and split this logic up into smaller files to make it more manageable?

For example, we could have something like this

insertion_device/apple2_models.py
insertion_device/utils.py
insertion_device/phase_motors.py
insertion_device/apple2_controllers.py
insertion_device/energy.py
insertion_device/polarisation.py

plus the others we have

insertion_device/energy_motor_lookup.py
insertion_device/id_enums.py
insertion_device/lookup_table_modeles.py (soon to be moved to config server)

It will help focus up the logic so you know exactly what area of the insertion device logic you are looking at as it is a complicated device and make it easier to review in the future. 

Now that I have typed it out, probably for another issue to do and not for this ticket...

tests/devices/insertion_device/test_polarisation.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture
+async def mock_id_polarisation(
+    mock_id_controller: DummyApple2Controller,
+) -> InsertionDevicePolarisation:
+    async with init_devices(mock=True):
+        mock_id_polarisation = InsertionDevicePolarisation(
+            id_controller=mock_id_controller,
+        )
+    return mock_id_polarisation
```
oliwenmandiamond: This seems to be duplicated from test_energy.py. Is it possible we can create a conftest.py inside dodal/tests/devices/insertion_device/conftest.py to reduce duplicate device setup between tests?

src/dodal/devices/insertion_device/apple2_controller.py
```diff
@@ -0,0 +1,359 @@
...
+    soft_signal_rw,
+)
+
+from dodal.devices.insertion_device.energy_motor_lookup import (
+    EnergyMotorLookup,
+)
+from dodal.devices.insertion_device.enum import Pol
+from dodal.log import LOGGER
+
+from .apple2_undulator import Apple2, Apple2PhasesVal, Apple2Val, UndulatorPhaseAxes
```
oliwenmandiamond: Use full import

oliwenmandiamond (approved):

Issue #634, opened 2024-06-18
Add an example to showcase device creation process


stan-dot:
There's no worked examples in this repository of how to produce a device. I think it makes a lot of sense to defer that [to existing ophyd-async docs](https://blueskyproject.io/ophyd-async/main/how-to/make-a-simple-device.html) ([& docs](https://blueskyproject.io/ophyd-async/main/how-to/make-a-standard-detector.html)).

Some of the process of writing a device should then be captured there (enum validation, testing PVs) , and then this can be simplified to "when to create/not create a new device", "how to write a good issue that captures requirements" and "how to properly test a device".

_Originally posted by @DiamondJoseph in https://github.com/DiamondLightSource/dodal/pull/579#pullrequestreview-2124806100_

PR #1484, opened 2025-08-27
Example device creation process and write tests docs

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/634

1. Check tests pass
2. Check added documentation makes sense and relevant.

### Checks for reviewer


docs/how-to/write-tests.md
```diff
@@ -6,3 +6,130 @@ Testing is essential to maintain the integrity and reliability of the codebase.
...
+Once we have written the tests, we need to check that they pass and how much coverage there is. pytest should automatically detect your tests and display one of the following icons:
+- Grey circle (not run yet)
+- Green circle with a tick (test passed)
+- Red circle with a cross (test failed)
+
+You can also check code coverage by right-clicking on the test icon and selecting `Run with coverage`. This will re-run the test and highlight line numbers in your code:
+- Green = covered by tests
+- Red = not covered (and ideally should be tested).
+
+To view all tests inside dodal, use the `Test Explorer` panel (flask icon). It displays all of your tests in a hierarchy, which you can run individually or in groups. It also shows the pass/fail icons mentioned above. Clicking on an item will run all tests beneath it in the hierarchy.
```
noemifrisina, line 128: Please specify that this is in the vscode IDE and the next part from the command line - it may not be very clear to people new to vscode...

docs/how-to/write-tests.md
```diff
@@ -6,3 +6,130 @@ Testing is essential to maintain the integrity and reliability of the codebase.
...
+
+You can also run all tests via the command:
+`tox -e tests`
+
+This will also display each Python module’s percentage of code coverage achieved.
+
+If you find that your tests are being skipped or not recognised by pytest, check for any syntax errors as this will block the tests being found. You can also check to see if there is an error output at the very bottom of the `Test Explorer` panel. This is usually caused by invalid syntax in your test file, or by circular dependencies in the code you are testing.
```
noemifrisina, line 135: Nice addition

noemifrisina (approved):
Just a couple of comments, it's good overall, just a bit clunky in a couple of places. 
Example is nice, I don't think we need anything else to give a good idea what to do.

docs/how-to/write-tests.md
```diff
@@ -6,3 +6,130 @@ Testing is essential to maintain the integrity and reliability of the codebase.
...
+
+    @AsyncStatus.wrap
+    async def unstage(self):
+        """
+        Once device is finished, set signal_b back to OFF.
+        """
+        await asyncio.gather(super().stage(), self.signal_b.set(OnOff.OFF))
+```
+
+In this example, we need to test the `stage`, `unstage`, and `read` methods. The `read` method needs to be tested because we are adding `signal_a` and `signal_b` as readables to the device, so we need to confirm that we get the expected read signals back when this method is called.
```
DiamondJoseph, line 56: I wouldn't enforce that `read` requires to be tested, as it's just using the behaviour of the StandardReadable method. As a reviewer I might want a test if there's complexity (e.g. some signals in `with add_children_as_readables()`, others in `add_children_as_readables(ConfigSignal)` or `with add_children_as_readables(HintedSignal)`) just to check that the developer is happy with which signals are where.

docs/how-to/write-tests.md
```diff
@@ -6,3 +6,130 @@ Testing is essential to maintain the integrity and reliability of the codebase.
...
+
+We use [pytest](https://docs.pytest.org/en/stable/contents.html) for writing tests in dodal. A core part of this library is the use of fixtures. A fixture is a function decorated with `@pytest.fixture` that provides setup/teardown or reusable test data for your tests. It is defined once and can be reused across multiple tests. Fixtures are mainly used to define devices and then inject them into each test.
+
+To help set up a device, they are created with the `init_devices(mock=True)` function from `ophyd_async.core`, which automatically initialises the device in mock mode. The `RunEngine` is present when creating a device to ensure there is an event loop available when connecting signals. The `RunEngine` fixture itself is defined in `dodal/tests/conftest.py`, which is a special file recognised by `pytest`. It defines fixtures that are automatically available to any test in the same directory (and its subdirectories) without needing to import them. This is useful for defining common setup code once without duplicating it across test files.
```
DiamondJoseph, line 60:
```suggestion
To help set up a device, they are created with the `init_devices(mock=True)` function from `ophyd_async.core`, which automatically initialises the device in mock mode. The `RunEngine` fixture is passed when creating a device to ensure there is an event loop available when connecting signals. This fixture is defined in `dodal/tests/conftest.py`, which is a special file recognised by `pytest`. It defines fixtures that are automatically available to any test in the same directory (and its subdirectories) without needing to import them. This is useful for defining common setup code once without duplicating it across test files.
```

docs/how-to/write-tests.md
```diff
@@ -6,3 +6,130 @@ Testing is essential to maintain the integrity and reliability of the codebase.
...
+    get_mock_put(sim_my_device.signal_b).assert_called_once_with(OnOff.OFF, wait=True)
+
+
+async def test_my_device_read(sim_my_device: MyDevice, RE: RunEngine) -> None:
+    prefix = sim_my_device.name
+    await assert_reading(
+        sim_my_device,
+        {
+            f"{prefix}-signal_a": partial_reading(ANY),
+            f"{prefix}-signal_b": partial_reading(ANY),
```
DiamondJoseph, line 100:
```suggestion
            f"{prefix}-signal_a": partial_reading(0),
            f"{prefix}-signal_b": partial_reading(OnOff.OFF),
```

If you're going to test the read method, you must check a real expected value. As discussed above, I don't think it's necessary unless the reading behaviour is complex.

docs/how-to/write-tests.md
```diff
@@ -6,3 +6,130 @@ Testing is essential to maintain the integrity and reliability of the codebase.
...
+```Python
+    prefix = sim_my_device.name
+    await assert_reading(
+        sim_my_device,
+        {
+            f"{prefix}-signal_a" : {"value": ANY},
+            f"{prefix}-signal_b" : {"value": ANY},
+        }
+    )
+```
```
DiamondJoseph, line 117: If the above test for the read is removed, this then flows a little better explaining how to do it if necessary.

```suggestion
You should test the output of a device when the device has many signals read and you want to ensure the correct ones are read at the correct times, or when the `read` method of it or one of its signals (e.g. a DerivedSignal) requires testing. Functions are defined in `ophyd-async` to aid with this. `assert_reading` allows us to compare the readings generated from a `Readable` device to the expected results. `partial_reading` wraps the value given in a mapping like `{"value": ANY}`, so we are actually checking that the reading matches the expected structure.

```Python
    prefix = sim_my_device.name
    await assert_reading(
        sim_my_device,
        {
            f"{prefix}-signal_a" : {"value": 0},
            f"{prefix}-signal_b" : {"value": OnOff.OFF},
        }
    )
```
```

File: docs/how-to/write-tests.md [hunk 7]
```diff
@@ -6,3 +6,130 @@ Testing is essential to maintain the integrity and reliability of the codebase.
...
+- Red = not covered (and ideally should be tested)
+
+To view all tests inside dodal, use the `Test Explorer` panel (flask icon) in VSCode. It displays all of your tests in a hierarchy, which you can run individually or in groups. It also shows the pass/fail icons mentioned above. Clicking on an item will run all tests beneath it in the hierarchy.
+
+You can also run all tests in the command line via the command:
+`tox -e tests`
+
+This will also display each Python module’s percentage of code coverage achieved.
+
+If you find that your tests are being skipped or not recognised by `pytest`, check for any syntax errors as this will block the tests being found. You can also check to see if there is an error output at the very bottom of the `Test Explorer` panel. This is usually caused by invalid syntax in your test file, or by circular dependencies in the code you are testing.
```
DiamondJoseph, line 135: I think most of this is IDE specific and could be reduced or cut.

It should at least defer to these already existing docs and if required improve them instead
https://diamondlightsource.github.io/dodal/main/how-to/coverage.html
https://diamondlightsource.github.io/dodal/main/how-to/run-tests.html

DiamondJoseph (approved):

Issue #503, opened 2024-05-03
 Support for devices that are shared between endstations


DiamondJoseph:
Some beamlines with multiple endstations share hardware in the experiment hutch, with the device moving between them. This is easily handled with the @skip_device decorator to temporarily disable the device while it is not present.

However, other components (such as some optical components) are present for both endstations, will need to be read by both beamlines, and may need to be partially/wholly controlled by one/both during the course of a visit.

Options:
- ReadOnly version of device for non-controlling branch?
- ReadOnly signal when device instantiated on non-controlling branch?
- Push handling to AuthZ service?
- Good faith?

## Acceptance Criteria
* A pattern has been decided upon and documented for how to manage shared devices.

PR #1490, opened 2025-08-29
feat: Control access to devices

DiamondJoseph:
Provides an alternative solution to #503, by wrapping Movable devices set methods so that they are only operated when `condition` is True. This removes the need for an N+1th instance of blueapi when N instances of blueapi share a device, and the need for additional network traffic, potentially across beamline networks.

This is also configurable per-instance of a device class, meaning that potential diversification of implementations is avoided.

1. Test that the `set` method can be used when the HutchAccessControl device is set to the correct value
2. Test that the `set` method cannot be used when the HutchAccessControl device is set to the correct value

### Checks for reviewer


DominicOram:
This is obviously significantly neater and cleaner than the current solution of multiple BlueAPI deployments. However, my concern with it is that it is still possible to get around the access control. For example, there is nothing stopping me being on i19-1 and running `from dodal.beamlines.i19-2 import shutter` and now I can control the shutter as if I was i19-2. Under a multiple BlueAPI system you can do proper authentication and have a token per BlueAPI instance so that you can be sure where traffic is coming from.

Maybe this is overkill but if we can engineer a system where accidentally controlling someone else's shutter is basically impossible that seems better at this stage than relying on developers to do the right thing.

olliesilvester:
This PR has been hanging around for ages because we haven't agreed on https://github.com/DiamondLightSource/dodal/issues/503 . Next Athena drop-in we will come up with some proper actions

PR #1503, opened 2025-09-03
Add basic outlier rejection for murko

### LINKED ISSUE #1242: Do some outlier rejection when using Murko results to centre
We should apply some outlier rejection to results from Murko before using them to centre. In testing, we found that taking the smallest 25% of x values dramatically improved the results, as where Murko got it wrong it was almost always to the right hand side of the loop, further down the pin. However this also removes plenty of good results so a less blunt solution should be found.

There are algorithms like [RANSAC](https://medium.com/@saurabh.dasgupta1/outlier-detection-using-the-ransac-algorithm-de52670adb4a) which may work well, and @aragaod has some ideas for how to improve the outlier rejection for the x values.

## Acceptance Criteria
* Murko results undergo outlier rejection which successfully removes results where it got it wrong

DominicOram:
Fixes #1242

1. Confirm tests pass

### Checks for reviewer


jacob720:
All comments should be addressed, thanks!

src/dodal/devices/i04/murko_results.py
```diff
@@ -21,8 +22,8 @@
 
 NO_MURKO_RESULT = (-1, -1)
 
-MurkoResult = dict
-FullMurkoResults = dict[str, list[MurkoResult]]
+# MurkoResult = dict
+# FullMurkoResults = dict[str, list[MurkoResult]]
```
rtuck99, line 27: Can remove these commented out lines

PR #1506, opened 2025-09-04
Change tetramm statistic plugin to stop tetramms on i22 failing to connect

RJCD-Diamond:
The tetramms had some plugins on i22, which no longer connect or have PV's unavailable. I have removed this plugin and replaced it with the parent, with less PV's


DiamondJoseph (approved):
Approving because if they don't connect we need to get them working again, but is the prefix wrong [does it now automtically get the prefix appended?], or has the Stats plugin been disabled? We should get the prefix corrected if possible, the stats plugin is a useful one to have

PR #1511, opened 2025-09-08
Ophyd async motor limit check in smargon

### LINKED ISSUE #1505: Use motor limit checking from `ophyd-async` for the smargon
As of https://github.com/bluesky/ophyd-async/pull/1032 `ophyd-async`'s motor provides limit checking for us. This means that we shouldn't need to do any logic to check limits in `dodal`/`mx-bluesky` (i.e. we can remove https://github.com/DiamondLightSource/mx-bluesky/blob/main/src/mx_bluesky/hyperion/device_setup_plans/smargon.py). However, when we move all the axes in the smargon we do so with a `CombinedMove`, which as of https://github.com/DiamondLightSource/dodal/pull/1401, does not use the `set` of the motor anymore. So we should explicitly check the motor limits inside `smargon.set`.

## Acceptance Criteria
* `Smargon.set` explicitly calls the limit checking on each of the axes it moves
* There are unit tests to confirm that when we try and move the smargon with a combined move outside of limits we get an exception
* Smargon limit checking code is removed from `mx-bluesky` (maybe spin into new issue)

teoching0705:
Fixes #1505 

1. Check if the motor are initialized with the limits
2. Check if the motor limit check works
3. Check the testing make sense

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/smargon.py
+++ b/src/dodal/devices/smargon.py
@@ -177,6 +177,7 @@ class Smargon(XYZStage, Movable):
             for motor_name, new_setpoint in value.items():
                 if new_setpoint is not None and isinstance(new_setpoint, int | float):
                     axis: Motor = getattr(self, motor_name)
+                    await axis.check_motor_limit(axis.user_setpoint, new_setpoint)
                     put_completion = await set_and_wait_for_value(
                         axis.user_setpoint,
                         new_setpoint,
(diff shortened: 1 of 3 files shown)


teoching0705:
I can't seems to let the CI to take the latest version of ophyd-async, and I think thats what causing all test to fail as there is no check_motor_limit in the older ophyd_async.

teoching0705:
I changed the mock motor limit from -inf, +inf to -1999, +1999 for checking motor out of limit exception. @oliwenmandiamond

DominicOram (approved):
Great, thank you!

PR #1524, opened 2025-09-11
Class pydoc added - done partly as part of Docathon

### LINKED ISSUE #1360: Every device class should have a docstring
To aid people in finding other devices that could be useful for them each device should have a docstring explaining what it does. Specifically:
* What a scientist will use it for
* Any unusual logic that the device might need

## Acceptance Criteria
* Every dodal device has a class docstring
* There is a CI step enforcing this

fajinyuan:
Fixes #1360

1. check class pydoc added
2. check they meet pydoc format

### Checks for reviewer


oliwenmandiamond:
After doing a bit more research on this, it appears if stuff defined internally in the device e.g 

```Python
class MyDevice:
    """
    Example device.

    Attributes
    ----------
    name : str
        A friendly name for the device.
    value : int
        The numeric value associated with it.
    """
    def __init__(self):
        self.name = "hello"
        self.value = 5
```
You do have to explicitly define the attributes inside for it to appear in auto generated docs. Is this necessary for all devices?

Can I ask someone else from @DiamondLightSource/bluesky-reviewers with more experience with this to have a look as I'm unsure if the goal of this is to just add a doc string to each device or to go further and list everything e.g attributes, inheritance etc for the doc builder to pick up?

fajinyuan:
I close this pullrequest as I believe now I mis-interpreted #1360 requirements and acceptance

Issue #1245, opened 2025-05-27
Tetramm only saves the last trigger of TriggerInfo to hdf


RJCD-Diamond:
When tetramms are setup for a flyscan and using a mulit-trigger TriggerInfo, the triggers are sent and received by the Tetramms but only the data from the last trigger is saved to the hdf.

Unlike most detectors, the tetramms do not have the CAM plugin and therefore need to configured differently in order to handle n-triggers, because the acquire PV cannot be used for determining and IDLE state

## Acceptance Criteria
* Trigger the Tetramms with 1, 2 and n triggers, complete correctly, save data to hdf and end.

PR #1527, opened 2025-09-12
Update Tetramm for n TriggerInfo triggers and modern ophyd_async, to stop freezing on i22

RJCD-Diamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/1245

This allows the tetramm to accept multuple triggers, save them to the hdf and not hang when waiting for an idle state


tests/devices/test_tetramm.py
```diff
@@ -266,11 +266,6 @@ async def test_pilatus_controller(
 
 
 async def assert_armed(driver: TetrammDriver) -> None:
-    sample_time = await driver.sample_time.get_value()
-    samples_per_reading = int(VALID_TEST_EXPOSURE_TIME / sample_time)
-    exposure = samples_per_reading * sample_time
-
     assert (await driver.trigger_mode.get_value()) is TetrammTrigger.EXT_TRIGGER
-    assert (await driver.averaging_time.get_value()) == exposure
```
callumforrester, line 274: Should: Probably still assert that the averaging time is correct, both in the case of 0 exposures and >0 exposures

src/dodal/devices/tetramm.py
```diff
@@ -111,9 +110,21 @@ async def prepare(self, trigger_info: TriggerInfo) -> None:
...
+            )
+        else:
+            averaging_time = trigger_info.livetime
+
         await asyncio.gather(
-            self.driver.averaging_time.set(trigger_info.livetime),
-            self.set_exposure(trigger_info.livetime),
+            self.driver.averaging_time.set(averaging_time),
+            self.set_exposure(averaging_time),
+            self._file_io.num_capture.set(trigger_info.total_number_of_exposures),
```
callumforrester, line 127: Not sure why this is needed, or why it won't get overwritten by a later call to `prepare`?

callumforrester (changes requested):

src/dodal/devices/tetramm.py
RJCD-Diamond: Because the tetramm doesn't have the normal CAM plugin it cannot process more than one trigger. If the file_io isn't set to capture more than one trigger, then the subsequent triggers will overwrite the data for the the first trigger. This is one of the  most important part of this change.

If there is a better way of doing it, I'm happy to hear but I can't see how

callumforrester: There's a little expand button to fill in the blank lines

RJCD-Diamond: I've just realised what you were trying to say:

``` 
self.driver.averaging_time.set(averaging_time),
self.set_exposure(averaging_time)
```

Do the same thing, I removed the first bit, so it's not doing it twice per prepare

callumforrester (approved):
I think this is mergable for now. I did take a look to see if I could simplify it but I would need to take more time. @coretl might have some views for future iterations.

PR #1529, opened 2025-09-15
Add backlight device to i19-2

### LINKED ISSUE #1522: Add a backlight device on I19-2
The backlight device on I19-2 is a simple in/out switch. There is already a device for a backlight, see if possible to instantiate that one and if not create a new one.

As the controls side will take care of switching the light on and off with the change in position, the only thing this device should do is move it to the IN or OUT position.

shree-iyengar-dls:
Fixes #1522 

1. Run dodal connect 
2. Run tests

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/i19/backlight.py
+++ b/src/dodal/devices/i19/backlight.py
@@ -9,7 +9,7 @@ class BacklightPosition(StandardReadable, Movable[InOutUpper]):
     """Device moves backlight to the IN or OUT position since controls side manages switching the light on/off"""
 
     def __init__(self, prefix: str, name: str = "") -> None:
-        self.position = epics_signal_rw(InOutUpper, prefix + "AD1:choiceButton")
+        self.position = epics_signal_rw(InOutUpper, f"{prefix}AD1:choiceButton")
         super().__init__(name)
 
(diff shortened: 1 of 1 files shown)


src/dodal/beamlines/i19_2.py
```diff
@@ -75,3 +76,11 @@ def synchrotron() -> Synchrotron:
...
     """
     return Synchrotron()
+
+
+@device_factory()
+def backlight() -> BacklightPosition:
+    """Get the i19-2 backlight device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i19-2, it will return the existing object.
+    """
+    return BacklightPosition(prefix=(PREFIX.beamline_prefix))
```
noemifrisina, line 86:
```suggestion
    return BacklightPosition(prefix=f"{PREFIX.beamline_prefix}-EA-IOC-12:")
```
Nit: As a general rule, we tend to pass the full prefix here, instead of just the beamline bit, unless there's reasons for it like differently named PVs in the device

noemifrisina (changes requested):
Good start, thanks! Just a couple of comments to get everything to connect...

PR #1533, opened 2025-09-15
Aithre get move required so that beam is at pixel non smargon

teoching0705:
Fixes DiamondLightSource/mx-bluesky#1269

1. Check if anything use smargon breaks

### Checks for reviewer


src/dodal/beamlines/aithre.py
```diff
@@ -4,8 +4,11 @@
 from dodal.devices.oav.oav_detector import NullZoomController, OAVBeamCentreFile
 from dodal.devices.oav.oav_parameters import OAVConfigBeamCentre
 
-ZOOM_PARAMS_FILE = "/dls_sw/i23/software/aithre/aithre_oav.xml"
-DISPLAY_CONFIG = "/dls_sw/i23/software/aithre/aithre_display.configuration"
+# ZOOM_PARAMS_FILE = "/dls_sw/i23/software/aithre/aithre_oav.xml"
+# DISPLAY_CONFIG = "/dls_sw/i23/software/aithre/aithre_display.configuration"
+ZOOM_PARAMS_FILE = "/dls/science/users/wck38436/aithre_oav.xml"
```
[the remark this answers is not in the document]
teoching0705: Sorry for confusion, I was trying to test if the auto centering on the laser shaping was taking these as a parameter. @co2e14 mentioned he does not get permission for the config file in the path above for now. Will change back later so probably ignore this for now

src/dodal/devices/aithre_lasershaping/goniometer.py
```diff
@@ -16,10 +16,11 @@ class Goniometer(XYZStage):
     """
 
     def __init__(self, prefix: str, name: str = "") -> None:
-        self.sampy = Motor(prefix + "SAMPY")
-        self.sampz = Motor(prefix + "SAMPZ")
-        self.omega = Motor(prefix + "OMEGA")
+        super().__init__(name)
```
rtuck99, line 19: I think this should probably be specifying prefix and x,y infixes to the super init

src/dodal/devices/aithre_lasershaping/goniometer.py
```diff
@@ -1,9 +1,9 @@
...
 
-from dodal.devices.motors import XYZStage, create_axis_perp_to_rotation
+from dodal.devices.motors import XYZOmegaStage, create_axis_perp_to_rotation
 
 
-class Goniometer(XYZStage):
+class Goniometer(XYZOmegaStage):
     """The Aithre lab goniometer and the XYZ stage it sits on.
 
     `x`, `y` and `z` control the axes of the positioner at the base, while `sampy` and
```
rtuck99, line 9: Probably best to change the naming convention here to reflect the names of the attribs rather than the PV names

src/dodal/devices/oav/oav_detector.py
```diff
@@ -46,9 +46,11 @@ class NullZoomController(BaseZoomController):
     def __init__(self):
         self.level = soft_signal_rw(str, "1.0x")
         self.percentage = soft_signal_rw(float, 100)
+        super().__init__()
 
-    def set(self, value):
-        raise Exception("Attempting to set zoom level of a null zoom controller")
+    @AsyncStatus.wrap
+    async def set(self, value: str) -> None:
+        await self.level.set("1.0x", wait=True)
```
rtuck99, line 53: This seems like it violates the principle of least surprise. I think if we must support set and it doesn't support anything other than a fixed value, we let it set to "1.0x" and nothing else, rather than quietly throwing away the input and pretending we worked.

src/dodal/devices/oav/pin_image_recognition/utils.py
```diff
@@ -132,6 +134,8 @@ def __init__(
         """
 
         self.preprocess = preprocess
+        self.open_ksize = open_ksize
```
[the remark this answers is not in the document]
rtuck99: Would be interesting to know what kind of performance we get out of all these opencv operations, from what I remember from when I played with OpenCV several years ago, it absolutely loves allocating new arrays for everything.

src/dodal/devices/oav/oav_parameters.py
```diff
@@ -92,17 +92,19 @@ def update(name, param_type, default=None):
         self.preprocess_K_size: int = update(
             "preProcessKSize", int
         )  # length scale for blur preprocessing
+        self.preprocess_iter: int = update("preProcessIteration", int, default=5)
         self.detection_script_filename: str = update("filename", str)
+        self.open_ksize: int = update("open_ksize", int, default=11)
```
rtuck99: Any particular reason why we went for ksize=11 here but 5 in PinTipDetection ?

rtuck99: Also, how confident are we that these values are suitable for other beamlines? Are the cameras, resolutions etc, all comparable?

src/dodal/devices/oav/pin_image_recognition/utils.py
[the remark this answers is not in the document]
teoching0705: I am not sure about the performance, but should not increase by too much as I just added one open_morph in front of the original pre-processing and just after the grayscale. Will need more testing though

rtuck99 (approved):
Approved - pending some minor documentation changes. 

Also perhaps think about whether we have the best defaults with the denoise for the existing pin tip detection in Hyperion - it may be better to be more conservative given that I don't think we're currently having too many issues with no denoising.

PR #1534, opened 2025-09-16
Move i10 and i10-1 shared optics devices to i10_optics

DG-At-Diamond:
Fixes sim-bluesky Issue [#111](https://github.com/DiamondLightSource/sm-bluesky/issues/111), #1560 

@DG-At-Diamond is on holiday so @Relm-Arrowny completed the rest of the issue. I have split the rest of the i10 devices and fixed the file not found issue due to mocking the wrong folder as we have moved ids to i10-optics.

I have also added __all__ in device.i10 for easier import. 

This also includes merging i10-1.py into i10_1.py as we apparently had both.

Check shared config file can be loaded by both and all devices in i10_shared are actually shared

### DIFF
--- a/src/dodal/devices/tetramm.py
+++ b/src/dodal/devices/tetramm.py
@@ -24,7 +24,9 @@ from ophyd_async.epics.adcore import (
     NDFileHDFIO,
     NDPluginBaseIO,
 )
-from ophyd_async.epics.core import PvSuffix, epics_signal_r
+from ophyd_async.epics.core import PvSuffix, epics_signal_r, stop_busy_record
+
+from dodal.log import LOGGER
 
 
 class TetrammRange(StrictEnum):
@@ -110,6 +112,7 @@ class TetrammController(DetectorController):
         current_trig_status = await self.driver.trigger_mode.get_value()
 
         if current_trig_status == TetrammTrigger.FREE_RUN:  # if freerun turn off first
+            LOGGER.info("Disarming TetrAMM from free run")
             await self.disarm()
 
         # trigger mode must be set first and on its own!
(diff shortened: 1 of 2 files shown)


src/dodal/beamlines/i10_1.py
```diff
@@ -0,0 +1,27 @@
+from dodal.common.beamlines.beamline_utils import device_factory
```
rtuck99, line 1: Seems like we now have `i10-1.py` and `i10_1.py` which is not at all confusing. Can we get rid of the former?

src/dodal/devices/i10/diagnostics.py
```diff
@@ -149,15 +162,34 @@ def __init__(self, prefix, name: str = "") -> None:
...
+        self.d5 = create_positioner(CellPosition, f"{prefix}IONC-01:Y")
         self.d5A = create_positioner(D5APosition, f"{prefix}PHDGN-06:DET:X")
         self.d6 = FullDiagnostic(f"{prefix}PHDGN-05:", D6Position, "DET:X")
         self.d7 = create_positioner(D7Position, f"{prefix}PHDGN-07:Y")
 
         super().__init__(name)
 
 
+class I10JDiagnostic(Device):
+    """Collection of all the diagnostic stage on i10."""
```
rtuck99, line 182: Presumably the comment here should read i10-1?

src/dodal/devices/i10/slits.py
```diff
@@ -104,9 +109,39 @@ def __init__(self, prefix: str, name: str = "") -> None:
...
+        self, prefix: str, name: str = "", connector: DeviceConnector | None = None
+    ) -> None:
+        self.s4 = BladeDrainCurrents(prefix=prefix + "DI-IAMP-02:")
+        self.s5 = BladeDrainCurrents(prefix=prefix + "DI-IAMP-03:")
+        self.s6 = BladeDrainCurrents(prefix=prefix + "DI-IAMP-04:")
+        super().__init__(name, connector)
+
+
+class I10SlitsSharedDrainCurrent(Device):
+    """Collection of all the drain current from i10 slits."""
```
rtuck99, line 143: Should it be `I10SharedSlitsDrainCurrent`, presumably it's the slits that are shared not the drain current

Relm-Arrowny: it is the drain current from the slits.

PR #1536, opened 2025-09-16
Increase odin timeout to fix issue seen on beamline

### LINKED ISSUE #1555: D spacing should be optional for DCM
dodal.devices.common_dcm.BaseDCM adds a number of pvs by default including crystal_metadata_d_spacing_a (PV is prefix + "DSPACING:RBV").  This does not seem to exist in i07, they do not collect this metadata.  It should be made optional in some way (defaulting to included is probably best so we don't need to change existing dcm extensions).

## Acceptance Criteria
I can create a dcm for i07 (which does not have this pv) using BaseDCM.

DominicOram:
Fixes #1555

1. Confirm tests still pass
2. Confirm timeout is increased

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -23,6 +23,7 @@ class EigerTimeouts:
     meta_file_ready_timeout: int = 30
     all_frames_timeout: int = 120
     arming_timeout: int = 60
+    odin_stop_timeout: int = 30
 
 
 class InternalEigerTriggerMode(Enum):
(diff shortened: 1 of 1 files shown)


RJCD-Diamond:
Had the same issue in GDA and upped it to 10 or 15 seconds at the beginning of this run. I'd be interested to know when you last had this issue and if you get it next time

RJCD-Diamond (approved):
Let me know if this solves the issue and keep an eye on this

PR #1549, opened 2025-09-18
Add DeviceManager as alternative to make_all_devices

tpoliaw:
See https://github.com/DiamondLightSource/dodal/discussions/1574 for some context

### DIFF
--- a/tests/test_device_manager.py
+++ b/tests/test_device_manager.py
@@ -2,7 +2,6 @@ from unittest.mock import MagicMock, Mock, patch
 
 import pytest
 from ophyd.device import Device as OphydV1Device
-from ophyd_async.core import Device
 from ophyd_async.core import Device as OphydV2Device
 from ophyd_async.sim import SimMotor
 from pytest import RaisesExc, RaisesGroup
@@ -421,8 +420,8 @@ def test_kwargs_factory(dm: DeviceManager):
 
     # Factories can have kwargs but they're ignored by the device manager
     @dm.factory
-    def foo(**kwargs) -> Device:  # type: ignore
-        s1(**kwargs)
+    def foo(**kwargs):
+        return s1(**kwargs)
 
     dm.build_all(s1)
     s1.assert_called_once_with()


src/dodal/device_manager.py
```diff
@@ -0,0 +1,612 @@
...
+V2 = TypeVar("V2", bound=OphydV2Device)
+
+DeviceFactoryDecorator = Callable[[Callable[Args, V2]], "DeviceFactory[Args, V2]"]
+OphydInitialiser = Callable[Concatenate[V1, ...], V1 | None]
+
+_EMPTY = object()
+"""Sentinel value to distinguish between missing values and present but null values"""
+
+
+class LazyFixtures(Mapping[str, Any]):
```
rtuck99: Wouldn't it be easier to do something like subclass `UserDict` with a prepopulated set of keys, and override `__getitem__`?

src/dodal/device_manager.py
```diff
@@ -0,0 +1,612 @@
...
+    def __contains__(self, key: Any) -> bool:
+        return key in self.ready or key in self.lazy
+
+    def __len__(self) -> int:
+        # Can just add the lengths as the keys are distinct by construction
+        return len(self.ready.keys()) + len(self.lazy.keys())
+
+    def __getitem__(self, key: str) -> Any:
+        if (value := self.ready.get(key, _EMPTY)) is not _EMPTY:
+            return value
```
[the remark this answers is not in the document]
DiamondJoseph: Looking at [Tiled's equivalent](https://github.com/bluesky/tiled/blob/98466e248036a43a1799659eda1fd1fc86519700/tiled/utils.py#L212) the pythonic way would be exceptions for flow control. I'm fine with any solution

tpoliaw:
> However, I think the typing for v1 is broken

Should be fixed

src/dodal/cli.py
```diff
@@ -62,16 +64,28 @@ def connect(beamline: str, all: bool, sim_backend: bool) -> None:
...
+    # Don't connect devices as they're built and do connection as an extra step,
+    # because the alternatives is handling the fact that only some devices may
+    # be lazy.
+
+    if (manager := getattr(mod, device_manager, None)) and isinstance(
+        manager, DeviceManager
+    ):
+        devices, instance_exceptions, connect_exceptions = manager.build_and_connect(
+            mock=sim_backend,
+        )
```
DiamondJoseph: Don't we need a path provider on this path? Or will beamlines that need one have a pathprovider fixture that is overwritten by blueapi?

DiamondJoseph (dismissed):

src/dodal/cli.py
tpoliaw: That was the intention. I didn't want the CLI to assume anything about what was required by the beamline.

tests/test_device_manager.py
```diff
@@ -0,0 +1,745 @@
...
+        dm.build_all()
+
+
+def test_missing_fixture_ok_if_not_required(dm: DeviceManager):
+    s1 = Mock()
+    s2 = Mock()
+
+    @dm.factory
+    def foo(unknown):
+        return s1(bar)
```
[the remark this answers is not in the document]
tpoliaw: Fair point, I'll replace `bar` with `unknown` instead of adding the second parameter though

Issue #1500, opened 2025-09-02
Add Data to Tetramm's


RJCD-Diamond:
Tetramms devices have the raw readouts from the 4 inputs, which are displayed under the "Data" section in the IOC.

Beamlines such as i22, b21 used these inputs to read from diodes which are then used to align samples etc.

The PV's that contains this data are currently not implemented in the ophyd_async devices. 

PVs such as: BL22I-EA-XBPM-02:Cur1:MeanValue_RBV

There should be an extra class called TetrammData created, which contains all these PV's and added to the TetrammDetector class

PR #1561, opened 2025-09-22
Add data PV's to Tetramm so the values can be used for alignment scans, read etc

RJCD-Diamond:
Fixes [#ISSUE](https://github.com/DiamondLightSource/dodal/issues/1500)

Tests dodal connect i22 (or other beamline works)
Test it can be used in a step scan

### DIFF
--- a/src/dodal/devices/temperture_controller/__init__.py
+++ b/src/dodal/devices/temperture_controller/__init__.py
@@ -1,0 +1,3 @@
+from .lakeshore.lakeshore import Lakeshore
+
+__all__ = ["Lakeshore"]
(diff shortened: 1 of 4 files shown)


DominicOram (changes requested):
Thanks, just the minor typo. 

Each of these stats are presumably an instance of something like https://github.com/bluesky/ophyd/blob/4585e1a0285c8898001dfe0aa90b905895a8f611/ophyd/areadetector/plugins.py#L355 and have `Min`/`Max` etc. too. @coretl does it make sense to put something in ophyd-async like the one in ophyd? (Happy to merge this w/o that though)

DominicOram (approved):
Looks good to me

PR #1563, opened 2025-09-22
1259 test oav plans

srishtysajeev:
### Checks for reviewer
- check src/dodal/devices/scintillator.py


olliesilvester:
I think this can be closed now because of https://github.com/DiamondLightSource/dodal/pull/1583/files, is that right?

PR #1590, opened 2025-10-01
1589 decople polarisation and energy from apple2 motors.

### LINKED ISSUE #1589: decople polarisation and energy from apple2 motors
### Goals
- [x] Make apple2 base class with gap and phase motor.
- [x] All test must still pass.
- [x] I10Apple2 function as before. 
 etc etc..

### Tasks
- [ ] 
- [ ]

Relm-Arrowny:
Fixes #1589 

Refactored Apple2 to use Apple2Motors and abstract motor setting.
Created Apple2Motors class to encapsulate the gap and phase motors.
Updated Apple2 to use an instance of Apple2Motors as an attribute.
Modified the set method to call the new _set method and handle energy updates.

1. Check the new design is what we want.
2. Run test and confirm everything still work.

### Checks for reviewer


src/dodal/devices/i10/i10_apple2.py
```diff
@@ -359,14 +360,16 @@ def __init__(
...
+                apple2_motors=Apple2Motors(
+                    id_gap=UndulatorGap(name="id_gap", prefix=prefix),
+                    id_phase=UndulatorPhaseAxes(
+                        name="id_phase",
+                        prefix=prefix,
+                        top_outer="RPQ1",
+                        top_inner="RPQ2",
+                        btm_inner="RPQ3",
+                        btm_outer="RPQ4",
+                    ),
```
oliwenmandiamond, line 372: Are the names necessary here? If you want the naming to always be id_gap and id_phase, why not just call them that in `Apple2Motors` so that the naming is done automatically correctly?

```suggestion
                apple2_motors=Apple2Motors(
                    id_gap=UndulatorGap(prefix),
                    id_phase=UndulatorPhaseAxes(
                        prefix=prefix,
                        top_outer="RPQ1",
                        top_inner="RPQ2",
                        btm_inner="RPQ3",
                        btm_outer="RPQ4",
                    ),
```

oliwenmandiamond (changes requested):
I think looks good! Just some minor comments on formatting

src/dodal/devices/apple2_undulator.py
```diff
@@ -420,24 +475,32 @@ async def _set_pol(
...
+        LOGGER.info(f"Energy set to {value} eV successfully.")
+
+    @abc.abstractmethod
+    async def _set(self, value: float) -> None:
+        """
+        This method should be implemented by the beamline specific ID class as the
+         motor positions will be different for each beamline depending on the
+         undulator design and the lookup table used. The set method can be
+         used to set the motor positions for the given energy and polarisation
+         provided that all motors can be moved at the same time.
```
oliwenmandiamond, line 502:
```suggestion
        This method should be implemented by the beamline specific ID class as the
        motor positions will be different for each beamline depending on the
        undulator design and the lookup table used. The set method can be
        used to set the motor positions for the given energy and polarisation
        provided that all motors can be moved at the same time.
```

oliwenmandiamond (approved):
One minor formatting change for doc string but other than all good!

src/dodal/devices/apple2_undulator.py
```diff
@@ -363,8 +420,7 @@ class Apple2(abc.ABC, StandardReadable, Movable):
 
     def __init__(
         self,
-        id_gap: UndulatorGap,
-        id_phase: UndulatorPhaseAxes,
+        apple2_motors: Apple2Motors,
         energy_motor_convertor: EnergyMotorConvertor,
```
Villtord, line 424: I thought we agreed on not having any energy conversion in undulator class?

Villtord (commented):
Sorry I am a bit lost - are we not making it pure gap and phase setting device or is it a separate ticket?

src/dodal/devices/i10/i10_apple2.py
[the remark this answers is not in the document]
Villtord: Feel like I'm being late for the party, in which case go ahead and just merge it and we can discuss on a design later in more details including maybe Fajin and Oli. I was also trying to see common case with hard-x-ray ID but it is currently quite tricky

src/dodal/devices/apple2_undulator.py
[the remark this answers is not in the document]
Villtord: yeah, this I get - but can you not just call function directly?
```
    async def _set_pol(
        self,
        value: Pol,
    ) -> None:
        # This changes the pol setpoint and then changes polarisation via set energy.
        self._polarisation_setpoint_set(value)
        await self.set(await self.energy.get_value())
```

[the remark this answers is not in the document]
Relm-Arrowny: o you are 100% right, it was historical and I forgot to remove it

src/dodal/devices/apple2_undulator.py
Villtord: Right, ok then apple2_motors looks like what in my understanding of recent agreement should be an ID device. I guess I was confused by docstring to your apple2 class that still say it is insertion device.

src/dodal/devices/i10/i10_apple2.py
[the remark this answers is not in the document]
Villtord: Ok, anyway I see there are lots of similarities in hard-x-ray ID and apple as well as differeences, so probably better to move forward and use it separately for now, later this week/next week we can try to refactor

Villtord, line 34: Minor - Perhaps better to use absolute import? `from dodal.devices.apple2_undulator `

src/dodal/devices/i10/i10_apple2.py
```diff
@@ -23,6 +23,7 @@
 
 from ..apple2_undulator import (
     Apple2,
+    Apple2Motors,
     Apple2Val,
     EnergyMotorConvertor,
     Pol,
```
oliwenmandiamond, line 29: Can you change this to absolute import please

oliwenmandiamond (changes requested):

src/dodal/devices/i10/i10_apple2.py
```diff
@@ -23,6 +23,7 @@
 
 from ..apple2_undulator import (
```
Villtord, line 24: `from dodal.devices.apple2_undulator`?

src/dodal/devices/apple2_undulator.py
Villtord, line 91: Could you please add docs to the "set" method as well?

Villtord: Sorry it's really bit difficult to read (for me of course), would it be possible, despite possible repetitions, make it simpler though, e.g.
1. do not pass set_move signal at all, but leave it in child classes set() methods
2. make this class do just safe checking and common signals wiring.
3. this class then could be either Movable or not depending on individual preference.

Relm-Arrowny: I can add docs but it kind of already there as class docstring.

I think what you asking for is this?:
https://github.com/DiamondLightSource/dodal/blob/f4873aacadaa3d11c02bf9b9712dec306a928561/src/dodal/devices/apple2_undulator.py#L76

If it is not quite what you after, I can make an issue.
I really do not want to introduce more confict here, as this issue is just to decople polarisation and energy from apple2.

PR #1593, opened 2025-10-01
Fix the Zebra prefixes for I19

noemifrisina:
The prefixes for the Zebra devices on I19 hutches were both connecting to the wrong device.
The correct prefixes are: 
- for eh1: `BL19I-EA-ZEBRA-02:`
- for eh2: `BL19I-EA-ZEBRA-03:`

1. Run dodal connect on i19-1 and i19-2 and confirm it still connect correctly

### Checks for reviewer


DominicOram:
>but it's not in use

Have we asked about the possibility of turning off the IOC then? It seems mad that we have it hanging around

noemifrisina:
> > but it's not in use
> 
> Have we asked about the possibility of turning off the IOC then? It seems mad that we have it hanging around

I haven't yet - mainly because it's easy to just forget it exists until we run into something like this - likely will next week at the beamline catch up.

PR #1594, opened 2025-10-01
Murko: Add metadata on whether an image was used for centring back to redis

jacob720:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/1303

Requires merging https://github.com/DiamondLightSource/mx-bluesky/pull/1323 too

1. Check metadata containing whether or not results have been used is added back into redis
2. Check tests make sense

### Checks for reviewer


DominicOram (changes requested):
Thanks, I think the logic is maybe getting a bit messy. I would do:
* Keep `self.results` as just all the results (make it private too)
* Don't keep the `discarded_results` separately
* `filter_outliers` should just return the good ones and they can be used in the rest of the `trigger` function

Does that make sense?

Advantages of this:
* The meaning of what is in `results` doesn't change
* When we set the metadata back to redis we can just do it for all `results`, we don't need to worry about separate lists

DominicOram (approved):
Great, thank you!

PR #1600, opened 2025-10-02
Fix binding energy

oliwenmandiamond:
Fixes #ISSUE

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/electron_analyser/abstract/base_driver_io.py
+++ b/src/dodal/devices/electron_analyser/abstract/base_driver_io.py
@@ -153,11 +153,12 @@ class AbstractAnalyserDriverIO(
             self.energy_source.selected_source.set(region.excitation_energy_source)
         excitation_energy = await self.energy_source.energy.get_value()
 
-        # Copy region so doesn't alter the actual region and switch to kinetic energy
-        ke_region = region.model_copy()
-        ke_region.switch_energy_mode(EnergyMode.KINETIC, excitation_energy)
-
+        # Switch to kinetic energy as epics doesn't support BINDING.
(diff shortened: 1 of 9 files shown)


src/dodal/devices/electron_analyser/abstract/base_driver_io.py
```diff
@@ -153,11 +153,12 @@ async def set(self, region: TAbstractBaseRegion):
             self.energy_source.selected_source.set(region.excitation_energy_source)
         excitation_energy = await self.energy_source.energy.get_value()
 
-        # Copy region so doesn't alter the actual region and switch to kinetic energy
-        ke_region = region.model_copy()
-        ke_region.switch_energy_mode(EnergyMode.KINETIC, excitation_energy)
-
+        # Switch to kinetic energy as epics doesn't support BINDING.
+        ke_region = region.switch_energy_mode(EnergyMode.KINETIC, excitation_energy)
```
Villtord, line 157: Nit: Would be nice to have this method name somehow state that it returns copy of itself, otherwise I expect it to just switch energy mode in a current region and return None

Villtord (changes requested):

src/dodal/devices/electron_analyser/abstract/base_driver_io.py
oliwenmandiamond: We could maybe instead just add a copy flag. It always returns region, but if copy=False, it alters it self so it can be ignored

e.g 

```Python
ke_region = region.switch_energy_mode(EnergyMode.KINETIC, excitation_energy, copy=True)
```
and
```Python
region.switch_energy_mode(EnergyMode.KINETIC, excitation_energy, copy=False)
```

PR #1609, opened 2025-10-06
1606 make generic energy device

Relm-Arrowny:
Fixes #ISSUE

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/apple2_undulator.py
+++ b/src/dodal/devices/apple2_undulator.py
@@ -5,9 +5,10 @@ from math import isclose
 from typing import Generic, Protocol, TypeVar
 
 import numpy as np
-from bluesky.protocols import Movable
+from bluesky.protocols import Movable, Status
 from ophyd_async.core import (
     AsyncStatus,
+    Reference,
     SignalR,
     SignalW,
     StandardReadable,
@@ -15,10 +16,12 @@ from ophyd_async.core import (
     StrictEnum,
     derived_signal_rw,
     soft_signal_r_and_setter,
+    soft_signal_rw,
     wait_for_value,
 )
 from ophyd_async.epics.core import epics_signal_r, epics_signal_rw, epics_signal_w
 
+from dodal.devices.pgm import PGM
 from dodal.log import LOGGER
 
 T = TypeVar("T")
@@ -40,12 +43,8 @@ class Apple2PhasesVal:
 
 
 @dataclass
-class Apple2Val:
+class Apple2Val(Apple2PhasesVal):
     gap: str
-    top_outer: str
-    top_inner: str
-    btm_inner: str
-    btm_outer: str
 
 
 class Pol(StrictEnum):
@@ -301,9 +300,9 @@ class UndulatorJawPhase(SafeUndulatorMover[float]):
         )
 
 
-class Apple2Motors(StandardReadable, Movable):
+class Apple2(StandardReadable, Movable):
     """
-    Device representing the combined motor controls for an Apple2 undulator.
+    Device representing the combined motor controls for an Apple2Controller undulator.
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


src/dodal/devices/apple2_undulator.py
```diff
@@ -605,3 +579,90 @@ def determine_phase_from_hardware(
...
+        self.add_readables(
+            [
+                self._Id_energy().energy(),
+                self._mono_energy().user_readback,
+            ],
+            StandardReadableFormat.HINTED_SIGNAL,
+        )
+
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            self.energy_offset = soft_signal_rw(float, initial_value=0)
```
Villtord, line 629: I would rename it to e.g. `id_energy_offset `as it applies only to `id`

src/dodal/devices/apple2_undulator.py
```diff
@@ -366,141 +365,116 @@ def __call__(self, energy: float, pol: Pol) -> tuple[float, float]:
...
-    Apple2 Undulator Device
+Apple2Type = TypeVar("Apple2Type", bound="Apple2")
 
-    The `Apple2` class represents an Apple 2 insertion device (undulator) used in synchrotron beamlines.
-    This device provides additional degrees of freedom compared to standard undulators, allowing independent
-    movement of magnet banks to produce X-rays with various polarisations and energies.
 
-    The class is designed to manage the undulator's gap, phase motors, and polarisation settings, while
-    abstracting hardware interactions and providing a high-level interface for beamline operations.
+class Apple2Controller(abc.ABC, StandardReadable, Generic[Apple2Type]):
```
Villtord, line 371: I think it is important as well to highlight that instance of this class must be a singleton. Maybe we need to force it like apply a singleton decorator.

src/dodal/devices/apple2_undulator.py
```diff
@@ -605,3 +579,90 @@ def determine_phase_from_hardware(
...
+
+        self.add_readables(
+            [
+                self._Id_energy().energy(),
+                self._mono_energy().user_readback,
+            ],
+            StandardReadableFormat.HINTED_SIGNAL,
+        )
+
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
```
Villtord, line 628: Do you actually need to add anything here as readables? it looks like controller defines logic, but later you use Reference to either energy or polarisation.

src/dodal/devices/apple2_undulator.py
```diff
@@ -605,3 +579,90 @@ def determine_phase_from_hardware(
...
+            self._Id_energy().set(energy=energy + await self.energy_offset.get_value()),
+            self._mono_energy().set(energy),
+        )
+
+
+class IdEnergy(IdEnergyBase):
+    """Apple2 ID energy movable device."""
+
+    def __init__(self, id_controller: Apple2Controller, name: str = "") -> None:
+        self.energy = Reference(id_controller.energy)
```
Villtord, line 644: Does this need to be added as readables (again?) as it is a Reference?

src/dodal/devices/apple2_undulator.py
```diff
@@ -366,141 +365,116 @@ def __call__(self, energy: float, pol: Pol) -> tuple[float, float]:
...
+        name: str
+            Name of the device.
         """
+        self.energy_to_motor: EnergyMotorConvertor
+        self.apple2 = Reference(apple2)
 
-        self.motors = apple2_motors
-        self.energy_to_motor = energy_motor_convertor
+        # Store the set energy for readback.
+        self._energy = soft_signal_rw(float, initial_value=None)
```
Villtord, line 428: why here it is `soft_signal_rw `and cached polarisation is `soft_signal_r_and_setter`?

Villtord (commented):
I think it has much better structure now, thanks. Had a few comments/questions, although I didn't make it to the end so might have more soon.

PR #1625, opened 2025-10-10
If scintillator is already out do nothing when moved out

DominicOram:
Fixes DiamondLightSource/mx-bluesky#1283

1. Confirm test passes

### Checks for reviewer


tests/devices/test_scintillator.py
```diff
@@ -79,7 +82,7 @@ async def test_given_aperture_scatterguard_parked_when_set_to_out_position_then_
     scintillator, ap_sg = scintillator_and_ap_sg
     ap_sg.return_value.selected_aperture.get_value.return_value = ApertureValue.PARKED  # type: ignore
 
-    await scintillator.selected_pos.set(InOut.OUT)
+    await scintillator.selected_pos.set(InOut.OUT, wait=True)
```
rtuck99: wait=True is the default, which makes it strange that it's specified here but not elsewhere in this file

PR #1630, opened 2025-10-16
Fix typing for new pillow release

DominicOram:
1. Confirm `pyright tests` passes on this branch but fails on `main`

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -29,6 +29,7 @@ from dodal.devices.fast_grid_scan import PandAFastGridScan, ZebraFastGridScanThr
 from dodal.devices.fluorescence_detector_motion import FluorescenceDetector
 from dodal.devices.flux import Flux
 from dodal.devices.focusing_mirror import FocusingMirrorWithStripes, MirrorVoltages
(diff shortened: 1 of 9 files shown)


olliesilvester (approved):

tests/devices/oav/test_snapshots.py
```diff
@@ -116,8 +116,8 @@ def image_data_coro(image_bytes: BytesIO) -> AsyncMock:
 
 
 def assert_images_identical(left: Image.Image, right: Image.Image):
-    left_data = left.getdata()
-    right_data = right.getdata()
+    left_data = list(left.getdata())  # type:ignore # Can remove once https://github.com/python-pillow/Pillow/pull/9261 is released
```
tpoliaw: Have commented on the PR but I think the simpler change that doesn't skip type checking would be 


```suggestion
    left_data = list(iter(left.getdata()))
```

tpoliaw: Although looking at what the test is doing, I think the whole thing could be 

```python
assert all(l == r for l,r in zip(left.getdata(), right.getdata()), strict=True))
```

Issue #1639, opened 2025-10-17
Create dodal module for VMXi


olliesilvester:
VMXi (i02-2) have requested a dodal module with a few simple devices so that they can start to familiarise themselves a bit with the Bluesky stack

## Acceptance Criteria
- Dodal module created for i02-2
- A few simple devices added
- dodal connect i02_2 works

PR #1641, opened 2025-10-17
Add i02-2 dodal module

olliesilvester:
Fixes [#1639](https://github.com/DiamondLightSource/dodal/issues/1639)

To review:
Check `dodal connect i02_2` works

### Checks for reviewer


src/dodal/devices/i02_2/sample_motors.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __init__(
+        self,
+        prefix: str,
+        name: str = "",
+    ):
+        # See https://github.com/DiamondLightSource/mx-bluesky/issues/1212
+        # regarding a potential motion issue with omega
+        with self.add_children_as_readables():
+            self.x = Motor(f"{prefix}X")
```
oliwenmandiamond: You're inheriting from XYStage but redefine it as x, this can be removed.

src/dodal/devices/i02_2/sample_motors.py
```diff
@@ -0,0 +1,20 @@
...
+            self.z = Motor(f"{prefix}Z")
```
oliwenmandiamond: z should be removed and instead the base class should be `XYZStage`

src/dodal/devices/i02_2/sample_motors.py
```diff
@@ -0,0 +1,20 @@
+from ophyd_async.epics.motor import Motor
```
DominicOram: Should: I don't think you're using this?

src/dodal/beamlines/i02_2.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    return Synchrotron()
+
+
+@device_factory()
+def sample_motors() -> XYStage:
+    """Get the i02-2 goniometer device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i02-2, it will return the existing object.
+    """
+    return XYStage(f"{PREFIX.beamline_prefix}-MO-GONIO-01:SAMPLE:")
```
oliwenmandiamond: This is returning `XYStage` and not `SampleMotors` defined in `dodal.devices.i02_2.sample_motors`?

oliwenmandiamond (changes requested):

src/dodal/devices/i02_2/sample_motors.py
olliesilvester: Sry, meant to delete that `sample_motors` file

src/dodal/devices/i02_2/sample_motors.py
olliesilvester: Sry, meant to delete that `sample_motors` file

PR #1644, opened 2025-10-20
Tidy up oav for i04

### LINKED ISSUE #1619: Murko: Tidy up OAV for i04
There are two issues with the OAVs on i04:
* There is a full screen and a ROI OAV. The fullscreen OAV still points to the `MJPG` stream, rather than the `XTAL` one that is actually where the full screen images is. Therefore taking an image with the full screen OAV would give the ROI one
* The `OAVToRedisForwarder` contains PV names for the fullscreen/ROI OAVs. Given we have ROI and full screen OAVs it should just have a reference to these. In this way we can have a reference to the `OAVToRedisForwarder` and ask it for metadata about the selected OAV (like the beam centre)

## Acceptance Criteria
* Taking an image with the full screen OAV will actually use the full screen
* I can get the metadata of the OAV that is being used out of the `OAVToRedisForwarder`

jacob720:
Fixes #1619 

Required by https://github.com/DiamondLightSource/mx-bluesky/pull/1383

1. Initialise OAV with different mjpeg prefixes
2. Confirm stream URL is updated accordingly.
3. Confirm new test makes sense and passes

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -44,7 +44,7 @@ from dodal.devices.scintillator import Scintillator
 from dodal.devices.smargon import Smargon
 from dodal.devices.synchrotron import Synchrotron
 from dodal.devices.thawer import Thawer
-from dodal.devices.undulator import Undulator
+from dodal.devices.undulator import UndulatorInKeV
 from dodal.devices.webcam import Webcam
 from dodal.devices.xbpm_feedback import XBPMFeedback
 from dodal.devices.xspress3.xspress3 import Xspress3
@@ -256,11 +256,11 @@ def synchrotron() -> Synchrotron:
 
 
 @device_factory()
-def undulator(daq_configuration_path: str | None = None) -> Undulator:
+def undulator(daq_configuration_path: str | None = None) -> UndulatorInKeV:
     """Get the i03 undulator device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i03, it will return the existing object.
     """
-    return Undulator(
+    return UndulatorInKeV(
         f"{BeamlinePrefix(BL).insertion_prefix}-MO-SERVC-01:",
         # evaluate here not as parameter default to enable post-import mocking
         id_gap_lookup_table_path=f"{daq_configuration_path or DAQ_CONFIGURATION_PATH}/lookup/BeamLine_Undulator_toGap.txt",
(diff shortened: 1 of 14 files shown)


src/dodal/devices/oav/oav_to_redis_forwarder.py
```diff
@@ -35,13 +37,10 @@ class Source(IntEnum):
...
-    def __init__(
-        self,
-        prefix: str,
-        oav_name: str,
-    ):
-        self.url = epics_signal_r(str, f"{prefix}MJPG_URL_RBV")
-        self.oav_name = oav_name
+    def __init__(self, oav: OAV, label: str):
+        self.url_ref = Reference(oav.grid_snapshot.url)
+        self.oav_ref = Reference(oav)
```
DominicOram: Could: `oav_ref` is never used, just the `url_ref` is so I would remove `oav_ref`

DominicOram (approved):
Minor comment but otherwise looks good. Thanks!

src/dodal/devices/oav/oav_to_redis_forwarder.py
DominicOram: Hang on, I'm wrong, this is used in the `mx-bluesky` side. Let me review that side too and then approve this

tests/devices/i04/test_murko_results.py
```diff
@@ -795,3 +799,30 @@ def test_results_with_tiny_x_pixel_value_are_filtered_out(
...
+        (10, 350, True, True),
+        (350, 10, False, True),
+        (350, 10, True, False),
+        (None, 10, True, False),
+        (None, 350, True, False),
+        (None, 10, False, False),
+        (None, 350, False, False),
+    ],
+)
+async def test_check_passes_and_fails_correctly_based_on_last_omega_and_stop_angle_and_inverted(
```
DominicOram: Nicely parametrized test, thanks!

tests/devices/i04/test_murko_results.py
```diff
@@ -21,6 +21,10 @@
 @patch("dodal.devices.i04.murko_results.StrictRedis")
 async def murko_results(mock_strict_redis: MagicMock) -> MurkoResultsDevice:
     murko_results = MurkoResultsDevice(name="murko_results")
+    murko_results.stop_angle.get_value = AsyncMock()
+    murko_results.stop_angle.get_value.return_value = 350
+    murko_results.invert_stop_angle.get_value = AsyncMock()
+    murko_results.invert_stop_angle.get_value.return_value = False
```
DominicOram: Could: If you make this as a mock device then all of this is done for you:
```suggestion
    with init_devices(mock=True):
        murko_results = MurkoResultsDevice(name="murko_results")
```
Then in your tests you can do `set_mock_value(murko_results.stop_angle, 100)`. Leaving this as a could because I know these signals are going to be removed shortly anyway but good to know for future.

Issue #1384, opened 2025-07-24
Create motor group ( on behalf of the i19 attenuation system )


CoePaul:
I19 have several motors which drive absorber sub-elements of a notional group which forms the beamline attenuator.

A natural first step in building up to a fully functional "Attenuator" is to merely create the motor group and ask it to move to a set of positions ( one set element per motor ).

## Acceptance Criteria
That an i19 optics blue API Device shall exist which 
* which receives via calls to "set" instructions as a dict of {string, float} or {string, int} pairings
   e.g. [ "x": 25.7, "y":-3.8, "w", 4 ]
* which applies the hutch access control protocols (initially defined for AccessControlledShutter)
* which - when access control permits - passes on approrpiate move position requests to the corresponding PVs

CoePaul:
With the completion of #1787 - this task is now complete

PR #1654, opened 2025-10-22
Add attenuator motor squad to i19 access controlled devices

CoePaul:
* Add attenuator motor squad class to represent a set of mutually cooperating motors which are not co-mounted ( therefore do not form a physical stage ); but rather act in concert

* Add initimately related position demand class used as driving argument to command motor squad to alter the posture of a beamline's transmission system

* Add tests for both new classes: AttenuatorMotorPositionDemands AttenuatorMotorSquad

* Adds access controlled subdirectory in the i19 device tests (to match the src devices file system layout)

* Moves established test class for access controlled I19 shutters into the access controlled subdirectory

DiamondLightSource/dodal#1384

1. Continues PR 1652 in a new branch
2. Confirm all tests pass 

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i19_1.py
+++ b/src/dodal/beamlines/i19_1.py
@@ -32,6 +32,8 @@ set_log_beamline(BL)
 set_utils_beamline(BL)
 
 
+I19_1_COMMISSIONING_INSTR_SESSION: str = "cm40638-5"
+
 I19_1_ZEBRA_MAPPING = ZebraMapping(
     outputs=ZebraTTLOutputs(TTL_PILATUS=1),
     sources=ZebraSources(),
@@ -45,7 +47,9 @@ DISPLAY_CONFIG = "/dls_sw/i19-1/software/daq_configuration/domain/display.config
 
 @device_factory()
 def attenuator_motor_squad() -> AttenuatorMotorSquad:
-    return AttenuatorMotorSquad(hutch=HutchState.EH1)
+    return AttenuatorMotorSquad(
+        hutch=HutchState.EH1, instrument_session=I19_1_COMMISSIONING_INSTR_SESSION
+    )
 
 
 # Needs to wait until enum is fixed on the beamline
@@ -74,7 +78,7 @@ def shutter() -> AccessControlledShutter:
     return AccessControlledShutter(
         prefix=f"{PREFIX.beamline_prefix}-PS-SHTR-01:",
         hutch=HutchState.EH1,
-        instrument_session="cm40638-4",
+        instrument_session=I19_1_COMMISSIONING_INSTR_SESSION,
     )
 
 
--- a/src/dodal/beamlines/i19_2.py
+++ b/src/dodal/beamlines/i19_2.py
@@ -46,6 +46,7 @@ set_path_provider(
     )
 )
 
+I19_2_COMMISSIONING_INSTR_SESSION: str = "cm40639-5"
 
 I19_2_ZEBRA_MAPPING = ZebraMapping(
     outputs=ZebraTTLOutputs(),
@@ -55,7 +56,9 @@ I19_2_ZEBRA_MAPPING = ZebraMapping(
 
 @device_factory()
 def attenuator_motor_squad() -> AttenuatorMotorSquad:
-    return AttenuatorMotorSquad(hutch=HutchState.EH2)
+    return AttenuatorMotorSquad(
+        hutch=HutchState.EH2, instrument_session=I19_2_COMMISSIONING_INSTR_SESSION
+    )
 
 
 @device_factory()
@@ -114,7 +117,7 @@ def shutter() -> AccessControlledShutter:
     return AccessControlledShutter(
         prefix=f"{PREFIX.beamline_prefix}-PS-SHTR-01:",
         hutch=HutchState.EH2,
-        instrument_session="cm40639-4",
+        instrument_session=I19_2_COMMISSIONING_INSTR_SESSION,
     )
 
 


src/dodal/devices/i19/access_controlled/attenuator_motor_squad.py
```diff
@@ -0,0 +1,65 @@
...
+class AttenuatorMotorPositionDemands(BaseModel):
+    continuous_demands: dict[PermittedKeyStr, float] = {}
+    indexed_demands: dict[PermittedKeyStr, PositiveInt] = {}
+
+    @model_validator(mode="after")
+    def no_keys_clash(self) -> Self:
+        common_key_filter = filter(
+            lambda k: k in self.continuous_demands, self.indexed_demands
+        )
+        common_key_count = sum(1 for _ in common_key_filter)
```
rtuck99: if you had done a for-comprehension instead of using `filter` then you could just do 
`if filtered_list:`

Alternatively, since you have an iterator, you could do
`next(common_key_filter, False)`, that way you don't have to iterate through the whole set (although this is not going to make any practical performance difference in this case).

src/dodal/devices/i19/access_controlled/blueapi_device.py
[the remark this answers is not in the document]
CoePaul: I think You're reviewing code which is already merged here - because it was too late to review it before?

rtuck99: If we fail to set the worker task as active, then this effectively returns and the `Status` will be successful

tests/devices/i19/access_controlled/test_attenuator_motor_squad.py
```diff
@@ -0,0 +1,128 @@
...
+@patch("dodal.devices.i19.access_controlled.blueapi_device.ClientSession.post")
+async def test_when_rest_post_unsuccessful_that_error_raised(
+    unsuccessful_post, invoking_hutch
+):
+    motors: AttenuatorMotorSquad = await given_a_squad_of_attenuator_motors(
+        invoking_hutch
+    )
+    with pytest.raises(ClientConnectionError):
+        restful_response: AsyncMock = given_an_unhappy_restful_response()
+        unsuccessful_post.return_value.__aenter__.return_value = restful_response
```
rtuck99: This test doesn't fail in the expected way because you haven't mocked the PUT following the POST.
`aiohttp` raises ClientConnectionError because it does actually attempt a connect, but the POST error status is never evaluated.

src/dodal/devices/i19/access_controlled/blueapi_device.py
rtuck99: This response HTTP status code is never checked

src/dodal/devices/i19/access_controlled/attenuator_motor_squad.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class AttenuatorMotorPositionDemands(BaseModel):
+    continuous_demands: dict[PermittedKeyStr, float] = {}
+    indexed_demands: dict[PermittedKeyStr, PositiveInt] = {}
+
+    @model_validator(mode="after")
+    def no_keys_clash(self) -> Self:
+        common_key_filter = filter(
+            lambda k: k in self.continuous_demands, self.indexed_demands
```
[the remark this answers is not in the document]
tpoliaw: I think it does do what it means to do, it's just confusingly formatted. `lambda k: k in self.continuous_demands, self.indexed_demands` are two arguments to the `filter` function, not a single `lambda` trying to look up a key in two dicts.

Either way it would be simpler to use `&` to get the common keys so the whole function becomes

```python
if common := self.continuous_demands.keys() & self.indexed_demands:
    raise ValueError(f"duplicate key(s) found: {common}")
return self
```

tests/devices/i19/access_controlled/test_attenuator_motor_squad.py
```diff
@@ -0,0 +1,128 @@
...
+    response_to_put = AsyncMock()
+    response_to_put.ok = False
+    restful_put.return_value.__aenter__.return_value = response_to_put
+
+    motors: AttenuatorMotorSquad = await given_a_squad_of_attenuator_motors(
+        invoking_hutch
+    )
+    position_demands: AttenuatorMotorPositionDemands = given_position_demands()
+
+    logger.error.assert_not_called()
```
rtuck99: The await should raise an exception instead of succeeding.

src/dodal/devices/i19/access_controlled/attenuator_motor_squad.py
```diff
@@ -0,0 +1,65 @@
...
+        if common_key_count < 1:
+            return self
+        else:
+            ks: str = "key" if common_key_count == 1 else "keys"
+            error_msg = (
+                f"{common_key_count} common {ks} found in distinct motor demands"
+            )
+            raise ValueError(error_msg)
+
+    def restful_format(self) -> dict[PermittedKeyStr, Any]:
```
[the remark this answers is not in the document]
CoePaul: Whatever it's called - I want it to be blatantly "Thing needed because REST needs something" 
rather than "What the heck is this for?"
and no I don't want the reader to have to know REST is involved prior to reading that code - I want the variable name to hint at that fact if the reader is new to the whole system and doesn't yet know this piece of code is building a packet to be sent by REST

tests/devices/i19/access_controlled/test_attenuator_position_demand.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def test_that_attenuator_position_demand_can_be_created_with_only_one_wedge():
+    wedge_position_demands = {"x": 0.5}
+    wheel_position_demands = {}
+    position_demand = AttenuatorMotorPositionDemands(
+        continuous_demands=wedge_position_demands,
+        indexed_demands=wheel_position_demands,
+    )
+    assert position_demand is not None
```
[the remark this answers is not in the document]
CoePaul: Tests don't just test - that is a side function 
the primary purpose of tests is documenting actual behaviour - 
parameterising tends to obfuscate that function - it can be used when You're just hammering the same nail over and over ( but that strikes me as a waste of testing time in general ) 
Different scenarios should be indicated-as-having-been-thought-about-distinctly not by being folded and knotted up into the laundry basket of a scruffy parameterised bundle 
but should instead be carefully laid out for the reader one well named test at a time 

if You want to know what code does, reading the test function names should tell You everything 
one bullet point at a time 
it also makes clear what the developers haven't thought about 

testing is for the CIs benefit and is a very important but far inferior aspect IMHO

src/dodal/devices/i19/access_controlled/attenuator_motor_squad.py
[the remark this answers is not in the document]
tpoliaw: But naming it after what it does also explains what it does. `restful_format` explains what it's used for but you need to read the code to figure out what it does.

src/dodal/devices/i19/access_controlled/blueapi_device.py
```diff
@@ -29,11 +29,18 @@ class OpticsBlueAPIDevice(StandardReadable, Movable[D]):
...
+    def __init__(
+        self, hutch: HutchState, instrument_session: str = "", name: str = ""
+    ) -> None:
+        self.hutch_request = hutch
+        self.instrument_session = instrument_session
         self.url = OPTICS_BLUEAPI_URL
         self.headers = HEADERS
         super().__init__(name)
 
+    def _get_invoking_hutch(self) -> str:
```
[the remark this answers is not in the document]
CoePaul: If I read / understand notes on property annotation / decoration of functions - it's primary purpose seems to be to guard public facing methods with the effect of invisible getter and setter methods ( and a free del while we're at it ).

Here the getter itself is not supposed to be public.  Is it still obvious that property annotation covers that scenario?

src/dodal/devices/i19/access_controlled/blueapi_device.py
noemifrisina: Yeah, as weird as that may be, that was on purpose.  We usually handle the error in the case the plan didn't run on the plan/gda side as needed (mainly for testing reasons).

src/dodal/devices/i19/access_controlled/attenuator_motor_squad.py
```diff
@@ -0,0 +1,61 @@
...
+        else:
+            ks: str = "key" if common_key_count == 1 else "keys"
+            error_msg = f"Common {ks} found in distinct motor demands: {common_keys}"
+            raise ValueError(error_msg)
+
+    def validated_complete_demand(self) -> dict[PermittedKeyStr, Any]:
+        return self.continuous_demands | self.indexed_demands
+
+
+class AttenuatorMotorSquad(OpticsBlueAPIDevice):
```
noemifrisina: Must: this device still needs to be instantiated in i19_1 and i19_2 so that the respective blueapi instances can pick it up

noemifrisina (changes requested):
Thanks, looks like one way or another the comments have been addressed (some of them by discussion in the office). Just one more thing that's missing...

PR #1655, opened 2025-10-23
Add Slits with Individual blade control. Add those devices to i22

RJCD-Diamond:
This allows slits to either controlled via x_gap / y_gap nomenclature or x_plus / x_minus. Apparently necessary on i22 and requested


DominicOram (commented):
Could: Did you try just adding these to the `Slits` and running `dodal connect` to see what broke? I think most of the slits on my beamlines have these PVs too and I don't see any harm in adding them across the board

EmsArnold:
From a quick look at current the current synoptics, it seems like 
``` 
x_minus: str = "X:MINUS", x_plus: str = "X:PLUS", y_minus: str = "Y:MINUS", y_plus: str = "Y:PLUS"
```
is most widely in use, with some other naming conventions having a few instruments using them: 

- inboard/outboard/bottom/top
- ring/hall/minus/plus
- neg/pos/neg/pos

and a few with only a single instrument:

- 2/1/2/1
- nearside/offside/bottom/top
- nearside/offside/lower/upper

DominicOram (approved):
I'm going to get this merged as I think it's good.

> This is the 2nd time we making slit blade for slits, should we think about moving it to common and standardise what it call?

From @EmsArnold comment it sounds like the de facto standard is +/- and i10 is the odd one out. It would be good if i10 could follow the same standard but I will leave you to discuss that with the scientists

PR #1660, opened 2025-10-23
i09 energy to gap function

Villtord:
Fixes https://github.com/DiamondLightSource/sm-bluesky/issues/211

1. Make sure tests pass

### Checks for reviewer


oliwenmandiamond (commented):
Looks good. However, shouldn't `hard_undulator_functions.py` live inside `i09_1_shared` as `i09` will also being using it? This way were have a clear distinction on what devices are specific to i09-1 and shared with i09 and i09-1

PR #1672, opened 2025-10-28
1663 i09 energy

### LINKED ISSUE #1663: i09-2 energy
# Goals
Acceptance criteria :
- [ ] To be able to control pgm and apple2 devices via set.
- [ ] To be able to read and set polarisation.  
- [ ] To perform energy scan with both pgm and insertion devices.
- [ ] A set of dodal devices are needed that acheve the goal. 

### Tasks
- [x] Create apple2 device.
- [x] Create an energy to motor (ETM) position function. 
- [ ] Use the apple2 and the ETM function to make a apple2 controller that control energy and polarisation.
- [ ] Using the controller to make a id energy and polarisation device.
- [ ] Combine the id energy and PGM to from beam energy device to control the beamline energy.
- [ ] Add all these device to beamline config.

Relm-Arrowny:
Fixes #1663 
- Added  `_generate_phase_lut` to `EnergyMotorLookup` to enable generation of phase lookup table for i09-2/i21 ID.
- Moved paths into lookuptableConfig as optional field.
- Introduced J09Apple2Controller to manage Apple II polarization and the conversion of energy to motor movement.
- Refactored test setup by adding conftest.py to both i10 and i09_2_share folder to reuse fixtures.
- Updated the i09_2 configuration to include devices for beam energy, ID energy, and polarization.

1. Check tests cover use cases.
2. run tests
3. Confirm tests pass.

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -13,6 +13,7 @@ from dodal.devices.aperturescatterguard import (
 )
 from dodal.devices.attenuator.attenuator import BinaryFilterAttenuator
 from dodal.devices.backlight import Backlight
+from dodal.devices.baton import Baton
 from dodal.devices.detector import DetectorParams
 from dodal.devices.detector.detector_motion import DetectorMotion
 from dodal.devices.diamond_filter import DiamondFilter, I04Filters
@@ -143,11 +144,19 @@ def transfocator() -> Transfocator:
 
 
 @device_factory()
+def baton() -> Baton:
+    """Get the i04 baton device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return Baton(f"{PREFIX.beamline_prefix}-CS-BATON-01:")
+
+
+@device_factory()
 def xbpm_feedback() -> XBPMFeedback:
     """Get the i04 xbpm_feedback device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i04, it will return the existing object.
     """
-    return XBPMFeedback(f"{PREFIX.beamline_prefix}-EA-FDBK-01:")
+    return XBPMFeedback(f"{PREFIX.beamline_prefix}-EA-FDBK-01:", baton=baton())
 
 
 @device_factory()
@@ -236,6 +245,7 @@ def undulator() -> UndulatorInKeV:
     return UndulatorInKeV(
         prefix=f"{PREFIX.insertion_prefix}-MO-SERVC-01:",
         id_gap_lookup_table_path="/dls_sw/i04/software/daq_configuration/lookup/BeamLine_Undulator_toGap.txt",
+        baton=baton(),
     )
 
 


src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -71,6 +82,8 @@ class LookupTableConfig(BaseModel):
     max_energy: str = "MaxEnergy"
     poly_deg: list[str] = Field(default_factory=lambda: DEFAULT_POLY_DEG)
     mode_name_convert: dict[str, str] = Field(default_factory=lambda: MODE_NAME_CONVERT)
+    gap_path: Path | None = None
+    phase_path: Path | None = None
```
oliwenmandiamond: Should: I don't think we should be adding back the `gap_path` and `phase_path` to the `LookupTableConfig`. As discussed before, we pass this into `convert_csv_to_lookup` but we also pass in the file contents, after the file path has already been used and read. 

It makes the design confusing as were not using all of the parameters from `LookupTableConfig` and we're not sure which file contents we're looking at. It makes much more sense to decouple them or make LookupTableConfig take a single path and then we provide the ID's two `LookupTableConfig`, one for gap and one for phase.

src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -357,10 +370,30 @@ def update_lookuptables(self):
...
+                    generate_lookup_table_entry(
+                        min_energy=self.lookup_tables.gap.root[
+                            Pol(key.lower())
+                        ].limit.minimum,
+                        max_energy=self.lookup_tables.gap.root[
+                            Pol(key.lower())
+                        ].limit.maximum,
+                        poly1d_param=(PhasePoly1dParameters[Pol(key.lower())]),
+                    )
+                )
```
oliwenmandiamond:
```suggestion
        for key in self.lookup_tables.gap.root.keys():
            if key is not None:
                self.lookup_tables.phase.root[key] = (
                    generate_lookup_table_entry(
                        min_energy=self.lookup_tables.gap.root[
                            key
                        ].limit.minimum,
                        max_energy=self.lookup_tables.gap.root[
                            key
                        ].limit.maximum,
                        poly1d_param=(PhasePoly1dParameters[key]),
                    )
                )
```


I personally think this is overcomplicating the standard `EnergyMotorLookup` class. If we need specialised behaviour for the `EnergyMotorLookup`, we should be creating another class that extends / overrides the logic. 

So for i09's case, we should create J09EnergyMotorLookup which overrides `_update_phase_lut` to provide the custom phase values. 

I don't even think we need to do this either though because we should be able to just add these limits inside a lookup table file like the rest? (correct me if I'm wrong).

Relm-Arrowny: I think it is ok to establish a default phase lookup table within the base class, particularly since this default value is functional for all Apple II  and most beamlines currently lack their own phase lookup table.

tests/devices/i10/test_i10_apple2.py
```diff
@@ -181,17 +115,7 @@ async def mock_id_controller(
...
-    set_mock_value(mock_id_controller.apple2().phase().gate, UndulatorGateStatus.CLOSE)
-    set_mock_value(
-        mock_id_controller.apple2().jaw_phase().gate, UndulatorGateStatus.CLOSE
-    )
-    set_mock_value(mock_id_controller.apple2().gap().velocity, 1)
-    set_mock_value(mock_id_controller.apple2().jaw_phase().jaw_phase.velocity, 1)
-    set_mock_value(mock_id_controller.apple2().phase().btm_inner.velocity, 1)
-    set_mock_value(mock_id_controller.apple2().phase().top_inner.velocity, 1)
-    set_mock_value(mock_id_controller.apple2().phase().btm_outer.velocity, 1)
-    set_mock_value(mock_id_controller.apple2().phase().top_outer.velocity, 1)
```
oliwenmandiamond: Why is this removed?

src/dodal/beamlines/i10_optics.py
```diff
@@ -125,7 +123,9 @@ def idd_controller() -> I10Apple2Controller:
     """I10 downstream insertion device controller."""
     idd_energy_motor_lut = EnergyMotorLookup(
         config_client=I10_CONF_CLIENT,
-        lut_config=LookupTableConfig(source=("Source", "idd")),
+        lut_config=LookupTableConfig(
+            source=("Source", "idd"),
+        ),
```
oliwenmandiamond: Undo this format change

src/dodal/beamlines/i10_optics.py
```diff
@@ -190,7 +190,9 @@ def idu_controller() -> I10Apple2Controller:
     """I10 upstream insertion device controller."""
     idu_energy_motor_lut = EnergyMotorLookup(
         config_client=I10_CONF_CLIENT,
-        lut_config=LookupTableConfig(source=("Source", "idu")),
+        lut_config=LookupTableConfig(
+            source=("Source", "idu"),
+        ),
```
oliwenmandiamond: Undo this format change

src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -321,11 +332,11 @@ def __init__(
         phase_path:
             File path to the phase lookup table.
         """
+        self.gap_path = gap_path
+        self.phase_path = phase_path
```
oliwenmandiamond: Please undo this format change so it is last again

tests/devices/i10/conftest.py
```diff
@@ -0,0 +1,8 @@
+from tests.devices.test_apple2_undulator import (
+    mock_id_gap,
+    mock_jaw_phase,
+    mock_phase_axes,
+)
```
oliwenmandiamond: This isn't the correct way to do this, you should not import pytest fixtures directly. They are designed to be made discoverable via defining them in a higher level `conftest.py` file or via `pytest_plugins`.

As this may require some restructuring and is a little outside the scope of this change, I would suggest revert this by removing this + undo the removing of mock_id_gap. mock_jaw_phase, and mock_config_client and do this in a separate change to make it easier to review

tests/devices/i10/test_i10_apple2.py
```diff
@@ -105,35 +64,6 @@ async def mock_pgm(prefix: str = "BLXX-EA-DET-007:") -> PlaneGratingMonochromato
...
-
-    def my_side_effect(file_path, reset_cached_result) -> str:
-        assert reset_cached_result is True
-        with open(file_path) as f:
-            return f.read()
-
-    mock_config_client.get_file_contents.side_effect = my_side_effect
-    return mock_config_client
-
-
```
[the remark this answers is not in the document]
oliwenmandiamond: Please see my comment above about reverting this for now

oliwenmandiamond (changes requested):
Almost there. There's some formatting changes that need to be addressed and also some changes that are outside the scope of this one

tests/devices/i10/test_i10_apple2.py
```diff
@@ -54,46 +54,10 @@
...
-    set_mock_value(mock_phase_axes.top_outer.user_setpoint_readback, 0)
-    set_mock_value(mock_phase_axes.top_inner.user_setpoint_readback, 0)
-    set_mock_value(mock_phase_axes.btm_outer.user_setpoint_readback, 0)
-    set_mock_value(mock_phase_axes.btm_inner.user_setpoint_readback, 0)
-    set_mock_value(mock_phase_axes.status, EnabledDisabledUpper.ENABLED)
-    return mock_phase_axes
+pytest_plugins = [
+    "tests.devices.util.test_lookup_tables_apple2",
+    "tests.devices.test_apple2_undulator",
+]
```
oliwenmandiamond: Can you just add a comment which fixtures you are after for this test? I've just checked and we can't add individual fixtures and it has to be the whole module rather than fixture for a plugin. I dislike this as you saying the whole test is a plugin which isn't correct, we should separate this out so we have a clear distinction between re usable fixtures. Can be this change or just link a ticket to do it as another bit of work

PR #1679, opened 2025-10-29
Create apple2 energy motor look up

Relm-Arrowny:
Part of #1663 

- Extracted all the shared functionality in I10EnergyMotorLookup to EnergyMotorLookup. 
- Make phase file optional and added helper functions to generate phase lookuptable from given parameters. 

## Instructions to reviewer on how to test:
1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -0,0 +1,378 @@
...
+from daq_config_server.client import ConfigServer
+from pydantic import BaseModel, ConfigDict, RootModel
+
+from dodal.devices.apple2_undulator import Pol
+
+
+@dataclass
+class LookupPath:
+    Gap: Path | None
+    Phase: Path | None
```
oliwenmandiamond:
```suggestion
@dataclass
class LookupPath:
    gap: Path | None
    phase: Path | None
```

oliwenmandiamond (commented):
Looks good, but you need to update your `BaseModels` to use `snake_case`.

I think you expect them to come in as captials though right? You can do this by adding an alias e.g

```Python
from pydantic import Field
...
limit = Field(alias="Limit")
```

You also have many constants littered every where in the code e.g 

"Limit"
"Minimum"
"Maximum"
"Low"
"High"
"Energies"
"Poly"

I would suggest maybe defining these as constants

so you can then do this

```Python
limit = Field(alias=LIMIT)
```

src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -0,0 +1,378 @@
...
+    root: dict[str, LookupTableEntries]
+
+
+def convert_csv_to_lookup(
+    file: str,
+    source: tuple[str, str] | None = None,
+    mode: str | None = "Mode",
+    mode_name_convert: dict[str, str] | None = None,
+    min_energy: str | None = "MinEnergy",
+    max_energy: str | None = "MaxEnergy",
```
oliwenmandiamond: Does this need to be changed to "min_energy" and "max_energy"?

Relm-Arrowny: no those are the column name/str in the actual lookuptable

PR #1682, opened 2025-10-30
1680 fix run engine filehandle leak

### LINKED ISSUE #1680: RunEngine is causing file handler leak for tests
When running the tests in dodal, it is possible to get this error:

```
File "/scratch/wto16239/github/dodal/src/dodal/testing/fixtures/run_engine.py", line 28,
in run_engine
yield RunEngine ()
File "/scratch/wto16239/github/i19-bluesky/.venv/lib/python3.11/site-packages/bluesky/run_engine.py", line 417, in _init
loop = asyncio.new event loop()
File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/asyncio/events.py", line 806, in new event_loop
eturn get event loop policy) .new event loop()
File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/asyncio/events.py", line 695, in new event_loop return self.
loop factory)
File"/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/asyncio/unix_events.py", line 64, in init File"/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/asyncio/selector_events.py", line 56,
self. make self pipe()
File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/asyncio/selector_events.py", line 107, in _make_self_pipe
self._ssock, self._csock = socket. socketpair()
File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/socket.py", line 608, in socketpair a, b =
OSError: [Errno 24] Too many open files
```
However, this is only appears to happen on certain work station and not others. 

Examining the above further, it seems RunEngine tries to be created which tries to create the event loop which tries to create the socket which then fails.

I finally managed to replicate this error myself when forcefully limiting the number of open files possible in my environment  

My current settings
```
ulimit -n
262144
```

New settings
```
(venv) root ➜ /workspaces/dodal (main) 
$ ulimit -n 100
```
Then running all dodal tests results in the error above. 

To prove this is the run_engine causing this issue, I added this to pytests to track the number of open files

```Python
@pytest.hookimpl()
def pytest_sessionfinish(session, exitstatus):
    """
    Called after the whole pytest run (all tests in the session).
    """
    import os
    import resource

    pid = os.getpid()
    open_files = len(os.listdir(f"/proc/{pid}/fd"))
    soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)

    print(f"Open files: {open_files}/{soft} (soft limit)")
```

I then ran created some tests

```python
def test():
    pass
```
```
tests/devices/electron_analyser/specs/test_detector.py::test PASSED      [100%]
Open files: 39/262144 (soft limit)
```
The 39 is from files I already have open on my computer

```Python
def test_my_run_engine(run_engine):
    pass
```
Increased when calling run_engine fixture
```
tests/devices/electron_analyser/specs/test_detector.py::test_my_run_engine PASSED [100%]
Open files: 42/262144 (soft limit)
```

Then when calling a subsection of tests e.g 

tests/devices/electron_analyser
Open files: 171/262144 (soft limit)

all tests
Open files: 1233/262144 (soft limit)

It is increasing more and more which relates to the number of tests need a fresh RunEngine, causing a memory leak as the resources are never freed!

## Acceptance Criteria

- The run_engine fixture is rewritten to make sure it closes the event loop and socket between RunEngine instances
- A test is written which calls the run_engine fixture multiple times and checks at each stage the number of open files stays the same to prevent memory leaks.

rtuck99:
Fixes #1680 

* Requires #1681

* Properly closes the run engine event loop after use
* Fixes a number of places where the run engine was constructed instead of using the fixture
* Patches usage of the run engine in tests where production code created a run engine directly, in order that these tests don't leak
* Cleans up watcher utils tests where a signal observer was left dangling
* Fixture that can be enabled to locate a leaking test

1. Tests pass with `ulimit -n 100`

### Checks for reviewer

### DIFF
--- a/src/dodal/testing/fixtures/run_engine.py
+++ b/src/dodal/testing/fixtures/run_engine.py
@@ -6,7 +6,7 @@ import asyncio
 import os
 import threading
 import time
-from collections.abc import Generator, Mapping
+from collections.abc import AsyncGenerator, Mapping
 
 import pytest
 import pytest_asyncio
@@ -17,7 +17,7 @@ from bluesky.simulators import RunEngineSimulator
 _run_engine = RunEngine()
 
 _ENABLE_FILEHANDLE_LEAK_CHECKS = (
-    os.getenv("ENABLE_FILEHANDLE_LEAK_CHECKS", "").lower() == "true"
+    os.getenv("DODAL_ENABLE_FILEHANDLE_LEAK_CHECKS", "").lower() == "true"
 )
 
 
@@ -33,7 +33,7 @@ async def _ensure_running_bluesky_event_loop(_global_run_engine):
 
 
 @pytest.fixture()
-async def run_engine(_global_run_engine: RunEngine) -> Generator[RunEngine, None, None]:
+async def run_engine(_global_run_engine: RunEngine) -> AsyncGenerator[RunEngine, None]:
     try:
         yield _global_run_engine
     finally:
@@ -41,7 +41,7 @@ async def run_engine(_global_run_engine: RunEngine) -> Generator[RunEngine, None
 
 
 @pytest_asyncio.fixture(scope="session", loop_scope="session")
-async def _global_run_engine() -> Generator[RunEngine, None, None]:
+async def _global_run_engine() -> AsyncGenerator[RunEngine, None]:
     """
     Obtain a run engine, with its own event loop and thread.
 


src/dodal/testing/fixtures/run_engine.py
```diff
@@ -3,32 +3,70 @@
...
+
+    On closure of the scope, the run engine is stopped and the event loop closed
+    in order to release all resources it consumes.
+    """
+    run_engine = RunEngine({}, call_returns_result=True)
+    yield run_engine
+    try:
+        run_engine.halt()
+    except Exception as e:
+        print(f"Got exception while halting RunEngine {e}")
```
oliwenmandiamond: I personally think just do 

```suggestion
    run_engine.halt()
```
If there's an error halting, we shouldn't just print it, we should let it fail naturally and let test crash so very obvious something has gone wrong

tests/common/test_watcher_utils.py
```diff
@@ -61,6 +61,7 @@ async def test_log_on_percentage_complete_value_error_on_bad_input():
         match="Percent interval on class _LogOnPercentageProgressWatcher must be a positive number, but received 0",
     ):
         log_on_percentage_complete(status, "", 0)
+    test_watchable.complete_event.set()
```
oliwenmandiamond: Can you please add comment on why this is needed?

tests/common/test_watcher_utils.py
```diff
@@ -75,3 +76,4 @@ async def do_log():
         log_on_percentage_complete(status, "")
 
     await asyncio.gather(update_signal(), do_log())
+    test_watchable.complete_event.set()
```
oliwenmandiamond: Same here

tests/devices/test_gridscan.py
```diff
@@ -139,14 +139,12 @@ async def test_waits_for_running_motion(grid_scan: FastGridScanCommon):
     ],
 )
 async def test_given_different_step_numbers_then_expected_images_correct(
-    zebra_fast_grid_scan: ZebraFastGridScanThreeD, steps, expected_images
+    run_engine, zebra_fast_grid_scan: ZebraFastGridScanThreeD, steps, expected_images
```
oliwenmandiamond: Not directly related, but could you add the types for steps and expected_images?

tests/test_cli.py
```diff
@@ -19,6 +19,12 @@
 EXAMPLE_BEAMLINE = "i22"
 
 
+@pytest.fixture(autouse=True)
+def patch_run_engine_in_cli_to_avoid_leaks(run_engine):
+    with patch("dodal.cli.RunEngine", return_value=run_engine):
+        yield
```
oliwenmandiamond:
```suggestion
@pytest.fixture(autouse=True)
def patch_run_engine_in_cli_to_avoid_leaks(run_engine: RunEngine):
    with patch("dodal.cli.RunEngine", return_value=run_engine):
        yield
```

tests/conftest.py
```diff
@@ -79,3 +80,18 @@ async def baton_in_commissioning_mode() -> AsyncGenerator[Baton]:
...
+def check_for_filehandle_leaks(request: FixtureRequest):
+    pid = os.getpid()
+    _baseline_n_open_files = len(os.listdir(f"/proc/{pid}/fd"))
+    try:
+        yield
+    finally:
+        _n_open_files = len(os.listdir(f"/proc/{pid}/fd"))
+        assert _n_open_files == _baseline_n_open_files, (
+            f"Function {request.function.__name__} leaked some filehandles"
+        )
```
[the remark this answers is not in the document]
rtuck99: I will move it to dodal fixtures. 
It's commented out because normally you don't want to turn it on due to the big performance hit. But when you do want it to be used, you want it everywhere because the thing you are trying to find was allocated somewhere in the test suite but you don't know where.
I will add some kind of comment to this effect.

src/dodal/testing/fixtures/run_engine.py
```diff
@@ -3,32 +3,73 @@
...
         if time.monotonic() > timeout:
             raise TimeoutError("This really shouldn't happen but just in case...")
 
 
 @pytest.fixture()
-def run_engine():
-    global _run_engine
-    _run_engine.reset()
-    return _run_engine
+async def run_engine(_global_run_engine: RunEngine):
```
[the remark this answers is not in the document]
rtuck99: The return type is Generator[RunEngine, None, None]

tests/testing/fixtures/test_run_engine.py
```diff
@@ -31,3 +32,27 @@ def test_run_engine_fixture_has_no_file_handler_leak(
                 f"{_baseline_n_open_files} to {n_open_files} when calling the "
                 "run_engine fixture",
             ) from exc
+
+
+@pytest.fixture(
+    # autouse=True
+)
```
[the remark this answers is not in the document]
rtuck99: I had to move it as well because it was accidentally in the wrong file

Issue #1498, opened 2025-09-02
Create wrapped versions of all standard bluesky plan and step_scan version of scan


RJCD-Diamond:
In the folder:

[dodal](https://github.com/DiamondLightSource/dodal/tree/main)/[src](https://github.com/DiamondLightSource/dodal/tree/main/src)/[dodal](https://github.com/DiamondLightSource/dodal/tree/main/src/dodal)/plans/

We have a wrapped.py which contains a wrapped version of the standard bluesky count plan. In bluesky we have lots of additional standard plans such as:

scan
rel_scan
list_scan
rel_list_scan

etc. These are unable to be used in blueapi, due to type hinting issues. We should create new, wrapped and type hinted versions of all the standard bluesky plans and ones that also reimplement gda style start stop step scans

RJCD-Diamond:
@EmsArnold Perhaps a good one to start with?

PR #1683, opened 2025-10-30
Add wrapped list_scan

EmsArnold:
Part of #1498

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/plans/wrapped.py
+++ b/src/dodal/plans/wrapped.py
@@ -55,3 +55,28 @@ def count(
     metadata = metadata or {}
     metadata["shape"] = (num,)
     yield from bp.count(tuple(detectors), num, delay=delay, md=metadata)
+
+
+@attach_data_session_metadata_decorator()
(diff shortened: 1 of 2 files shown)


src/dodal/plans/wrapped.py
```diff
@@ -55,3 +55,28 @@ def count(
     metadata = metadata or {}
     metadata["shape"] = (num,)
     yield from bp.count(tuple(detectors), num, delay=delay, md=metadata)
+
+
+@attach_data_session_metadata_decorator()
+@validate_call(config={"arbitrary_types_allowed": True})
+def list_scan(
+    detectors: Annotated[
+        set[Readable] | list[Readable],
```
RJCD-Diamond: Use: from collections.abc import Iterable instead of set, and also change it in the count plan above.

sets can't be serialised, but lists and tuples can, so using the generic one will be best.

Also make it Readable or AsyncReadable. Then it covers our ophyd_async devices like the StandardDetector class too

PR #1689, opened 2025-10-31
Make sure fast grid scan parameters are applied

DominicOram:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/1360

The fast grid scan parameters get rounded on the readback value so we need to check that they get set with a tolerance. Additionally, sets the parameters on the gridscan asynchronously.

1. Confirm tests pass

### Checks for reviewer


DominicOram:
Tests fixed in https://github.com/DiamondLightSource/dodal/pull/1693

src/dodal/devices/fast_grid_scan.py
```diff
@@ -307,13 +309,24 @@ async def prepare(self, value: ParamType):
...
 
         LOGGER.info("Applying gridscan parameters...")
+
         # Create arguments for bps.mv
         for key, signal in self._movable_params.items():
             param_value = value.__dict__[key]
-            set_statuses.append(await set_and_wait_for_value(signal, param_value))  # type: ignore
+
+            matcher = partial(isclose, param_value, abs_tol=0.001)
+            matcher.__name__ = "does_parameter_match"  # type: ignore # Remove when https://github.com/bluesky/ophyd-async/pull/1123 deployed
```
rtuck99: Looks like this is now merged so can remove

PR #1690, opened 2025-11-03
Add wrapped plan stubs to init

EmsArnold:
Part of #1498

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/plan_stubs/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .wrapped import move, move_relative, set_absolute, set_relative, sleep, wait
+
+__all__ = ["set_absolute", "set_relative", "move", "move_relative", "sleep", "wait"]
```
tpoliaw: nit: Could the order here match the order in the imports to make it obvious everything is included?

PR #1695, opened 2025-11-05
Add devices required for Hyperion beamstop check

rtuck99:
Add gain control to IPin device
Add `OUT` position to the beamstop

Required for  
* DiamondLightSource/mx-bluesky#351

1. `dodal connect i03` shows devices as connected

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -461,7 +461,7 @@ def collimation_table() -> CollimationTable:
 
 @device_factory()
 def ipin() -> IPin:
-    """Get the i04 ipin device, instantiate it if it hasn't already been.
+    """Get the i03 ipin device, instantiate it if it hasn't already been.
     If this is called when already instantiated in i04, it will return the existing object.
     """
     return IPin(f"{PREFIX.beamline_prefix}-EA-PIN-01:")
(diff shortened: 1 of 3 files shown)


src/dodal/beamlines/i03.py
```diff
@@ -456,3 +457,11 @@ def collimation_table() -> CollimationTable:
     If this is called when already instantiated in i03, it will return the existing object.
     """
     return CollimationTable(prefix=f"{PREFIX.beamline_prefix}-MO-TABLE-01")
+
+
+@device_factory()
+def ipin() -> IPin:
+    """Get the i04 ipin device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
```
DominicOram: Should: I know these docstrings are rubbish but this is i03, not i04

src/dodal/devices/mx_phase1/beamstop.py
```diff
@@ -28,6 +30,7 @@ class BeamstopPositions(StrictEnum):
     """
 
     DATA_COLLECTION = "Data Collection"
+    OUT = "Out"
```
DominicOram: Nit: For the aperture/scatterguard we've used `OUT_OF_BEAM` might be good to use the same convention here

DominicOram (changes requested):
Thanks, mostly minor comments

src/dodal/devices/mx_phase1/beamstop.py
DominicOram: Yh, I'd be up for a common set if we can think of ones that work.

PR #1707, opened 2025-11-13
Update ID lookup with type checking

oliwenmandiamond:
Improves type checking across the id lookup table code. Now uses the pydantic models rather unchecked dictionaries so it is statically typed and much easier to read.  

1. Check tests pass
2. Implementation makes sense

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/apple2_undulator.py
+++ b/src/dodal/devices/apple2_undulator.py
@@ -29,6 +29,7 @@ from dodal.log import LOGGER
 T = TypeVar("T")
 
 DEFAULT_MOTOR_MIN_TIMEOUT = 10
+MAXIMUM_MOVE_TIME = 550  # There is no useful movements take longer than this.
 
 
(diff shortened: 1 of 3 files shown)


src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -77,127 +104,113 @@ class EnergyMinMax(BaseModel):
...
 
 class LookupTableEntries(BaseModel):
     energies: EnergyCoverage
     limit: EnergyMinMax
 
 
-class Lookuptable(RootModel):
-    root: dict[str, LookupTableEntries]
+class LookupTable(RootModel[dict[Pol, LookupTableEntries]]):
+    # Allow to auto speficy a dict if one not provided
```
Relm-Arrowny:
```suggestion
    # Allow to auto specify a dict if one not provided
```

PR #1714, opened 2025-11-17
Replace i10 ID tests that use pickle files with json

oliwenmandiamond:
Updates test files to json so they are human readable and easier to debug

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


Relm-Arrowny (approved):
This look fine, but I wonder if we we can just store the poly parameter rather than poly1d object and remove those serializaer completely, probably an other time.

oliwenmandiamond:
Yes, using a list directly in the model would make this simpler. I will try and do this in another change

PR #1723, opened 2025-11-19
Create a device that will give the max pixel from an AD camera

### LINKED ISSUE #1699: Max Pixel device
As part of https://github.com/DiamondLightSource/mx-bluesky/issues/1318 we need an ophyd_async device that we can use to get the max pixel value from an optical image.

This device should let us do something like:
```python
with init_devices():
    max_pixel_device = MaxPixelDevice("MY_OAV_PREFIX")

await max_pixel_device.trigger()
max_intensity = await max_pixel_device.max_pixel.get_value()
```

https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/oav/pin_image_recognition/__init__.py does something very similar to this, we should start from there.

## Acceptance Criteria
* There is a device I can use to get the max pixel value from an OAV stream

srishtysajeev:
Fixes #1699

- Get tests to work

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/i04/max_pixel.py
+++ b/src/dodal/devices/i04/max_pixel.py
@@ -33,6 +33,6 @@ class MaxPixel(StandardReadable, Triggerable):
     @AsyncStatus.wrap
     async def trigger(self):
         arr = await self._convert_to_gray_and_blur()
-        max_val = np.max(arr)
-        assert isinstance(max_val, int)
+        max_val = float(np.max(arr))  # np.int64
+        assert isinstance(max_val, float)
         self._max_val_setter(max_val)
--- a/tests/devices/i04/test_max_pixel.py
+++ b/tests/devices/i04/test_max_pixel.py
@@ -18,7 +18,12 @@ async def max_pixel() -> AsyncGenerator[MaxPixel]:
 
 @pytest.mark.parametrize(
     "preprocessed_data, expected",
-    [([1, 2, 3], 3), ([[0, 0, 0], [0, 0, 0], [0, 0, 0]], 0), ([-9, -8, 0], 0)],
+    [
+        ([1, 2, 3], 3),  # test can handle standard input
+        ([[0, 0, 0], [0, 0, 0], [0, 0, 0]], 0),  # test can handle all 0's
+        ([-9, -8, 0], 0),  # test can handle negatives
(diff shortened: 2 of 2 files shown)


tests/devices/i04/test_max_pixel.py
```diff
@@ -0,0 +1,47 @@
...
+    device.array_data = cv2.cvtColor(  # type: ignore
+        np.stack([mock_array] * 3, axis=-1),  # make it 3-channel BGR
+        cv2.COLOR_BGR2GRAY,
+    )
+    blurred = device.preprocessed_data()
+    assert isinstance(blurred, np.ndarray)
+    assert blurred.shape == mock_array.shape
+
+
+def test_trigger_sets_max_pixel_val():
```
DominicOram: Should: This should reuse the fixture

tests/devices/i04/test_max_pixel.py
```diff
@@ -0,0 +1,47 @@
...
+    # Create a simple mock array
+    mock_array = np.array(
+        [[0, 29, 60, 80], [100, 100, 110, 200], [100, 100, 110, 200], [0, 50, 50, 50]],
+        dtype=np.uint8,
+    )
+
+    # Make it 3-channel so cv2.COLOR_BGR2GRAY works
+    mock_image = np.stack([mock_array] * 3, axis=-1)
```
DominicOram: Should: We should check if the data coming off the hardware is 3 channel. If it is then the mocked read data should be 3 channel too, if not then this conversion needs to happen in the device

DominicOram (changes requested):
Thanks, some comments in code. I will try and get us some example data for the tests

src/dodal/devices/i04/max_pixel.py
```diff
@@ -0,0 +1,35 @@
...
+    epics_signal_r,
+)
+
+# kernal size describes how many of the neigbouring pixels are used for the blur,
+# higher kernal size means more of a blur effect
+KERNAL_SIZE = (7, 7)
+
+
+class MaxPixel(StandardReadable, Triggerable):
+    """Gets the max pixel from the image"""
```
DominicOram: Could: Would be good to mention the blur here, and why you do it.

DominicOram (changes requested):
Great, thanks. A couple of final comments, sorry. Otherwise looks good

PR #1733, opened 2025-11-25
Create ConfigServerEnergyMotorLookup and decouple gap and phase

oliwenmandiamond:
Refactored `EnergyMotorLookup` class to decouple the gap and phase. You now provide an `EnergyMotorLookup` for gap and phase now rather than just one. EnergyMotorlookup is only LookupTable as configuration, ConfigServerEnergyMotorLookup will get file contents from config server and convert it to LookupTable. ID's can provide any way to calculate the gap and phase as it can be any implementation that is of type EnergyMotorLookup, giving more flexibility e.g a pre configured lookup table, reading from a file, auto generated or from epics. Have also update the `Apple2Controller` methods to be more modular so it reuses more logic between devices that implement it.

1. Check refactoring makes sense
2. Confirm tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -23,6 +23,7 @@ from dodal.devices.flux import Flux
 from dodal.devices.i03.dcm import DCM
 from dodal.devices.i04.beamsize import Beamsize
 from dodal.devices.i04.constants import RedisConstants
+from dodal.devices.i04.max_pixel import MaxPixel
 from dodal.devices.i04.murko_results import MurkoResultsDevice
 from dodal.devices.i04.transfocator import Transfocator
 from dodal.devices.ipin import IPin
@@ -387,6 +388,16 @@ def scintillator() -> Scintillator:
         f"{PREFIX.beamline_prefix}-MO-SCIN-01:",
         Reference(aperture_scatterguard()),
         get_beamline_parameters(),
+    )
+
+
+@device_factory()
+def max_pixel() -> MaxPixel:
+    """Get the i04 max pixel device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return MaxPixel(
+        f"{PREFIX.beamline_prefix}-DI-OAV-01:",
     )
 
 
--- a/src/dodal/beamlines/i22.py
+++ b/src/dodal/beamlines/i22.py
@@ -19,7 +19,7 @@ from dodal.devices.i22.dcm import DCM
 from dodal.devices.i22.fswitch import FSwitch
 from dodal.devices.i22.nxsas import NXSasMetadataHolder, NXSasOAV, NXSasPilatus
 from dodal.devices.linkam3 import Linkam3
-from dodal.devices.motors import XYPitchStage
+from dodal.devices.motors import XYPitchStage, XYRollStage, XYStage
 from dodal.devices.slits import Slits
 from dodal.devices.synchrotron import Synchrotron
 from dodal.devices.tetramm import TetrammDetector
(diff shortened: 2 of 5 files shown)


src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -347,51 +312,7 @@ def available_pol(self) -> list[Pol]:
...
-        for key in self.lookup_tables.gap.root.keys():
-            if key is not None:
-                self.lookup_tables.phase.root[key] = generate_lookup_table_entry(
-                    min_energy=self.lookup_tables.gap.root[key].limit.minimum,
-                    max_energy=self.lookup_tables.gap.root[key].limit.maximum,
-                    poly1d_param=(PhASE_POLY1D_PARAMETERS[key]),
-                )
-
-    def get_motor_from_energy(self, energy: float, pol: Pol) -> tuple[float, float]:
+    def get_motor_from_energy(self, energy: float, pol: Pol) -> float:
```
Relm-Arrowny: Could: I am not 100% what it should be but get_motor_form_energy seem wrong now.
```suggestion
    def get_value_from_lookuptable(self, energy: float, pol: Pol) -> float:
```

src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -408,12 +329,61 @@ def get_motor_from_energy(self, energy: float, pol: Pol) -> tuple[float, float]:
...
+        poly = get_poly(lookup_table=self.lut, energy=energy, pol=pol)
+        return poly(energy)
+
+
+TAbstractEnergyMotorLookup = TypeVar(
+    "TAbstractEnergyMotorLookup", bound=AbstractEnergyMotorLookup
+)
+
+
+class FileReadingEnergyMotorLookup(AbstractEnergyMotorLookup):
```
Relm-Arrowny: Should: We not realy doing file reading here, we getting the data from DAQ server.
```suggestion
class EnergyMotorLookupConfigServer(AbstractEnergyMotorLookup):
```

src/dodal/devices/util/lookup_tables_apple2.py
```diff
@@ -408,12 +329,61 @@ def get_motor_from_energy(self, energy: float, pol: Pol) -> tuple[float, float]:
...
+    ):
+        """
+        Parameters:
+        -----------
+        config_client:
+            The config server client to fetch the look up table data.
+        lut_config:
+            Configuration that defines how to process file contents into a LookupTable
+        path:
+            File path to the gap lookup table.
```
Relm-Arrowny: Should:
```suggestion
            File path to the lookup table.
```

src/dodal/devices/util/lookup_tables_apple2.py
Relm-Arrowny: May be we can just replace this with a helper function and let the controller handle the update if updateis needed. That would allow everything just run off EnergyMotorLookup
```python 
def get_lookup_from_server (
        config_client: ConfigServer,
        lut_config: LookupTableConfig,
        path: Path,
    ):
        file_contents = config_client.get_file_contents(
            path, reset_cached_result=True
        )
        return convert_csv_to_lookup(file_contents, lut_config=self.lut_config)
```

src/dodal/devices/apple2_undulator.py
```diff
@@ -496,24 +494,39 @@ def __init__(
...
-        undulator design and the lookup table used.
+        undulator design.
         """
 
+    async def _set_apple2(self, id_motor_values: Apple2Val, pol: Pol) -> None:
+        """Logic to set the apple2 with the motor values. Specialised logic can also extend this method."""
+        await self.apple2().set(id_motor_values=id_motor_values)
+
+    async def _set_motors_from_energy_and_polarisation(
+        self, energy: float, pol: Pol
```
Relm-Arrowny: I am not sure about this, with ` def _set_motors_from_energy(self, value: float)` The task is quite clear, we have to define how the ID should move. Now we break the task into 4 parts?
It is quite hard to know what I needed to do and I wrote the original.....

Relm-Arrowny (changes requested):
Thanks for doing this, it look good, I just have a feeling we may have gone too far.

Relm-Arrowny (changes requested):
This kind of grow on me..... just one last point.

PR #1702, opened 2025-11-11
Wrap plans and plan stubs

EmsArnold:
Closes #1498

Wraps bluesky plans (`scan`, `rel_scan`, `grid_scan`, `rel_grid_scan`) and plan stubs (`rd`, `stop`) for use in blueapi. Intended as a replacement for gda-style `start stop step` scans until `spec_scan` is stable and supports `start stop step` rather than `start stop num`. 

### Instructions to reviewer on how to test:
1. Do thing x
2. Confirm thing y happens

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes

PR #1734, opened 2025-11-25
Wrap plans and plan stubs with lists of tuples

### LINKED ISSUE #1498: Create wrapped versions of all standard bluesky plan and step_scan version of scan

EmsArnold:
Closes #1498, and as an alternative to [#1702](https://github.com/DiamondLightSource/dodal/pull/1702). Scan syntax is more similar to both GDA and bluesky.

Wraps bluesky plans (`scan`, `rel_scan`, `grid_scan`, `rel_grid_scan`) and plan stubs (`rd`, `stop`) for use in blueapi. Intended as a replacement for gda-style start stop step scans until spec_scan is stable and supports start stop step rather than start stop num. Start stop step scans are achieved through generating a list for use with `list_scan`, `rel_list_scan`, ...

Concurrent trajectory scans (`scan`, etc.) and independent trajectory scans (`grid_scan`, etc.) are both accessed through the same plan and differentiated by number of parameters given for each movable - similar to current GDA-style scans. 

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/plan_stubs/wrapped.py
+++ b/src/dodal/plan_stubs/wrapped.py
@@ -1,6 +1,6 @@
 import itertools
 from collections.abc import Mapping
-from typing import Annotated, TypeVar
+from typing import Annotated, Any, TypeVar
 
 import bluesky.plan_stubs as bps
 from bluesky.protocols import Movable, Readable, Stoppable
@@ -133,7 +133,7 @@ def wait(
     return (yield from bps.wait(group, timeout=timeout))
 
 
-def rd(readable: Readable) -> MsgGenerator:
+def rd(readable: Readable) -> MsgGenerator[Any]:
     """Reads a single-value non-triggered object, wrapper for `bp.rd`.
 
     Args:
--- a/src/dodal/plan_stubs/wrapped.py
+++ b/src/dodal/plan_stubs/wrapped.py
@@ -14,7 +14,7 @@ T = TypeVar("T")
 
 
 def set_absolute(
-    movable: Movable[T], value: T, group: Group | None = None, wait: bool = False
+    movable: Movable[T], value: T, group: Group | None = None, wait: bool = True
 ) -> MsgGenerator:
     """Set a device, wrapper for `bp.abs_set`.
 
@@ -36,7 +36,7 @@ def set_absolute(
 
 
 def set_relative(
-    movable: Movable[T], value: T, group: Group | None = None, wait: bool = False
+    movable: Movable[T], value: T, group: Group | None = None, wait: bool = True
 ) -> MsgGenerator:
     """Change a device, wrapper for `bp.rel_set`.
 
--- a/src/dodal/plan_stubs/wrapped.py
+++ b/src/dodal/plan_stubs/wrapped.py
@@ -3,7 +3,7 @@ from collections.abc import Mapping
 from typing import Annotated, TypeVar
 
 import bluesky.plan_stubs as bps
-from bluesky.protocols import Movable
+from bluesky.protocols import Movable, Readable, Stoppable
 from bluesky.utils import MsgGenerator
 
 """Wrappers for Bluesky built-in plan stubs with type hinting."""
@@ -131,3 +131,27 @@ def wait(
         Iterator[MsgGenerator]: Bluesky messages.
     """
     return (yield from bps.wait(group, timeout=timeout))
+
+
+def rd(readable: Readable) -> MsgGenerator:
+    """Reads a single-value non-triggered object, wrapper for `bp.rd`.
+
+    Args:
+        readable (Readable): The device to be read
+
+    Returns:
+        Iterator[MsgGenerator]: Bluesky messages
+    """
+    return (yield from bps.rd(readable))
+
+
+def stop(stoppable: Stoppable) -> MsgGenerator:
+    """Stop a device, wrapper for `bp.stop`.
+
+    Args:
+        stoppable (Stoppable): Device to be stopped
+
+    Returns:
+        Iterator[MsgGenerator]: Bluesky messages
+    """
+    return (yield from bps.stop(stoppable))
--- a/src/dodal/plan_stubs/wrapped.py
+++ b/src/dodal/plan_stubs/wrapped.py
@@ -134,7 +134,7 @@ def wait(
 
 
 def rd(readable: Readable) -> MsgGenerator[Any]:
-    """Reads a single-value non-triggered object, wrapper for `bp.rd`.
+    """Reads a single-value non-triggered object, wrapper for `bps.rd`.
 
     Args:
         readable (Readable): The device to be read
@@ -149,7 +149,7 @@ def rd(readable: Readable) -> MsgGenerator[Any]:
 
 
 def stop(stoppable: Stoppable) -> MsgGenerator:
-    """Stop a device, wrapper for `bp.stop`.
+    """Stop a device, wrapper for `bps.stop`.
 
     Args:
         stoppable (Stoppable): Device to be stopped
--- a/src/dodal/plan_stubs/wrapped.py
+++ b/src/dodal/plan_stubs/wrapped.py
@@ -140,7 +140,10 @@ def rd(readable: Readable) -> MsgGenerator[Any]:
         readable (Readable): The device to be read
 
     Returns:
-        Iterator[MsgGenerator]: Bluesky messages
+        MsgGenerator: Plan.
+
+    Yields:
+        Iterator[MsgGenerator]: Bluesky messages.
     """
     return (yield from bps.rd(readable))
 
@@ -152,6 +155,9 @@ def stop(stoppable: Stoppable) -> MsgGenerator:
         stoppable (Stoppable): Device to be stopped
 
     Returns:
-        Iterator[MsgGenerator]: Bluesky messages
+        MsgGenerator: Plan.
+
+    Yields:
+        Iterator[MsgGenerator]: Bluesky messages.
     """
     return (yield from bps.stop(stoppable))
--- a/src/dodal/plans/wrapped.py
+++ b/src/dodal/plans/wrapped.py
@@ -147,8 +147,8 @@ def num_grid_scan(
 ) -> MsgGenerator:
     """Scan independent multi-motor trajectories.
 
-    Snakes all fast axes by default. The scan is defined by number of points along scan
-    trajectories. Wraps bluesky.plans.grid_scan(det, *args, snake_axes, md=metadata).
+    The scan is defined by number of points along scan trajectories. Snakes all fast
+    axes by default. Wraps bluesky.plans.grid_scan(det, *args, snake_axes, md=metadata).
     """
     # TODO: move to using Range spec and spec_scan when stable and tested at v1.0
     args, shape = _make_num_scan_args(params)
@@ -215,8 +215,8 @@ def num_grid_rscan(
 ) -> MsgGenerator:
     """Scan independent trajectories, relative to current positions.
 
-    Snakes all fast axes by default. The scan is defined by number of points along scan
-    trajectories. Wraps bluesky.plans.rel_grid_scan(det, *args, md=metadata).
+    The scan is defined by number of points along scan trajectories. Snakes all fast
+    axes by default. Wraps bluesky.plans.rel_grid_scan(det, *args, md=metadata).
     """
     # TODO: move to using Range spec and spec_scan when stable and tested at v1.0
     args, shape = _make_num_scan_args(params)
@@ -302,8 +302,8 @@ def list_grid_scan(
 ) -> MsgGenerator:
     """Scan independent trajectories.
 
-    Snakes slow axes by default. The scan is defined by providing a list of points for
-    each scan trajectory. Wraps bluesky.plans.list_grid_scan(det, *args, md=metadata).
+    The scan is defined by providing a list of points for each scan trajectory. Snakes
+    slow axes by default. Wraps bluesky.plans.list_grid_scan(det, *args, md=metadata).
     """
     args, shape = _make_list_scan_args(params=params, grid=True)
     metadata = metadata or {}
@@ -370,8 +370,9 @@ def list_grid_rscan(
 ) -> MsgGenerator:
     """Scan independent trajectories, relative to current positions.
 
-    The scan is defined by providing a list of points for each scan trajectory.
-    Wraps bluesky.plans.rel_list_grid_scan(det, *args, md=metadata).
+    The scan is defined by providing a list of points for each scan trajectory. Snakes
+    all fast axes by default. Wraps bluesky.plans.rel_list_grid_scan(det, *args,
+    md=metadata).
     """
     args, shape = _make_list_scan_args(params=params, grid=True)
     metadata = metadata or {}
@@ -520,8 +521,9 @@ def step_grid_scan(
 ) -> MsgGenerator:
     """Scan independent trajectories with specified step size.
 
-    Snakes all slow axes by default. Generates list(s) of points for each trajectory,
-    used with bluesky.plans.list_grid_scan(det, *args, md=metadata).
+    Generates list(s) of points for each trajectory, used with
+    bluesky.plans.list_grid_scan(det, *args, md=metadata). Snakes all fast axes by
+    default.
     """
     # TODO: move to using Linspace spec and spec_scan when stable and tested at v1.0
     args, shape = _make_step_scan_args(params, grid=True)
@@ -589,8 +591,9 @@ def step_grid_rscan(
 ) -> MsgGenerator:
     """Scan independent trajectories with specified step size, relative to position.
 
-    Snakes all slow axes by default. Generates list(s) of points for each trajectory,
-    used with bluesky.plans.list_grid_scan(det, *args, md=metadata).
+    Generates list(s) of points for each trajectory, used with
+    bluesky.plans.list_grid_scan(det, *args, md=metadata). Snakes all fast axes by
+    default.
     """
     # TODO: move to using Linspace spec and spec_scan when stable and tested at v1.0
     args, shape = _make_step_scan_args(params, grid=True)
--- a/src/dodal/plans/wrapped.py
+++ b/src/dodal/plans/wrapped.py
@@ -96,7 +96,7 @@ def num_scan(
             "numN]}'."
         ),
     ],
-    snake_axes: list | bool | None = None,
+    snake_axes: list | bool = True,
     metadata: dict[str, Any] | None = None,
 ) -> MsgGenerator:
     """Scan over concurrent or independent multi-motor trajectories.
@@ -134,7 +134,7 @@ def num_rscan(
             "numN]}'."
         ),
     ],
-    snake_axes: list | bool | None = None,
+    snake_axes: list | bool = True,
     metadata: dict[str, Any] | None = None,
 ) -> MsgGenerator:
     """Scan over concurrent or independent trajectories relative to current position.
@@ -184,7 +184,7 @@ def list_scan(
         ),
     ],
     grid: bool = False,
-    snake_axes: bool = False,  # Currently specifying axes to snake is not supported
+    snake_axes: bool = True,  # Currently specifying axes to snake is not supported
     metadata: dict[str, Any] | None = None,
 ) -> MsgGenerator:
     """Scan over concurrent or independent trajectories relative to current position.
@@ -223,7 +223,7 @@ def list_rscan(
         ),
     ],
     grid: bool = False,
-    snake_axes: bool = False,  # Currently specifying axes to snake is not supported
+    snake_axes: bool = True,  # Currently specifying axes to snake is not supported
     metadata: dict[str, Any] | None = None,
 ) -> MsgGenerator:
     """Scan over concurrent or independent trajectories relative to current position.
@@ -257,6 +257,10 @@ def _make_stepped_list(
     if len(params) == 3:
         stop = params[1]
         step = params[2]
+        if start == stop:
+            raise ValueError(
+                f"Start ({start}) and stop ({stop}) values cannot be the same."
+            )
         if abs(step) > abs(stop - start):
             step = stop - start
         step = abs(step) * np.sign(stop - start)
@@ -339,7 +343,7 @@ def step_scan(
             "[startN, stopN, stepN]}'."
         ),
     ],
-    snake_axes: bool = False,  # Currently specifying axes to snake is not supported
+    snake_axes: bool = True,  # Currently specifying axes to snake is not supported
     metadata: dict[str, Any] | None = None,
 ) -> MsgGenerator:
     """Scan over multi-motor trajectories with specified step size.
@@ -378,7 +382,7 @@ def step_rscan(
             "[startN, stopN, stepN]}'."
         ),
     ],
-    snake_axes: bool = False,  # Currently specifying axes to snake is not supported
+    snake_axes: bool = True,  # Currently specifying axes to snake is not supported
     metadata: dict[str, Any] | None = None,
 ) -> MsgGenerator:
     """Scan over multi-motor trajectories with specified step size.
--- a/src/dodal/plans/wrapped.py
+++ b/src/dodal/plans/wrapped.py
@@ -50,6 +50,7 @@ def count(
     metadata: dict[str, Any] | None = None,
 ) -> MsgGenerator:
     """Reads from a number of devices.
+
     Wraps bluesky.plans.count(det, num, delay, md=metadata) exposing only serializable
     parameters and metadata.
     """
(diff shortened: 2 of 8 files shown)


RJCD-Diamond:
whatt else does this need until it can not be a draft and be merged?

DominicOram (dismissed):
>until spec_scan is stable and supports start stop step rather than start stop num

Does this mean you intend for these to be removed when https://github.com/bluesky/scanspec/issues/186 is done? If so I would put a comment in each plan you intend to retire saying that it will be replaced and pointing to that issue.

> Concurrent trajectory scans (scan, etc.) and independent trajectory scans (grid_scan, etc.) are both accessed through the same plan and differentiated by number of parameters given for each movable - similar to current GDA-style scans.

Is this a hard requirement? I think I would rather we have them separate as:
* We can then just point people to https://blueskyproject.io/bluesky/main/plans.html as documentation
* They seem to actually be quite different things and lead to weirdness e.g. `snake_axes` meaning nothing in some contexts

On that note, having different behaviour based on whether we provide a `num` for each axis or not is also not intuitive. Again, I would just stick to the `bluesky` way of being more explicit about what you actually want to do

RJCD-Diamond:
@EmsArnold Been using this for commissioning purposes on i22 - and strongly suggest we make sure that all of these plans /stubs wait for the plan to finish by default, and have an optional wait bool to control this

fajinyuan:
Sorry for comments so late due to holiday leaves. I support @DominicOram suggestion that all scans should be explicitly named so its behaviour doesn't depend on the number of parameters in the input - I regard this feature in GDA scan is a bad design as users need to read doc or use help to discover what each scan syntax does. in the new system we should make these explicitly and then enforce the number of parameters checking for each command in command pre-processor or parser.

src/dodal/plan_stubs/wrapped.py
```diff
@@ -131,3 +131,27 @@ def wait(
...
+
+
+def rd(readable: Readable) -> MsgGenerator[Any]:
+    """Reads a single-value non-triggered object, wrapper for `bp.rd`.
+
+    Args:
+        readable (Readable): The device to be read
+
+    Returns:
+        Iterator[MsgGenerator]: Bluesky messages
```
tpoliaw: Do these return an `Iterator[MsgGenerator]` or just a `MsgGenerator`?

src/dodal/plan_stubs/wrapped.py
```diff
@@ -131,3 +131,27 @@ def wait(
         Iterator[MsgGenerator]: Bluesky messages.
     """
     return (yield from bps.wait(group, timeout=timeout))
+
+
+def rd(readable: Readable) -> MsgGenerator[Any]:
+    """Reads a single-value non-triggered object, wrapper for `bp.rd`.
```
tpoliaw:
```suggestion
    """Reads a single-value non-triggered object, wrapper for `bps.rd`.
```

src/dodal/plan_stubs/wrapped.py
```diff
@@ -131,3 +131,27 @@ def wait(
...
+        readable (Readable): The device to be read
+
+    Returns:
+        Iterator[MsgGenerator]: Bluesky messages
+    """
+    return (yield from bps.rd(readable))
+
+
+def stop(stoppable: Stoppable) -> MsgGenerator:
+    """Stop a device, wrapper for `bp.stop`.
```
tpoliaw:
```suggestion
    """Stop a device, wrapper for `bps.stop`.
```

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
     metadata = metadata or {}
     metadata["shape"] = (num,)
     yield from bp.count(tuple(detectors), num, delay=delay, md=metadata)
+
+
+def _make_num_scan_args(
+    params: list[tuple[Movable | Motor, list[float | int]]], num: int | None = None
+):
+    shape = []
+    if num:
```
tpoliaw, line 69: Does there need to be more validation of `num`? -3 or 0 etc.

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+                )
+
+    args = []
+    for param in params:
+        args.append(param[0])
+        args.extend(param[1])
+    return args, shape
+
+
+@attach_data_session_metadata_decorator()
```
tpoliaw: Pretty sure this decorator is a no-op since the move to device manager.

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+                pass
+            else:
+                raise ValueError("You must provide 'start stop' for each motor.")
+    else:
+        for param in params:
+            if len(param[1]) == 3:
+                shape.append(param[1][-1])
+            else:
+                raise ValueError(
+                    "You must provide 'start stop step' for each motor in a grid scan."
```
tpoliaw: Is this `start, stop, step` or `start, stop, num`? If the former, doesn't that make the `num_scan` run a non-num_scan if you don't set num?

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+            [start2, stop2, num2]), ... , (movableN, [startN, stopN, numN])]'."
+        ),
+    ],
+    snake_axes: list | bool = True,
+    metadata: dict[str, Any] | None = None,
+) -> MsgGenerator:
+    """Scan independent multi-motor trajectories.
+
+    The scan is defined by number of points along scan trajectories. Snakes all fast
+    axes by default. Wraps bluesky.plans.grid_scan(det, *args, snake_axes, md=metadata).
```
tpoliaw: How does it determine which are the fast axes?

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+    ],
+    params: Annotated[
+        list[tuple[Movable | Motor, list[float | int]]],
+        Field(
+            description="List of tuples (device, parameter). For concurrent \
+            trajectories, provide '[(movable1, [start1, stop1]), (movable2, [start2, \
+            stop2]), ... , (movableN, [startN, stopN])]'."
+        ),
+    ],
+    num: int | None = None,
```
tpoliaw: If `num=None` is there a difference between this and `num_grid_scan`?

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+            description="Set of readable devices, will take a reading at each point",
+            min_length=1,
+        ),
+    ],
+    params: Annotated[
+        list[tuple[Movable | Motor, list[float | int]]],
+        Field(
+            description="List of tuples (device, parameter). For concurrent \
+            trajectories, provide '[(movable1, [start1, stop1]), (movable2, [start2, \
+            stop2]), ... , (movableN, [startN, stopN])]'."
```
tpoliaw: I think using line continuations on an indented paragraph ends up with weird spaces in the descriptions, eg `For concurrent             trajectories, provide`

```suggestion
            description=("List of tuples (device, parameter). For concurrent "
            "trajectories, provide '[(movable1, [start1, stop1]), (movable2, [start2, "
            "stop2]), ... , (movableN, [startN, stopN])]'.")
```

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+            [start2, stop2, num2]), ... , (movableN, [startN, stopN, numN])]'."
+        ),
+    ],
+    snake_axes: list | bool = True,
+    metadata: dict[str, Any] | None = None,
+) -> MsgGenerator:
+    """Scan independent trajectories, relative to current positions.
+
+    The scan is defined by number of points along scan trajectories. Snakes all fast
+    axes by default. Wraps bluesky.plans.rel_grid_scan(det, *args, md=metadata).
```
tpoliaw: Not sure if this is needed but it's mentioned in the `num_grid_scan` docs

```suggestion
    axes by default. Wraps bluesky.plans.rel_grid_scan(det, *args, snake_axes, md=metadata).
```

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+    metadata = metadata or {}
+    metadata["shape"] = shape
+
+    yield from bp.rel_grid_scan(
+        tuple(detectors), *args, snake_axes=snake_axes, md=metadata
+    )
+
+
+def _make_list_scan_args(
+    params: list[tuple[Movable | Motor, list[float | int]]], grid: bool | None = None
```
tpoliaw: Can all the `float | int` hints be `float`? I think most type checkers allow ints to count as floats

tpoliaw: Is there any reason to allow `None` instead of `False`?

```suggestion
    params: list[tuple[Movable | Motor, list[float | int]]], grid: bool = False
```

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+        Sequence[Readable | AsyncReadable],
+        Field(
+            description="Set of readable devices, will take a reading at each point",
+            min_length=1,
+        ),
+    ],
+    params: Annotated[
+        list[tuple[Movable | Motor, list[float | int]]],
+        Field(
+            description="List of tuples (device, parameter). For concurrent \
```
tpoliaw:
```suggestion
            description="List of tuples (device, positions). For concurrent \
```

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+            [point1, point2, ...]), ... , (movableN, [point1, point2, ...])]'."
+        ),
+    ],
+    snake_axes: bool = True,  # Currently specifying axes to snake is not supported
+    metadata: dict[str, Any] | None = None,
+) -> MsgGenerator:
+    """Scan independent trajectories.
+
+    The scan is defined by providing a list of points for each scan trajectory. Snakes
+    slow axes by default. Wraps bluesky.plans.list_grid_scan(det, *args, md=metadata).
```
tpoliaw: Which axes are the slow axes?

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+    metadata = metadata or {}
+    metadata["shape"] = shape
+
+    yield from bp.rel_list_grid_scan(
+        tuple(detectors), *args, snake_axes=snake_axes, md=metadata
+    )
+
+
+def _make_stepped_list(
+    params: list[Any] | Sequence[Any],
```
tpoliaw: Isn't a list also a sequence?

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+    return rounded_stepped_list, len(rounded_stepped_list)
+
+
+def _make_step_scan_args(
+    params: list[tuple[Movable | Motor, list[float | int]]], grid: bool | None = None
+):
+    args = []
+    shape = []
+    stepped_list_length = None
+    for param, movable_num in zip(params, range(len(params)), strict=True):
```
tpoliaw:
```suggestion
    for movable_num, param in enumerate(params):
```

or


```suggestion
    for movable_num, (movable, parameters) in enumerate(params):
```

which would save you having to use `param[0]` and `param[1]` everywhere

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+        if movable_num == 0:
```
tpoliaw: Rather than checking indices for each parameter, I think you could do something like

```python
if not params: return [], []
first, *rest = params
# the movable_num == 0 case
for param in rest:
    # the other case
```

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+            if len(param[1]) == 3:
+                stepped_list, stepped_list_length = _make_stepped_list(params=param[1])
+                args.append(param[0])
+                args.append(stepped_list)
+                shape.append(stepped_list_length)
+            else:
+                raise ValueError(
+                    f"You provided {len(param[1])} parameters, rather than 3."
```
tpoliaw: nit: Could be useful to include which movable had the wrong number of args

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+            raise ValueError(
+                f"Start ({start}) and stop ({stop}) values cannot be the same."
+            )
+        if abs(step) > abs(stop - start):
+            step = stop - start
+        step = abs(step) * np.sign(stop - start)
+        stepped_list = np.arange(start, stop, step).tolist()
+        if abs((stepped_list[-1] + step) - stop) <= abs(step * 0.05):
+            stepped_list.append(stepped_list[-1] + step)
+        rounded_stepped_list = round_list_elements(stepped_list=stepped_list, step=step)
```
tpoliaw: Do we want to round the positions? With this included, running `_make_stepped_list([0.123, 1.123, 0.1])` gives `([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1], 11)` which is unlikely to be the intention.

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+    elif len(params) == 2 and num:
+        step = params[1]
+        stepped_list = [start + (n * step) for n in range(num)]
+        rounded_stepped_list = round_list_elements(stepped_list=stepped_list, step=step)
+    else:
+        raise ValueError(
+            f"You provided {len(params)}, rather than 3, or 2 and number of points."
```
tpoliaw: Should this check for 3 and a number of points?

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+def _make_stepped_list(
+    params: list[Any] | Sequence[Any],
+    num: int | None = None,
+):
+    def round_list_elements(stepped_list, step):
+        d = Decimal(str(step))
+        exponent = d.as_tuple().exponent
+        decimal_places = -exponent  # type: ignore
+        return np.round(stepped_list, decimals=decimal_places).tolist()
+
```
tpoliaw: I think this would be clearer as two functions - there's nothing in common between the two cases

`def _stepped_list_step(start, stop, step) -> list:`
and
`def _stepped_list_num(start, step, num) -> list`

It would also enforce the need for 'two and num' or 'three and no num'

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
...
+    else:
+        raise ValueError(
+            f"You provided {len(params)}, rather than 3, or 2 and number of points."
+        )
+
+    return rounded_stepped_list, len(rounded_stepped_list)
+
+
+def _make_step_scan_args(
+    params: list[tuple[Movable | Motor, list[float | int]]], grid: bool | None = None
```
tpoliaw: This might be clearer as two functions as well. It's doing two almost completely different things

src/dodal/plans/wrapped.py
```diff
@@ -56,3 +61,545 @@ def count(
     metadata = metadata or {}
     metadata["shape"] = (num,)
     yield from bp.count(tuple(detectors), num, delay=delay, md=metadata)
+
+
+def _make_num_scan_args(
+    params: list[tuple[Movable | Motor, list[float | int]]], num: int | None = None
+):
```
tpoliaw, line 67: Worth adding return types?

src/dodal/plans/wrapped.py
EmsArnold: For either [`bp.grid_scan`](https://blueskyproject.io/bluesky/main/generated/bluesky.plans.grid_scan.html#bluesky.plans.grid_scan) or `bp.rel_grid_scan`, you can give a list of motors which will be snaked, `False` for no snaking, or `True` for snaking all motors. It is always assumed that the first motor is the slow axis, and thus will not be snaked if `snake_axes=True` and will raise an error if the motor is present in the list of axes to be snaked.

src/dodal/plan_stubs/wrapped.py
EmsArnold: Good spot - just a `MsgGenerator`. Fixed now

src/dodal/plan_stubs/wrapped.py
EmsArnold: Fixed

src/dodal/plan_stubs/wrapped.py
EmsArnold: Fixed

src/dodal/plans/wrapped.py
EmsArnold: Removed now

src/dodal/plans/wrapped.py
EmsArnold: I think for `num_scan`, `num` is necessary. Likely a hold-over from when I had `num_scan` and `num_grid_scan` together. Removed now so `num` can only be an integer.

src/dodal/plans/wrapped.py
EmsArnold: Fixed

src/dodal/plans/wrapped.py
EmsArnold: Added

src/dodal/plans/wrapped.py
EmsArnold: Fixed for all variations of `list_scan`

src/dodal/plans/wrapped.py
EmsArnold: Likely a copy and paste error - changed to `start, stop, num`

src/dodal/plans/wrapped.py
EmsArnold: I've tried it, and tests are failing: 
```FAILED tests/plans/test_wrapped.py::test_num_grid_scan[x_list0-y_list0] - TypeError: 'float' object cannot be interpreted as an integer```

Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: I don't think so, probably just a hold-over from when non-grid and grid scans were together. Changed now.

File: src/dodal/plans/wrapped.py (hunk 13 above)
Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: Oops - looks like this was just plain wrong. Fixed and gave further context for what 'fast axes' are considered to be for all scan variations.

File: src/dodal/plans/wrapped.py (hunk 14 above)
Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: Indeed it is, I'm not sure why I did that. Removed now with tests passing...

File: src/dodal/plans/wrapped.py (hunk 15 above)
Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: Addressed by changes addressing next comment.

File: src/dodal/plans/wrapped.py (hunk 16 above)
Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: Thanks, that's much cleaner! Done

File: src/dodal/plans/wrapped.py (hunk 17 above)
Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: Good idea, done

File: src/dodal/plans/wrapped.py (hunk 18 above)
Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: If memory served, without rounding it would producing slightly variable numbers with a large number of trailing zeroes, such as `1.0000000001` or `0.9999999999`. Perhaps the better way would be to take number of significant digits expected from the largest of the `start`, `stop`, or `step`, rather than just relying on `step`.

File: src/dodal/plans/wrapped.py (hunk 19 above)
Comment (EmsArnold) [reply src/dodal/plans/wrapped.py]: It could, but I've left it open as the user could potentially provide any number of parameters. As it's only being called by another non-public function, theoretically this shouldn't be a problem at all.

File: src/dodal/plans/wrapped.py [hunk 23]
```diff
@@ -56,3 +61,538 @@ def count(
...
+def num_scan(
+    detectors: Annotated[
+        Sequence[Readable | AsyncReadable],
+        Field(
+            description="Set of readable devices, will take a reading at each point",
+            min_length=1,
+        ),
+    ],
+    params: Annotated[
+        list[tuple[Movable | Motor, list[float | int]]],
```
oliwenmandiamond: Why do we need to type hint `Motor` for every scan? A `Motor` is `Movable`, so `Movable` alone should be sufficient for the type hint

EmsArnold: The reason for this is because [`ophyd-async.epics.motor` is not actually a `Movable`](https://github.com/bluesky/ophyd-async/blob/195b6a2894501f13634c9f7b6a0d84f7a866c84e/src/ophyd_async/epics/motor.py#L109). This will shortly be rectified when [`StandardMovable`](https://github.com/bluesky/ophyd-async/pull/1200) is done. I'll add a comment and an issue to update this when `StandardMovable` comes in.

oliwenmandiamond: Ah interesting, thanks! 

I still would of thought it wouldn't be needed because `Movable` is a Protocol, and `Motor` follows the `Movable` protocol even if it doesn't directly inherit it. Correct me if wrong though?

EmsArnold: My impression is that if you tried to use it, it would probably work. However, as `Motor` doesn't directly (or indirectly) inherit `Movable`, the type checking fails. I may also be wrong though.
I suppose it's relatively academic at this point, but if something walks, talks, and quacks like a duck, if there's not a sticker on it saying duck then there's a chance it could be something else...

oliwenmandiamond: Okay, just tested this

```Python
from bluesky.protocols import Movable

def my_move(motor: Movable, value: float):
    yield from ()


def motor() -> Motor:
    with init_devices(mock=True):
        motor = Motor("TEST:")
    return motor


def test_motor(motor: Motor, run_engine: RunEngine):
    run_engine(my_move(motor, 5))
```
It passes type checking fine.

Motor is currently this:
```Python
class Motor(
    StandardReadable,
    Locatable[float],
    Stoppable,
    Flyable,
    Preparable,
    Subscribable[float],
):
```
It doesn't look like it is implements Movable here, however it implements `Locatable` which extends `Movable`

```Python
@runtime_checkable
class Locatable(Movable[T], Protocol):
    @abstractmethod
    def locate(self) -> SyncOrAsync[Location[T]]:
        """Return the current location of a Device.

        While a ``Readable`` reports many values, a ``Movable`` will have the
        concept of location. This is where the Device currently is, and where it
        was last requested to move to. This protocol formalizes how to get the
        location from a ``Movable``.
        """
        ...
```
So it does directly implement Movable via `Locatable`! It is safe to remove the Motor type hints from the scans. There is a sticker confirming it is a duck

src/dodal/plans/wrapped.py
EmsArnold: Agreed, these have now been split for added clarity.

src/dodal/plans/wrapped.py
EmsArnold: That's odd... At one point it was necessary (I remember being confused, though not much else). 
I'll remove it as it's working!

src/dodal/plans/wrapped.py
EmsArnold: I'm not sure why this would be improved by separating out into two separate functions? Currently this just builds the list of parameters passed to the bluesky plan. You'd still have > ~50% of code shared/repeated if this became two functions.

src/dodal/plans/wrapped.py
EmsArnold: Resolved as a by-product of addressing the below comment

src/dodal/plans/wrapped.py
EmsArnold: Added

src/dodal/plans/wrapped.py
EmsArnold: This validation happens in `bp.scan` already, but I've added a test to ensure this fails in the expected way.

src/dodal/plans/wrapped.py
EmsArnold: I've added `start` and `stop` to the consideration of number of significant figures used for rounding. @tpoliaw , perhaps you could have a look and see if you're happy with this?

oliwenmandiamond (approved):
I will approve of this. I think this has been blocked long enough and keeps growing in size and scope. I think any other issues can be addressed in other (and smaller) PRs.

PR #1737, opened 2025-11-26
Allow demand energy to be None for eiger parameters

olliesilvester:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/1471

If the expected energy is None, we should use the current energy instead.

Typing should have caught this issue, but we had it disabled for the whole file - we disabled it because old ophyd-devices don't play nicely with typing. I've added individual type ignores instead so that real typing errors still get noticed.


DominicOram (approved):
Great, thanks!

PR #1745, opened 2025-12-01
add hybrid fastcs/adodin eiger. moved from ophyd_async branch

RJCD-Diamond:
After discussions with @coretl and given the fact that we will be removing the adodin in order to only support fascs odin, the adodin module has been moved to dls-dodal in order to maintain compatibility until all EIger's have been moved to fascs odin

https://github.com/bluesky/ophyd-async/pull/1128

### DIFF
--- a/tests/devices/test_async_eiger.py
+++ b/tests/devices/test_async_eiger.py
@@ -280,14 +280,14 @@ async def test_observe_indices_written(odin_driver_and_writer: OdinDriverAndWrit
 
     indices = writer.observe_indices_written(timeout=0.5)
 
-    c = 0
+    trigger_count = 0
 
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -19,5 +19,7 @@
     "python.analysis.extraPaths": [
         "./src"
     ],
-    "python.analysis.typeCheckingMode": "basic"
+    "python.analysis.typeCheckingMode": "basic",
+    "python-envs.defaultEnvManager": "ms-python.python:system",
+    "python-envs.pythonProjects": []
```
DominicOram: Nit: Why were these added?

PR #1748, opened 2025-12-01
Use config server for OAV config files

### LINKED ISSUE #1448: Migrate OAV configuration to use config server
The OAV gets configuration from the following files:
- `jCameraManZoomLevels.xml`
- `display.configuration`
- `OAVCentring_hyperion.json` - this is supported by the config server but we don't seem to call it yet

## Acceptance Criteria
- Hyperion / mx-bluesky no longer read OAV configuration directly from the filesystem

jacob720:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/1448
Required by https://github.com/DiamondLightSource/mx-bluesky/pull/1521

1. Check OAV config is read through config server

### Checks for reviewer


tests/devices/oav/test_data/jCameraManZoomLevels.json
```diff
@@ -0,0 +1,45 @@
+{
+    "JCameraManSettings": {
+        "levels": {
```
rtuck99, line 3: This isn't an xml file any more, so it should have a different extension

src/dodal/devices/oav/oav_parameters.py
```diff
@@ -119,18 +124,21 @@ class OAVConfigBase(Generic[ParamType]):
...
+        ]
 
     def _read_zoom_params(self) -> dict:
         um_per_pix = {}
-        for node in self.zoom_params:
-            zoom = str(_get_element_as_float(node, "level"))
-            um_pix_x = _get_element_as_float(node, "micronsPerXPixel")
-            um_pix_y = _get_element_as_float(node, "micronsPerYPixel")
-            um_per_pix[zoom] = (um_pix_x, um_pix_y)
+        zoom_levels: list[dict] = self.zoom_params["levels"]["zoomLevel"]
```
rtuck99, line 131: I think I would probably have gone down the pydantic model route, but seeing as in either case we have to do this dict build to get the data structure we want from the file, I don't think it makes much difference in the end.

src/dodal/devices/oav/oav_parameters.py
```diff
@@ -119,18 +124,21 @@ class OAVConfigBase(Generic[ParamType]):
...
+        for level in zoom_levels:
+            zoom = level["level"]
+            um_per_pix[zoom] = (
+                float(level["micronsPerXPixel"]),
+                float(level["micronsPerYPixel"]),
+            )
```
rtuck99, line 137: I would just like to point out here that I think `OAVBeamCentreFile` is really badly named because it inherits from `OAV` not `OAVConfigBase` or `OAVParameters`.

src/dodal/devices/oav/oav_parameters.py
```diff
@@ -161,19 +169,14 @@ def __init__(
         self.display_config = self._get_display_config(display_config_file)
         super().__init__(zoom_params_file)
 
-    def _get_display_config(self, display_config_file: str):
-        with open(display_config_file) as f:
-            file_lines = f.readlines()
-        return file_lines
+    def _get_display_config(self, display_config_file: str) -> DisplayConfig:
```
rtuck99: Must: This needs to be injected from device factory because i03 will pick up the wrong configuration otherwise.

rtuck99 (changes requested):
Just needs that one change to make sure that the OAV config uses the right config server.

src/dodal/devices/oav/oav_parameters.py
jacob-williamson: Thanks, missed this when updating this branch

src/dodal/devices/oav/oav_parameters.py
jacob-williamson: I think you're right, will write an issue and do this separately so we can get this merged

src/dodal/beamlines/i03.py
```diff
@@ -205,11 +205,13 @@ def panda_fast_grid_scan() -> PandAFastGridScan:
...
 def oav(
+    config_client: ConfigClient,
     params: OAVConfigBeamCentre | None = None,
 ) -> OAVBeamCentreFile:
     return OAVBeamCentreFile(
         prefix=f"{PREFIX.beamline_prefix}-DI-OAV-01:",
-        config=params or OAVConfigBeamCentre(ZOOM_PARAMS_FILE, DISPLAY_CONFIG),
+        config=params
+        or OAVConfigBeamCentre(ZOOM_PARAMS_FILE, DISPLAY_CONFIG, config_client),
     )
```
rtuck99, line 220: When doing `dodal connect i03` I get the following about 1/5 the time

```
aperture_scatterguard: NameError: name 'CONFIG_CLIENT' is not defined
beamstop: NameError: name 'CONFIG_CLIENT' is not defined
beamsize: ValueError: Errors building dependencies: {'aperture_scatterguard'}
scintillator: ValueError: Errors building dependencies: {'aperture_scatterguard'}
fastcs_eiger: NotConnectedError:
```

The device factories for these devices are still calling get_beamline_parameters - we should be initialising them all using the fixture

rtuck99: Should go through the other beamlines as well

rtuck99 (approved):
Approved

PR #1749, opened 2025-12-01
Add i04 device which finds the centre of the beam from OAV camera by fitting ellipses

### LINKED ISSUE #1726: OAV Beam Center Device
Create a dodal device which calculates the beam centre.

srishtysajeev:
Fixes #1726 

1. Check that the tests work

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i04.py
+++ b/src/dodal/beamlines/i04.py
@@ -21,6 +21,7 @@ from dodal.devices.eiger import EigerDetector
 from dodal.devices.fast_grid_scan import ZebraFastGridScanThreeD
 from dodal.devices.flux import Flux
 from dodal.devices.i03.dcm import DCM
+from dodal.devices.i04.beam_centre import CentreEllipseMethod
 from dodal.devices.i04.beamsize import Beamsize
 from dodal.devices.i04.constants import RedisConstants
 from dodal.devices.i04.max_pixel import MaxPixel
@@ -409,4 +410,14 @@ def beamsize() -> Beamsize:
     return Beamsize(
         transfocator=transfocator(),
         aperture_scatterguard=aperture_scatterguard(),
+    )
+
+
+@device_factory()
+def beam_centre() -> CentreEllipseMethod:
+    """Get the i04 centring device, instantiate it if it hasn't already been.
+    If this is called when already instantiated in i04, it will return the existing object.
+    """
+    return CentreEllipseMethod(
+        f"{PREFIX.beamline_prefix}-DI-OAV-01:",
     )


src/dodal/devices/i04/beam_centre.py
```diff
@@ -0,0 +1,72 @@
...
+from ophyd_async.core import AsyncStatus, StandardReadable, soft_signal_r_and_setter
+from ophyd_async.epics.core import (
+    epics_signal_r,
+)
+
+from dodal.devices.i04.max_pixel import convert_to_gray_and_blur
+from dodal.log import LOGGER
+
+
+async def binary_img(img, img_name="Threshold"):
```
olliesilvester: Should: Give types for all our parameters and return values


```suggestion
async def binary_img(img: cv2.typing.MatLike, img_name: str = "Threshold"):
```

src/dodal/devices/i04/beam_centre.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+    Function which creates a binary image from a beamline image. This is where the
+    pixels of an image are converted to one of two values (a high and a low value).
+    Otsu's method is used for automatic threshholding.
+    See https://docs.opencv.org/4.x/d7/d4d/tutorial_py_thresholding.html.
+    The threshold is increased by 10 (brightness taken from image in grayscale)
+    in order to get more the centre of the beam.
+    """
+    blurred = await convert_to_gray_and_blur(img)
```
olliesilvester: `convert_to_gray_and_blur` doesn't have any `await` statements so it shouldn't be an async function, and we shouldn't await it here. Can just call it like a regular function

src/dodal/devices/i04/beam_centre.py
```diff
@@ -0,0 +1,72 @@
...
+    thresh_img = cv2.threshold(blurred, thresh, 255, cv2.THRESH_BINARY)[1]
+
+    LOGGER.info(f"[INFO] thresholding value (otsu with blur): {thresh}")
+    return thresh_img
+
+
+class CentreEllipseMethod(StandardReadable, Triggerable):
+    """
+    Fits an ellipse a binary image of the beam. The centre of the fitted ellipse is taken
+    to be the centre of the beam.
```
olliesilvester:
```suggestion
    Upon triggering, fits an ellipse to a binary image of the OAV. The scintillator should be in position when reading the image. The centre of the fitted ellipse is taken
    to be the centre of the beam, and these pixel values are stored in the device.
```

src/dodal/devices/i04/beam_centre.py
```diff
@@ -0,0 +1,72 @@
...
+    The threshold is increased by 10 (brightness taken from image in grayscale)
+    in order to get more the centre of the beam.
+    """
+    blurred = await convert_to_gray_and_blur(img)
+    assert blurred is not None, "Image is None before thresholding"
+    print(blurred.shape, blurred.dtype)
+
+    (thresh, thresh_img) = cv2.threshold(
+        blurred, 0, 255, cv2.THRESH_BINARY + cv2.THRESH_OTSU
+    )
```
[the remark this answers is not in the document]
srishtysajeev: was just looking into this and it is confusing. they're int flags which tell the threshold method what kind of thresholding to use . Using cv2.THRESH_BINARY | cv2.THRESH_OTSU also works because cv2.THRESH_BINARY = 0. All examples from openCV do it this way, and I think the logic is hidden away within cv2 so it's probably best to keep it like this incase the numbers change in a later version of the library

PR #1768, opened 2025-12-08
Use config server for undulator config

jacob720:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/1494

Required by https://github.com/DiamondLightSource/mx-bluesky/pull/1497

1. Confirm config is read through the config server

### Checks for reviewer


src/dodal/devices/undulator.py
```diff
@@ -185,6 +186,7 @@ def __init__(
             baton (optional): Baton object if provided.
             name (str, optional): Name for device. Defaults to "".
         """
+        self.config_server = ConfigServer(url="https://daq-config.diamond.ac.uk")
```
Villtord: I am not sure if this hardcoded config server is a  good solution long term. I understand that it works fine now, but what if for some reason you need to divert your config server to a different endpoint - you would have to change the path here.
It feels like we need  some king of an interface or a lookuptable provider which should probably be configured separately.

Issue #1772, opened 2025-12-09
Investigate if we should restructure electron analyser devices


oliwenmandiamond:
As of current, the electron analyser devices are structured as follows:

```
devices/electron_analyser/
  ├── abstract/
  │     ├── base_region.py
  │     └── base_driver_io.py
  │     └── base_detector.py
  ├── vgscienta/
  │     ├── region.py
  │     └── driver_io.py
  │     └── detector.py
  ├── specs/
  │     ├── region.py
  │     └── driver_io.py
  │     └── detector.py
   ...
```

It might be better to restructure it as follows:
```
devices/electron_analyser/
  ├── region/
  │     ├── base_region.py
  │     └── vscienta_region.py
  │     └── specs_region.py
  ├── driver_io/
  │     ├── base_driver_io.py
  │     └── vgscienta_driver_io.py
  │     └── specs_driver_io.py
  ├── detector/
  │     ├── base_detector.py
  │     └── vgscienta_detector.py
  │     └── specs_detector.py
   ...
```
This will make it more like a plugin structure. The first structure doesn't make it easy to create any new set of supporting devices that don't fit into abstract, vgscienta or specs. @Villtord @Relm-Arrowny thoughts?

Villtord:
We (or me) would need to add MBS analyser as well, so whichever structure is easier to do that.

oliwenmandiamond:
Yep am aware, so think better to do this now before it gets any larger! I think the new structure is a big improvement and helps to group together the logic of devices better

PR #1777, opened 2025-12-10
Restructure electron analyser module like ophyd_async

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/1772

1. Check the structure makes sense
2. Tests pass

### Checks for reviewer


src/dodal/devices/electron_analyser/base/base_driver_io.py
Relm-Arrowny: nit: Should driver_io only contain signal and the rest of the logic live in controller?

PR #1778, opened 2025-12-11
I16 insertion device

DG-At-Diamond:
Fixes https://github.com/DiamondLightSource/sm-bluesky/issues/126

This adds the undulator for I16, setting the gap in mm.  It does not currently include the energy calculations or lookup table, these are performed in GDA at the moment and will be implemented in bluesky in a later change.

run dodal connect i16 and check the id is in the list 

### Checks for reviewer


DG-At-Diamond:
I have been looking into the energy calculations and think this should be merged as is and we can add that later on as a separate PR as there are some unanswered questions about how it's going to work.

PR #1781, opened 2025-12-11
Add function to get RunEngine metadata

### LINKED ISSUE #1779: Add get RunEngine metadata util function and callback
Sometimes our bluesky plans require information that is only stored in the `RunEngine`'s metadata, this is a bit fiddly to access when we don't have access to the `RunEngine` directly.

For example, in MX we need to know the visit in our plan so we can send it to ispyb, but when using BlueAPI, the visit information is only stored in the `RunEngine`

This issue is to create a utility function and callback to get specified metadata from the RunEngine. The callback should wait for a "get metadata" event document and then return the specified metadata. The function should subscribe to this callback, trigger the read, and unsubscribe to the callback, then return the metadata.

olliesilvester:
Fixes #1779 

I think this should allow us to get instrument_session within our plans when we're using BlueAPI with numtracker.

1. Check function does what it says it does
2. Check tests

### Checks for reviewer

### DIFF
--- a/src/dodal/plans/get_run_engine_metadata.py
+++ b/src/dodal/plans/get_run_engine_metadata.py
@@ -1,14 +1,9 @@
-from typing import Generic, Protocol, TypeVar
+from typing import Generic, TypeVar
 
 import bluesky.plan_stubs as bps
 import bluesky.preprocessors as bpp
 from bluesky.callbacks import CallbackBase
 from bluesky.utils import MsgGenerator
-
-
-class FromStr(Protocol):
-    def __init__(self, value: str) -> None: ...
-
 
 T = TypeVar("T", str, int, float)
 
@@ -43,6 +38,8 @@ def get_run_engine_metadata(
     Creates a callback and a dummy run event to read metadata from the RunEngine. Useful
     for plans which don't have reference to the RunEngine, for example when running through
     BlueAPI
+
+    Example usage: visit = get_run_engine_metadata("instrument_session")
 
     Args:
     metadata_name: The string name for requested metadata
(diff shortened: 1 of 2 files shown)


tests/plans/test_get_run_engine_metadata.py
```diff
@@ -0,0 +1,54 @@
...
+    test_float = 4.4
+
+    run_engine.md["test_string"] = test_string
+    run_engine.md["test_int"] = test_int
+    run_engine.md["test_float"] = test_float
+
+    def test_plan():
+        assert (yield from get_run_engine_metadata("test_string")) == test_string
+        assert (yield from get_run_engine_metadata("test_int", int)) == test_int
+        assert (yield from get_run_engine_metadata("test_float", int)) == test_float
```
oliwenmandiamond: I think you miss typed here causing your test to fail
```suggestion
        assert (yield from get_run_engine_metadata("test_float", float)) == test_float
```

src/dodal/plans/get_run_engine_metadata.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Args:
+    metadata_name: The string name for requested metadata
+    expected_datatype: The datatype you expect the data to be in. String by default.
+    """
+
+    callback = _GetRunEngineMetaDataCallback[T](metadata_name, expected_datatype)
+
+    @bpp.subs_decorator(callback)
+    @bpp.run_decorator()
```
tpoliaw: Won't making this a run in its own right end up with it getting its own scan numbers/run IDs etc with a call to numtracker for each call? Does creating nested runs cause issues in any of our downstream handling (tiled/nexus writer etc)?

If exposing run_engine metadata is something that is not possible with the run engine builtins, would it be worth registering a new `metadata` message handler so you could do something like

```
instrument = yield Msg("metadata", key="instrument")
```
Is there anything in the metadata that we don't want plans to be able to access?

olliesilvester:
Closing this for now since it misleading will increment scan id when using numtracker

olliesilvester:
Leaving this closed since we still don't need it in MX, but if anyone still wants this, it would work by:
- Requiring that you're already inside an open run when calling this function
- The function uses bps.create with a specific event name which the callback checks against to give the metadata

PR #1788, opened 2025-12-16
Converting MX Beamlines to device manager

### LINKED ISSUE #1765: Convert MX beamlines to new device manager
Following https://github.com/DiamondLightSource/dodal/pull/1549 being merged we should convert all the MX beamlines to use the new device manager. https://github.com/DiamondLightSource/dodal/pull/1684 is an example of how to do it for one beamline.

## Acceptance Criteria
* All MX beamlines use the new device manager

adaudon:
Fixes #1765 

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i09_2.py
+++ b/src/dodal/beamlines/i09_2.py
@@ -70,18 +70,18 @@ def jid_phase() -> UndulatorPhaseAxes:
 
 
 @device_factory()
-def jid() -> Apple2:
+def jid() -> Apple2[UndulatorPhaseAxes]:
     """I09 soft x-ray insertion device."""
-    return Apple2(
+    return Apple2[UndulatorPhaseAxes](
         id_gap=jid_gap(),
         id_phase=jid_phase(),
     )
 
 
 @device_factory()
-def jid_controller() -> Apple2EnforceLHMoveController:
+def jid_controller() -> Apple2EnforceLHMoveController[UndulatorPhaseAxes]:
     """J09 insertion device controller."""
-    return Apple2EnforceLHMoveController(
+    return Apple2EnforceLHMoveController[UndulatorPhaseAxes](
         apple2=jid(),
         gap_energy_motor_lut=ConfigServerEnergyMotorLookup(
             lut_config=LookupTableColumnConfig(poly_deg=J09_GAP_POLY_DEG_COLUMNS),
--- a/src/dodal/beamlines/i21.py
+++ b/src/dodal/beamlines/i21.py
@@ -74,18 +74,18 @@ def id_phase() -> UndulatorPhaseAxes:
 
 
 @device_factory()
-def id() -> Apple2:
+def id() -> Apple2[UndulatorPhaseAxes]:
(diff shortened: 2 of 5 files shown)


src/dodal/beamlines/i23.py
tpoliaw: This should use `path_provider` from a method parameter

tpoliaw: This should probably be replaced with a local fixture to provide this as a fallback if one is not provided. `i03.py` has had it done as an example

src/dodal/beamlines/i24.py
```diff
@@ -73,7 +73,7 @@ def aperture() -> Aperture:
     )
```
DominicOram: Could: Whilst we're here the formatting of a lot of these device instantiations is a bit odd. If we remove the trailing comma the save on format will make this one line. Same with many other devices in this file.

src/dodal/beamlines/i03.py
```diff
@@ -157,7 +157,7 @@ def detector_motion() -> DetectorMotion:
     )
 
 
-@devices.v1_init(EigerDetector, prefix="BL03I-EA-EIGER-01:", wait=False)
+@devices.v1_init(EigerDetector, prefix="BL04I-EA-EIGER-01:", wait=False)
```
DominicOram: Must: This is i03, the correct prefix is `BL03I`, again https://github.com/DiamondLightSource/dodal/pull/1788#discussion_r2673503510 would probably be a better fix

src/dodal/beamlines/i23.py
```diff
@@ -53,18 +52,12 @@ class I23DetectorPositions(StrictEnum):
 
 
 def _is_i23_machine():
-    """
-    Devices using PVA can only connect from i23 machines, due to the absence of
-    PVA gateways at present.
-    """
```
DominicOram: Should: This comment is still useful. It's just the generic "Get the i04 centring device, instantiate it if it hasn't already been. If this is called when already instantiated in i04, it will return the existing object." ones that aren't

src/dodal/beamlines/i24.py
```diff
@@ -180,57 +130,39 @@ def zebra() -> Zebra:
...
 
 
 @devices.factory()
 def commissioning_jungfrau(
+    path_provider: PathProvider,
     path_to_dir: str = "/tmp/jf",  # Device factory doesn't allow for required args,
     filename: str = "jf_output",  # but these should be manually entered when commissioning
 ) -> CommissioningJungfrau:
-    """Get the commissionning Jungfrau 9M device, which uses a temporary filewriter
-    device in place of Odin while the detector is in commissioning.
```
DominicOram: Should: The first part of this comment is still useful. It's just the generic "Get the i04 centring device, instantiate it if it hasn't already been. If this is called when already instantiated in i04, it will return the existing object." ones that aren't

DominicOram (changes requested):
Great, thank you! Some more comments in code to address.

PR #1791, opened 2025-12-17
Wait for smargon enable in robot load and unload

jacob720:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/1525

Instead of waiting on the pin sensor, the robot waits on the smargon disable PV to be low. This makes the robot logic more robust as before the robot could block the sensor and cause false positives (thinking a pin is loaded when it is not). This should also help make the robot load plans in mx-bluesky more generic (see https://github.com/DiamondLightSource/mx-bluesky/issues/1478).

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/robot.py
```diff
@@ -85,6 +90,9 @@ def __init__(self, prefix: str, name: str = "") -> None:
 
             self.current_puck = epics_signal_r(float, prefix + "CURRENT_PUCK_RBV")
             self.current_pin = epics_signal_r(float, prefix + "CURRENT_PIN_RBV")
+            self.smargon_enabled = epics_signal_r(
+                SmargonStatus, prefix + "ROBOT_OP_16_BITS.B8"
+            )
```
DominicOram: Should: I think it's going to confuse someone one day that we call this `smargon_enabled` but the synoptic calls it `Cryojet Retract Request`. Particularly as there's no reason this robot couldn't be used on a beamline with a different gonio to a smargon. I think a more generic name like `beamline_enabled` and a comment like "We assume that a robot action is finished, and that we are safe to proceed once the cryoject retract request is complete"

DominicOram: Must: I think the logic is also inverted. When this is high the smargon is disabled. So the name should be something like `beamline_disabled` or we change it in the enum (see above)

DominicOram: Must: If you run `dodal connect i03` you will see this fails. This is because a `StrictEnum` only works when the PV is an enum type with the same enum names. In this case the PV is a binary 0/1 with no mapping:
```
(.venv) [ffv81422@ws455 dodal]$ caget BL03I-MO-ROBOT-01:ROBOT_OP_16_BITS.B8
BL03I-MO-ROBOT-01:ROBOT_OP_16_BITS.B8 0
(.venv) [ffv81422@ws455 dodal]$ caget -d31 BL03I-MO-ROBOT-01:ROBOT_OP_16_BITS.B8
BL03I-MO-ROBOT-01:ROBOT_OP_16_BITS.B8
    Native data type: DBF_CHAR
    Element count:    1
    Value:            Enum Index Overflow (0)
    Status:           NO_ALARM
    Severity:         NO_ALARM
    Enums:            ( 0)
```
There actually doesn't seem to be a great way to do this, see https://github.com/bluesky/ophyd-async/issues/1168. So I think we will have to leave it as an `int` for now.

PR #1808, opened 2026-01-07
Convert adsim module to use device manager

tpoliaw:


src/dodal/beamlines/adsim.py
```diff
@@ -38,44 +37,47 @@
...
+@devices.fixture
+def path_provider():
+    from pathlib import Path
+
+    from ophyd_async.core import StaticPathProvider, UUIDFilenameProvider
+
+    return StaticPathProvider(
+        UUIDFilenameProvider(),
+        Path("/tmp"),  # The directory for `det` to write to- may be mounted as a volume
+    )
```
abbiemery: Doesn't adsim use numtracker and therefore not need this? I think it was in the comment in case the  infrastructure wasn't there. Perhaps put it back in the comment but with a note saying why its there.

abbiemery (approved):
The comment is helpful, thanks :)

Approved so it can be merged once the tests are happy.

PR #1815, opened 2026-01-14
Add rectangle class to math module

### LINKED ISSUE #1818: Make a helper rectangle class
For the purpose of Apple Knot insertion device undulator I need to define several rectangular zones in a 2D gap-phase space. It would be very helpful to have then a simple Rectangle class that has:
1. min/max for phase and gap methods.
2. can check if a given (phase,gap) point falls inside Rectangle.

Suggested existing` scanspec.regions.Rectangle` class in the imported `scanspec `package is too complicated and does not have required methods while having other parameters and methods not required/not known for this use case.

Villtord:
Fixes #1818 
IT might be handy to have it - specifically I will use it in the upcoming Apple Knot class which has rectangular exclusion zones.

1. Do look at the class
2. Confirm tests path

### Checks for reviewer


src/dodal/common/maths.py
```diff
@@ -48,3 +48,61 @@ def in_micros(t: float) -> int:
     if t < 0:
         raise ValueError(f"Expected a positive time in seconds, got {t!r}")
     return int(np.ceil(t * 1e6))
+
+
+class Rectangle2D:
```
tpoliaw: Can you get non 2D rectangles?

DominicOram (changes requested):
> Emm, no - I didn't, but for such simple thing I don't think it's necessary to rely on "yet another external python package" which can change with time

I disagree, [`scanspec`](https://github.com/bluesky/scanspec) is:
* Already in the bluesky ecosystem
* Actively maintained by people inside DLS
* Already imported by `dodal`

If it doesn't fit the usecase that's fine but then can you link to a PR/issue explaining the usecase? I'm worried that this is a slippery slope to reinventing existing python libraries (which is particularly egregious if we're reinventing something written in the next office to ours)

src/dodal/common/maths.py
```diff
@@ -48,3 +48,61 @@ def in_micros(t: float) -> int:
...
+            float: The larger of the two y-coordinates (y1, y2).
+        """
+        return max(self.y1, self.y2)
+
+    def contains(self, x: float, y: float) -> bool:
+        """
+        Check if a point is contained within the rectangle.
+
+        Args:
+            point (list[float]): A list containing [x, y] coordinates of the point.
```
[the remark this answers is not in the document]
Villtord: Yes, thanks for spotting it!

src/dodal/common/maths.py
Villtord: do you mean higher dimensions or lower? This class is only for 2D rectangles as of now

Villtord:
With all respect - having a package in bluesky ecosystem does not guarantee it won't break downstream packages in a next release as proven countless times by other dependent packages in bluesky ecosystem (ophyd-async - dodal - daq-server and so on)

Also, I don't see what can be slippery in writing a simplest possible internal util Rectangle class. Sometimes reinventing simple things is not bad and gives more control and stability.

With this being said - I don't think scanspec.region.Rectangle class fit into my simple purpose - I will write an issue explaining why. Thanks for looking at this PR!

PR #1819, opened 2026-01-15
Mx bluesky 1354 scintillator safe move

rtuck99:
Fixes
* DiamondLightSource/mx-bluesky#1354
* DiamondLightSource/mx-bluesky#1540

See also mx-bluesky PR
* DiamondLightSource/mx-bluesky#1551

1. The aperture scatterguard is now moved sideways to the corresponding SCIN_MOVE position for its current position and the position restored on successful scintillator move.
2. Beamstop position is checked that it is either in data collection or out-of-beam position, any other unrecognised or unsupported position prevents a scintillator move.

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i03.py
+++ b/src/dodal/beamlines/i03.py
@@ -1,6 +1,6 @@
 from functools import cache
 
-from ophyd_async.core import PathProvider
+from ophyd_async.core import PathProvider, Reference
 from ophyd_async.fastcs.eiger import EigerDetector as FastEiger
 from ophyd_async.fastcs.panda import HDFPanda
 from yarl import URL
@@ -342,10 +342,14 @@ def fluorescence_det_motion() -> FluorescenceDetector:
 
 
 @devices.factory()
-def scintillator() -> Scintillator:
+def scintillator(
+    aperture_scatterguard: ApertureScatterguard, beamstop: Beamstop
+) -> Scintillator:
     return Scintillator(
         f"{PREFIX.beamline_prefix}-MO-SCIN-01:",
         get_beamline_parameters(),
+        Reference(aperture_scatterguard),
+        Reference(beamstop),
     )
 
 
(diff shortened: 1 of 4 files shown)


src/dodal/devices/scintillator.py
```diff
@@ -77,32 +80,48 @@ def _get_selected_position(self, y: float, z: float) -> InOut:
...
-                f"Cannot move scintillator if aperture/scatterguard is not parked. Position is currently {await self._aperture_scatterguard().selected_aperture.get_value()}"
-            )
+    async def _check_beamstop_position(self):
+        position = await self._beamstop().selected_pos.get_value()
+        match position:
+            case BeamstopPositions.OUT_OF_BEAM | BeamstopPositions.DATA_COLLECTION:
+                return
+            case _:
+                raise ValueError(
+                    f"Scintillator cannot be moved due to beamstop position {position}, must be in either in DATA_COLLECTION or OUT_OF_BEAM position."
```
olliesilvester: Our scintillator and aperture scatterguard devices are currently in the "common device" section of dodal. Are we sure that the collision checks here are valid for a generic beamline? Maybe @DominicOram has some idea?

src/dodal/devices/scintillator.py
```diff
@@ -77,32 +80,48 @@ def _get_selected_position(self, y: float, z: float) -> InOut:
...
-                    return
-                await self._check_aperture_parked()
+        current_y = await self.y_mm.user_readback.get_value()
+        current_z = await self.z_mm.user_readback.get_value()
+        if self._get_selected_position(current_y, current_z) == position:
+            return
+
+        await self._check_beamstop_position()
+
+        async def move_to_new_position():
```
olliesilvester: I might have missed something but I can't see why this needs to be an inner function which is passed to `do_with_ap_sg_in_safe_pos`? I think this can just be an outer class method, which takes `position` as a parameter, so the code only ever defines it once

PR #1841, opened 2026-01-20
Ensure path_providers are available when required

tpoliaw:
Several beamline modules assume a path provider has been set. This worked until
recently but as more beamlines convert to using the device manager, these tests
are becoming unreliable.

For now, this adds a fixture to ensure a clean state at the start of each test
and sets a path provider if one has not already been set. This should allow them
to continue working with a path provider set by blueapi while letting the tests
pass more reliably in CI.

### DIFF
--- a/src/dodal/device_manager.py
+++ b/src/dodal/device_manager.py
@@ -360,6 +360,37 @@ class DeviceManager:
         self._fixtures[func.__name__] = func
         return func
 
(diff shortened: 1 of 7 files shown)


DominicOram (changes requested):
Actually, on a second look. This presumably needs to be done on every beamline where `get_path_provider` is used but it doesn't use the new device manager? In which case, there's some missing?

Issue #1481, opened 2025-08-26
Write a style guide


DominicOram:
Whilst we follow PEP8, there are more general style questions that we do not have written down. Particularly relevant to people coming from Java are things like when to use objects and when to use getters/setters. We should write a guide to this that highlights major differences in style when coming from Java.

## Acceptance Criteria
* There is a style guide in the docs that covers some of these broader questions

## Notes
* I would personally just link to https://google.github.io/styleguide/pyguide.html. I think I agree with most of it and saves us having to write something or argue over it. I think it would still be good to just highlight some of the major differences to java next to that link though

DiamondJoseph:
I don't agree with "Use import statements for packages and modules only, not for individual types, classes, or functions.", but I care more about standardising than I do about specifics.

DominicOram:
I agree with enabling more ruff rules as we discussed at https://github.com/DiamondLightSource/dodal/pull/1112#issuecomment-2830383051. I think the main reason that didn't get merged was it just got out of date and was such a big change. We should revisit how to do it.

The main two things that I've come across recently, that ruff can't catch are:
* Don't use getters and setters if they're not needed - https://google.github.io/styleguide/pyguide.html#2134-decision
* Use utility functions over `staticmethod` - https://google.github.io/styleguide/pyguide.html#2174-decision

These are the kinds of things that really trip up people coming from Java. 

I think I agree with your list of exceptions though.

callumforrester:
I agree with `staticmethod`s, on the fence with getters and setters, for example I really wish I'd added them to the `BlueskyContext` in blueapi.

> For example, using a property to simply both get and set an internal attribute isn’t allowed: there is no computation occurring, so the property is unnecessary ([make the attribute public instead](https://google.github.io/styleguide/pyguide.html#getters-and-setters)).

The Java argument is that just because there's no computation occurring _now_, there might be one day, and you're less likely to have to make a breaking change if you've already encapsulated your data. Currently the `BlueskyContext` will just give you mutable references to dicts of plans and devices to do with as you please, it's a bit horrible. The only place I like to just have public attributes is pure data classes.

PR #1844, opened 2026-01-21
Add glossary and style guide

### LINKED ISSUE #876: Add a dodal glossary and style guide
There are some things which we have discussed which are not easily enforceable through linting, we should add a document about best practices, e.g.

- how to use the new `Reference` to add private references to other devices or signals
- naming devices and their associated enums
  - enums that live in a device file don't need to include the device name
  - composite devices which exist to expose a synthetic signal should be named after that synthetic signal and not the component devices
-

EmsArnold:
Adds a glossary of terms, and style guide.

Closes #876 and #1481.

1. Do explanations make sense
2. Are all acronyms/abbreviation/terms included which are currently in use
3. Is guidance in style guide well reasoned


Villtord (commented):
Thanks it looks great!
I am not a native English speaker so please double check my suggestions :)

PR #1873, opened 2026-01-28
Add i05 sample stage

oliwenmandiamond:
Fixes #ISSUE

1. Check dodal connect works
2. Check tests pass

### Checks for reviewer


oliwenmandiamond:
The only other beamline I can see using it is i05-1 which creates this class
```
	<bean id="smvirtual" class="gda.device.scannable.scannablegroup.ScannableGroup">
		<property name="groupMembers">
			<list>
				<ref bean="smvert" />
				<ref bean="smhor" />
				<ref bean="smlong" />
				<ref bean="smperp" />
			</list>
		</property>
	</bean>
```
Seems to be a object design to strictly rotate the object using these transformations and also adds horizontal and vertical rotation to it as well. If we also want to add this, what we can do is add a helper function in `i05_shared` to create them and then we can add them to any device e.g for i05 we use it for `Gonio` and for i05-1 we create `smvirtual` object. Unless we don't follow gda way and instead just add them directly onto the device e.g `XYZPolarAzimuthDefocusStage`

Villtord:
Thanks! It does look much nicer using rotation matrices.
I personally think the way it is now is a bit of an overkill.
1. Logically 2D vector class should probably not have a matrix parameter inside for coordinate transformation or at least it does not look like it is a property of 2D vector IMO.
2. Since we have rotation angle fixed and unlikely to change often, and rotation  matrix coefficients probably not needed to be generalized as callables as they are anyway (at the moment at least) used only at i05 - would be much easier just define a constant rotation matrix.
3. As a consequence you probably don't need a 2D vector class at all but just use a couple of methods from it (rotate and reverse) as util functions?

Once we gain more use cases we could try to make it more generalised for sure.

src/dodal/devices/beamlines/i05_shared/rotation_signals.py
```diff
@@ -0,0 +1,89 @@
...
+    if isinstance(angle_deg, SignalR):
+        return await angle_deg.get_value()
+    return angle_deg
+
+
+def create_rotational_ij_component_signals(
+    i_read: SignalR[float],
+    j_read: SignalR[float],
+    i_write: Movable[float],
+    j_write: Movable[float],
```
Villtord: Why is there a separation of read and write if there could be just one SignalRW ?

oliwenmandiamond: No, it can't unfortunately. I tried this initially but it doesn't work because our signals are `Motors`. `Motors` are `Movable[float]` and also have` user_readback` signal of type `SignalR[float]'. So to make this compatible with all devices / Signals, we have to separate the read and write parts to make it work.

oliwenmandiamond: Just to add on to this, I can't use `Motor` as the type either because one of the derived signals also uses another derived signal which is not  a `Motor`. So separate write and read inputs is only way to make it generic enough to work for each case.

PR #1888, opened 2026-02-04
Add i21 sample stage

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/1964

1. Check tests pass
2. Logic makes sense

### Checks for reviewer

### DIFF
--- a/tests/devices/beamlines/i21/test_i21_motors.py
+++ b/tests/devices/beamlines/i21/test_i21_motors.py
@@ -1,5 +1,6 @@
 import pytest
 from ophyd_async.core import init_devices
+from ophyd_async.testing import assert_reading, partial_reading
 
 from dodal.devices.beamlines.i21 import I21SampleManipulatorStage
 
@@ -11,5 +12,18 @@ def smp() -> I21SampleManipulatorStage:
     return smp
 
 
-def test_smp_read(smp: I21SampleManipulatorStage) -> None:
-    pass
+async def test_smp_read(smp: I21SampleManipulatorStage) -> None:
+    await assert_reading(
+        smp,
+        {
+            smp.x.name: partial_reading(0),
+            smp.y.name: partial_reading(0),
+            smp.z.name: partial_reading(0),
+            smp.azimuth.name: partial_reading(0),
+            smp.tilt.name: partial_reading(0),
+            smp.polar.name: partial_reading(0),
+            smp.difftth.name: partial_reading(0),
+            smp.para.name: partial_reading(0),
+            smp.perp.name: partial_reading(0),
+        },
+    )


src/dodal/devices/beamlines/i21/toolpoint_motion.py
```diff
@@ -0,0 +1,227 @@
...
+    return np.array(
+        [
+            [cos(tilt), sin(tilt) * sin(azimuth), sin(tilt) * cos(azimuth)],
+            [0.0, cos(azimuth), -sin(azimuth)],
+            [-sin(tilt), cos(tilt) * sin(azimuth), cos(tilt) * cos(azimuth)],
+        ]
+    )
+
+
+def toolpoint_to_xyz(
```
fajinyuan: would it better to name it as uvw_to_xyz?

src/dodal/devices/beamlines/i21/toolpoint_motion.py
```diff
@@ -0,0 +1,227 @@
...
+    axes u, v, and w as signals.
+    """
+
+    def __init__(
+        self,
+        smp: XYZAzimuthTiltStage,
+        zero: tuple[float, float, float] = DEFAULT_AXES_ZERO,
+        name: str = "",
+    ):
+        self.smp_ref = Reference(smp)
```
fajinyuan: please tell me why you have to create smp_ref here, why it cannot use 'smp' directly?

src/dodal/devices/motors.py
```diff
@@ -116,16 +124,16 @@ def __init__(
...
         with self.add_children_as_readables():
-            self.polar = Motor(prefix + polar_infix)
+            self.azimuth = Motor(prefix + azimuth_infix)
         super().__init__(prefix, name, x_infix, y_infix, z_infix)
 
 
-class XYZPolarAzimuthStage(XYZPolarStage):
-    """Five-axis stage with a standard xyz stage and two axis of rotation: polar and
-    azimuth.
+class XYZAzimuthPolarStage(XYZAzimuthStage):
```
fajinyuan: a general comment on motor class here - why choose to use inheritance instead of composition?

src/dodal/beamlines/i21.py
```diff
@@ -129,3 +133,15 @@ def energy(
...
+
+@devices.factory()
+def smp() -> XYZAzimuthTiltPolarParallelPerpendicularStage:
+    return XYZAzimuthTiltPolarParallelPerpendicularStage(
+        prefix=f"{PREFIX.beamline_prefix}-EA-SMPL-01:"
+    )
+
+
+@devices.factory()
+def uvw(smp: XYZAzimuthTiltPolarParallelPerpendicularStage) -> ToolPointMotion:
```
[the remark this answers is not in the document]
oliwenmandiamond: Yes, this is new Athena way. Devices are created as functions which are given to the `DeviceManager`. If first time function called, create the instance. If already called before, use the cahce. `DeviceManager` then gives these to BlueAPI to tell user what the available devices are and managers the api to interact with them.

Users will never directly interact with dodal, only through blueapi. The only time they need visibility of dodal is for type checking of their plans/scripts or to get the enums a PV takes

src/dodal/devices/beamlines/i21/toolpoint_motion.py
```diff
@@ -0,0 +1,227 @@
...
+    AsyncStatus,
+    Reference,
+    SignalRW,
+    StandardReadable,
+    derived_signal_rw,
+)
+
+from dodal.devices.motors import XYZAzimuthTiltStage
+
+DEFAULT_AXES_ZERO = (5.667, 0.2408, 1.221)
```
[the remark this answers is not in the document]
oliwenmandiamond: As far as I can tell, these values never change and is just configuration given to the device when creating the device for the offset. 

If beamline scientists need access to this, we could make it a soft signal. I don't see any evidence in GDA that they do though.

Issue #1902, opened 2026-02-05
Convert the final beamlines to new device factory and remove old device factory


DominicOram:
Following https://github.com/DiamondLightSource/dodal/issues/1900 there are only a few beamlines left that use the old device factory:

* Aithre - I can do this one as it's MX
* b01_1
* b16
* b18
* i13_1
* i18
* k11
* p38
* p45
* p51

Once these are converted we can then remove any code that references `device_factory` and `set_path_provider`/`get_path_provider`

## Acceptance Criteria
* All beamlines are using the new device factory
* All old device factory code is removed

PR #1903, opened 2026-02-05
Move aithre to new device factory

DominicOram:
Part of #1902 

1. Confirm the change looks correct (annoyingly you cannot remote connect to aithre)

### Checks for reviewer


src/dodal/beamlines/aithre.py
tpoliaw: could: You could create a `params` fixture for these and leave this as `config = params`

tpoliaw (approved):

PR #1917, opened 2026-02-11
Add logging around global path provider use

tpoliaw:
Intended to give us some insight into who is using what before we remove the
global path provider.


src/dodal/common/beamlines/beamline_utils.py
```diff
@@ -26,6 +27,9 @@
 BL = ""
 
 
+LOGGER = logging.getLogger(__name__)
```
DominicOram: Should: I think you probably want to use the logger in `log.py`. Then you'll get the beamline tag etc. e.g. `from dodal.log import LOGGER`

PR #1921, opened 2026-02-13
add energy and id to i05-shared

Villtord:
Add energy and insertion device control (AppleKnot) for i05 beamline.
Fixes https://github.com/DiamondLightSource/sm-bluesky/issues/259

1. Do dodal connect i05 and dodal connect i05-1
2. Confirm tests pass

### Checks for reviewer


src/dodal/devices/beamlines/i05_shared/apple_knot.py
```diff
@@ -0,0 +1,92 @@
+import numpy.polynomial as poly
```
oliwenmandiamond: Nit: Why not just import Polynomial directly as that is only thing it is used for?

oliwenmandiamond: Maybe rename file to apple_knot_constants/definitions/config.py?

Villtord: done

PR #1929, opened 2026-02-19
Support for mod-360 omega

rtuck99:
Required for

* DiamondLightSource/mx-bluesky#1598

Introduces a `wrapped_omega` `Readable` to `XYZOmegaStage` which delegates motion to the underlying omega motor.

`wrapped_omega` has two signals:
   * `phase` which is read-write and represents the phase angle with respect to an offset, which is a multiple of 360 degrees.
   * `offset_and_phase` which returns a numpy array containing the offset and phase values together. This can be retrieved in plans and events and used to instantiate an instance of `AngleWithPhase` which can be used as a convenient value-object helper for conversion and manipulation of phase angles, while retaining information about the mapping to the underlying unwrapped real coordinate.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


tests/devices/test_motors.py
```diff
@@ -358,3 +363,120 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+        [-10000 * 360 - 0.001, 359.999],
+    ],
+)
+async def test_mod_360_read(real_value: float, expected_phase):
+    stage = XYZOmegaStage("BL03I-MO-SGON-01:")
+    await stage.connect(mock=True)
+    set_mock_value(stage.omega.user_readback, real_value)
+    offset, phase = await stage.omega_axis.offset_and_phase.get_value()
+    assert 0 <= expected_phase < 360
+    assert math.isclose(phase, expected_phase, abs_tol=1e-6)
```
jacob720: Could: add expected offset to this test

jacob720 (commented):
Looks good! Logic is clean and easy to follow

DominicOram (dismissed):
Thanks, some comments

src/dodal/devices/motors.py
```diff
@@ -102,7 +110,10 @@ def __init__(
         omega_infix: str = _OMEGA,
     ) -> None:
         with self.add_children_as_readables():
-            self.omega = Motor(prefix + omega_infix)
+            real_motor = Motor(prefix + omega_infix)
+            self.omega = real_motor
+            self.omega_axis = WrappedAxis(self.omega)
```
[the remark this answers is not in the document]
oliwenmandiamond: Okay, I see. I was referring to the current implementation.

Yes a protocol could work. If this could wait for `[StandardMovable](https://github.com/bluesky/ophyd-async/pull/1200)`, you could make the protocol

```Python
class XYZOmega(Protocol):
    x: StandardMovable
    y: StandardMovable
    z: StandardMovable
    omega: StandardMovable
```

Then you could implement your own `MovableLogic` rather than use the `MotorMovableLogic`, then it doesn't matter what the implementation is for the plan.

src/dodal/devices/smargon.py
```diff
@@ -104,6 +105,15 @@ def __init__(self, prefix: str, name: str = ""):
...
+        if motor_name == "omega":
+            current_angle = AngleWithPhase.from_offset_and_phase(
+                await self.omega_axis.offset_and_phase.get_value()
+            )
+            return current_angle.nearest_to_phase(value).unwrap()
+        else:
+            return value
+
     @AsyncStatus.wrap
     async def set(self, value: CombinedMove):
```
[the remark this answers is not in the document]
rtuck99: I think making the underlying motor object represent phase is a bad idea for reasons I will state elsewhere in this PR, however I agree that having the omega coordinate as the only wrapped value in `CombinedMove` is somewhat ugly. The way I see, it the options are:

* `CombinedMove.omega` is unwrapped and the plan unwraps the coordinate before specifying it - also ugly, and means you have to read the current position in the plan.
* `CombinedMove.omega_phase` becomes a thing and you can specify the phase there, specifying `omega_phase`
* We remove support for setting omega from `CombinedMove` - following #1998 it seems the consensus is that omega moves should be serialized anyway because we've decided there are likely hardware limitations that mean we can't reliably do simultaneous fast omega moves at the same time as other axes. I think this is what I will ultimately do in the follow-on PR
* We implement a separate move_to_robot_load which just moves to 0,0,0,0,0,mod-360 - since this is the only time we need to do the combined move apart from moving to the next xtal. We could do this in conjunction with # 3 if we want to make this an explicit smargon feature, but I think it's not strictly necessary.

src/dodal/common/maths.py
```diff
@@ -144,3 +154,100 @@ def rotate_clockwise(theta: float, x: float, y: float) -> tuple[float, float]:
...
+        target_value = current_position.nearest_to_phase(value).unwrap()
+        await self._real_motor().set(target_value)
+
+    def distance(self, theta1_deg: float, theta2_deg: float) -> float:
+        """Obtain the shortest distance between theta2 and theta1 in degrees.
+        This will be the shortest distance in mod360 space (i.e. always <= 180 degrees).
+        """
+        return AngleWithPhase.wrap(theta1_deg).phase_distance(
+            AngleWithPhase.wrap(theta2_deg)
+        )
```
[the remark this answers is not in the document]
rtuck99: I have moved this

src/dodal/devices/smargon.py
```diff
@@ -114,11 +124,15 @@ async def set(self, value: CombinedMove):
         only come back after the motion on that axis finished.
         """
         await self.defer_move.set(DeferMoves.ON)
+        # TODO something something i03 broken smargon serialise moves workaround
```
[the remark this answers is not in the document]
rtuck99: See #1998

src/dodal/common/maths.py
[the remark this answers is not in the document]
oliwenmandiamond: Thanks for separating it out from the maths. However, the wrapped_axis.,py still needs to live under `dodal/devices/wrapped_axis.py` as it is a device.

src/dodal/devices/motors.py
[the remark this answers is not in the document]
rtuck99: I think that making `omega` motor use a wrapped coordinate system is a Bad Idea - this would be a potentially breaking change for plans that use this motor, and the changes might not be obvious to users that are not consciously considering the implications.

I have updated the PR and now all the plans that are in use by Hyperion are wrapped-omega aware, so a change to make omega a wrapped axis would be relatively trivial. However there would potentially be an impact on other beamlines if they inherit from `XYZOmegaStage`.

I don't think that `Motor`s with wrapped coordinate systems are appropriate, since speed limits and motor limits don't make sense within the wrapped coordinate space. Certainly it is not a drop-in replacement for the current unwrapped coordinate motor for all the reasons I have listed in wrapped_axis.py and I would not be in favour of it.

If you really want to replace `omega` with the wrapped equivalent, then I would strongly advise doing it in a device class that is not going to be used outside of MX to minimise the impact. Currently some of the MX plans are used by e.g. the aithre pin tip centring which do not have a standard gonio, therefore the base class we accept is `XYZOmegaStage` so at a minimum they would need to use whatever base device class we decide to use, and they would have to ensure  their plans weren't affected by a change to a wrapped omega. Since having access to only the phase angle represents a loss of information, unless all impacted plans can be adapted to use only the phase angle, it is likely that it will still be necessary to provide access to the underlying unwrapped angle.

As it currently stands adding a new wrapped axis is essentially free to any device and has no implications other than requiring the underlying axis to have unconstrained rotation, and the possibility that wrapped-unaware plans calling wrapped-aware plans may sometimes find the axis 360 degrees from where they might have expected it, which for most cases will not be an issue.

On the other hand, it is also possible to argue that we could do away with WrappedAxis altogether since you can always compute everything from the unwrapped coordinate space. However I do think it is nice to be able to have the phase angle accessible at the device level as a convenience to set directly (since for the majority of moves, this is the coordinate space of interest), and to expose as a `Readable` in bluesky events, as well as making it more obvious in the code when we are considering phase angle as opposed to absolute angle.

src/dodal/devices/wrapped_axis.py
```diff
@@ -0,0 +1,73 @@
...
+    StandardReadableFormat,
+    derived_signal_rw,
+)
+from ophyd_async.epics.motor import Motor
+
+from dodal.common.maths import AngleWithPhase, MotorOffsetAndPhase
+
+
+class WrappedAxis(StandardReadable):
+    """This device is a wrapper around a rotational Motor that presents the unwrapped coordinate system of the
```
oliwenmandiamond, line 14: This is a `Device`, it must be moved to `dodal/devices/wrapped_axis.py`

src/dodal/devices/motors.py
```diff
@@ -113,7 +117,10 @@ def __init__(
         omega_infix: str = _OMEGA,
     ) -> None:
         with self.add_children_as_readables():
-            self.omega = Motor(prefix + omega_infix)
+            real_motor = Motor(prefix + omega_infix)
+            self.omega = real_motor
+            self.wrapped_omega = WrappedAxis(self.omega)
+
         super().__init__(prefix, name, x_infix, y_infix, z_infix)
```
oliwenmandiamond, line 136: You are editing a class which is common and meant to be generic for all beamlines to use. This is extended off for `SixAxisGonio` and `SixAxisGonioKappaPhi` which is used by i15, i23. It is also extended off of `Smargon` for i03 and i04 (your desired MX beamlines?) and `I05Goniometer` used on i05.

You should not be altering the base class that will effect all other beamlines that don't need this behaviour. You need to create an MX specific device that doesn't effect others or roll this into `Smargon`.

oliwenmandiamond (changes requested):

tests/devices/test_motors.py
```diff
@@ -358,3 +363,120 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+        [-0.001, 359.999],
+        [180.001, 180.001],
+        [360.001, 0.001],
+        [719.999, 359.999],
+        [720.001, 0.001],
+        [10000 * 360 + 0.001, 0.001],
+        [-10000 * 360 - 0.001, 359.999],
+    ],
+)
+async def test_mod_360_read(real_value: float, expected_phase):
```
oliwenmandiamond:
```suggestion
async def test_mod_360_read(real_value: float, expected_phase: float):
```

tests/devices/test_motors.py
```diff
@@ -358,3 +363,120 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+        [-720.001, 270, 359.999],
+        [10000 * 360 + 0.001, 360000 - 90, 360000.001],
+        [-10000 * 360 - 0.001, -360000 + 90, -360000.001],
+    ],
+    ids=lambda values: f"input={values[0]}, current={values[1]}, expected={values[2]}",
+)
+def values_for_rotation(request):
+    input_value, current_real_value, expected_real_value = request.param
+    yield input_value, current_real_value, expected_real_value
+
```
oliwenmandiamond:
```suggestion
def values_for_rotation(request: pytest.FixtureRequest) -> tuple[float, float, float]:
    input_value, current_real_value, expected_real_value = request.param
    yield input_value, current_real_value, expected_real_value
```

tests/devices/test_motors.py
```diff
@@ -358,3 +363,120 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+
+@pytest.fixture()
+async def stage_in_initial_state(values_for_rotation):
+    input_value, current_real_value, expected_real_value = values_for_rotation
+
+    stage = XYZOmegaStage("BL03I-MO-SGON-01:")
+    await stage.connect(mock=True)
+    set_mock_value(stage.omega.user_readback, current_real_value)
+    return stage
```
oliwenmandiamond:
```suggestion
@pytest.fixture()
async def stage_in_initial_state(values_for_rotation: tuple[float, float, float]) -> XYZOmegaStage:
    input_value, current_real_value, expected_real_value = values_for_rotation
    with init_devices(mock=True):
          stage_in_initial_state = XYZOmegaStage("BL03I-MO-SGON-01:")
    set_mock_value(stage_in_initial_state.omega.user_readback, current_real_value)
    return stage_in_initial_state
```

tests/devices/test_motors.py
```diff
@@ -358,3 +363,120 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+
+
+async def test_mod_360_expected_direction_of_rotation_same_as_apparent_for_moves_apparently_less_than_180(
+    values_for_rotation, stage_in_initial_state
+):
```
oliwenmandiamond:
```suggestion

async def test_mod_360_expected_direction_of_rotation_same_as_apparent_for_moves_apparently_less_than_180(
    values_for_rotation: tuple[float, float, float], stage_in_initial_state: XYZOmegaStage
) -> None:
```

tests/devices/test_motors.py
```diff
@@ -358,3 +363,120 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+        expected_real_value < current_real_value
+    )
+    assert is_no_move or is_ge_than_180 or motion_in_same_direction_as_apparent, (
+        f"expected ({input_value} - {current_real_value}) same sign as ({expected_real_value} - {current_real_value})"
+    )
+
+
+async def test_mod_360_expected_actual_movement_never_more_than_180(
+    values_for_rotation, stage_in_initial_state
+):
```
oliwenmandiamond, line 531:
```suggestion
async def test_mod_360_expected_actual_movement_never_more_than_180(
    values_for_rotation: tuple[float, float, float], stage_in_initial_state: XYZOmegaStage
) -> None:
```

tests/devices/test_motors.py
```diff
@@ -358,3 +363,120 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+    input_value, current_real_value, expected_real_value = values_for_rotation
+    assert abs(expected_real_value - current_real_value) <= 180
+
+
+async def test_mod_360_unwrap_computes_expected(
+    values_for_rotation, stage_in_initial_state, run_engine: RunEngine
+):
```
oliwenmandiamond:
```suggestion

async def test_mod_360_unwrap_computes_expected(
    values_for_rotation: tuple[float, float, float], stage_in_initial_state: XYZOmegaStage, run_engine: RunEngine
) -> None:
```

tests/devices/test_smargon.py
```diff
@@ -215,3 +245,7 @@ async def test_given_motor_does_not_change_setpoint_then_deferred_move_times_out
 
     with pytest.raises(TimeoutError):
         await smargon.set(CombinedMove(x=10))
+
+
+def test_smargon_deferred_moves_move_omega_phase_not_absolute():
+    pass
```
oliwenmandiamond: Remove?

src/dodal/devices/smargon.py
```diff
@@ -104,6 +105,15 @@ def __init__(self, prefix: str, name: str = ""):
 
         super().__init__(prefix, name)
 
+    async def _get_target_value(self, motor_name: str, value: float):
+        if motor_name == "omega":
```
oliwenmandiamond:
```suggestion
    async def _get_target_value(self, motor_name: str, value: float) -> float:
        if motor_name == "omega":
```

src/dodal/common/maths.py
```diff
@@ -146,3 +152,101 @@ def rotate_clockwise(theta: float, x: float, y: float) -> tuple[float, float]:
...
+        Args:
+            offset (float | Iterable[float]): the offset in degrees, or the
+            offset and phase as a list or other iterable
+            phase (float): the phase in degrees
+        """
+        if isinstance(offset, Iterable):
+            offset, phase = offset
+        correction = 360 * (phase // 360)
+        self.offset: float = offset + correction
+        self.phase: float = phase - correction
```
oliwenmandiamond: I think this is a better fit for a dataclass


```suggestion
@dataclass
class AngleWithPhase:
    """Represents a point in an absolute rotational space which is defined by a phase where 0<=phase<360
     and an offset from an origin where the absolute coordinate is the sum of the phase and the offset.

    Attributes:
        offset: The offset of 0 phase from some other unwrapped rotational coordinate space
        phase: The phase in degrees relative to this offset.
    """

    offset: float
    phase: float = 0.0

    def __post_init__(self) -> None:
        correction = 360 * (self.phase // 360)
        self.offset += correction
        self.phase -= correction

    @classmethod
    def from_iterable(cls, values: Iterable[float]) -> Self:
        """Construct a normalised representation of the offset and phase, such that
        0 <= phase < 360.

        Args:
            offset Iterable[float]: the offset and phase as a list or other iterable
        """
        offset, phase = values
        return cls(offset, phase)
```

src/dodal/devices/motors.py
```diff
@@ -13,7 +13,11 @@
 )
 from ophyd_async.epics.motor import Motor
 
-from dodal.common.maths import rotate_clockwise, rotate_counter_clockwise
+from dodal.common.maths import (
+    rotate_clockwise,
+    rotate_counter_clockwise,
+)
```
oliwenmandiamond:
```suggestion
from dodal.common.maths import rotate_clockwise, rotate_counter_clockwise
```
Undo import change

tests/devices/test_motors.py
```diff
@@ -66,6 +69,16 @@ async def xyzpyr_stage() -> XYZPitchYawRollStage:
     return xyzpyr_stage
 
 
+@pytest.fixture()
+async def xyz_wrapped_omega_stage(values_for_rotation):
```
oliwenmandiamond:
```suggestion
async def xyz_wrapped_omega_stage(values_for_rotation: Iterable[float]) -> XYZWrappedOmegaStage:
```

tests/devices/test_motors.py
```diff
@@ -358,3 +388,110 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+        [719.999, -270, -360.001],
+        [-720.001, 270, 359.999],
+        [10000 * 360 + 0.001, 360000 - 90, 360000.001],
+        [-10000 * 360 - 0.001, -360000 + 90, -360000.001],
+    ],
+    ids=lambda values: f"input={values[0]}, current={values[1]}, expected={values[2]}",
+)
+def values_for_rotation(request):
+    input_value, current_real_value, expected_real_value = request.param
+    yield input_value, current_real_value, expected_real_value
```
oliwenmandiamond:
```suggestion
def values_for_rotation(request: pytest.FixtureRequest) -> tuple[float, float, float]:
    input_value, current_real_value, expected_real_value = request.param
    yield input_value, current_real_value, expected_real_value
```

tests/devices/test_motors.py
```diff
@@ -358,3 +388,110 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+        [-0.001, 359.999],
+        [180.001, 180.001],
+        [360.001, 0.001],
+        [719.999, 359.999],
+        [720.001, 0.001],
+        [10000 * 360 + 0.001, 0.001],
+        [-10000 * 360 - 0.001, 359.999],
+    ],
+)
+async def test_mod_360_read(real_value: float, expected_phase):
```
oliwenmandiamond:
```suggestion
async def test_mod_360_read(xyz_wrapped_omege_stage: XYZWrappedOmegaStage, real_value: float, expected_phase: float):
```
Why don't we use the already created stage?

tests/devices/test_motors.py
```diff
@@ -66,6 +69,16 @@ async def xyzpyr_stage() -> XYZPitchYawRollStage:
     return xyzpyr_stage
 
 
+@pytest.fixture()
+async def xyz_wrapped_omega_stage(values_for_rotation):
+    input_value, current_real_value, expected_real_value = values_for_rotation
+
+    async with init_devices(mock=True):
+        gonio = XYZWrappedOmegaStage("BL03I-MO-SGON-01:")
```
oliwenmandiamond:
```suggestion
        xyz_wrapped_omega_stage = XYZWrappedOmegaStage("BL03I-MO-SGON-01:")
```

tests/devices/test_motors.py
```diff
@@ -311,6 +324,23 @@ async def test_reading_six_axis_gonio(six_axis_gonio: SixAxisGonio):
...
+):
+    await assert_reading(
+        xyz_wrapped_omega_stage,
+        {
+            "gonio-omega": partial_reading(0.0),
+            "gonio-z": partial_reading(0.0),
+            "gonio-y": partial_reading(0.0),
+            "gonio-x": partial_reading(0.0),
+            "gonio-wrapped_omega-phase": partial_reading(0.0),
+            "gonio-wrapped_omega-offset_and_phase": partial_reading(np.array([0, 0])),
```
oliwenmandiamond: It's a little confusing when the name of the device doesn't match, please make it so name is also xyz_wrapped_omega_stage so it is consistent.

tests/devices/test_motors.py
```diff
@@ -311,6 +324,23 @@ async def test_reading_six_axis_gonio(six_axis_gonio: SixAxisGonio):
     )
 
 
+@pytest.mark.parametrize("values_for_rotation", [[0, 0, 0]], indirect=True)
+async def test_reading_xyz_wrapped_omega_gonio(
+    xyz_wrapped_omega_stage, values_for_rotation
```
oliwenmandiamond:
```suggestion
    xyz_wrapped_omega_stage: XYZWrappedOmegaStage, values_for_rotation: Iterable[float]
```

tests/devices/test_motors.py
```diff
@@ -358,3 +388,110 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+        [360.001, 0.001],
+        [719.999, 359.999],
+        [720.001, 0.001],
+        [10000 * 360 + 0.001, 0.001],
+        [-10000 * 360 - 0.001, 359.999],
+    ],
+)
+async def test_mod_360_read(real_value: float, expected_phase):
+    stage = XYZWrappedOmegaStage("BL03I-MO-SGON-01:")
+    await stage.connect(mock=True)
```
oliwenmandiamond:
```suggestion

```

tests/devices/test_motors.py
```diff
@@ -358,3 +388,110 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+    ids=lambda values: f"input={values[0]}, current={values[1]}, expected={values[2]}",
+)
+def values_for_rotation(request):
+    input_value, current_real_value, expected_real_value = request.param
+    yield input_value, current_real_value, expected_real_value
+
+
+async def test_mod_360_expected_direction_of_rotation_same_as_apparent_for_moves_apparently_less_than_180(
+    values_for_rotation, xyz_wrapped_omega_stage
+):
```
oliwenmandiamond, line 513:
```suggestion
async def test_mod_360_expected_direction_of_rotation_same_as_apparent_for_moves_apparently_less_than_180(
    values_for_rotation: Iterable[float], xyz_wrapped_omega_stage: XYZWrappedOmegaStage
):
```

tests/devices/test_motors.py
```diff
@@ -358,3 +388,110 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+    motion_in_same_direction_as_apparent = (input_value < current_readback) == (
+        expected_real_value < current_real_value
+    )
+    assert is_no_move or is_ge_than_180 or motion_in_same_direction_as_apparent, (
+        f"expected ({input_value} - {current_real_value}) same sign as ({expected_real_value} - {current_real_value})"
+    )
+
+
+async def test_mod_360_expected_actual_movement_never_more_than_180(
+    values_for_rotation, xyz_wrapped_omega_stage
```
oliwenmandiamond:
```suggestion
    values_for_rotation: Iterable[float], xyz_wrapped_omega_stage: XYZWrappedOmegaStage
```

tests/devices/test_motors.py
```diff
@@ -358,3 +388,110 @@ async def test_vertical_in_lab_space_for_default_axes(
...
+):
+    input_value, current_real_value, expected_real_value = values_for_rotation
+    assert abs(expected_real_value - current_real_value) <= 180
+
+
+async def test_mod_360_unwrap_computes_expected(
+    values_for_rotation, xyz_wrapped_omega_stage, run_engine: RunEngine
```
oliwenmandiamond:
```suggestion
async def test_mod_360_unwrap_computes_expected(
    values_for_rotation: Iterable[float], xyz_wrapped_omega_stage: XYZWrappedOmegaStage, run_engine: RunEngine
```

oliwenmandiamond: Why is the run_engine added here if it is not used?

tests/devices/test_wrapped_axis.py
```diff
@@ -0,0 +1,105 @@
+import numpy as np
+import pytest
+from ophyd_async.core import get_mock_put, init_devices, set_mock_value
+from ophyd_async.epics.motor import Motor
+
+from dodal.devices.wrapped_axis import WrappedAxis
+
+
+@pytest.fixture
+async def omega():
```
oliwenmandiamond:
```suggestion
async def omega() -> Motor:
```

tests/devices/test_wrapped_axis.py
```diff
@@ -0,0 +1,105 @@
...
+    async with init_devices(mock=True, connect=True):
+        motor = Motor("BL03I-MO-SGON-01:OMEGA")
+    return motor
+
+
+@pytest.fixture
+def wrapped_omega(omega: Motor):
+    return WrappedAxis(omega, "wrapped_omega")
```
oliwenmandiamond:
```suggestion
@pytest.fixture
def wrapped_omega(omega: Motor) -> WrappedAxis:
    with init_devices(mock=True):
          wrapped_omega = WrappedAxis(omega)
    wrapped_omega
```

tests/devices/test_wrapped_axis.py
```diff
@@ -0,0 +1,105 @@
...
+        [181, 360, 360],
+        [270, 0, 360],
+        [-270, 0, -360],
+        [0, 450, 90],
+    ],
+)
+async def test_wrapped_axis_set_applies_phase_with_current_set(
+    omega: Motor,
+    wrapped_omega: WrappedAxis,
+    initial_unwrapped,
```
oliwenmandiamond:
```suggestion
    initial_unwrapped: float,
```

tests/devices/test_wrapped_axis.py
```diff
@@ -0,0 +1,105 @@
...
+
+async def test_wrapped_axis_read_returns_phase_and_offset_phase(
+    omega: Motor, wrapped_omega: WrappedAxis
+):
+    set_mock_value(omega.user_readback, 450)
+    reading = await wrapped_omega.read()
+    assert all(
+        reading["wrapped_omega-offset_and_phase"]["value"] == np.array([360, 90])
+    )
+    assert reading["wrapped_omega-phase"]["value"] == 90
```
oliwenmandiamond: The standard way to check read is below

`from ophyd_async.testing import assert_reading, partial_reading`

```suggestion
     await assert_reading(
         wrapped_omega, 
         {
              "wrapped_omega-offset_and_phase": partial_reading(np.array([360, 90]))
              "wrapped_omega-phase": partial_reading(90),
         }
    )
```

src/dodal/devices/wrapped_axis.py
```diff
@@ -0,0 +1,73 @@
...
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.offset_and_phase = derived_signal_rw(
+                self._get_motor_offset_and_phase,
+                self._set_motor_offset_and_phase,
+                motor_pos=real_motor,
+            )
+        with self.add_children_as_readables():
+            self.phase = derived_signal_rw(
+                self._get_phase, self._set_phase, offset_and_phase=self.offset_and_phase
+            )
```
oliwenmandiamond, line 53: We have a signal that has `offset_and_phase`, why do we need phase as well which gives the same information?

src/dodal/devices/smargon.py
```diff
@@ -78,7 +79,7 @@ class CombinedMove(TypedDict, total=False):
     chi: float | None
 
 
-class Smargon(XYZOmegaStage, Movable):
+class Smargon(XYZWrappedOmegaStage, Movable):
```
oliwenmandiamond, line 82: This is the only device that uses the wrapped omega axis. Rather than creating `XYZWrappedOmegaStage`, you should change it so `Smargon`  still uses `XYZOmegaStage` and then just wrap the omega axis directly here.

```Python
class Smargon(XYZWrappedOmegaStage, Movable[float]):

               self.wrapped_omega = WrappedAxis(self.omega)
               # Rest of smargon logic
               ...
```

oliwenmandiamond (changes requested):
This is looking better, thanks. Still needs a bit more work though. Please see my comments but in summary, `XYZWrappedOmegaStage` appears to be an unnecessary layer. Instead just wrap the axis directly on the device you want it on such as `Smargon`. Type checking on the tests also need to be improved

src/dodal/devices/aithre_lasershaping/goniometer.py
```diff
@@ -1,9 +1,9 @@
 from ophyd_async.epics.motor import Motor
 
-from dodal.devices.motors import XYZOmegaStage, create_axis_perp_to_rotation
+from dodal.devices.motors import XYZWrappedOmegaStage, create_axis_perp_to_rotation
 
 
-class Goniometer(XYZOmegaStage):
+class Goniometer(XYZWrappedOmegaStage):
```
oliwenmandiamond, line 6:
```suggestion
class Goniometer(XYZOmegaStage):

     def(
         self,
         prefix: str,
         name: str = ""
         ...
     ):
        super().__init__(prefix=prefix, name=name, ...)
        with self.add_children_as_readables():
             self.wrapped_omega = WrappedAxis(self.omega)
             ...
```

src/dodal/devices/wrapped_axis.py
rtuck99: phase is very convenient to directly read and write from plans, most plans are not interested in absolute position, but the extra information is useful when you are.

src/dodal/devices/smargon.py
rtuck99: `Goniometer` also inherits from XYZWrappedOmegaStage.

`Goniometer` is used in the same pin tip centring plan that is used with the Smargon, and which following DiamondLightSource/mx-bluesky#1640 does use the wrapped axis, so we need a common abstraction.

src/dodal/devices/aithre_lasershaping/goniometer.py
rtuck99: See my comment below re Smargon

src/dodal/devices/wrapped_axis.py
oliwenmandiamond: Would it not make more sense to just have one signal for `offset` and one for `phase`?

src/dodal/devices/smargon.py
oliwenmandiamond: Hmmm. My preference would be to still use `XYZOmegaStage`, and then have your plan include a parameter called `wrapped_axis` or `wrapped_omega`. So it looks like this:

```Python
@pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
class PinTipCentringComposite:
    """All devices which are directly or indirectly required by this plan"""

    oav: OAV
    gonio: XYZOmegaStage
    gonio_wrapped_axis: WrappedAxis
    pin_tip_detection: PinTipDetection
```

The reason being is this makes it way more generic. Any axis can now be wrapped. This means we don't need to create a new subclass for every possible stage combination which needs an axis wrapped.

However, looking at it more I'm not sure this would be supported by BlueAPI as it is a child device and from what I can tell, composites are extracted from being a unique type (correct me if wrong?).

oliwenmandiamond (commented):
This looks much better now, thanks! I just have a couple questions

src/dodal/devices/smargon.py
rtuck99: BlueAPI constructs injected composites by locating devices in the context with names corresponding to the field names in the composite type, so you could have "gonio_wrapped_axis" in the composite, but in the beamline module it would have to be a standalone device. I think this would be problematic because it would need to have been instantiated in a separate factory and this would cause a circular dependency if it was still a child device of the gonio.

oliwenmandiamond: Yeah, that's what I thought. It would be good if blueapi supported the ability to inject stuff via a composite via the device name so then you do individual device names from child devices, maybe something like this:

```Python
@pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
class PinTipCentringComposite:
    """All devices which are directly or indirectly required by this plan"""

    oav: OAV
    gonio: XYZOmegaStage
    wrapped_axis: WrappedAxis = inject("gonio-wrapped_omega")
    pin_tip_detection: PinTipDetection
```

src/dodal/devices/wrapped_axis.py
rtuck99: We could have a separate `offset` signal, however it's not terribly useful on its own, I can't think of any particular reason you'd want to set it independently, or read it without the phase. Supporting a non-atomic write to `offset` might have undesirable behaviour if that's not what they are explicitly trying to do - they would be better off in that case to write to raw `omega` with the unwrapped value.

The one case where I think it _might_ be useful is if you have some device which exposes wrapped coordinate space directly, and there is some offset in the hardware you can twiddle.

If we ever had such a device, I think it would be natural to abstract `WrappedAxis` into a base class where one implementation is the current one for motors with unwrapped coordinates, and the other is a new implementation for motors with wrapped coordinates. Then setting the offset sets the underlying offset in the device for those using wrapped coordinates natively. Having said that, I think potentially in that scenario, the semantics of setting the offset are unclear - does it actually move the motor in this case, or does it just affect some other calculation?

oliwenmandiamond (approved):
Spoken in IRL, we had a play around with this using protocols too but it became too complex. Sub plans required XYZStage and didn't know how to convert XYZWrappedOmega protocol to just XYZ without having to create an XYZ protocol and subclassing it all which just duplicates all the stages. I think this is fine for now, and hopefully will be simplified further in future when ophyd-async implements https://github.com/bluesky/ophyd-async/issues/1016?issue=bluesky%7Cophyd-async%7C1182

PR #1939, opened 2026-02-24
Update fast shutter and selectable source device names

oliwenmandiamond:
Override set_name method on fast shutter and selectable source devices. This is similar to motor where you name the primary signal to be same name as the device so when reading it makes more sense

e.g for a shutter called `fsj1` it now reads as 

`{"fsj1" : {"value": InOut.IN, ...}}`

rather than

`{"fsj1-shutter_value" : {"value": InOut.IN}}`

1. Check tests pass

### Checks for reviewer


DominicOram:
Have you discussed with @coretl about the possibility of putting this in `ophyd-async`? I feel like we might always want the hinted signal to behave like this?

PR #1940, opened 2026-02-25
Add optics shutters

### LINKED ISSUE #1937: Add experimental shutters to surface beamlines
Current `HutchShutter `class has hardcoded interlock from PSS, however most surface beamlines use interlock from the shutter itself. Need to refactor this class to make it more flexible.
Add experimental shutter fo I05, B07, I09

Villtord:
Fixes #1937 

Here I made interlock an optional argument when creating `HutchShutter`. If it exists then interlock will be checked before shutter open, if not then it won't be checked.
Interlock now has a base class interface that declares `shutter_safe_to_operate` method which may differ in internal logic. It now gives much more flexibility in constructing interlock PV. It also declares a signal_type of the interlock as it can be Enum(for epics edm interlock) or float (for PSS interlock), however we do not currenlty use Enum-based interlock at all.

1. Do dodal connect beamlines (i03, i19-optics, i24, i05, b07, i09)
2. Confirm no bugs, tests pass

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ description = "Ophyd devices and other utils that could be used across DLS beaml
 dependencies = [
     "click",
     "ophyd",
-    "ophyd-async[ca,pva]>=v0.15.0",
+    "ophyd-async[ca,pva]>=v0.16.0",
     "bluesky>=1.14.5",
     "pyepics",
     "pillow",
@@ -206,7 +206,7 @@ commands = [
     [
         "pytest",
         "-m",
-        "not requires(instrument=\"i04\")",
+        "not requires(instrument=\"i04\") and not requires(instrument=\"i15-1\")",
         "--cov-report",
         "xml:system_cov.xml",
         "--timeout=10",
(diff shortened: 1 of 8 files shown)


src/dodal/devices/hutch_shutter.py
```diff
@@ -52,43 +92,44 @@ async def shutter_safe_to_operate(self) -> bool:
...
-        self.interlock = HutchInterlock(bl_prefix)
-
+    def __init__(
+        self,
+        bl_prefix: str,
+        interlock: BaseHutchInterlock | None = None,
+        shtr_infix: str = EXP_SHUTTER_1_INFIX,
+        name: str = "",
+    ) -> None:
+        self.interlock = interlock
```
noemifrisina: Should: This should at the very least be inside `add_children_as_readables` as otherwise the device factory won't be able to find it when not None

Villtord: Thanks, good point - actually I thought about it and I looked at the original class - there interlock was not included to readables as far as I can see, do you want it now to be included? Although interlock itself is StandardReadable so you probably want it to be read somewhere...  I am fine with any answer :)

noemifrisina: Honestly, I think it wasn't included because this device was implemented before `add_children_as_readables` was a thing - and no one has refactored the class in awhile. I am actually not convinced it would have kept working as it is now with the new DeviceManager (at least for I19 and I24 I know the version of dodal on the beamline precedes it). We might as well make the change in here to make sure it still works :) In any case it will be an improvement

Villtord: Added to child readables

PR #1975, opened 2026-03-11
Concept of checkbeam device

oliwenmandiamond:
Fixes #ISSUE

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i09.py
+++ b/src/dodal/beamlines/i09.py
@@ -11,6 +11,7 @@ from dodal.devices.electron_analyser.vgscienta import VGScientaDetector
 from dodal.devices.fast_shutter import DualFastShutter, GenericFastShutter
 from dodal.devices.hutch_shutter import EXP_SHUTTER_2_INFIX, HutchShutter
 from dodal.devices.motors import XYZAzimuthPolarStage
+from dodal.devices.pause_plan_device import PausePlanDevice
 from dodal.devices.pgm import PlaneGratingMonochromator
 from dodal.devices.selectable_source import SourceSelector
 from dodal.devices.synchrotron import Synchrotron
@@ -126,3 +127,24 @@ def lakeshore() -> Lakeshore336:
 def smpm() -> XYZAzimuthPolarStage:
     """Sample Manipulator."""
     return XYZAzimuthPolarStage(prefix=f"{I_PREFIX.beamline_prefix}-MO-SMPM-01:")
+
+
+@devices.factory()
+def checkbeam(
+    synchrotron: Synchrotron, dual_fast_shutter: DualFastShutter
+) -> PausePlanDevice:
+    async def _close_shutters():
+        await dual_fast_shutter.set(dual_fast_shutter.close_state)
+
+    async def _open_shutters():
+        await dual_fast_shutter.set(dual_fast_shutter.open_state)
+
+    checkbeam = PausePlanDevice(
+        signals_to_condition={
+            synchrotron.current: lambda rc: rc > 190,
+            synchrotron.top_up_start_countdown: lambda topup: topup < 5,
+        },
+        callable_when_paused=_close_shutters,
+        callable_on_resume=_open_shutters,
+    )
+    return checkbeam
--- a/src/dodal/devices/pause_plan_device.py
+++ b/src/dodal/devices/pause_plan_device.py
@@ -1,0 +1,92 @@
+import asyncio
+import contextlib
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+from bluesky.protocols import Readable, Stageable
+from ophyd_async.core import (
+    AsyncStatus,
+    Device,
+    SignalR,
+    observe_value,
+)
+
+
+class PausePlanDevice(Device, Stageable, Readable):
+    def __init__(
+        self,
+        signals_to_condition: dict[SignalR[Any], Callable[[Any], bool]],
+        callable_when_paused: Callable[[], Awaitable[None]] | None = None,
+        callable_on_resume: Callable[[], Awaitable[None]] | None = None,
+        seconds_to_wait_before_resume: float = 5,
+        name: str = "",
+    ):
+        self._signals_to_condition = signals_to_condition
+        self._callable_when_paused = callable_when_paused
+        self._callable_on_resume = callable_on_resume
+        self._seconds_to_wait_before_resume = seconds_to_wait_before_resume
+        super().__init__(name)
+
+    async def _pause(self):
+        """Pause until all signal conditions are met, calling hooks as needed."""
+        # Check if we actually need to pause
+        values = await asyncio.gather(
+            *(sig.get_value() for sig in self._signals_to_condition)
+        )
+        all_met = all(
+            pred(value)
+            for value, (_, pred) in zip(
+                values, self._signals_to_condition.items(), strict=True
+            )
+        )
+        if all_met:
+            return  # no need to pause
+
+        # Call pause hook
+        if self._callable_when_paused:
+            await self._callable_when_paused()
+
+        latest = {}
+        event = asyncio.Event()
+
+        async def watch(signal, predicate):
+            async for value in observe_value(signal):
+                latest[signal] = predicate(value)
+                if len(latest) == len(self._signals_to_condition) and all(
+                    latest.values()
+                ):
+                    event.set()
+                    return
+
+        tasks = [
+            asyncio.create_task(watch(sig, pred))
+            for sig, pred in self._signals_to_condition.items()
+        ]
+
+        await event.wait()
+
+        # Cancel watchers
+        for task in tasks:
+            task.cancel()
+            with contextlib.suppress(asyncio.CancelledError):
+                await task
+
+        await asyncio.sleep(self._seconds_to_wait_before_resume)
+        # Call resume hook
+        if self._callable_on_resume:
+            await self._callable_on_resume()
+
+    @AsyncStatus.wrap
+    async def stage(self):
+        await self._pause()
+
+    @AsyncStatus.wrap
+    async def unstage(self):
+        pass
+
+    async def read(self):
+        await self._pause()
+        return {}
+
+    async def describe(self):
+        return {}
--- a/tests/devices/test_pause_plan_device.py
+++ b/tests/devices/test_pause_plan_device.py
@@ -1,0 +1,119 @@
+import asyncio
+
+import pytest
+from bluesky import RunEngine
+from bluesky import plan_stubs as bps
+from ophyd_async.core import InOut, SignalRW, init_devices, soft_signal_rw
+
+from dodal.devices.fast_shutter import GenericFastShutter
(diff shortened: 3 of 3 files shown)


rtuck99:
I wonder if there would be any benefit in supporting `WatchableAsyncStatus` with this - I have done something similar for hyperion waiting-for-beam, because it enables the supervisor to monitor it for progress to determine whether the plan is genuinely stuck or merely waiting for beam.

DominicOram:
Have you looked at using [suspenders](https://nsls-ii.github.io/bluesky/state-machine.html?highlight=suspenders#automated-suspension) for this? I think it's the more Bluesky-native way of doing it

oliwenmandiamond:
Yeah, I've had a look and they do look to be what we want. My only concern is that they are installed globally on the RunEngine so applied to every plan. This is the opposite of what GDA currently does where it is only applied when user specifies it in the scan. We would also need BlueAPI to support being able to add them.

e.g

```python
@devices.suspender
def checkbeam() -> MySuspender:
     ...
```

then on blueapiclient side we could maybe do this

```python
bc = BlueapiClient()
devs = bc.devices
plans = bc.plans

# Installed suspenders by default
`>>> plans.scan(...)`

#Run a plan without the suspender
>>> bc.remove_suspender("checkbeam")
>>> bc.my_plan()

>>> bc.install_suspender("checkbeam")
>>> bc.scan(...)
```
As I mentioned in one of the previous athena drop in sessions, it feels like `DeviceManager` will have to evolve to a `BeamlineManager` to being able to support things that aren't devices. Or maybe we create lots of sub managers to manage specific things e.g `SuspenderManager` which blueapi can also register like `DeviceManager`. However, suspenders need access to device and their signals to construct one.

DominicOram:
>My only concern is that they are installed globally on the RunEngine so applied to every plan.

You can do it inside the plan. Using [install](https://nsls-ii.github.io/bluesky/generated/bluesky.plan_stubs.install_suspender.html#bluesky.plan_stubs.install_suspender) or [remove](https://nsls-ii.github.io/bluesky/generated/bluesky.plan_stubs.remove_suspender.html#bluesky.plan_stubs.remove_suspender) e.g.

```python
def my_plan():
   yield from bps.install_suspender(checkbeam)
   ...
   yield from bps.remove_suspender(checkbeam)
```

or more cleanly (and to cover the `finally` case):

```python
@bpp.suspend_wrapper(checkbeam)
def my_plan():
    ...
```

oliwenmandiamond:
> > My only concern is that they are installed globally on the RunEngine so applied to every plan.
> 
> You can do it inside the plan. Using [install](https://nsls-ii.github.io/bluesky/generated/bluesky.plan_stubs.install_suspender.html#bluesky.plan_stubs.install_suspender) or [remove](https://nsls-ii.github.io/bluesky/generated/bluesky.plan_stubs.remove_suspender.html#bluesky.plan_stubs.remove_suspender) e.g.
> 
> ```python
> def my_plan():
>    yield from bps.install_suspender(checkbeam)
>    ...
>    yield from bps.remove_suspender(checkbeam)
> ```
> 
> or more cleanly (and to cover the `finally` case):
> 
> ```python
> @bpp.suspend_wrapper(checkbeam)
> def my_plan():
>     ...
> ```
Okay I didn't know this. My issue with doing it like this though is that you now have to wrap every plan with this where as with it being a device like GDA, it can be used with any plan whenever required dynamically as you just specify the device to the plan arguments. This way is less flexible.

Also, how do we apply this with BlueAPI?

Do we have to define a suspender instance in dodal and then have our plan repos import the suspender and add this as a decorator?

oliwenmandiamond:
To make it generic, I think we would have to have blueapi support it as server object so then from client side we could do
```Python
>>> plans.install_suspender(bc.suspenders.checkbeam)
```
So we can turn on / off whenever user requires it.

If it is a server object too, again we can use inject for plans to auto include them. Another use case for https://github.com/DiamondLightSource/dodal/issues/1997 and https://github.com/DiamondLightSource/blueapi/issues/1462

DominicOram:
> My issue with doing it like this though is that you now have to wrap every plan with this where as with it being a device like GDA, it can be used with any plan whenever required dynamically as you just specify the device to the plan arguments. This way is less flexible.

Ok, in which case having an endpoint or configuration option on BlueAPI to add it to the RE seems like a better way to go.

> Also, how do we apply this with BlueAPI?

I'm not sure I understand, BlueAPI runs your plan, your plan adds the suspender. Do you mean how do we apply it to the out of the box scans? In which case, yh, creating a way of asking BlueAPI to hook it into the RE would be the way to do it.

> Do we have to define a suspender instance in dodal and then have our plan repos import the suspender and add this as a decorator?
> Another use case for https://github.com/DiamondLightSource/dodal/issues/1997 and https://github.com/DiamondLightSource/blueapi/issues/1462

Yh, I agree we should do this.

I really think we should go down the suspender route if possible though, this is literally what they're designed for. And using a device like this seems like we're going back down the route of having "Scannables are everything". @DiamondLightSource/developers-daq-core how do you envisage suspenders being added to plans?

oliwenmandiamond:
> I really think we should go down the suspender route if possible though, this is literally what they're designed for. And using a device like this seems like we're going back down the route of having "Scannables are everything". @DiamondLightSource/developers-daq-core how do you envisage suspenders being added to plans?

Agreed, we should find a way to support suspenders to do checkbeam and any other conditions.

fajinyuan:
Agreed that we should make use of bluesky offered mechanism - suppender - for this use case. BlueAPI client has to support this requirement to enable user control of this. As to the implementation as Device or not is debateable, in my personal view Ophyd device can be a virtual device implementing special logics using other devices. After all, suspender is to pause data collection under certain detected conditions. This can be done by suspend Run Engine or making a virtual device busy (at least for step scan this will work).

DominicOram:
I attempted to use suspenders for a similar problem on i22 today and found some issues:
* Suspenders do not currently work with ophyd-async (https://github.com/bluesky/ophyd-async/issues/508). I worked around this on the beamline
* There seems to be something odd where the RE will go into "suspending" briefly then back into "running" even though the suspender is still asking it to be paused - I wonder if this is BlueAPI doing something wacky on top of the RE, needs more investigation

oliwenmandiamond:
> I attempted to use suspenders for a similar problem on i22 today and found some issues:
> 
>     * Suspenders do not currently work with ophyd-async ([Implement support for bluesky suspenders bluesky/ophyd-async#508](https://github.com/bluesky/ophyd-async/issues/508)). I worked around this on the beamline
> 
>     * There seems to be something odd where the RE will go into "suspending" briefly then back into "running" even though the suspender is still asking it to be paused - I wonder if this is BlueAPI doing something wacky on top of the RE, needs more investigation

Interesting, can you link me the code you used to test with? How did you get this to work with BlueAPI? Did you build the suspender inside a plan from devices and then install it to run engine?

DominicOram:
> 
> Interesting, can you link me the code you used to test with? How did you get this to work with BlueAPI? Did you build the suspender inside a plan from devices and then install it to run engine?

Initial implementation is at https://github.com/DiamondLightSource/saxs-bluesky/pull/104. I just added it to the plan I needed it for using `bpp.suspend_decorator`. I appreciate that won't work for the out of the box `scan` commands but it's enough for i22 for now.

src/dodal/beamlines/i09.py
```diff
@@ -126,3 +127,24 @@ def lakeshore() -> Lakeshore336:
 def smpm() -> XYZAzimuthPolarStage:
     """Sample Manipulator."""
     return XYZAzimuthPolarStage(prefix=f"{I_PREFIX.beamline_prefix}-MO-SMPM-01:")
+
+
+@devices.factory()
+def checkbeam(
```
fajinyuan, line 133: this checkbeam is missing front end shutter checking. Is this done somewhere else?

oliwenmandiamond: This was just a proof of concept so not everything is added yet

oliwenmandiamond: However, I've made it flexible enough so that you can add any custom logic. So you would just need to inject the front end shutter here, add check to see if closed is signals_to_condions

oliwenmandiamond: To be clear though, I agree with @DominicOram and I don't think a device is the way to go about this. This was just a proof of concept on how we could possibly translate the GDA code to bluesky with this being a possibility. If we make it a device and it "pauses" a scan, it won't pause the RunEngine and therefore BlueAPI will not be able to tell the scan is paused either. It will probably look like the plan is hanging as it is blocking and it may not be obvious why to the user. Using the Suspenders will give BlueAPI a lot more natural context and feedback as to what is happening with the plan.

fajinyuan: with checkbeam for data collection in user mode, I would ensure it is built into RE directly as Suspender for any plans to be run as default as users data collection will always need these. However for developers or beamline staffs who will work during shutdown this must be not as deafult in RE.

oliwenmandiamond: It is very flexible. the RunEngine has `install_suspender` and `remove_suspender` hooks to remove it globally. E.g by default, `run_engine.install_suspender(checkbeam) `and in dummy we can easily do `run_engine.remove_suspender(checkbeam)` for shutdown.

We also have the option for the suspender to only suspender if `synchrotron.synchtron_mode` is not `"Shutdown"` or `"Mach. Dev"` as well. So this will not be a problem. 

The two things we need from core team though will be for this to be supported in BlueAPI. We will need https://github.com/DiamondLightSource/dodal/issues/1997 to be supported, `HasName` on objects or something equivalent.

We also need the run engine hooks to be present so that from blueapi client we can do something like so

dodal side

```python
@devices.fixture
def checkbeam(
     synchrotron: Synchrotron,
     shutter: Shutter,
     # Anything else,
) -> MySuspender:
     return MySuspender(synchrotron, shutter, ...)
```

Blueapi client side

```python
bc = BlueapiClient(...)
plans = bc.plans
devs = bc.devices
fixtures = bc.fixtures

# Applied globally
bc.install_suspender(fixtures.checkbeam)
...
bc.remove_suspender(fixtures.checkbeam)
```

and also inside plans, we can control it locally (inject has to be able to support things with names that aren't devices https://github.com/DiamondLightSource/blueapi/issues/1462)

```python
from bluesky.plan_stubs import install_suspender, remove_suspender

def my_plan(a, b, checkbeam: MySuspender = inject("checkbeam")):
      # Applied only for this plan
      yield from install_suspender(checkbeam)
      ...
      yield from remove_suspender(checkbeam)
```

fajinyuan: I prefer your 2nd offer to auto detect if synchrotron in user mode or not. This will eliminate the need for users to know how to install and remove suspender as checkbeam will be permenantly installed in run engine which can be configured by beamline support software engineers. 
However we still need to extend blueapi to support this instal/remove suspender feature for other use cases.

PR #1984, opened 2026-03-20
Add test to ensure all functions in beamline modules are devices or fixtures

### LINKED ISSUE #1981: Add test criterion to confirm decorators correct in beamline device definition script
**Background**
In a typical beamline definition script, a sequence of devices are defined and a device factory decorator is applied to each.

**Issue**
The issue is that if a decorator is left out from a single device, there is currently nothing to detect this 
( except for the Mark 1 human eyeball ).

**Task**
Make some form of automated test that will detect any discrepancies.

_**Caveat**_
This should not require a separate source of ground truth to provide a set of expected devices.  
That would merely create the potential for two sources of error and the classic "Which one do You believe?" problem.

One option for a simple solution might be to insist that every python def line in such definition files comes with a decorator.

## Acceptance Criteria
* That a beamline script with decorators proper and correct passes any new test
* That a beamline script with the decorator forgotten ( or deliberately removed in a local repo just to exercise the test )
   should cause the same test to fail

jacob720:
Fixes #1981

1. Delete fixture or device decorator in a beamline module.
2. Confirm test for that beamline fails.

### Checks for reviewer

### DIFF
--- a/tests/common/beamlines/test_device_instantiation.py
+++ b/tests/common/beamlines/test_device_instantiation.py
@@ -1,3 +1,4 @@
+import inspect
 from typing import Any
 
 import pytest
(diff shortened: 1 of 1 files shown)


DominicOram (commented):
I was going to suggest moving the i23 helper function to utils but on looking at it again I believe it should actually be entirely removed as it will mean thinks won't connect on the beamline cluster, where we should generally be running.

tests/common/beamlines/test_device_instantiation.py
```diff
@@ -63,3 +64,37 @@ def test_devices_are_identical(module_and_devices_for_beamline):
...
+)
+def test_all_functions_in_beamline_modules_are_devices_or_fixtures(
+    module_and_devices_for_beamline,
+):
+    module, _, _ = module_and_devices_for_beamline
+    devices: DeviceManager | None = getattr(module, "devices", None)
+
+    def func_is_device(func) -> bool:
+        return hasattr(func, "_devices_decorator")
+
```
tpoliaw, line 82: Where is `"_devices_decorator"` coming from?

PR #1996, opened 2026-03-27
i15-1: Add cobra and blower devices

jacob-williamson:
Fixes #[ISSUE](https://github.com/DiamondLightSource/crystallography-bluesky/issues/3)
Required by https://github.com/DiamondLightSource/crystallography-bluesky/pull/24
Adds cobra, blower and cryostream devices

1. Confirm tests pass

### Checks for reviewer


src/dodal/beamlines/i15_1.py
```diff
@@ -49,15 +65,23 @@ def base_y() -> Motor:
 
 
 @devices.factory()
-def blower_y() -> Motor:
+def blower_y(config_client: ConfigClient) -> Blower:
```
DominicOram: Should: It feels like a weird holdover from GDA to have two of these, can we just have the Z?

src/dodal/beamlines/i15_1.py
```diff
@@ -184,6 +208,23 @@ def attenuator() -> Attenuator:
...
+def cobra(config_client: ConfigClient) -> Cobra:
+    return Cobra(
+        f"{PREFIX.beamline_prefix}-MO-TABLE-01:ENV:X",
+        config_client,
+        XPDF_PARAMETERS_FILEPATH,
+    )
+
+
+@devices.factory()
+def cryostream(config_client: ConfigClient) -> Cryostream:
```
DominicOram: Should: Did we work out that these are interchangeable devices that go on the same rail? Can we add this into the comments here or in the device classes? If you're not sure can you double check with the scientists?

src/dodal/devices/beamlines/i15_1/temperature_controller.py
```diff
@@ -0,0 +1,36 @@
...
+)
+from ophyd_async.epics.motor import Motor
+
+
+class TemperatureControllerPosition(StrictEnum):
+    SAFE = "Safe"
+    BEAM = "Beam"
+
+
+class TemperatureController(StandardReadable, Movable[TemperatureControllerPosition]):
```
DominicOram: Could: `TemperatureController` isn't a great name here because we're not doing anything with the fact they're temperature controllers. Maybe something more like `InBeamOrSafePositioner`? That's clunky too though but you get the idea

DominicOram (changes requested):
There's a lot of repeated code in all the different `TemperatureController`s, we should move more into the base class:
```python
class TemperatureController(StandardReadable, Movable[TemperatureControllerPosition]):
    def __init__(self, config_client: ConfigClient, xpdf_parameters_path: str, prefix: str):
        self.config_client = config_client
        self.xpdf_parameters_path = xpdf_parameters_path        
            self.motor = Motor(prefix=prefix)
        super().__init__(prefix)

    def get_full_config():
       return self.config_client.get_file_contents(
            self.xpdf_parameters_path,
            desired_return_type=TemperatureControllersConfig,
            force_parser=TemperatureControllersConfig.from_xpdf_parameters,
        )

    @abstractmethod
    def get_config(): ...

    @property
    def _safe_position(self) -> float:
         return self.get_config().safe_position

    @property
    def _beam_position(self) -> float:
        return self.get_config().beam_position

    @AsyncStatus.wrap
    async def set(self, value: TemperatureControllerPosition):
        if value == TemperatureControllerPosition.SAFE:
            await self.motor.set(self._safe_position)
        elif value == TemperatureControllerPosition.BEAM:
            await self.motor.set(self._beam_position)
```
then:
```python
class Blower(TemperatureController):
    def get_config(self) -> TemperatureControllerParams:
        return self.get_full_config().blower
```

It might then make more sense to just have one set of tests that confirms against the base class that the movement works for different sets of safe/in beam and then just have smaller sets of tests that confirm the child classes get the config from the right place

src/dodal/beamlines/i15_1.py
jacob-williamson: Yes they're interchangeable and on the same rail, there's a couple details about safety positions which I'll double check.

src/dodal/beamlines/i15_1.py
```diff
@@ -184,6 +198,23 @@ def attenuator() -> Attenuator:
 
 
 @devices.factory()
+def cobra(config_client: ConfigClient) -> Cobra:
```
Mark-Booth: Since moving stuff around later makes merges difficult to follow, and existing devices should all be in alphanumeric order, could we start with these new device in a consistent position?

src/dodal/beamlines/i15_1.py
```diff
@@ -49,15 +65,13 @@ def base_y() -> Motor:
...
 
 @devices.factory()
-def blower_y() -> Motor:
-    """Same motor as blowerZ."""
-    return Motor(f"{PREFIX.beamline_prefix}-EA-BLOWR-01:TLATE")
-
-
-@devices.factory()
-def blower_z() -> Motor:
+def blower_z(config_client: ConfigClient) -> Blower:
```
Mark-Booth, line 69: I was hoping we might have some way to tell if this motor was configured in Y mode or Z mode before we added devices for both. I worry that people may rely on the metadata of which motor was moved, and then assume that the motor was moving in Y mode when it was actually configured to move in Z, which is why I made a comment rather than a duplicate device.

src/dodal/devices/beamlines/i15_1/safe_or_beam_positioner.py
```diff
@@ -0,0 +1,55 @@
...
+        if value == SafeOrBeamPosition.SAFE:
+            await self.motor.set(self._safe_position)
+        elif value == SafeOrBeamPosition.BEAM:
+            await self.motor.set(self._beam_position)
+
+    def get_full_config(self):
+        return self.config_client.get_file_contents(
+            self.xpdf_parameters_path,
+            desired_return_type=TemperatureControllersConfig,
+            force_parser=TemperatureControllersConfig.from_xpdf_parameters,
```
DominicOram: Should: Can we have a comment in here that we can drop this when a release of the config server is made with a link to which PR is needed in that release for this to be dropped

src/dodal/testing/fixtures/config_server.py
```diff
@@ -13,16 +15,19 @@ def fake_config_server_get_file_contents(
     filepath: str | Path,
     desired_return_type: type[T] = str,
     reset_cached_result: bool = True,
+    force_parser: Callable[[str], Any] | None = None,
```
DominicOram, line 18: Could: If this is only useful for interims when the config server hasn't been released can we add a docstring here about that?

src/dodal/beamlines/i15_1.py
DominicOram: > existing devices should all be in alphanumeric order

I don't think this is documented anywhere as the standard. I'm ambivalent about it but if you feel strongly @Mark-Booth could you add to the docs and probably make a ticket to automate it?

DominicOram (approved):
Thanks, minor comment but otherwise looks good.

src/dodal/devices/beamlines/i15_1/safe_or_beam_positioner.py
jacob-williamson: A new version of the config server has just been deployed so I'll remove these lines

Issue #1998, opened 2026-03-30
Smargon combined moves should perform omega moves separately


rtuck99:
Following discussions around i03's replacement smargon, it was concluded that the issues around combined moves and slow y-axis caused by the faulty smargon notwithstanding, there are likely to still be issues even with correctly performing smargon.

These are thought to possibly be caused by a) the higher omega speed limit impacting timing calculations for the move  and b) real vs virtual axis causing omega moves to make additional smaller moves on the other axes which might already be at their speed limit.

Therefore the plan is even when the new smargon is in place we should perform the omega move first.

PR #1999, opened 2026-03-30
Fix for smargon combined moves with current slow-smargon workaround

rtuck99:
Fixes 
* #1998

Requires:
* #1929 

This removes support for omega in CombinedMove as described in the accompanying mx-bluesky PR

See also mx-bluesky PR
* DiamondLightSource/mx-bluesky#1692

1. Tests pass

### Checks for reviewer


src/dodal/devices/smargon.py
```diff
@@ -105,15 +104,6 @@ def __init__(self, prefix: str, name: str = ""):
...
-        if motor_name == "omega":
-            current_angle = AngleWithPhase.from_iterable(
-                await self.wrapped_omega.offset_and_phase.get_value()
-            )
-            return current_angle.nearest_with_phase(value).unwrap()
-        else:
-            return value
-
     @AsyncStatus.wrap
     async def set(self, value: CombinedMove):
```
noemifrisina, line 120: Nit: may be useful for future reference/future people looking at this mentioning either in this docstring or in the CombinedMove why it does not include the omega axis

noemifrisina (approved):
lgtm, only have one comment about the docstring because if I hadn't read the ticket I would be wondering why omega is not part of the CombinedMove - feel free to ignore if it's written somewhere else and I missed it

PR #2005, opened 2026-04-10
Add i21 shutter with lateral motor

oliwenmandiamond:
Added fastshutter to i21. Extended base to add the additional motor that i21 requires. Updated existing classes to reuse the same enum `OpenClosed`

1. Check dodal connect
2. Check tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/beamlines/i21/fast_shutter.py
+++ b/src/dodal/devices/beamlines/i21/fast_shutter.py
@@ -1,0 +1,20 @@
+from typing import Generic
+
+from ophyd_async.epics.motor import Motor
+
+from dodal.devices.fast_shutter import EnumTypesT, FastShutter
+
+
+class FastShutterWithLateralMotor(FastShutter[EnumTypesT], Generic[EnumTypesT]):
+    def __init__(
+        self,
+        prefix: str,
+        open_state: EnumTypesT,
+        close_state: EnumTypesT,
+        shutter_suffix: str = "CON",
+        lateral_suffix: str = "LAT",
+        name: str = "",
+    ):
+        with self.add_children_as_readables():
+            self.x = Motor(prefix + lateral_suffix)
+        super().__init__(prefix + shutter_suffix, open_state, close_state, name)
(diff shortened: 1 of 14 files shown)


src/dodal/devices/beamlines/i21/fast_shutter.py
```diff
@@ -0,0 +1,34 @@
+from typing import Generic
+
+from ophyd_async.epics.motor import Motor
+
+from dodal.devices.fast_shutter import EnumTypesT, FastShutter
+
+
+class FastShutterWithLateralMotor(FastShutter[EnumTypesT], Generic[EnumTypesT]):
```
Relm-Arrowny, line 8: We need to ask @fajinyuan how this fast shutter actually should behave, looking at it as an outsider, it look to me this is a externally triggered fast shutter. Couple of things concern me:
1. The ERIO output is the status we trying to change as if it is a standard shutter, I am not 100% sure open will mean open or it just set the default state to open and a trigger signal will close it. 
2. There are 4 more outputs and a source signals on that fast shutter controller we likely want to be able to access them depending of the experimental condition.

src/dodal/devices/beamlines/i21/fast_shutter.py
```diff
@@ -0,0 +1,34 @@
...
+    the open and close state of the shutter. This also has the lateral motor x.
+
+    Args:
+        pv (str): The pv to connect to the shutter device.
+        open_state (EnumTypesT): The enum value that corresponds with opening the
+            shutter.
+        close_state (EnumTypesT): The enum value that corresponds with closing the
+            shutter.
+        shutter_suffix (str, optional): Shutter suffix state. Defaults to CON.
+        lateral_suffix (str, optional): Lateral mottor suffix. Defaults to LAT.
```
Relm-Arrowny:
```suggestion
        lateral_suffix (str, optional): Lateral motor suffix. Defaults to LAT.
```

Relm-Arrowny (commented):
I am not 100% sure about the i21 fastshutter should behave like a standard  shutter the other change looks good to me.

PR #2016, opened 2026-04-17
Implement EPICS-backed Polynomial ID Device for I06

### LINKED ISSUE #2003: Add apple2 energy device to i06
see DiamondLightSource/sm-bluesky#195 for detail and task list.

Unlike other Apple2 insertion devices that rely on static lookup tables for energy-to-gap conversion, I06 stores its conversion polynomials directly in EPICS PVs. Since I06 utilises a single polynomial for the entire energy range and does not require polarization changes via LH, we will implement a dynamic EnergyMotorConvertor that reads parameters from EPICS on demand.

# Goals
Acceptance criteria :
- [x] To be able to control pgm and apple2 devices via set.
- [x] To be able to read and set polarisation.  
- [x] To perform energy scan with both pgm and insertion devices.

### Tasks
- [x] Implement I06EnergyMotorConverter: Create a device that reads polynomial parameters from EPICS and performs the conversion calculation.
- [x] Combine I06EnergyMotorConverter with Apple2 controller and the PGM into a single BeamEnergy device for synchronized motion.
- [x] Add the new devices to the I06 beamline configuration within dodal.
- [x] Unit Tests: Verify the polynomial calculation matches expected gap outputs for given energies.

###Implementation note
- No Lookup Table: EnergyMotorLookup is not required. This avoids the overhead of generating and storing large tables at startup.
- Standardization: By using the Apple2Controller with a custom EnergyMotorConvertorDevice, we maintain a common API while handling I06's unique hardware interface.

Relm-Arrowny:
Fixes #2003 

-  Introduced I06EpicsPolynomialDevice to construct EnergyMotorLookup tables dynamically from EPICS PVs.
-  Added  an optional inverse EnergyMotorConverter to Apple2Controller for bidirectional gap/energy calculations.
-  Renamed EnergyMotorConverter parameters from energy to value to better support generic bidirectional conversions (e.g., Motor ↔ Energy).

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/insertion_device/apple2_controller.py
```diff
@@ -102,20 +109,12 @@ def __init__(
         self.apple2 = Reference(apple2)
         self.gap_energy_motor_converter = gap_energy_motor_converter
         self.phase_energy_motor_converter = phase_energy_motor_converter
+        self.energy_gap_converter = energy_gap_coverter
```
oliwenmandiamond: The name makes more sense if something like  pol_gap_to_energy

oliwenmandiamond (commented):
This looks good. Just think the variable name needs to be changed + need to add the config

src/dodal/devices/beamlines/i06_shared/i06_apple2_controller.py
```diff
@@ -0,0 +1,58 @@
...
+        apple2: Apple2[UndulatorPhaseAxes],
+        gap_energy_motor_lut: EnergyMotorLookup,
+        phase_energy_motor_lut: EnergyMotorLookup,
+        inverse_gap_energy_motor_lut: EnergyMotorLookup,
+        units: str = "eV",
+        name: str = "",
+    ) -> None:
+        self.gap_energy_motor_lut = gap_energy_motor_lut
+        self.phase_energy_motor_lut = phase_energy_motor_lut
+        self.inverse_gap_energy_motor_lut = inverse_gap_energy_motor_lut
```
oliwenmandiamond: Is this needed? These are never used anywhere and parent class will keep the method to find the lut

src/dodal/devices/beamlines/i06_shared/i06_apple2_controller.py
```diff
@@ -0,0 +1,58 @@
...
+    def _get_apple2_value(self, gap: float, phase: float, pol: Pol) -> Apple2Val:
+        return Apple2Val(
+            gap=gap,
+            phase=Apple2PhasesVal(
+                top_outer=phase,
+                top_inner=0.0,
+                btm_inner=phase,
+                btm_outer=0.0,
+            ),
+        )
```
oliwenmandiamond, line 55: This shares the same return as `Apple2EnforceLHMoveController`. I wonder if there is a better way we could do this or just accept the duplication. Maybe create a another controller called FourArmApple2Controller or UnlockedApple2Controller which this and `Apple2EnforceLHMoveController` inherit from.

src/dodal/beamlines/i06_shared.py
```diff
@@ -24,6 +27,22 @@ def synchrotron() -> Synchrotron:
...
+        prefix=f"{PREFIX.beamline_prefix}-OP-PGM-01:",
+        grating=I06Grating,
+        grating_pv="NLINES2",
+    )
+
+
+# Insertion Device
+# ------------- Downstream Insertion Device --------------------
+@devices.factory()
+def i06_idd_epics_polynomial_device() -> I06EpicsPolynomialDevice:
```
oliwenmandiamond:
```suggestion
def idd_polynomial() -> I06EpicsPolynomialDevice:
```
We don't need beamline, epics and device on the instance

src/dodal/beamlines/i06_shared.py
```diff
@@ -44,6 +63,25 @@ def idd(idd_gap: UndulatorGap, idd_phase: UndulatorLockedPhaseAxes) -> Apple2:
...
+        apple2=idd,
+        gap_energy_motor_lut=i06_idd_epics_polynomial_device.energy_gap_motor_lookup,
+        phase_energy_motor_lut=i06_idd_epics_polynomial_device.energy_phase_motor_lookup,
+        inverse_gap_energy_motor_lut=i06_idd_epics_polynomial_device.gap_motor_energy_lookup,
+    )
+
+
+# -------------------- Upstream Insertion Device -------------------
+@devices.factory()
+def i06_idu_epics_polynomial_device() -> I06EpicsPolynomialDevice:
```
oliwenmandiamond:
```suggestion
def idu_polynomial() -> I06EpicsPolynomialDevice:
```

src/dodal/devices/beamlines/i06_shared/i06_epics_lookup.py
```diff
@@ -0,0 +1,170 @@
...
+    Pol.LH: [0],
+    Pol.LV: [MAXIMUM_ROW_PHASE_MOTOR_POSITION],
+    Pol.PC: [ROW_PHASE_CIRCULAR],
+    Pol.PC3: [ROW_PHASE_CIRCULAR],
+    Pol.NC: [-ROW_PHASE_CIRCULAR],
+    Pol.NC3: [-ROW_PHASE_CIRCULAR],
+}
+
+
+class I06EpicsPolynomialDevice(Device, Triggerable):
```
oliwenmandiamond, line 26: This class seems confused to me. This should be inheriting from `EnergyMotorLookup`. 

I think this possibly needs to be restructured so that we have three classes of this, one for `EnergyGapEpicsMotorLookup`, `EnergyPhaseEpicsMotorLookup`, and `GapMotorEpicsEnergyLookup` and then create an instance of each one. This way each logic is more focused and smaller.

Then we can also update base class method `EnergyMotorLookup.update_lookup` to be async and then create each one to then give to the controller. We could even make it automatically update by adding a monitor and then forcing an update table if detects a change.

src/dodal/devices/insertion_device/apple2_controller.py
```diff
@@ -102,20 +103,12 @@ def __init__(
...
-            float, initial_value=None, units=units
-        )
-        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
-            self.energy = derived_signal_rw(
-                raw_to_derived=self._read_energy,
-                set_derived=self._set_energy,
-                energy=self._energy,
-                derived_units=units,
-            )
+        self._energy = soft_signal_rw(float, initial_value=100)
```
oliwenmandiamond: Why have we given a random value of 100 as the initial value. Where are the units?

oliwenmandiamond (changes requested):
Thanks, is a good first start but I think the design needs a bit more work. Please see my comments

src/dodal/devices/beamlines/i06_shared/i06_apple2_controller.py
Relm-Arrowny: I think this will be moot when we add LA and extra lookup correct i06 has.

src/dodal/devices/beamlines/i06_shared/i06_epics_lookup.py
Relm-Arrowny: `EpicsPolynomialEnergyMotorLookup` now inherits from `EnergyMotorLookup` via a new shared `StaticPolynomialEnergyMotorLookup `class. I opted not to create three separate classes since their functions are identical, instead I make `I06EpicsPolynomialDevice` as a composition of EnergyMotorLookup.

We do not really want a autoupdate to prevent energy calibrations from changing without an explicit request through the triggerable interface. Not 100% if we want triggerable or not but it seem to me it much easier to do `yield from bps.trigger(device)` Then working out how to make blueapi do await device.update_lookup().

src/dodal/beamlines/i06_shared.py
```diff
@@ -24,6 +27,22 @@ def synchrotron() -> Synchrotron:
...
+        grating=I06Grating,
+        grating_pv="NLINES2",
+    )
+
+
+# Insertion Device
+# ------------- Downstream Insertion Device --------------------
+@devices.factory()
+def idd_epics_polynomial_device() -> I06EpicsPolynomialDevice:
+    return I06EpicsPolynomialDevice(prefix=f"{PREFIX.beamline_prefix}-OP-IDD-01:")
```
oliwenmandiamond:
```suggestion
def idd_polynomial() -> I06EpicsPolynomialDevice:
    return I06EpicsPolynomialDevice(prefix=f"{PREFIX.beamline_prefix}-OP-IDD-01:")
```

src/dodal/beamlines/i06_shared.py
```diff
@@ -44,6 +63,25 @@ def idd(idd_gap: UndulatorGap, idd_phase: UndulatorLockedPhaseAxes) -> Apple2:
...
+        gap_energy_motor_lut=idd_epics_polynomial_device.energy_gap_motor_lookup,
+        phase_energy_motor_lut=idd_epics_polynomial_device.energy_phase_motor_lookup,
+        inverse_gap_energy_motor_lut=idd_epics_polynomial_device.gap_motor_energy_lookup,
+    )
+
+
+# -------------------- Upstream Insertion Device -------------------
+@devices.factory()
+def idu_epics_polynomial_device() -> I06EpicsPolynomialDevice:
+    return I06EpicsPolynomialDevice(prefix=f"{PREFIX.beamline_prefix}-OP-IDU-01:")
```
oliwenmandiamond:
```suggestion
def idu_polynomial() -> I06EpicsPolynomialDevice:
    return I06EpicsPolynomialDevice(prefix=f"{PREFIX.beamline_prefix}-OP-IDU-01:")
```

src/dodal/devices/beamlines/i06_shared/i06_epics_lookup.py
```diff
@@ -0,0 +1,98 @@
...
+            Pol.NC3: "NC:HAR3",
+        }
+        self._inv_pol_map = {
+            Pol.LH: "BHZ",
+            Pol.LV: "BVT",
+            Pol.PC: "BPC",
+            Pol.NC: "BNC",
+            Pol.PC3: "BPC:HAR3",
+            Pol.NC3: "BNC:HAR3",
+        }
```
oliwenmandiamond: Is this necessary? inv_pol_map is the exact same as pol_map but has a B in front.

src/dodal/devices/insertion_device/apple2_controller.py
```diff
@@ -32,8 +30,8 @@
 
 
 class EnergyMotorConvertor(Protocol):
-    def __call__(self, energy: float, pol: Pol) -> float:
-        """Protocol to provide energy to motor position conversion."""
+    def __call__(self, value: float, pol: Pol) -> float:
+        """Protocol to provide energy, motor position conversion."""
```
oliwenmandiamond, line 35: Why change this? Surely energy makes more sense? It also makes this change a lot bigger than it really is as you then you have to propagate energy to value throughout the devices and tests.

src/dodal/devices/insertion_device/apple2_controller.py
```diff
@@ -247,57 +250,38 @@ def determine_phase_from_hardware(
...
-        ):
+        tol = ROW_PHASE_MOTOR_TOLERANCE
+        max_p = self.maximum_phase_motor_position
+
+        zero = {
+            "to": isclose(top_outer, 0.0, abs_tol=tol),
+            "ti": isclose(top_inner, 0.0, abs_tol=tol),
+            "bi": isclose(btm_inner, 0.0, abs_tol=tol),
+            "bo": isclose(btm_outer, 0.0, abs_tol=tol),
+        }
```
oliwenmandiamond: I think a dataclass would make more sense here

src/dodal/devices/insertion_device/energy_motor_lookup.py
```diff
@@ -79,3 +86,159 @@ def read_lut(self) -> LookupTable:
...
+        self,
+        mock: bool | DeviceMock = False,
+        timeout: float = DEFAULT_TIMEOUT,
+        force_reconnect: bool = False,
+    ) -> None:
+        await super().connect(
+            mock=mock, timeout=timeout, force_reconnect=force_reconnect
+        )
+        # Highjack connect to update lookup tables on connection.
+        await self.update_lookup()
```
oliwenmandiamond, line 244: Have you tested this with dodal connect? One issue I can see here is would this make dodal connect fail as it isn't connected to the daq-config-server service

src/dodal/beamlines/i06_shared.py
```diff
@@ -44,6 +63,25 @@ def idd(idd_gap: UndulatorGap, idd_phase: UndulatorLockedPhaseAxes) -> Apple2:
...
+    """I06 downstream insertion device controller."""
+    return I06Apple2Controller(
+        apple2=idd,
+        gap_energy_motor_lut=idd_epics_polynomial_device.energy_gap_motor_lookup,
+        phase_energy_motor_lut=idd_epics_polynomial_device.energy_phase_motor_lookup,
+        inverse_gap_energy_motor_lut=idd_epics_polynomial_device.gap_motor_energy_lookup,
+    )
+
+
+# -------------------- Upstream Insertion Device -------------------
```
oliwenmandiamond:
```suggestion
```

src/dodal/beamlines/i06_shared.py
```diff
@@ -24,6 +27,22 @@ def synchrotron() -> Synchrotron:
...
+def pgm() -> PlaneGratingMonochromator:
+    return PlaneGratingMonochromator(
+        prefix=f"{PREFIX.beamline_prefix}-OP-PGM-01:",
+        grating=I06Grating,
+        grating_pv="NLINES2",
+    )
+
+
+# Insertion Device
+# ------------- Downstream Insertion Device --------------------
```
oliwenmandiamond:
```suggestion
```
I don't think comment is needed

oliwenmandiamond (commented):
Thanks, I think this looks much better. Couple comments to help refine it and some questions I have

src/dodal/devices/insertion_device/energy_motor_lookup.py
oliwenmandiamond: Actually ignore this. This doesn't use daq-config-server as it is the epics version (duh). Tested and dodal connect works fine!

src/dodal/devices/insertion_device/apple2_controller.py
Relm-Arrowny: The original naming was single  direction of flow. Since the EnergyMotorConvertor now supports mapping in both directions, I’ve moved to value, which avoids the strange syntax of passing motor values into energy labelled parameters.

oliwenmandiamond: Ah that makes a lot more sense. Thanks for explaining

oliwenmandiamond (approved):
LGTM, thanks!

PR #2019, opened 2026-04-21
i15-1: add goniometer interlock 

EmsArnold:
Fixes crystallography_bluesky [#26](https://github.com/DiamondLightSource/crystallography-bluesky/issues/26)

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/beamlines/i15_1/gonio_interlock.py
+++ b/src/dodal/devices/beamlines/i15_1/gonio_interlock.py
@@ -24,6 +24,10 @@ class GonioInterlock(BaseHutchInterlock):
         )
 
     async def shutter_safe_to_operate(self) -> bool:
-        """Description."""
+        """Check device is safe to operate.
+
+        If the status value is not 65535 (healhty), the interlock has been tripped and
+        the device cannot be operated.
+        """
         interlock_state = await self.status.get_value()
         return isclose(float(interlock_state), GONIO_SAFE_FOR_OPERATIONS, abs_tol=5e-2)
--- a/src/dodal/devices/beamlines/i15_1/gonio_interlock.py
+++ b/src/dodal/devices/beamlines/i15_1/gonio_interlock.py
@@ -1,5 +1,7 @@
 from math import isclose
 
+from ophyd_async.core import EnumTypes
+
 from dodal.devices.hutch_shutter import BaseHutchInterlock
 
 GONIO_SAFE_FOR_OPERATIONS = 65535  # All 16 bits are true
@@ -23,11 +25,10 @@ class GonioInterlock(BaseHutchInterlock):
             name=name,
         )
 
-    async def shutter_safe_to_operate(self) -> bool:
+    def _safe_to_operate(self, status: float | EnumTypes) -> bool:
         """Check device is safe to operate.
 
         If the status value is not 65535 (healhty), the interlock has been tripped and
         the device cannot be operated.
         """
-        interlock_state = await self.status.get_value()
-        return isclose(float(interlock_state), GONIO_SAFE_FOR_OPERATIONS, abs_tol=5e-2)
+        return isclose(float(status), GONIO_SAFE_FOR_OPERATIONS, abs_tol=5e-2)
--- a/src/dodal/devices/beamlines/i15_1/gonio_interlock.py
+++ b/src/dodal/devices/beamlines/i15_1/gonio_interlock.py
@@ -1,0 +1,31 @@
+from math import isclose
+
+from dodal.devices.hutch_shutter import BaseHutchInterlock
+
+# TODO: Make this the right number
+GONIO_SAFE_FOR_OPERATIONS = 65535
+# GONIO_SAFE_FOR_OPERATIONS = 65439
+
+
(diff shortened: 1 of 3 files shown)


src/dodal/devices/beamlines/i15_1/gonio_interlock.py
```diff
@@ -0,0 +1,29 @@
+from math import isclose
+
+from dodal.devices.hutch_shutter import BaseHutchInterlock
+
+GONIO_SAFE_FOR_OPERATIONS = 65535  # All 16 bits are true
+
+
+class GonioInterlock(BaseHutchInterlock):
```
DominicOram, line 10: Should: `BaseHutchInterlock` implies that it's only for interlocks on the hutch. I think this is expanding it so maybe we should rename to `BaseInterlock`

src/dodal/devices/beamlines/i15_1/gonio_interlock.py
```diff
@@ -0,0 +1,29 @@
...
+            bl_prefix=bl_prefix,
+            interlock_infix=shtr_infix,
+            interlock_suffix=interlock_suffix,
+            name=name,
+        )
+
+    async def shutter_safe_to_operate(self) -> bool:
+        """Description."""
+        interlock_state = await self.status.get_value()
+        return isclose(float(interlock_state), GONIO_SAFE_FOR_OPERATIONS, abs_tol=5e-2)
```
jacob-williamson: Out of interest, why do we convert to float and have a tolerance here? The comment about 16 bits makes it sound like the `self.status` is a 16 bit value

src/dodal/devices/beamlines/i15_1/gonio_interlock.py
```diff
@@ -0,0 +1,29 @@
...
+        super().__init__(
+            signal_type=int,
+            bl_prefix=bl_prefix,
+            interlock_infix=shtr_infix,
+            interlock_suffix=interlock_suffix,
+            name=name,
+        )
+
+    async def shutter_safe_to_operate(self) -> bool:
+        """Description."""
```
jacob-williamson: Should: Fill out the docstring or remove

jacob720 (approved):
Looks good, just a small comment and question

DominicOram:
I am partway through reviewing this but ended up gettign sidetracked, more comments incoming...

src/dodal/devices/beamlines/i15_1/gonio_interlock.py
EmsArnold: This is emulating the [hutch interlock logic](https://github.com/DiamondLightSource/dodal/blob/cb0a654a86d716aeb1698e9c6518f82c7e7325d2/src/dodal/devices/hutch_shutter.py#L95) to check if the device is safe to operate. In practice, I think the conversion to float is necessary, as the other day during testing I did get an AssertionError that 65535 was not equal to 65535 before I added this in to my check.

With regards to the 16 bits, `65535` is simply the float equivalent of the binary `1111111111111111` - or all 16 flags of the interlock being healthy. This may be something that we ask controls to look at, to see if this can return an [enum like some other interlocks](https://github.com/DiamondLightSource/dodal/blob/cb0a654a86d716aeb1698e9c6518f82c7e7325d2/src/dodal/devices/hutch_shutter.py#L46).

DominicOram (changes requested):
Thanks. As I said in the other PR, I think https://github.com/DiamondLightSource/dodal/pull/2022 might help with this?

src/dodal/devices/beamlines/i15_1/gonio_interlock.py
EmsArnold: I agree, though I think this could also be expanded to the other interlocks as well - for example `PLCShutterInterlock` would also be useful for fast vacuum valves. I would lean towards re-naming these all at the same time.

DominicOram: Sure, feel free to spin into a new issue if you'd rather

EmsArnold: Will do - I think it would also warrant moving the interlocks out of `hutch_shutter.py` with similar reasoning. I wouldn't immediately look in `hutch_shutter` if I only needed an interlock.

DominicOram: Agreed

EmsArnold:
> Thanks. As I said in the other PR, I think #2022 might help with this?

That's all done now - updated to use `interlock.is_safe`

src/dodal/devices/beamlines/i15_1/gonio_interlock.py
DominicOram: For archaeology, issue is https://github.com/DiamondLightSource/dodal/issues/2024

DominicOram (approved):
Great, thanks. Might make sense to wait until https://github.com/DiamondLightSource/crystallography-bluesky/pull/31 is approved to merge both at the same time.

PR #2028, opened 2026-04-24
Update electron analyser to use single detector rather than mutlple

### LINKED ISSUE #1899: Data saved from read_configuration() is wrong on a device if using multiple streams.
When a device is used in a plan with multiple streams, the read_configuration() and read_description() method only get called once, for the first stream. It caches the data from the first stream and adds it all other streams. This leads to unexpected results in the data collected. Please see below for an example.

**Example code to reproduce**
Create dummy and very simple device that saves data per point and config data. 

```
class ExampleDriverIO(StandardReadable):
            self.region_name = soft_signal_rw(str, initial_value="null")
            self.excitation_energy = soft_signal_rw(float, initial_value=0, units="eV")

        super().__init__(name)


class ExampleDetector(Device, Stageable, Triggerable, AsyncReadable, AsyncConfigurable):
    def __init__(self, prefix: str, name: str):
        self.driver = ExampleDriverIO(prefix + CAM_SUFFIX)

        self.per_scan_metadata: list[SignalR] = [
            self.driver.region_name,
        ]
        self.per_point_metadata: list[SignalR] = [
            self.driver.excitation_energy,
        ]

        super().__init__(name)

    @AsyncStatus.wrap
    async def trigger(self) -> None:
        pass

    @AsyncStatus.wrap
        pass

    @AsyncStatus.wrap
        pass

    async def read(self) -> dict[str, Reading]:
        data = await merge_gathered_dicts(
            [signal.read() for signal in self.per_point_metadata]
        )
        return data

    async def describe(self) -> dict[str, DataKey]:
        data = await merge_gathered_dicts(
            [signal.describe() for signal in self.per_point_metadata]
        )
        return data

    async def read_configuration(self) -> dict[str, Reading]:
        return await merge_gathered_dicts(
            [signal.read() for signal in self.per_scan_metadata]
        )

    async def describe_configuration(self) -> dict[str, DataKey]:
        return await merge_gathered_dicts(
            [signal.describe() for signal in self.per_scan_metadata]
        )

```
**Next make a plan that switches stream on each iteration.**

```
def examplescan(analyser: ExampleDetector, region_list: list[str]):
    yield from bps.open_run()

    for i, region in enumerate(region_list):
        # setup analyser to look at the region
        yield from bps.mv(analyser.driver.region_name, region)
        yield from bps.mv(analyser.driver.excitation_energy, i)

        LOGGER.info(f"Acquiring region {region}")

        yield from bps.trigger_and_read([analyser], name=region)
        LOGGER.info("Finished acquiring region.")

    LOGGER.info("Closing run")
    yield from bps.close_run()
```

**Now run the plan with the detector**
```
det = ExampleDetector("TEST", "example")
RE(examplescan(det, ["test1", "test2", "test3"]))
```

**Examine data**
The data in the table is correct, the excitation energy field correctly incremented
```
>>> header = db[-1]
>>> header.table("test1")
                                 time  example-driver-excitation_energy
seq_num                                                                
1       2025-04-11 09:21:46.203407526                               0.0
>>> header.table("test2")
                                 time  example-driver-excitation_energy
seq_num                                                                
1       2025-04-11 09:21:46.210328102                               1.0
>>> header.table("test3")
                                 time  example-driver-excitation_energy
seq_num                                                                
1       2025-04-11 09:21:46.215673923                               2.0
```
However, the config_data is wrong. You can see the region_name is wrong for each new stream, it cached the first one and used it for all of them.
```
>>> header.config_data(det.name)["test1"]
[{'example-driver-region_name': 'test1'}]
>>> header.config_data(det.name)["test2"]
[{'example-driver-region_name': 'test1'}]
>>> header.config_data(det.name)["test3"]
[{'example-driver-region_name': 'test1'}]
```

Looking bluesky.bundles.py, it seems the issue is coming from here:

```
    async def _ensure_cached(self, obj, collect=False):
        coros = []
        if not collect and obj not in self._describe_cache:
            coros.append(self._cache_describe(obj))
        elif collect and obj not in self._describe_collect_cache:
            coros.append(self._cache_describe_collect(obj))
        if obj not in self._config_desc_cache:
            coros.append(self._cache_describe_config(obj))
            coros.append(self._cache_read_config(obj))
        await asyncio.gather(*coros)
```
It is checking if the obj (detector) exists in this dicts without considering if it's a new stream or not. The change below seems to fix the issue for me:
 
```
    # Check to see if we have done this stream before
    def _use_cache_for_obj(self, obj) -> bool:
        if self._bundle_name in self._descriptor_objs:
            if obj in self._descriptor_objs[self._bundle_name]:
                return True
        return False

    async def _ensure_cached(self, obj, collect=False):
        coros = []
        if not collect and not self._use_cache_for_obj(obj):
            coros.append(self._cache_describe(obj))
        elif collect and not self._use_cache_for_obj(obj):
            coros.append(self._cache_describe_collect(obj))
        if not self._use_cache_for_obj(obj):
            coros.append(self._cache_describe_config(obj))
            coros.append(self._cache_read_config(obj))
        await asyncio.gather(*coros)
```
Redoing test, it gives the expected result:
```
>>> header = db[-1]
>>> header.table("test1")
                                 time  example-driver-excitation_energy
seq_num                                                                
1       2025-04-11 09:37:14.495726824                               0.0
>>> header.table("test2")
                                 time  example-driver-excitation_energy
seq_num                                                                
1       2025-04-11 09:37:14.503090143                               1.0
>>> header.table("test3")
                                 time  example-driver-excitation_energy
seq_num                                                                
1       2025-04-11 09:37:14.508798121                               2.0
>>> header.config_data(det.name)["test1"]
[{'example-driver-region_name': 'test1'}]
>>> header.config_data(det.name)["test2"]
[{'example-driver-region_name': 'test2'}]
>>> header.config_data(det.name)["test3"]
[{'example-driver-region_name': 'test3'}]
```
However, I do not know the wider consequences of the suggested change above so will need to be investigated further.

## Acceptance Criteria
- When using multiple streams, read_configuration() and read_description() should be called once per new stream. Not once and then use same cache in every new stream.

oliwenmandiamond:
With the release of Bluesky 1.15.0 https://github.com/bluesky/bluesky/releases, we can now remove dynamically creating new electron analyser detectors (1 per region) to force the run engine to save new configuration. Details found here and now fixed https://github.com/bluesky/bluesky/issues/1899. Instead, we can just use a single detector now and the `run_engine` will correctly save the configuration for each new stream for each region.

Related sm-bluesky change https://github.com/DiamondLightSource/sm-bluesky/pull/245.

1. Check tests make sense
2. Check tests pass

### Checks for reviewer


src/dodal/devices/electron_analyser/base/base_detector.py
```diff
@@ -18,13 +19,39 @@
...
+    def __init__(self):
+        self.data: AbstractBaseSequence[AbstractBaseRegion] | None = None
+
+    @AsyncStatus.wrap
+    async def prepare(self, value: AbstractBaseSequence[AbstractBaseRegion] | None):
+        self.data = value
+
+    @AsyncStatus.wrap
+    async def stage(self):
+        pass
```
Relm-Arrowny, line 45: Should we catch staging empty data?
```suggestion
        if self.data is None:
            raise RuntimeError(
               "Bah bah bah"
            )
```

src/dodal/devices/electron_analyser/base/base_detector.py
```diff
@@ -18,13 +19,39 @@
...
+    AbstractBaseRegion,
+    AbstractBaseSequence,
     GenericRegion,
     TAbstractBaseRegion,
 )
 
 
-class BaseElectronAnalyserDetector(
+class SequenceHolder(Stageable, Preparable):
+    """Wrapper to hold the sequence data for an electron analyser.
```
Relm-Arrowny, line 30: Not sure this is a wrapper, Stateholder, dataclass maybe?

Relm-Arrowny (changes requested):
I really like how little change in bluesky make this so much cleaner. Some minor point otherwise, it almost looked cool.

src/dodal/devices/electron_analyser/base/base_detector.py
oliwenmandiamond: So this is the debate I've kinda had with myself, do I make it Bluesky like and wrap it in an object to properly interact with it in plans, or do i just make it an internal state that I set inside the plan

and in plans instead of doing
```Python
yield from bps.prepare(analyser.sequence, sequence)
```
I just do
```Python
analyser.sequence = sequence
```
I guess an argument to just remove this wrapper/state holder/dataclass or whatever it is that the user should never try to prepare the device themselves because it can only be used with analyserscan plans which take the sequence object as an argument anyway to setup the analyser. 

I'm kinda leaning towards just removing the wrapper entirely and just do option 2

src/dodal/devices/electron_analyser/base/base_detector.py
oliwenmandiamond: I think the answer is no. So we have two ways to use the electron analyser.

We use an `analyserscan` plan. This takes the sequence data that holds a list of regions as an argument and configures the analyser with it. At every point, we do an additional loop for each region configured via the sequence data on the analyser (will cover Specs and VGScienta case). This is also why the sequence needs to be attached to the analyser and this is the only way to get a reference to it when inside the scan engine at each step using a custom [per_step](https://github.com/DiamondLightSource/sm-bluesky/pull/245/changes#diff-ba00994d773a5d80411fb32742e36f7a8941986584361095972000f899f10a8aR24) in a normal scan:

```Python
>>> bc.plans.analyserscan(bc.devices.analyser, load_sequence("/path/to/data"), ...) 
# load_sequence returns a data structure which is then serialisable with blueapi and passed to the plan to configure the analyser with
```

We use a standard scan. The user can configure the detector first by setting a single region. (Will cover Mbs case)
```Python
>>> r = load_sequence("/path/to/data").get("My Region")
```
or
```Python
>>> r = load_region("path/to/data")
```
And then configure before scan
```Python
>>> bc.plans.move(bc.devices.analyser, r)
>>> bc.plans.scan([bc.devices.analyser], ...)
```

So if we try to do scan 2, then it will raise an error as we didn't configure a sequence. So the check we need to do is inside analyserscan plan which is done [here](https://github.com/DiamondLightSource/sm-bluesky/pull/245/changes#diff-ba00994d773a5d80411fb32742e36f7a8941986584361095972000f899f10a8aR48). 

@Villtord thoughts?

src/dodal/devices/electron_analyser/base/base_detector.py
Relm-Arrowny: I probably missing something can't we just make analyser prepareable and do both of what you trying to do here? 

```Python
yield from bps.prepare(analyser, sequence)
and
analyser.sequence = sequence
```
something like this?
```Python
async def prepare(self, value:):
    self.sequence = value

    if self.sequence is None:
        raise bah
    await self._controller.disarm()

    await self._controller.disarm()
    self._sequence = None
```
With this the other comment is moot, it will fail as soon as we try to run a scan without having regions defined.

src/dodal/devices/electron_analyser/base/base_detector.py
Relm-Arrowny: I think both would benefit by having the earlier fail. as it fails on stage which happens before we do anything else in a scan?

src/dodal/devices/electron_analyser/base/base_detector.py
Villtord: If your last code is for analyser object then it won't work as analyser can't cycle regions through the sequence and start scan for each region - currently it can only do so for 1 region hence the need for custom scan plan - i.e. `anayserscan`
Another question is maybe we can also convert single region into a sequence such that even single region can be used in `analyserscan` plan?

src/dodal/devices/electron_analyser/base/base_detector.py
Villtord: Can we convert single region that is configured via `bc.plans.move(bc.devices.analyser, r)` into a some kind of default sequence internally? perhaps it could be a more unified approach, would be great if we can have one logic
If we do that then we can use this suggestion of Raymond.

Villtord: Is it actually true that `stage` is called later than `analysercan` checks if sequence is none? I thought `analyserscan` first checks sequence and then pass it on normal `nd_scan` which makes staging .

Relm-Arrowny: `analysercan` only call `prepare` and `scan` and I think the check comes in when `per_step` is call during a scan, that normally happen after `stage` and `run` is called. e.g. [scan](https://github.com/bluesky/bluesky/blob/4390e3dce0b04472157664b7de3111b18469f59c/src/bluesky/plans.py#L493)

Villtord: agreed

Relm-Arrowny (approved):
As mention in person, if you want the keep detector region state you can ignore the stage comment otherwise looks good to me.

PR #2033, opened 2026-04-28
Decouple interlocks from shutter

### LINKED ISSUE #2024: Interlocks should be de-coupled from hutch shutters
Interlocks are not only useful for shutters, but for other valves and devices. Currently, interlocks are tied to hutches and shutters; base interlocks should be renamed and moved to be separate from shutters.

## Acceptance Criteria
* Base interlocks do not mention shutters / hutches
* Base interlocks are moved out of `hutch_shutter.py` to increase discoverability

EmsArnold:
Fixes #2024

Interlocks should be decoupled from hutch shutters. This moves all potentially useful interlocks into `interlocks.py`. 

1. Check naming of interlocks
2. Confirm relevant beamlines connect (i03, i15_1, i19_optics, i24)

### Checks for reviewer


DominicOram (approved):
Looks good to me, thanks. Might be worth a second pair of eyes from @DiamondLightSource/developers-mx-daq given it changes their beamlines a little too.

src/dodal/devices/interlocks.py
```diff
@@ -0,0 +1,140 @@
...
+    def _safe_to_operate(self, status: float | EnumTypes) -> bool:
+        """If the status value is OK or Run Ilk OK, shutter is open or safe to operate.
+
+        If the status value is "OK", valve or shutter is open and interlocks are OK to
+        operate. If the status value is "Run Ilks Ok", the opening action can be
+        started. If the status value is not OK ("Failed" or "Disarmed"), interlocks have
+        failed and the shutter cannot be operated. Disarmed status applies only to fast
+        shutters.
+        """
+        return (status == PLCInterlockState.OK) | (
```
CoePaul, line 138: it seems counterintuitive here to have an OR 

a) usually something safety related will only say yes after a compound AND 
( I mean that's basic / fundamental / vital )

b) the implication is that this one piece of code 
is maybe covering two types of hardware and doesn't know which hardware type ( sub-genre whatever ) 
it is representing so has to go and ask OR type questions to cover all bases 

In a situation where we are using a bit of OO 
this is exactly where I'd expect one child / grandchild class to cover each hardware type in a 1:1 mapping even if it were only to override this one method 

Software people will throw up their hands in horrror 
but really the world should not revolve around the axis of software convenience
IMHO

EmsArnold:
@CoePaul mentioned offline:

> I take that as a sign that my wondering about names like PSSInterlock
> or IntPLCSomething are ill-positioned in time and scope
> 
> I thought camel case was all like
> 
> PssInterlock
> and
> IntPlcSomething
> 
> acronyms were supposed to be only capital on the first letter
> a constant bug bear in GDA is how much UDCSomething there is when it should have been
> UdcSomething ... are we really repeating that inconsistency-for-the-heck-of-it in dodal ?

I wasn't aware of the history, but am happy to change if that's the convention we're keeping to. It seems to go [against pep8](https://peps.python.org/pep-0008/#descriptive-naming-styles), so we may want think about adding this decision (and the reasoning) to the dodal style guide.

src/dodal/devices/interlocks.py
EmsArnold: From my understanding, `OK` is when the valve/shutter is open AND the interlocks are all okay, whereas `Run Ilks OK` is when the interlocks are all okay and thus the valve/shutter can be operated. The only enum specific to a device is actually `Failed`, which is only used for fast valves. 

I think this is then a third option c) the idea is to check whether the device is safe to operate (open or closed), whereas the enum is specific to either opening or closing? I suppose the fix then would be to check if something is open and can be closed, or closed and can be opened?

DominicOram:
>It seems to go [against pep8](https://peps.python.org/pep-0008/#descriptive-naming-styles), so we may want think about adding this decision (and the reasoning) to the dodal style guide.

I would strongly advocate for maintaining the PEP8 rules and having acronyms as all caps. This also matches the style of the standard library e.g. [HTTPStatus](https://docs.python.org/3/library/http.html#http.HTTPStatus) or [BytesIO](https://docs.python.org/3/library/io.html#io.BytesIO)

EmsArnold:
A further note that this change will require any plans which reference `HutchInterlock` to be renamed to `PSSInterlock`, and any which reference `BaseHutchInterlock` to be renamed to `BaseInterlock`

It looks like `HutchInterlock` pops up a few times in `mx-bluesky`

EmsArnold:
With fixes ready in both [crystallography-bluesky](https://github.com/DiamondLightSource/crystallography-bluesky/pull/40) and [mx-bluesky](https://github.com/DiamondLightSource/mx-bluesky/pull/1743), @noemifrisina and @CoePaul, are you both happy for this to go forward?

src/dodal/devices/interlocks.py
CoePaul: I think I was confused then: 
My assumption had been that this method was making safety-like decisions 
but instead it is only reporting on the decision which has already been made by a different system.
Maybe a name more like reported_safe_to_operate was what I was expecting in that case - but that's just my over-literal mindedness when I read code

CoePaul:
@EmsArnold - all looks good - now I've understood that it wasn't a camel after all but a PEP8 ( obscura lex pythonica )

CoePaul (approved):

PR #2036, opened 2026-04-30
Update scan plans to allow zero detectors

### LINKED ISSUE #2035: When using wrapped dodal scans, you must provide a detector
When testing the functionality of the dodal wrapped commands, I wanted to run the following command

```Python
>>> RE(list_grid_scan([], [(source1, [1, 5, 10, 50])]))
  File "<stdin>", line 1, in <module>
  File "/scratch/bluesky_development/dodal/.venv/lib64/python3.11/site-packages/pydantic/_internal/_validate_call.py", line 40, in wrapper_function
    return wrapper(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/scratch/bluesky_development/dodal/.venv/lib64/python3.11/site-packages/pydantic/_internal/_validate_call.py", line 137, in __call__
    res = self.__pydantic_validator__.validate_python(pydantic_core.ArgsKwargs(args, kwargs))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for list_grid_scan
0
  Value should have at least 1 item after validation, not 0 [type=too_short, input_value=[], input_type=list]
    For further information visit https://errors.pydantic.dev/2.13/v/too_short
```
However, I was getting this error message. I worked out it was because the detectors was empty. 

The first issue is that the error message is not obvious the detector is the issue as it doesn't point to which argument is wrong. The second and main one is that we shouldn't be enforcing the need to add a detector. It's perfectly legitimate to do a scan without detectors and is very useful for testing to not need to add it.

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/2035

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


Villtord (approved):
LGTM but would be good to have a warning somewhere downstream that there were no detectors provided for this scan.

Villtord:
yep would be good to add a test for 0 detectors then as well

PR #2043, opened 2026-05-01
Adds basic unit conversions for attenuator in general_maths folder

### LINKED ISSUE #2038: Add basic conversions ( used by- / useful for- attenuation ) to arithmetic module
Assuming an arithmetic module is ( or is about to exist ) in dodal common 
(see parent issue #2037  ) ...

Make cohesive set of unit tested functions for useful conversions
https://confluence.diamond.ac.uk/x/56rSEQ
Contains a full list of conversions written for I19 in Jython ( as part of GDA )

We can ignore polynomial ( as numpy will do that )
but the methods for 
range checking ( might be delegate-able to external library )
and
convert_something_to_something should be represented

and all equivalent unit test cases should be exercised 
ie all tests mentioned here
https://confluence.diamond.ac.uk/x/6arSEQ
and all tests mentioned here
https://confluence.diamond.ac.uk/x/IqvSEQ

## Advice

1. Start with the happy path tests
2. prove a test fails
3. write the bare minimum code to make that test pass
4. write a new test that fails - prove it does
5. go back to 3 - until all happy path tests have failed and were then made to pass

For the unhappy path tests - some of the exceptions might be different in python 3

Also if some checks are better covered using PyDantic then feel free to delegate some aspects to it

## Acceptance Criteria
* That code exists to perform the requested conversions
* That unit test cases listed in the aforementioned confluence docs are present 
   ( unless we have a good reason to exclude them )

Matt-Carre:
Closes #2038 

Begins to work on translating over the calculations behind the attenuator to be available in Dodal. Separates conversions and boundary checks into two different files. Also translates over unit tests.


src/dodal/common/general_maths/arithmetic_conversions.py
```diff
@@ -0,0 +1,43 @@
+def calculate(num: float, multiplier: float):
```
CoePaul: calculate is not an appropriate name

plus we already have the * operator in the basic language for this purpose 

not necessary therefore to break out to another method 
( although that general instinct is sound - You've hit the natural limit when You reach an internal symbol of the language )

src/dodal/common/general_maths/arithmetic_conversions.py
```diff
@@ -0,0 +1,43 @@
...
+    return num * multiplier
+
+
+def convert_percentage_to_factor(pc: float):
+    # Takes a percentage value and converts it the corresponding multiplication factor
```
CoePaul: I believe there is a dodal standard for documentation 
and this is not it 

I'll find a link

CoePaul: https://google.github.io/styleguide/pyguide.html section 3.8 in particular

src/dodal/common/general_maths/check_bounds.py
```diff
@@ -0,0 +1,22 @@
+def subordinate_range_test(lower_bound: float, upper_bound: float, x: float):
+    outside_range = x < lower_bound or x > upper_bound
+    return not outside_range
+
+
+def is_within_range(lower_bound: float, upper_bound: float, tested_value: float):
+
+    if upper_bound < lower_bound:
```
CoePaul: We have been using PyDantic for this sort of validation.
I'll let Noemi decide how mandatory using PyDantic ( vs simple operators ) is.

CoePaul (changes requested):

src/dodal/common/general_maths/arithmetic_conversions.py
jacob-williamson: There is a vs code extension called autoDocstring that I find helpful for this, it auto-generates a docstring template that follows the standard.

src/dodal/common/general_maths/check_bounds.py
noemifrisina: Yes, as mentioned yesterday, I think this is a good use case for it
Note that when changing this, the tests should be updated accordingly...

src/dodal/common/general_maths/arithmetic_conversions.py
```diff
@@ -0,0 +1,93 @@
+def convert_percentage_to_factor(pc: float):
```
noemifrisina: Should: add return types (same for the other file)

src/dodal/common/general_maths/arithmetic_conversions.py
noemifrisina, line 1: Could: Add a small docstring on top of both files giving an idea of what can be found in here

noemifrisina (changes requested):

oliwenmandiamond:

noemifrisina (approved):

PR #2044, opened 2026-05-01
Remove device_factory and supporting methods

tpoliaw:
Requires matching changes to avoid breakage:

* DiamondLightSource/blueapi#1534
* mx-bluesky changes DiamondLightSource/mx-bluesky#1732

All beamlines are now using a device manager to mark the device factories so the previous decorator and supporting code is no longer required.


DominicOram (changes requested):
`clear_devices` is used by `mx-bluesky` so this will require some changes there too. I'm not sure how significant they will be @DiamondLightSource/developers-mx-daq?

src/dodal/common/beamlines/beamline_utils.py
```diff
@@ -33,134 +13,6 @@ def set_beamline(beamline: str):
...
-            use_factory_name,
-            timeout,
-            mock,
-            skip,
-        )
-
-    return decorator
-
-
 def set_path_provider(provider: PathProvider):
```
DominicOram, line 16: Should: Do we still need this? And the get/clear? It was my understanding that BlueAPI now handles the path provider a different way?

DominicOram (changes requested):

src/dodal/common/beamlines/beamline_utils.py
rtuck99: Hyperion is still using it because it sets its own path provider

DominicOram: That's annoying. Can we have BlueAPI inject the path provider into the composite so that we can avoid having the globals around?

oliwenmandiamond:
@DominicOram can this be merged now? The blocking change seems merged now https://github.com/DiamondLightSource/mx-bluesky/pull/1731/changes?

DominicOram (approved):
Yep, fine by me

oliwenmandiamond:
@tpoliaw do you want to coordinate merging this when https://github.com/DiamondLightSource/blueapi/pull/1534 is merged?

tpoliaw:
Yep, will do

src/dodal/common/beamlines/beamline_utils.py
tpoliaw: Where is hyperion setting the path provider? I can see a call to `get_path_provider` in `setup_panda` but no call to `set_path_provider` anywhere.

PR #2047, opened 2026-05-05
Adds attenuation transmission interconversion calculations

### LINKED ISSUE #2040: Extend beamline-relevant arithmetic module with attenuation-transmission interconversion
In the context of the I19 attenuation system
we require 4 numerical functions which can 
a) provide attenuation from the natural log of transmission
b) provide attenuation from transmission ( as fraction ie in the > 0.0 up to and including 1.0 range )
c) in the opposite direction to a)
d) in the opposite direction to b)

Note that the units for the output of a) and b) will be Barnett units 
( see https://confluence.diamond.ac.uk/x/U4c0EQ )

( secondary reading https://confluence.diamond.ac.uk/x/V4c0EQ and 
https://confluence.diamond.ac.uk/x/bAA6Dw )

We also require unit test coverage of functionality as listed in tests mentioned in 
https://confluence.diamond.ac.uk/x/OqrSEQ
https://confluence.diamond.ac.uk/x/84xFEw
https://confluence.diamond.ac.uk/x/8IxFEw
https://confluence.diamond.ac.uk/x/6oxFEw

## Acceptance Criteria
* That the four requested maths functions exist 
* That the corresponding unit test cases ( as documented ) 
a) exist 
b) meet specifications 
c) can be demonstrated to fail 
d) eventually pass

failure testing should include trying them with slightly off values to make sure the calculation is suitably sensitive

e.g. 
write a happy path test with expected input -> expected output

You could:
write a second test with expected input + tiny shift and make the test criterion that the originally expected output is not the result 
and name the test "original_test_name_sensitivity_check" or similar

Matt-Carre:
Fixes #2040 

This adds calculations to convert between the:
attenuation and natural log of transmission and back,
transmission fraction and attenuation and back

Also includes tests for these functions.


src/dodal/common/general_maths/transmission_interconversion.py
```diff
@@ -0,0 +1,56 @@
+import math
+
+canonical_barnett_conversion = -1.0e3
```
CoePaul: I feel like I've commented on this in another review 
( so excuse any comment duplication )

In the original Jython these constants were inside a class
Now ( here ) they are not - so it would be good to start the names with an underscore or in some similar manner indicate that these are meant to be private to this script only

tests/common/general_maths/test_transmission_interconversion.py
```diff
@@ -0,0 +1,76 @@
+import math
+
+import pytest
+
+from dodal.common.general_maths import transmission_interconversion
+
+
+# happy:
+@pytest.mark.parametrize(
+    "ln_t,result", [(-1, 1000), (0, 0), (-0.4367, 436.7), (-5.9017, 5901.7)]
```
CoePaul: As commented in parallel reviews 
the original Jython test names gave a narrative context regarding "why" specific numbers were chosen, which were arbitrary and which had specific significance ( even if the significance is artificial / human factor only - ie not really important in the grand scheme )

but that's all lost now without any labelling

Please find a way to explain the purpose of the number choices 
lest a future developer just modify them arbitrarily and lose the specific tests

tests/common/general_maths/test_transmission_interconversion.py
```diff
@@ -0,0 +1,76 @@
...
+    assert transmission_interconversion.transmission_from_attenutation(
+        attenuation_bn
+    ) == pytest.approx(result)
+
+
+# inauspicious:
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_attenuation_from_natural_log_of_transmission_errors(bad_input):
```
CoePaul: perhaps test_x_raises_error is a better name than test_x_errors
to error is human
to raise an error is computer science

CoePaul (changes requested):

src/dodal/common/general_maths/transmission_interconversion.py
```diff
@@ -0,0 +1,55 @@
...
+    """Converts from natural log of transmission fraction into Barnett attenuation units
+    .
+
+    Args:
+        ln_t (float): natural log of transmission fraction
+
+    Returns:
+        (float): Barnett attenuation units
+    """
+    _canonical_barnett_conversion = -1.0e3
```
noemifrisina: Should: if these are constants for the module/file, they may as well be defined as such, so outside the functions in all caps

src/dodal/common/general_maths/transmission_interconversion.py
```diff
@@ -0,0 +1,55 @@
+import math
+
+
+def attenuation_from_natural_log_of_transmission(ln_t: float):
```
noemifrisina: Should: return types

noemifrisina (changes requested):

oliwenmandiamond:

noemifrisina (approved):
lgtm, just update branch before merging please

PR #2048, opened 2026-05-05
Adds photon mass attenuation calculations

### LINKED ISSUE #2039: Extend arithmetic module with calculations library script for photon mass attenuation
Based on equivalent mathematical functions and unit tests written for GDA in Jython 
( see https://confluence.diamond.ac.uk/x/JYc0EQ )

1. Create function for photon_mass_attenuation_per_unit_length( energy_kev, photon_absorption_factor_per_unit_length, energy_dependence_exponent)

Note the agnosticism to unit length choice - this function doesn't care which length scale You are using

2. Recreate unit test cases and be sure to include relevant energy values ( in the happy path cases )
that match i19 typical choices 

See gerrit 42287 and the Jython script test_photon_mass_attenuation.py

feel free to write Your own example test cases

## Acceptance Criteria
* That a mathematical function exists which can perform the requested attenuation calculation
* That unit tests cover its performance over typical energies relevant for at least the I19 beamlines

Matt-Carre:
Fixes #2039  - adds ability to calculate the photon mass attenuation per unit length, and tests this functionality.


src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,19 @@
...
+
+    Args:
+        energy_kev (float): energy
+        photon_absorption_factor_per_unit_length (float): photon absorption factor per
+        unit length
+        energy_dependence_exponent (float): energy dependence exponent
+
+    Returns:
+        (float): mass attenuation per unit length.
+    """
```
CoePaul, line 20: rather than write out everything-in-one-line 

( bearing in mind the original motivation for this entire project in GDA Jython 
  was to unpack an-all-in-one equation into its separate, easier to identify 
  and understand parts )

I'd rather retain the two lines 
roll_off = energy ** exponent 
return factor * roll_off

knowing that the first bit is a roll off helps with the physics understand / reader comprehension as to "Why do we have this calculation like this?"

src/dodal/common/general_maths/transmission_interconversion.py
```diff
@@ -0,0 +1,56 @@
+import math
+
+canonical_barnett_conversion = -1.0e3
```
CoePaul: In the original Jython - the constants were defined inside a class

Now that they are "exposed" it would better if the names began with an underscore
to indicate that they are not meant for external use / consumption / modification

tests/common/general_maths/test_transmission_interconversion.py
```diff
@@ -0,0 +1,76 @@
+import math
+
+import pytest
+
+from dodal.common.general_maths import transmission_interconversion
+
+
+# happy:
+@pytest.mark.parametrize(
+    "ln_t,result", [(-1, 1000), (0, 0), (-0.4367, 436.7), (-5.9017, 5901.7)]
```
CoePaul: As before 
in the original Jython the names of the specific test cases explained the narrative significance of the test values.

Now that narrative is lost and we just have " a bunch of numbers "

which means future developers will see no reason not to replace them with any other test examples.

If there is a way to label the numbers to explain the reason for using them - that would be great

CoePaul: Same comment for all tests here, there and everywhere

oliwenmandiamond:

src/dodal/common/general_maths/transmission_interconversion.py
```diff
@@ -0,0 +1,55 @@
+import math
```
CoePaul, line 1: As mentioned off-line 
if the input could be a StrictFloat that would help prevent boolean inputs

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,48 @@
...
+    photon_mass_attenuation_per_unit_length,
+)
+
+
+# happy path
+@pytest.mark.parametrize(
+    "energy_kev,photon_absorption_factor_per_unit_length,energy_dependence_exponent,"
+    "result",
+    [
+        (5.042, 1.98e2, -2.717, 2.44170544),  # Arbitrary
```
CoePaul: Arbitrary energy would be slightly clearer 
( the other params, ie not energy, are arbitrary on every line )

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,48 @@
...
+)
+def test_photon_mass_attenuation_per_unit_length(
+    energy_kev,
+    photon_absorption_factor_per_unit_length,
+    energy_dependence_exponent,
+    result,
+):
+    assert photon_mass_attenuation_per_unit_length(
+        energy_kev, photon_absorption_factor_per_unit_length, energy_dependence_exponent
+    ) == pytest.approx(result)
```
CoePaul: I think the default approx is only to 1 part in 10^6 but we are specc-ing the expected outputs to < counts d.places .. > 7 dp so maybe adjust the equality threshold for pytest.approx in this test to 5.0e-8 or similar

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+# inauspicious path
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
```
CoePaul: as mentioned off-line - include a boolean input too

tests/common/general_maths/test_transmission_interconversion.py
```diff
@@ -0,0 +1,124 @@
...
+        transmission_from_attenutation(bad_input)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_attenuation_from_transmission_raises_error(bad_input):
+    with pytest.raises(TypeError):
+        attenuation_from_transmission(bad_input)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
```
CoePaul: as mentioned elsewhere - include a boolean in the invalid set

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,18 @@
+def photon_mass_attenuation_per_unit_length(
+    energy_kev: float,
```
CoePaul: Change request to make inputs into StrictFloat types - sorry for not considering this before

CoePaul (dismissed):

Relm-Arrowny:
Got attracted to this issue seeing photon mass calculations!!! and got disappointed it is photon mass attenuation, May want to add attenuation to the pr title. As far as I know photon is massless at least at rest, E = m^2C^4+p^2c^2 and photon mass is refer to the m here which zero.

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,51 @@
...
+)
+
+
+# happy path
+@pytest.mark.parametrize(
+    "energy_kev,photon_absorption_factor_per_unit_length,energy_dependence_exponent,"
+    "result",
+    [
+        (5.042, 1.98e2, -2.717, 2.44170544),  # Arbitrary Energy
+        (8.3328, 2.5706e3, -2.83, 6.3708311),  # Arbitrary Nickel
```
noemifrisina: Nit: the wording "arbitrary {element}" is extremely unclear.

noemifrisina (approved):
Looks okay, thanks - just one comment but take it or leave it
Also please rebase before merging

CoePaul:
> Got attracted to this issue seeing photon mass calculations!!! and got disappointed it is photon mass attenuation, May want to add attenuation to the pr title. As far as I know photon is massless at least at rest, E = m^2C^4+p^2c^2 and photon mass is refer to the m here which zero.

@Relm-Arrowny - the LHS of Your equation should have read "E^2 = ..."

and it is indeed unfortunate that the name "mass" appears in the allegedly standardised phrase "mass attenuation" 
but I think that's referring to the mass of the absorber ( for the purposes of the overall modelling )

Photon states - when coupled to matter due gain an effective mass - e.g. plasmons etc but that's not our concern here

PR #2049, opened 2026-05-05
Adds conversion between material depth and attenuation

### LINKED ISSUE #2041: Extend beamline-relevant arithmetic to provide interconversion between attenuation and material depth
In the context of attenuation calculations for I19 ...
( see https://confluence.diamond.ac.uk/x/OIc0EQ )

Create a script with methods for converting between 

1. material thickness in cm and attenuation in Barnett units
2. the opposite direction

Create unit tests to exercise these functions 

The methods should delegate sub-steps of the calculation to the existing library methods created in related sub-issues off this common parent issue 

e.g. photon mass attenuation calculations - see existing Jython code for GDA 
Note also the design decision to reject negative material depth and transmission coefficients > 1.0
( the latter is valid - for a system with optical gain, e.g. a laser or the synchrotron insert device itself - but we're 
 not modelling optical gain here so we reject that input - if You want to use PyDantic to exclude / filter invalid values
 feel free )

See documentation on Jython versions of the unit tests 
https://confluence.diamond.ac.uk/x/J4c0EQ 
https://confluence.diamond.ac.uk/x/nYk0EQ

## Acceptance Criteria
* That the requested conversion functions exist 
* That the requested unit test cases are covered

Matt-Carre:
Fixes #2041 
Adds tests and functions to convert between depth (measured in cm) and attenuation


src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,85 @@
...
+    # TODO: check this - not sure if it really is the depth of the wedge?
+    """Calculates attenuation in Barnett units.
+
+    Args:
+        depth_cm (float): depth of wedge
+        absorption_coefficient_per_cm (float): absorption coefficient per cm
+
+    Raises:
+        ValueError: If either input value for either depth_cm or absorption_coefficient_
+        per_cm <0.0, raises a value error.
```
CoePaul: the comment would be better written in English ( rather than in code )
not 
[ Raises ] ValueError: If either input value for a or b < 0.0 raises a value error
but instead
[ Raises ] ValueError: If a or b are negative an error is raised

At present the comment is just a cut and paste of the code and is therefore duplication and so redundant - the comment is supposed to be the voice of one developer to another explaining the "Why" not the "what" ... the code already answers the "what" by itself

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    if depth_cm < 0.0:
+        raise ValueError(f"Negative depth is an invalid input: {depth_cm}")
+    if absorption_coefficient_per_cm < 0.0:
+        raise ValueError(
+            "Invalid absorption, this calculator is no for systems with\
+                         optical gain."
+        )
+    return transmission_interconversion.attenuation_from_natural_log_of_transmission(
+        -(depth_cm * absorption_coefficient_per_cm)
```
CoePaul: as in other comments: 
Please do not see this exercise as an invitation to pack all the operations onto one line 
The aim of the exercise is to do as little per line as possible in order to make the narrative purpose of each step more obvious
It's also benificial when debugging an error in live operations to know 
which step hit an error 

( If instead an error is raised on a line doing 6 things at once - You try answering a support call to debug it at 2 am )

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+    Raises:
+        ValueError: if attenuation is below zero, or absorption is below the minimum
+        meaningful absorption coefficient, a value error is raised
+
+    Returns:
+        (float): material depth in cm.
+    """
+    minimum_meaningful_absorption_coefficient = 1.0e-14
```
CoePaul: Good use of named constant

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def thickness_cm_required_to_attenuate(
+    target_attenuation_bn: float,
+    absorption_coefficient_per_cm: float,
+):
+    """Calculates material depth in cm.
+
+    Args:
+        target_attenuation_bn (float): Target attenuation to meet in Barnett attenuation
```
CoePaul: Target attenuation should use PositiveFloat type from PyDantic 
https://pydantic.dev/docs/validation/2.0/usage/types/number_types/ 

( and similarly throughout the code base where positive numbers are required )
 ( N.B. this type was not available to the original Jython implentation )

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,123 @@
...
+)
+def test_photon_mass_attenuation_per_unit_length(
+    energy_kev,
+    photon_absorption_factor_per_unit_length,
+    energy_dependence_exponent,
+    result,
+):
+    assert material_absorption_maths.photon_mass_attenuation_per_unit_length(
+        energy_kev, photon_absorption_factor_per_unit_length, energy_dependence_exponent
+    ) == pytest.approx(result)
```
CoePaul, line 32: Can we use the rel or abs flags ( prefer the former here ) 
to make it clear which precision criterion is in use for the approx 
( without future developers having to go look it up and reverse engineer 
 https://docs.pytest.org/en/4.6.x/reference.html#pytest-approx )

CoePaul: same for all uses of pytest approx through all these numerical tests

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,123 @@
...
+def test_attenuation_at_depth_cm(depth_cm, absorption_coefficient_per_cm, result):
+    assert material_absorption_maths.attenuation_at_depth_cm(
+        depth_cm, absorption_coefficient_per_cm
+    ) == pytest.approx(result)
+
+
+@pytest.mark.parametrize(
+    "target_attenuation_bn,absorption_coefficient_per_cm,result",
+    [
+        (0, 2.4, 0),
```
CoePaul: again the narrative purpose of the inputs is now obscure - where the original Jython tests explained the reason for particular choices

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,123 @@
...
+    with pytest.raises(ValueError):
+        material_absorption_maths.thickness_cm_required_to_attenuate(1, 1e-15)
+
+
+def test_attenuation_at_depth_cm_too_negative_target():
+    with pytest.raises(ValueError):
+        material_absorption_maths.thickness_cm_required_to_attenuate(-1, 1)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
```
CoePaul: doesn't always have to be math.sin
You could switch it up with cos etc if You like

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+@pytest.mark.parametrize(
+    "target_attenuation_bn,absorption_coefficient_per_cm,result",
+    [
+        (0, 2.4, 0),
+        (248.461, 2.13, 0.1166483568),
+    ],
+)
+def test_thickness_cm_required_to_attenuate(
```
CoePaul, line 46: Preference: 
in the original Jython tests - each tested method ( function ) 
has its own test script 

Here instead we have several (production code) methods being tested from the same script and the transition from testing one function to testing another is subtle ( a reader has to carefully read the specific names on the lines of a given test - when sweeping the eyes down the file - that subtle transition is easily missed ) 
Because I'm lazy - I would rather know just from which test file I have open 
that all the tests are about one ( production code ) method without having to scrutinise the details

CoePaul (changes requested):

oliwenmandiamond (dismissed):
The standard for dodal for tests is "mirror the file structure of the `src` folder with the corresponding code files. Use the test_*.py naming convention for test files."

You have created many test files that don't mirror the src structure e.g `test_photon_mass_attenutation.py`, `test_thickness_required_to_attenuate.py` etc.

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,82 @@
...
+    return photon_absorption_factor_per_unit_length * (
+        energy_kev**energy_dependence_exponent
+    )
+
+
+@validate_call()
+def attenuation_at_depth_cm(
+    depth_cm: NonNegativeFloat, absorption_coefficient_per_cm: NonNegativeFloat
+):
+    # TODO: check this - not sure if it really is the depth of the wedge?
```
CoePaul: Re the TODO: It is the depth of anything absorbing.
This calculation does not know what absorbing entity it is being used for 
All it knows is - we have d cm of material thickness 
( or from the photons persepective - material depth )

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -0,0 +1,82 @@
...
+    """Calculates attenuation in Barnett units.
```
CoePaul: Since this is the only place where the definition is applied 
and since nobody in the world has yet heard of these units 

change request: 
Calculates attenuation in Barnett units. ->

Calculates attenuation in natural log Barnett units,
where 1000 Bn equivalent to 1/e, 0 Bn to 1 and 2000 Bn to 1/(e^2).

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+)
+def test_attenuation_at_depth_cm(depth_cm, absorption_coefficient_per_cm, result):
+    assert attenuation_at_depth_cm(
+        depth_cm, absorption_coefficient_per_cm
+    ) == pytest.approx(result, rel=1e-6)
+
+
+# inauspicious path
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_photon_mass_attenuation_per_unit_length_errors_energy(bad_input):
```
CoePaul: understand the desire for shorthand method name but 
might be clearer if something like 
test_photon_mass_attenuation_per_length_error_with_invalid_energy

test_photon_mass_attenuation_per_length_error_with_invalid_factor

test_photon_mass_attenuation_per_length_error_with_invalid_exponent

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+        photon_mass_attenuation_per_unit_length(1.0, bad_input, 1.0)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_photon_mass_attenuation_per_unit_length_errors_expo(bad_input):
+    with pytest.raises(TypeError):
+        photon_mass_attenuation_per_unit_length(1.0, 1.0, bad_input)
+
+
+def test_thickness_cm_required_to_attenuate_too_small_coefficient():
```
CoePaul: test_thickness_cm_required_to_attenuate_transparent_medium
would be a more useful method name 
( otherwise You have to think what small coefficient corresponds to )

also inside the method You could 
( on line between 95 and 96 ) define
 transparent_medium = 1.0e-15
and use transparent_medium as the 2nd arg in the call to thickness..

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+    with pytest.raises(ValueError):
+        thickness_cm_required_to_attenuate(1, 1e-15)
```
CoePaul: for these unhappy path cases the 1st arg ( 1, )
would be more realistic if it were a common place request 
1 Bn = 99.9005 % transmission - an unlikely user request
numbers between 
around 300 (~74%) and 5300 (~0.5%) are more realistic

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def test_thickness_required_to_attenuate_raises_negative_error():
```
CoePaul: method name unclear w.r.t. underlying physics
maybe 
test_thickness_required_to_attenuate_raises_error_for_amplification

CoePaul: or _gain if that's clearer than _amplification
or something similar

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+    with pytest.raises(ValueError):
+        thickness_cm_required_to_attenuate(-1, 1)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_thickness_cm_required_to_attenuate_target_raises_error(bad_input):
```
CoePaul: suggest method name 
test_thickness_required_to_attenuate_raises_error_with_invalid_target_attenuation

the cm can be dropped for brevity - since it doesn't get as far as building a quantity ( and the future is unlikely to have parallel methods for other lengths like mm etc )

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+    with pytest.raises(ValueError):
+        thickness_cm_required_to_attenuate(bad_input, 1.0)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_thickness_cm_required_to_attenuate_absorption_raises_error(bad_input):
```
CoePaul: suggested method name
test_thickness_required_to_attenuate_raises_error_with_invalid_absorption

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+    with pytest.raises(ValueError):
+        thickness_cm_required_to_attenuate(1.0, bad_input)
+
+
+@pytest.mark.parametrize("bad_input", [-1, -5, -0.1])
+def test_attenuation_at_depth_cm_absorption_coefficient_error(bad_input):
```
CoePaul: suggested rename of method:
test_attenuation_at_depth_raises_error_with_invalid_absorption
or 
test_attenuation_at_depth_raises_error_with_invalid_absorption_coefficient

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+    with pytest.raises(ValueError):
+        attenuation_at_depth_cm(1.0, bad_input)
+
+
+@pytest.mark.parametrize("bad_input", [-1, -5, -0.1])
+def test_attenuation_at_depth_cm_depth_raises_negative_error(bad_input):
```
CoePaul: suggested method name 
test_attenuation_at_depth_raises_error_for_unphysical_depth

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+    with pytest.raises(ValueError):
+        attenuation_at_depth_cm(bad_input, 1.0)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_attenuation_at_depth_cm_depth_depth_raises_error(bad_input):
```
CoePaul: test_attenuation_at_depth_raises_error_with_invalid_depth

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -0,0 +1,137 @@
...
+    with pytest.raises(ValueError):
+        attenuation_at_depth_cm(bad_input, 1.0)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
+def test_attenuation_at_depth_cm_depth_absorption_raises_error(bad_input):
```
CoePaul: similar rename requested - I think by now You have the idea

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+    [
+        (0, 1),  # tests transparency from zero attenuation (canonical)
+        (
+            1e3,
+            0.3678794,
+        ),  # tests transmission from canonical attenuation is  1/e (canonical)
+        (
+            145.1,
+            0.8649358,
+        ),  # tests transmission from arbitrary weak attenuation is -1000*attenuation
```
CoePaul: the end of this comment ( after words "attenuation is" ) don't make sense 
since 145.1 * -1000 is not 0.865 ish

"...weak attenuation" would be sufficient

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+        (
+            7221.9,
+            7.3041331435179e-4,
+        ),  # tests transmission from arbitrary strong attenuation is -1000*attenuation
```
CoePaul: same comment as per weak attenuation above - end of sentence is misleading

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+        (0, 0),  # tests natural log of transparency from attenuation (canonical)
+        (
+            1e3,
+            -1,
+        ),  # tests natural log of transmission from canonical transmission is negative
+        # unity (canonical)
+        (
+            712.6,
+            -0.7126,
+        ),  # tests attenuation from natural log arbitrary high transmission is -1000 *
```
CoePaul: Having these details is useful - but not here - where no reader will seek them out - ( the reader here just wants the gist of what's going on )
So drop words after "transmission"

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+        # attenuation
+        (
+            8034.1,
+            -8.0341,
+        ),  # tests attenuation from natural log arbitrary low transmission is -1000 *
```
CoePaul: drop words after "low transmission"

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+    assert natural_log_of_transmission_from_attenuation(
+        attenuation_bn
+    ) == pytest.approx(result)
+
+
+@pytest.mark.parametrize(
+    "transmission_as_fraction,result",
+    [
+        (1, 0),  # tests attenuation from transparency is zero (canonical)
+        (0.37, 994.25227334),  # tests attenuation from 37% is close to 1000 (canonical)
```
CoePaul: I would drop canonical here ( line 66 ) as that strictly only applies to 0 Bn and 1000 Bn 
( in the same way that degrees C has only two defining points )

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+        (
+            0.871,
+            138.1133,
+        ),  # tests attenuation from arbitrary high transmission has expected value
```
CoePaul: drop words after "transmission" - the "has expected value" is implicit for any happy path test

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+        (
+            4.2e-4,
+            7775.2558,
+        ),  # tests attenuation from arbitrary high attenuation has expected value
```
CoePaul: similar to above drop "has expected value"

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+
+@pytest.mark.parametrize(
+    "ln_t,result",
+    [
+        (-1, 1000),  # tests negative unity log of transmission is 1000 (canonical)
+        (0, 0),  # tests natural log of transparency is zero (canonical)
+        (
+            -0.4367,
+            436.7,
+        ),  # tests log from arbitrary high transmission is -1000 * log
```
CoePaul: I would drop words after "transmission"

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+        (
+            -5.9017,
+            5901.7,
+        ),  # tests log from arbitrary low transmission is -1000 * log
```
CoePaul: I would drop words after "transmission"

tests/common/general_maths/test_transmission_inerconversion.py
```diff
@@ -0,0 +1,124 @@
...
+    ],
+)
+def test_attenuation_from_natural_log_of_transmission(ln_t, result):
+    assert attenuation_from_natural_log_of_transmission(ln_t) == pytest.approx(result)
+
+
+# inauspicious:
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object()])
```
CoePaul: I've just realised we should probably add in a boolean value to all these standard "bad input arrays"

CoePaul (changes requested):

Matt-Carre:
Currently, what I have works, but I could create a second model to check if the absorption coefficient is greater than 1.0e-14 with pydantic as well. Worth doing? Or do we want it to raise that specific value error?

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -1,4 +1,20 @@
-from pydantic import StrictFloat, validate_call
+from typing import Annotated
```
CoePaul, line 1: I think this script is mixing two ( or more ) colliding metaphors regarding use of 
pydantic, decorators, BaseModel and field validation 

You can remove the need for a class completely 

something like 
from typing import Annotated
from pydantic import Field, StrictFloat, validate_call

@validate_call
def func(
    arg1: Annotated(StrictFloat, Field(ge=0)]
    ) -> float: 
       return arg1 ** 2

should be enough to do the validation

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -11,12 +27,68 @@ def photon_mass_attenuation_per_unit_length(
 
     Args:
         energy_kev (StrictFloat): energy
-        photon_absorption_factor_per_unit_length (StrictFloat): photon absorption factor per
-        unit length
+        photon_absorption_factor_per_unit_length (StrictFloat): photon absorption factor
+        per unit length
         energy_dependence_exponent (StrictFloat): energy dependence exponent
```
CoePaul: once the validation is being done "properly" the energy dependence exponent needs to be checked to be negative

CoePaul: similarly the absorption factor should be zero or positive and never negative

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -11,12 +27,68 @@ def photon_mass_attenuation_per_unit_length(
...
+
+    Raises:
+        ValueError: if attenuation is below zero, or absorption is below the minimum
+        meaningful absorption coefficient, a value error is raised
+
+    Returns:
+        (float): material depth in cm.
+    """
+    minimum_meaningful_absorption_coefficient = 1.0e-14
+    if absorption_coefficient_per_cm < minimum_meaningful_absorption_coefficient:
```
CoePaul: there's a way to compactify this 
the gt notation for Field(gt=1.0e-14) should achieve the same effect

CoePaul: however I can see You were ( in the spirit of the original Jython code ) keeping the custom error message ( about transparent media ) 

in that case something like 

from pydantic_core import PydanticCustomError

def check_negligibility(v: float) -> float:
    # We let Pydantic handle the math check, but catch its error to rename it
    if v <= 1.0e-14:
        raise PydanticCustomError(
            'negligible',  # This is your internal error code/type
            ' < "explain about no optical gain nor transparent media" > .' # Your custom message
        )
    return v

< python comment hash > Now use it in your Annotated definition and NonNegligible is now a type
NonNegligible = Annotated[StrictFloat, AfterValidator(check_negligibility)]

def func( arg1: NonNegligible, arg2: NonNegligible ) -> float: etc

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -27,7 +29,48 @@ def test_photon_mass_attenuation_per_unit_length(
 ):
     assert photon_mass_attenuation_per_unit_length(
         energy_kev, photon_absorption_factor_per_unit_length, energy_dependence_exponent
-    ) == pytest.approx(result, 5.0e-8)
+    ) == pytest.approx(result)
+
+
+@pytest.mark.parametrize(
+    "target_attenuation_bn,absorption_coefficient_per_cm,result",
```
CoePaul: a space after the commas might be nice - if ruff and similar doesn't remove it again 
and I would rename "result" -> "required_cm" or similar

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -27,7 +29,48 @@ def test_photon_mass_attenuation_per_unit_length(
...
+    [
+        (0, 2.4, 0),  # tests attenuator thickness required for transparency is zero
+        (
+            248.461,
+            2.13,
+            0.1166483568,
+        ),  # tests attenuator thickness frequired for arbitrary attenuation
```
CoePaul: frequired is a typo

tests/common/general_maths/test_material_absorption_maths.py
```diff
@@ -49,3 +92,54 @@ def test_photon_mass_attenuation_per_unit_length_errors_with_invalid_exponent(
...
+        attenuation_at_depth_cm(1.0, bad_input)
+
+
+@pytest.mark.parametrize("bad_input", [-1, -5, -0.1])
+def test_attenuation_at_depth_raises_error_for_unphysical_depths(bad_input):
+    with pytest.raises(ValidationError):
+        attenuation_at_depth_cm(bad_input, 1.0)
+
+
+@pytest.mark.parametrize("bad_input", ["a", [], None, math.sin, object(), False])
```
CoePaul: I would change the boolean on this test and the next to True 
( as False = 0 might be rejected for the incorrect reason and this test would not catch that )

CoePaul (dismissed):

noemifrisina (approved):
lgtm, comment have been addressed, thanks

PR #2052, opened 2026-05-07
Fix misc eiger arming issues

rtuck99:
Fixes
* DiamondLightSource/mx-bluesky#1715
* DiamondLightSource/mx-bluesky#1689

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -5,7 +5,7 @@ from bluesky.protocols import Stageable
 from ophyd import Component, Device, EpicsSignalRO, Signal
 from ophyd.areadetector.cam import EigerDetectorCam
 from ophyd.signal import AttributeSignal
-from ophyd.status import AndStatus, Status, StatusBase, SubscriptionStatus
+from ophyd.status import Status, StatusBase, SubscriptionStatus
 
 from dodal.devices.detector import DetectorParams, TriggerMode
 from dodal.devices.eiger_odin import EigerOdin
@@ -248,7 +248,7 @@ class EigerDetector(Device, Stageable):
         )
         return status
 
-    def set_cam_pvs(self) -> AndStatus:
+    def set_cam_pvs(self) -> StatusBase:
         LOGGER.info("Eiger arming: Setting eiger camera PVs...")
         assert self.detector_params is not None
         self.cam.acquire_time.set(
--- a/src/dodal/devices/eiger.py
+++ b/src/dodal/devices/eiger.py
@@ -24,6 +24,7 @@ class EigerTimeouts:
     all_frames_timeout: int = 120
     arming_timeout: int = 60
     odin_stop_timeout: int = 30
+    odin_meta_active_timeout: int = 60
 
 
 class InternalEigerTriggerMode(Enum):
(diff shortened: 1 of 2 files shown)


src/dodal/devices/eiger.py
```diff
@@ -225,18 +225,27 @@ def change_roi_mode(self, enable: bool) -> StatusBase:
             if enable
             else self.detector_params.detector_size_constants.det_size_pixels
         )
+        LOGGER.info(f"Setting cam.roi_mode to {1 if enable else 0}")
         status = self.cam.roi_mode.set(
             1 if enable else 0, timeout=self.timeouts.general_status_timeout
         )
+        status.wait(timeout=self.timeouts.general_status_timeout)
```
DominicOram: Could: If we're going to wait on each individual status then we don't need to keep the status around, e.g.

```suggestion
self.cam.roi_mode.set(1 if enable else 0, timeout=self.timeouts.general_status_timeout).wait(timeout=self.timeouts.general_status_timeout)
```

src/dodal/devices/eiger.py
```diff
@@ -225,18 +225,27 @@ def change_roi_mode(self, enable: bool) -> StatusBase:
...
+        LOGGER.debug(f"Setting image_height to {detector_dimensions.height}")
         status &= self.odin.file_writer.image_height.set(
```
DominicOram: Should: The last status has already been waited on so no need to do `&=` here, `=` will do (or just don't keep it around as above)

src/dodal/devices/eiger.py
```diff
@@ -249,21 +258,30 @@ def set_cam_pvs(self) -> AndStatus:
...
             timeout=self.timeouts.general_status_timeout,
         )
+        status.wait(timeout=self.timeouts.general_status_timeout)
+        LOGGER.debug("Setting cam trigger_mode...")
         status &= self.cam.trigger_mode.set(
             InternalEigerTriggerMode.EXTERNAL_SERIES.value,
             timeout=self.timeouts.general_status_timeout,
         )
+        status.wait(timeout=self.timeouts.general_status_timeout)
         return status
```
DominicOram: Should: The only reason this returns the status is so that we can wait on it, if we've already waited on it then it seems a bit pointless

src/dodal/devices/eiger.py
```diff
@@ -379,7 +397,7 @@ def set_num_triggers_and_captures(self) -> StatusBase:
 
     def _wait_for_odin_status(self) -> StatusBase:
         self.forward_bit_depth_to_filewriter()
-        await_value(self.odin.meta.active, 1).wait(self.timeouts.general_status_timeout)
+        await_value(self.odin.meta.active, 1).wait(60)
```
DominicOram: Should: Can we pull this into a const?

tests/devices/test_eiger.py
```diff
@@ -355,7 +354,7 @@ def test_stage_runs_successfully(
     fake_eiger.stage()
     fake_eiger.arming_status.wait(1)  # This should complete long before 1s
     # One log message kicking off arming, then one for each of the 13 arming stages
-    assert len(caplog.messages) == 14
+    assert len(caplog.messages) == 18
```
DominicOram, line 340: Should: This no longer matches the comment above

DominicOram (changes requested):
Couple of suggestions. Now we're less parallel this presumably will affect the time it takes to arm/disarm the detector. Have you investigated how much it changes throughput and agreed that the stability is preferred?

rtuck99:
> Couple of suggestions. Now we're less parallel this presumably will affect the time it takes to arm/disarm the detector. Have you investigated how much it changes throughput and agreed that the stability is preferred?

Yes, we ran it with this hotfixed, it doesn't slow things down that much. I don't have specific numbers but we think the delay is probably of the order of ~1s, judging from the archiver logs for signals. however most of that is while waiting for the robot load.
The incidence of this issue seems to have become much more frequent to the extent that without this fix, hyperion (at least on the newer versions) is almost unusable.

src/dodal/devices/eiger.py
fajinyuan: Sorry for my question if it misses the point here. Is it possibl to improve IOC PV so the set call can rely on callback instead of wait on status, or the status already embedded callback inside?

DominicOram: Yes, the status is the callback from the IOC

DominicOram:
>The incidence of this issue seems to have become much more frequent to the extent that without this fix, hyperion (at least on the newer versions) is almost unusable.

Fair enough, it goes away with the fastCS Eiger anyway. Other comments still stand I think

src/dodal/devices/eiger.py
rtuck99: this all goes into the list of  run_functions_without_blocking() so to follow convention we return the status.

DominicOram (approved):

PR #2053, opened 2026-05-11
Added i15-1 zebra shutter

### LINKED ISSUE #10: i15-1: Add implementation for fast shutter
The GDA code for the fast shutter lives at /dls_sw/i15-1/scripts/Xpdf/Components/fs.py. We should add an implementation in dodal. It looks like the fast shutter is triggered/read through the zebra, the same is true in MX (https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/zebra/zebra_controlled_shutter.py) but in their case they have a set of PVs (and EDM screen) on top that makes the logic a bit smoother:

Image

It may be worth trying to copy this but we would need controls effort.

## Acceptance Criteria
- There is a way to open/close i15-1's fast shutter in dodal

DominicOram:
Fixes https://github.com/DiamondLightSource/crystallography-bluesky/issues/10

Note, also needs https://github.com/DiamondLightSource/mx-bluesky/pull/1742

1. Confirm tests pass

### Checks for reviewer


jacob-williamson (approved):
Looks good

PR #2056, opened 2026-05-13
Add Mbs driver and region classes

### LINKED ISSUE #2068: Add Mbs driver and region classes
Add implementations of the Mbs driver and region classes using the abstract electron analyser classes. Test dodal connect works with i05 and i05-1.

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/2068. This implements the `MbsRegion` and `MbsAnalyserDriverIO` using the base electron analyser classes. This has updated the testing framework to be much more flexible, allowing for sequence files to be loaded directly to given regions to tests and paired with other fixtures (e.g associated detector and driver) without needing the files to have same number and named regions. 

Note: I have removed the image and spectrum pv's from base (for now). This is because Mbs does not use the spectrum like Specs and VGScienta does. I also don't think they are needed from speaking with controls and understanding how the HDF file writer plugin works and will be next bit of work to do once this is complete.

1. Check tests pass
2. Check logic makes sense.

### Checks for reviewer


src/dodal/devices/electron_analyser/base/base_driver_io.py
```diff
@@ -29,10 +30,27 @@
     TPassEnergy,
 )
 
-AnyPsuMode: TypeAlias = SupersetEnum | StrictEnum
+AnyPsuMode: TypeAlias = SupersetEnum | StrictEnum | str
 TPsuMode = TypeVar("TPsuMode", bound=AnyPsuMode)
 
-_PSU = "PSU_MODE"
+
+@dataclass
```
Villtord: Since it's going to be a "true" class variable with potentially no overwriting in each child class - although it's unlikely that we will have both, e.g. specs_driver and mbs_driver in one python environment -  but for the sake of clarity this is still a mutable object. That means once changed on specs_driver will also change on mbs_driver. I think you can mark this class as 
```
@dataclass(frozen=True)
class ElectronAnalyserPVConfig:
```
that will prevent any mutations on it

Villtord (commented):
I started to review but this is a big change - I got only partially through it.

src/dodal/devices/electron_analyser/base/base_driver_io.py
oliwenmandiamond: yes good idea! I've add this. I'm hoping to remove it in future, it's just a temporarily measure so we can still have working devices without needing to adjust the `__init__` method

src/dodal/common/data_util.py
```diff
@@ -60,15 +64,40 @@ def update_config_from_file(self, new_file: str) -> None:
         self.default_path, self.default_file = split(new_file)
 
 
-def json_model_loader(
-    model: type[TBaseModel], config: JsonLoaderConfig | None = None
-) -> JsonModelLoader[TBaseModel]:
-    """Factory to create a function that loads a json file into a configured pydantic
-    model and with optional configuration for default path and file to use.
+class ModelLoader(Generic[TBaseModel]):
```
Villtord, line 67: I'm not sure how this change in model loader names and signatures is related to MBS - but should be fine I guess if tests pass, probably better to separate it next time.

src/dodal/beamlines/i05_1.py
```diff
@@ -35,3 +36,12 @@ def nano_shutter() -> HutchShutter:
 def sm() -> XYZAzimuthPolarDefocusStage:
     """Sample Manipulator."""
     return XYZAzimuthPolarDefocusStage(prefix=f"{PREFIX.beamline_prefix}-EA-SM-01:")
+
+
+@devices.factory
+def analyser_driver() -> MbsAnalyserDriverIO[LensMode, PassEnergy]:
+    return MbsAnalyserDriverIO[LensMode, PassEnergy](
+        prefix=f"{PREFIX.beamline_prefix}-EA-DET-04:CAM:",
```
Villtord, line 44: nit: I would go further and move `CAM:` to down to signals PVs and leave prefix as purely analyser base PV so you can reuse it in other epics plugins
`
prefix=f"{PREFIX.beamline_prefix}-EA-DET-04:"
 `
but that's optional.

Villtord (approved):
I think it looks fine, thanks.

src/dodal/common/data_util.py
oliwenmandiamond: I had to update model loader to be more flexible so it can work with xml files and json files. Yeah sorry for the size, hard to avoid. Why I moved detector to another change

PR #2059, opened 2026-05-15
Conversion for zoom and microns per pixel

adaudon:
In oav_parameters.py, there was an error when OAVConfigBeamCenter would run get_parameters() whenever the user clicked on the Move on Click button.
The um_xy and bc_xy had different types for their keys.
um_xy was giving integer strings while bc_xy was giving float strings.

So a change was made to convert um_xy to float strings which allowed the code to work as the keys now matched.

The issue might be related to the move to the daq config server?


jacob-williamson:
I think the type issues should also be fixed by this https://github.com/DiamondLightSource/daq-config-server/pull/174, which means the zoom config will have it's own pydantic model with specified types.

jacob-williamson (approved):

adaudon:
Once daq-config-server has that PR pushed, can remove this if needed to help prevent unintentional bugs. Not sure when it'll be pushed though so merging change for now.

PR #2060, opened 2026-05-18
Adds 'circular' tests for general_maths calculations

### LINKED ISSUE #2042: Extend beamline-relevant arithmetic with "Circularity" tests of conversions
a) Much of the arithmetic needed for attenuation related caculations involves conversions 
b) these conversion function usual appear in bi-directional pairings

If f(x) is the opposite function of g(x) 
then g( f(x) ) should equal x 
and likewise f( g(x) ) should equal x 

testing this is not a unit-test level check ( as it involves compound actions outside the scope of lowest level UNIT testing ) - but it is still a documentable, specification that the code should adhere to 

thus: Write a general integration test thing to check circularity of a pair of functions 

ie make the test code so that I can pass in as software arguments 
a) a numerical value x suitably within the valid range of the relelevant functions 
b) a conversion function f
c) the opposite conversion function g

Write a test suite which :
for each conversion-function-pairing created in parallel issues ( off this ticket's main parent )
creates a pair of tests
1) passes in the x, f, g 
2) passes in the x, g, f 

## Acceptance Criteria

Matt-Carre:
Fixes #2042 

This adds tests that check that f(g(x)) == g(f(x)), where f and g are inverses of each other (e.g., f=converting cm to mm, g=converting mm to cm), and x is an input value. 

Also adds in Bools when not included.


Matt-Carre:
This will be expanded once #2049 is merged into main.

tests/common/general_maths/test_arithmetic_conversions.py
```diff
@@ -56,6 +56,27 @@ def test_conversion_from_microns_to_centimetres(input, result):
...
+
+
+@pytest.mark.parametrize("input", [1.0, 10.0, 100.0])
+def test_circular_microns_to_mm_and_back(input):
+    assert convert_microns_to_mm(convert_mm_to_microns(input)) == input
+    assert convert_mm_to_microns(convert_microns_to_mm(input)) == input
+
+
+@pytest.mark.parametrize("input", [1.0, 10.0, 100.0])
+def test_circular_percentage_to_factor(input):
```
Villtord:
```suggestion
def test_circular_percentage_to_factor_and_back(input):
```

tests/common/general_maths/test_arithmetic_conversions.py
```diff
@@ -56,6 +56,27 @@ def test_conversion_from_microns_to_centimetres(input, result):
     assert convert_microns_to_cm(input) == pytest.approx(result)
 
 
+# Circular tests (all numbers here arbitrary)
+
+
+@pytest.mark.parametrize("input", [1.0, 10.0, 100.0])
+def test_circular_cm_to_mm_and_back(input):
+    assert convert_cm_to_mm(convert_mm_to_cm(input)) == input
+    assert convert_mm_to_cm(convert_cm_to_mm(input)) == input
```
Villtord: I would still use pytest.approx() with those float inputs in all tests added:
   `assert convert_cm_to_mm(convert_mm_to_cm(input)) == pytest.approx(input)
`

tests/common/general_maths/test_arithmetic_conversions.py
```diff
@@ -56,6 +56,27 @@ def test_conversion_from_microns_to_centimetres(input, result):
     assert convert_microns_to_cm(input) == pytest.approx(result)
 
 
+# Circular tests (all numbers here arbitrary)
+
+
+@pytest.mark.parametrize("input", [1.0, 10.0, 100.0])
```
Villtord: It's probably also good to add "0.0" here as an edge case.

oliwenmandiamond:
Could we also update the PR title. "Circular tests for two branches present" doesn't mean anything relevant

tests/common/general_maths/test_transmission_interconversion.py
```diff
@@ -98,6 +98,35 @@ def test_attenuation_from_natural_log_of_transmission(ln_t, result):
     assert attenuation_from_natural_log_of_transmission(ln_t) == pytest.approx(result)
 
 
+# Circular tests (all numbers here arbitrary)
+
+
+@pytest.mark.parametrize("input", [1.0, 10.0, 100.0])
+def test_circular_attenuation_from_log_and_back(input):
```
Villtord, line 105: Could you also use` pytest.approx()` here as well please?

Matt-Carre: Yep sorry! Just missed that one! Fixed now

Villtord (approved):
Thanks, looks good to me!

PR #2066, opened 2026-05-20
Adds DCM in optics, and read-only DCM in i19_1, i19_2

### LINKED ISSUE #2064: Add a dcm to i19
- Instantiate a DCM device in the i19-optics beamline definition. The dcm on I19 has one stationary and one PitchAndRollCrystal and should have dspacing as well.
- Create a read only dcm device in devices/beamlines/i19/access_controlled that has ro signals for energy and wavelength and instantiate it on i19-1 and i19-2. For the moment, do not make it an OpticsBlueAPIDevice, that will be done in a separate future PR.

## Acceptance Criteria
* There is a dcm device on i19-optics and it connects
* There is a read only dcm device on i19-1 and i19-2 and it connects

Matt-Carre:
closes #2064 

Adds a DCM. Not currently an optics blue api device, but does seem to connect

Makes wavelength and energy readable.


src/dodal/devices/beamlines/i19/access_controlled/dcm.py
```diff
@@ -0,0 +1,15 @@
+from ophyd_async.core import StandardReadable
+from ophyd_async.epics.core import epics_signal_r
+
+
+class DCM(StandardReadable):
```
noemifrisina: Should: please rename this to something that indicates this is just a read only version of the device - same goes for the name in the beamline definitions files

src/dodal/devices/beamlines/i19/access_controlled/read_only_dcm.py
```diff
@@ -0,0 +1,15 @@
+from ophyd_async.core import StandardReadable
+from ophyd_async.epics.core import epics_signal_r
+
+
+class DCM(StandardReadable):
+    def __init__(
```
noemifrisina, line 6: Should: can we have a small test that the values from the signals can be read please?

src/dodal/devices/beamlines/i19/access_controlled/dcm.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(
+        self,
+        prefix: str,
+        name: str = "",
+    ) -> None:
+        with self.add_children_as_readables():
+            self.energy_in_eV = epics_signal_r(float, f"{prefix}Energy")
```
noemifrisina: Must: The PV names are all caps. If you run dodal connect now on i19-1 and i19-2 it fails on the dcm because these two PVs do not exist...

src/dodal/beamlines/i19_optics.py
```diff
@@ -24,6 +29,15 @@
 devices = DeviceManager()
 
 
+@devices.factory()
+def dcm() -> DoubleCrystalMonochromatorWithDSpacing:
+    return DoubleCrystalMonochromatorWithDSpacing(
+        prefix=f"{PREFIX.beamline_prefix}-MO-DCM-01:",
+        xtal_1=PitchAndRollCrystal,
```
noemifrisina: Must: the crystals are swapped here, the Stationary should be the first one and the PitchAndRoll the second

noemifrisina (changes requested):
Thanks, a few things

tests/devices/beamlines/i19/access_controlled/test_read_only_dcm.py
```diff
@@ -0,0 +1,18 @@
...
+@pytest.fixture
+async def mock_dcm() -> ReadOnlyDCM:
+    mock_dcm = ReadOnlyDCM(prefix="FOO-MO")
+    await mock_dcm.connect(mock=True)
+    set_mock_value(mock_dcm.energy_in_eV, 1)
+    set_mock_value(mock_dcm.wavelength_in_a, 100)
+    return mock_dcm
+
+
+async def test_reading(mock_dcm: ReadOnlyDCM):
```
noemifrisina: Should:for a test like this, it would be better to use `assert_reading`, see https://diamondlightsource.github.io/dodal/main/how-to/write-tests.html

noemifrisina (changes requested):
Thanks, one more thing

noemifrisina (approved):

PR #2067, opened 2026-05-20
Adds config server to i19_2

### LINKED ISSUE #160: Add the config server to i19-2 and use it to read the detector LUT
Now that eh2 `daq_configuration` has been added to the daq-config-server whitelist, we should be able to use it to read the detector lookup table and find the beam centre. This will also need a first change to be made in the beamline definition in dodal.

1. Add the config_client to the i19-2 beamline definition in dodal
2. Add a small function that reads the file and returns the beam centre values for a given distance

The eiger LUT can be found in `/dls_sw/i19-2/software/daq_configuration/lookup`. Because these files have not yet been added to the converters in the config server, we'll need to use the `force_parcer` option to be able to get the data in a readable format:

```python
config_client.get_file_contents(bc_lut, DetectorXYLookupTable, force_parser=DetectorXYLookupTable.from_contents)
```

## Acceptance Criteria
- We can use the config server to read the beam centre values out of the lookup table

Matt-Carre:
Fixes https://github.com/DiamondLightSource/i19-bluesky/issues/160

Adds config server to i19_2


jacob-williamson (approved):
Looks good to me!

PR #2071, opened 2026-05-22
Add MbsDetector, remove unneeded EnergySource logic

### LINKED ISSUE #2070: Add Mbs detector
Add final device for Mbs which is the detector. It should use the Mbs driver and implement the region_logic, arm_logic and trigger_logic (data_logic will be done in another issue). This should also be added to the i05 and i05-1 beamlines.

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/2070

1. Check tests pass
2. Implementation makes sense

### Checks for reviewer


src/dodal/devices/electron_analyser/mbs/mbs_detector.py
```diff
@@ -0,0 +1,86 @@
...
+        shutter: GenericFastShutter | None = None,
+        source_selector: SourceSelector | None = None,
+        name: str = "",
+    ):
+        # Make attribute of class so connect applies to driver and populates parent.
+        self.driver = MbsAnalyserDriverIO[TLensMode, TPassEnergy](
+            prefix, lens_mode_type, pass_energy_type
+        )
+        region_logic = RegionLogic(self.driver, energy_source, source_selector)
+        self.close_shutter_idle = soft_signal_rw(bool, initial_value=True)
```
Villtord: I don't really get the purpose of this soft signal from variable name, would it be possible to make a comment or rename it? why is it necessary to have close shutter idle/not idle info??

Villtord (changes requested):
Thanks, looks good, just a minor comment

src/dodal/devices/electron_analyser/mbs/mbs_detector.py
oliwenmandiamond: The signal is needed so that we can control if shutters are closed between each region. This is how I've done it for Specs and VGScienta too. If this is not needed for Mbs, I can remove this signal. If something more fundamental, we should do this in another issue.

src/dodal/devices/electron_analyser/base/__init__.py
```diff
@@ -16,7 +16,7 @@
     TLensMode,
 )
 from .base_util import to_binding_energy, to_kinetic_energy
-from .energy_sources import AbstractEnergySource, DualEnergySource, EnergySource
+from .energy_sources import DualEnergySource
```
Villtord, line 19: I forgot if we agreed on using absolute imports as a standard or not?

Villtord (approved):

src/dodal/devices/electron_analyser/base/__init__.py
oliwenmandiamond: No, init has relative

PR #2072, opened 2026-05-26
Extend monochrometer to get energy from y position

jacob-williamson:
Fixes https://github.com/DiamondLightSource/crystallography-bluesky/issues/47

Requires https://github.com/DiamondLightSource/daq-config-server/pull/183

1. Check tests pass

### Checks for reviewer


src/dodal/devices/beamlines/i15_1/laue.py
```diff
@@ -0,0 +1,37 @@
...
+            self.config_client = config_client
+            self.crystal_lut_path = crystal_lut_path
+
+        super().__init__(name)
+
+    def _get_xtal_config(self) -> XpdfCrystalLookupTable:
+        return self.config_client.get_file_contents(
+            self.crystal_lut_path,
+            XpdfCrystalLookupTable,
+            # Remove once new config server is released + deployed
```
DominicOram: Should: Can you link to which PR the release needs for this

src/dodal/devices/beamlines/i15_1/laue.py
```diff
@@ -0,0 +1,37 @@
...
+        crystal_lut_path: str,
+        name: str = "",
+    ):
+        with self.add_children_as_readables():
+            self.bend = Motor(prefix + "BENDER")
+            self.bragg = Motor(prefix + "PITCH")
+            self.roll = Motor(prefix + "ROLL")
+            self.yaw = Motor(prefix + "YAW")
+            self.y = Motor(prefix + "Y")
+            self.energy_kev = derived_signal_r(self._get_energy, y=self.y)
```
DominicOram, line 23: Should: I think this needs to be a hinted signal with the others being config

src/dodal/devices/beamlines/i15_1/laue.py
```diff
@@ -0,0 +1,37 @@
...
+            self.config_client = config_client
+            self.crystal_lut_path = crystal_lut_path
```
DominicOram: Should: These aren't signals so don't need to be added as readables

tests/devices/beamlines/i15_1/test_laue_monochrometer.py
```diff
@@ -0,0 +1,45 @@
...
+        (50.6, 76.69),
+        (-48.845, 65.4),
+        (-50, 65.4),
+    ],
+)
+async def test_energy_kev_gets_value_based_on_y(
+    laue_monochrometer: LaueMonochrometer, y, expected_energy
+):
+    await laue_monochrometer.y.set(y)
+    assert await laue_monochrometer.energy_kev.get_value() == expected_energy
```
DominicOram, line 46: Could: Can you test that `read` in here also gives the energy? Should be able to use `assert_reading`

src/dodal/devices/beamlines/i15_1/laue.py
DominicOram: Also, they should probably be private

DominicOram (changes requested):
Great, thanks. Could you pin to the correct version of the daq config server so that the tests pass, with a comment saying to remove the pin once there's a release?

src/dodal/devices/beamlines/i15_1/laue.py
jacob-williamson: All the other signals being config? Or `_config_client` and `_crystal_lut_path` (which are now not signals)

DominicOram: I think what you've got is correct. Basically the one signal that is the most scientifically relevant should be hinted, everything else config

DominicOram (approved):
Great, thank you!

PR #2073, opened 2026-05-27
Add scaler card device for i09

oliwenmandiamond:
Add i09 scaler card devices for i09. GDA equivalent devices are `EpicsScalerSimple`.

1. Check tests pass
2. Check dodal connect

### Checks for reviewer


oliwenmandiamond:
Note: Unfortunately cannot check dodal connect at the moment due to devices being disconnected.

Relm-Arrowny:
I'm not 100% sure what the best course of action is, but it looks like the this is Struck Scaler we use around MMG. We may want to extract the [scaler card](https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/current_amplifiers/struck_scaler_counter.py. ) out of there while we are at it.

I worked on this a while back and have a few changes I like to add. Eventually, I want to create a StruckScaler class that includes all channels. Since triggering once captures all channels for "free," it makes sense to expose them. I have a specific use case where four detectors run on the same scaler card with current amplier, so we only want to trigger once per collection/autoscaling and collect everything.

P.S. If your scaler is connected to a current amplifier, you can just add the amplifier and use the scaler card there and it will handle the auto-scaling as well as count covertion back to voltage or even photon count.

oliwenmandiamond:
> I'm not 100% sure what the best course of action is, but it looks like the this is Struck Scaler we use around MMG. We may want to extract the [scaler card](https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/current_amplifiers/struck_scaler_counter.py.) out of there while we are at it.
> 
> I worked on this a while back and have a few changes I like to add. Eventually, I want to create a StruckScaler class that includes all channels. Since triggering once captures all channels for "free," it makes sense to expose them. I have a specific use case where four detectors run on the same scaler card with current amplier, so we only want to trigger once per collection/autoscaling and collect everything.
> 
> P.S. If your scaler is connected to a current amplifier, you can just add the amplifier and use the scaler card there and it will handle the auto-scaling as well as count covertion back to voltage or even photon count.

After speaking with beamline staff, it is fine to have a single scaler card device with the channels being just signals and all being read. I will simplify this tomorrow.

oliwenmandiamond:
@Relm-Arrowny I've redone this if you want to have another look

src/dodal/devices/beamlines/i09/scaler.py
```diff
@@ -0,0 +1,70 @@
...
+    StandardReadable,
+    Triggerable,
+    Movable,
+):
+    """Scaler controller that is triggerable. It will set the counting signal to True
+    and then waits for it to be False.
+    """
+
+    def __init__(self, prefix: str, name: str = ""):
+        self.counting = epics_signal_rw(bool, prefix + ".CNT")
```
Relm-Arrowny, line 47: I09 is power down so I cannot try, are you sure this is bool?
All I 10 scale are like this:
```python 
class CountState(StrictEnum):
    DONE = "Done"
    COUNT = "Count"
````

Comment (Relm-Arrowny) [reply src/dodal/devices/beamlines/i09/scaler.py:47]: an other i10 thing, do you want to check the mode before starting a scan, autocount mode may mess up the data?

```python

  @AsyncStatus.wrap
      await self.count_mode.set(CountMode.ONE_SHOT)

  @AsyncStatus.wrap
      await self.count_mode.set(CountMode.AUTO)

```

src/dodal/devices/beamlines/i09/scaler.py
```diff
@@ -0,0 +1,70 @@
...
+    @AsyncStatus.wrap
+    async def set(self, value: float):
+        await self.count_time.set(value)
+
+    @AsyncStatus.wrap
+    async def trigger(self):
+        acquire_status = await set_and_wait_for_value(
+            self.counting, True, wait_for_set_completion=True
+        )
+        await acquire_status
```
Relm-Arrowny: `wait_for_set_completion = True` mean it will return a completed status, making await acquire_status redundant. [see](https://github.com/bluesky/ophyd-async/blob/fab63a8a846acabfc1e2d3df4e29c7f57529c18f/src/ophyd_async/core/_signal.py#L638) 
```suggestion
        await set_and_wait_for_value(
            self.counting, True, wait_for_set_completion=True
        )
```

src/dodal/devices/detector/roi_stats_detector.py
Relm-Arrowny, line 1: Not sure why we have an empty file here.

tests/devices/beamlines/i09/test_scaler.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+async def test_scaler_controller_trigger_waits_for_counting_to_finish(
+    scaler1: ScalerController,
+) -> None:
+    start = time.monotonic()
+    await scaler1.trigger()
+    elapsed = time.monotonic() - start
+    assert elapsed >= 0.2
+    assert await scaler1.counting.get_value() is False
```
Relm-Arrowny, line 48: I think sleeping/waiting for a hardcoded 0.2 seconds makes the test slow and asking for it to be flaky.
Can we instead mock the underlying signal and not have the default DeviceMock behaviour of instantly setting the readback to the set?

Relm-Arrowny (changes requested):
A few things to check.

src/dodal/devices/beamlines/i09/scaler.py
oliwenmandiamond: So in GDA, it uses 0 and 1 for acquire. I've tried changing this to a bool, obviously can't test yet due to power being off

I've only ported the logic from GDA which as far as I can tell it doesn't do this

src/dodal/devices/beamlines/i09/scaler.py
oliwenmandiamond: Nice, have updated

tests/devices/beamlines/i09/test_scaler.py
oliwenmandiamond: So I've changed this to be 0.01 seconds. One of the first things I tested was if this would be flakey. I used the event_loop_fuzzing fixture and looped a 1000 times and it passed every one so it is not flakey.

Not sure what you mean, but I think the way I've done it accurately mocks the hardware. It acquires by setting a value to 1 (True) and then it has stopped acquiring when it becomes 0 (False) after a period of time (0.01 seconds)

src/dodal/devices/detector/roi_stats_detector.py
oliwenmandiamond: ahh,, that was something else I was experimenting with and accidentally added it to this. Removed!

tests/devices/beamlines/i09/test_scaler.py
Relm-Arrowny: I was thinking instead of waiting for a 0.2 seconds in _complete, we can make this test deterministic and much faster by manually controlling the `scaler1.counting` state

```python
status = scaler1.trigger() # this should make counting goes to Ture
await wait_for_pending_wakeups #not sure if we still need it.
assert await scaler1.counting.value() is True  #check it
assert status.done is False #it is blocking
mock_set_value(scaler1.counting,False) # unblock
await status
assert status.done
assert await scaler1.counting.get_value() is False
```

This eliminates the hardcoded sleep, makes the test run instantly, and effectively combines both tests since it validates both the blocking behaviour and the state transitions (False -> True -> False)."

oliwenmandiamond: This test is good, but I think misses the point of why I created the `MockScalerController`. The idea is that even if I connect my device in mock mode, it will still "work" as intended so it can work like GDA dummy mode. If I do a scan with this as the detector with mock=True, without the mock behaviour it will collect forever as nothing will set the counting back to False. The default mock behaviour does this, similar to how a MockMotor behaviour setups a motor correctly to be used in all other tests but also to just use it freely it use as a motor without connecting anything when using it as a user in a terminal to simulate how a user would use it.

Relm-Arrowny: o, you actually want it as a sim device, now that make sense and its useful, In which case 0.01s delay is probably fine for unit test if it does not fail but you may want longer delay for sim run, what if we keep the simulation behaviour, but allow the timing to be configurable? For example `MockScalerController` has self.delay?

Relm-Arrowny: something else cross my mind while thinking about this, [mock_puts_blocked](https://github.com/bluesky/ophyd-async/blob/fab63a8a846acabfc1e2d3df4e29c7f57529c18f/src/ophyd_async/core/_mock_signal_utils.py#L146) may work stopping it from setting counting to false. 
My be?
```python
with mock_puts_blocked(scaler1.counting):
    status = scaler1.trigger() 
    await wait_for_pending_wakeups #not sure if we still need it.
    assert status.done is False #it is blocking
# unblock
await status
assert status.done
assert await scaler1.counting.get_value() is False
```

oliwenmandiamond:
Some devices are still down, but as shown here scaler1 and 2 connect

```
$ dodal connect i09
Attempting connection to i09 (using dodal.beamlines.i09)
26 devices connected:
        dual_energy_source
        dual_fast_shutter
        fsi1
        fsj1
        ienergy
        ienergy_order
        iid
        iidenergy
        intensity_protection
        jenergy
        jgap
        jid
        jidcontroller
        jidenergy
        jphase
        jpolarisation
        lakeshore
        psi1
        psi2
        psj1
        psj2
        psk1
        scaler1
        scaler2
```

Relm-Arrowny (approved):

PR #2074, opened 2026-05-28
Mbs analyser slits

### LINKED ISSUE #2069: Add Mbs slit entrance device
Mbs analyser has extra information recorded for the device to do with the analyser slits e.g 
```
		completedRegion.setAnalyserSlitDirection(entranceSlitInformationProvider != null? entranceSlitInformationProvider.getDirection(): "vertical");
		completedRegion.setAnalyserSlitSetting(entranceSlitInformationProvider != null? (int) entranceSlitInformationProvider.getRawValue(): 0);
		completedRegion.setAnalyserSlitShape(entranceSlitInformationProvider != null? entranceSlitInformationProvider.getShape(): "unknown");
		completedRegion.setAnalyserSlitSize(entranceSlitInformationProvider != null? entranceSlitInformationProvider.getSizeInMM(): 0.0);
```
This should be implemented as it's own device and then passed to the analyser and then combine its configuration with this one (done by adding it as a child device).

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/2069

2. Check tests pass

### Checks for reviewer


src/dodal/devices/electron_analyser/mbs/mbs_analyser_slits.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class EntranceSlitInformation(BaseModel):
+    direction: str = "vertical"
+    setting: int = 100
+    size: float = 0.1
+    shape: str = "curved"
+
+    @classmethod
+    def from_slit_positions(cls, pos: SlitPosition):
```
Villtord: optional specify return class probably in quotes

src/dodal/devices/electron_analyser/mbs/mbs_analyser_slits.py
```diff
@@ -0,0 +1,80 @@
...
+            value: dict[str, Reading[SlitPosition]],
+        ) -> None:
+            val = value[self.slit_pos.name]["value"]
+            new_slit_info = EntranceSlitInformation.from_slit_positions(val)
+            self._direction_w(new_slit_info.direction)
+            self._setting_w(new_slit_info.setting)
+            self._size_w(new_slit_info.size)
+            self._shape_w(new_slit_info.shape)
+
+        self.slit_pos.subscribe(_sync_soft_signals_with_epics)
```
Villtord: IS there some sort of describe function as well when device disconnects or being deleted?

src/dodal/beamlines/i05.py
```diff
@@ -51,10 +54,24 @@ def sa() -> I05Goniometer:
...
+
+@devices.factory
+def analyser(
+    pgm: PlaneGratingMonochromator, analyser_slits: EntranceSlitInformationDevice
+) -> MbsDetector[LensMode, PassEnergy]:
+    config_sigs = (
+        analyser_slits.direction,
+        analyser_slits.size,
+        analyser_slits.shape,
+        analyser_slits.setting,
```
Villtord: It looks good, but I worry about users who will be tempted to make changes like:
`analyser_slits.size.set(300)`
in their terminal, then because you have only one directional syncing in EntranceSlitInformationDevice layer this soft signal will fall back to Epics value.
Maybe those soft signals should not be exposed at all?

src/dodal/devices/electron_analyser/mbs/mbs_detector.py
```diff
@@ -32,6 +33,7 @@ def __init__(
         energy_source: SignalR[float],
         shutter: GenericFastShutter | None = None,
         source_selector: SourceSelector | None = None,
+        config_sigs: Sequence[SignalR] = (),
```
Villtord, line 36: why allow here any arbitrary config signals, is there a use case for it or is it not a final class?

Villtord (changes requested):

src/dodal/beamlines/i05.py
oliwenmandiamond: These signals are read only. The write signals are all private. From blueapi side, I believe it doesn't expose private signals. They can only change it via epics (which is how it should be done). The read only config sigs are then given to the detector to record the data at the start of the scan whenever data collection is done. It is also impossible (very hard to do) to sync in the opposite direction as the valid combinations are defined by the epics enums.

src/dodal/devices/electron_analyser/mbs/mbs_detector.py
oliwenmandiamond: This is reflecting how `StandardDetector` adds config_sigs https://github.com/bluesky/ophyd-async/blob/main/src/ophyd_async/core/_detector.py#L421C2-L421C3 

and `AreaDetector` API https://github.com/bluesky/ophyd-async/blob/1e07267fb12ff511b248a77a5ed5b6ac50364ad9/src/ophyd_async/epics/adcore/_detector.py#L17

The electron analyser detectors currently use `StandardDetector`, but I am going to move to use `AreaDetector` soon. I was waiting for stable release of ophyd-async first to make it easier so was waiting for this to be merged first https://github.com/bluesky/ophyd-async/pull/1268

src/dodal/devices/electron_analyser/mbs/mbs_analyser_slits.py
oliwenmandiamond: So if the device is deleted, it would delete the slit_pos signal so there would be nothing to unsubscribe from. If the device is disconnected, I don't think this would make a difference. When it is connected again, it will sync the values up again.

src/dodal/devices/electron_analyser/mbs/mbs_analyser_slits.py
oliwenmandiamond: yeah good spot, will add

src/dodal/devices/electron_analyser/mbs/mbs_analyser_slits.py
oliwenmandiamond: I guess we can cache the subscribe id and then only subscribe if it is None

src/dodal/devices/electron_analyser/mbs/mbs_detector.py
Villtord: OK, but I mean you could specify here `analyser_slit` as input parameter and pass down all it's soft signals to 
```
config_sigs = (
            analyser_slit.direction,
            analyser_slit.setting,
            analyser_slit.size,
            analyser_slit.shape,
            self.driver.region_name,
            self.driver.energy_mode,
```

src/dodal/devices/electron_analyser/mbs/mbs_analyser_slits.py
oliwenmandiamond: Okay, I've added a test to check if multiple connects causes multiple subscribes. Updated code slightly and it now passes this test.

src/dodal/devices/electron_analyser/mbs/mbs_detector.py
oliwenmandiamond: I could, but then it makes it less generic when I can reuse what it already does rather than adding it to the constructor to do the exact same thing

Villtord: Right, but why do you need it to be generic for a end class like MBS analyser? Is it going to be changed often or are there going to be child classes from it?

oliwenmandiamond: It will need to work for i21 as well which may or may not have analyser slit. We will be adding additional logic in the `__init__` as it will have to be optional to construct the `config_sigs` which will be a pain and unnecessary. 

This way is much more natural as you just build the device from the dependencies you provide and specify what you want to add as config_sigs without any optionals.The detector is doing nothing with the slits other than outputting with the read_configuration

Villtord: ok, fine

Villtord (approved):

PR #2077, opened 2026-05-29
Pass the plugins on from i22's detectors

DominicOram:
From testing on i22 we found that the stats plugin wasn't actually getting associated with the detectors. This passes it on properly. This was also masking the issue in https://github.com/bluesky/ophyd-async/pull/1278

1. Run `dodal connect i22` on this branch and confirm it fails
2. Switch over your ophyd-async to use https://github.com/bluesky/ophyd-async/pull/1278, confirm dodal connect passes

### Checks for reviewer


RJCD-Diamond (approved):
Approving

DominicOram:
Probably worth doing https://github.com/DiamondLightSource/dodal/issues/2080 before the merge so as not to break things

PR #2079, opened 2026-06-01
Add a device to allow i19 hutches to change the energy

### LINKED ISSUE #2075: Add devices to I19 to set the energy
I19 wants to be able to change the energy on request. For this, we need a few devices to be instantiated in the i19-optics definition:
- dcm
- undulator
- focus mirrors with piezos
- focus mirrors stripes
(some of these may already be in there)

As this is an access controlled operation, we'll also need a device for i19-1 and i19-2 that can send a request to the optics blueapi to run the plan.

## Acceptance Criteria
* All the devices mentioned are instantiated and dodal connect i19-1/i19-2/i19-optics still works

noemifrisina:
Create an access controlled device that i19-1 and i19-2 can use to change the energy on the beamline.

Fixes #2075 

1. Verify tests pass
2. Run `dodal connect i19-{1,2}` and check it works

### Checks for reviewer


src/dodal/beamlines/i19_1.py
```diff
@@ -43,7 +46,7 @@
 set_utils_beamline(BL)
 
 
-I19_1_COMMISSIONING_INSTR_SESSION: str = "cm40638-5"
+I19_1_COMMISSIONING_INSTR_SESSION: str = "cm44168-3"
```
DominicOram, line 49: Should: Do we have a longer term solution to having to keep changing this? Can we make a ticket and reference it here?

src/dodal/devices/beamlines/i19/access_controlled/energy_device.py
```diff
@@ -0,0 +1,101 @@
...
+    HutchState,
+    OpticsBlueAPIDevice,
+)
+from dodal.devices.beamlines.i19.access_controlled.hutch_access import (
+    ACCESS_DEVICE_NAME,
+)
+from dodal.devices.beamlines.i19.mirror_stripes import StripeChoice
+
+
+class OutOfRangeEnergyRequestError(Exception):
```
DominicOram: Nit: It's probably a `OutOfRangeEnergyRequestError(ValueError)` but doesn't really matter

src/dodal/devices/beamlines/i19/access_controlled/energy_device.py
```diff
@@ -0,0 +1,101 @@
...
+            SI: [5KeV, 10KeV)
+            RH: [10KeV, 20KeV)
+            PT: [20KeV, 30KeV)
+        """
+        if 5 <= energy_in_kev < 10:
+            stripe = Stripes.SI
+        elif 10 <= energy_in_kev < 20:
+            stripe = Stripes.RH
+        elif 20 <= energy_in_kev < 30:
+            stripe = Stripes.PT
```
DominicOram: Could: It would be nice if these were in a config file I think

DominicOram (approved):
Great, thanks. Couple of minor things

src/dodal/beamlines/i19_1.py
noemifrisina: There was a ticket in blueapi to enable passing this information to it - can dig it out. 
For the moment I started trying to at least half automate the change of commissioning directory in a separate branch - although apparently it could be any string and the REST call would not complain

src/dodal/devices/beamlines/i19/access_controlled/energy_device.py
noemifrisina: I think there may be one, or there used to be (with just le lower limits in though - but that's fine, they can be worked out if that's the case), will check

CoePaul: Comment after the merge: Should not have used = here just 20 < energy_in_kev < 30

PR #2089, opened 2026-06-11
i15-1: Fix xtal device

jacob-williamson:
The `LaueMonochrometer` was had all of its signals as hinted signals, meaning an error was thrown when trying to read it as which signal to read could not be determined. This fixes it by making `energy_kev` the only hinted signal. 

Also, removes the `force_parser` option from the config server request, as the required PR has been merged + released.

1. Check tests pass

### Checks for reviewer


DominicOram (changes requested):
Great, thanks. Could you add a unit test that would have caught this?

tests/devices/beamlines/i15_1/test_laue_monochrometer.py
```diff
@@ -49,3 +49,10 @@ async def test_energy_kev_gets_value_based_on_y(
         {f"{laue_monochrometer.name}-energy_kev": partial_reading(expected_energy)},
         full_match=False,
     )
+
+
+async def test_reading_device_only_reads_energy_value(
+    laue_monochrometer: LaueMonochrometer,
+):
+    result = await laue_monochrometer.read()
+    assert list(result.keys()) == ["monocrometer-energy_kev"]
```
DominicOram: Nit: Might be better to use `assert_reading` here

DominicOram (approved):
One nit, take it or leave it. Thanks!

tests/devices/beamlines/i15_1/test_laue_monochrometer.py
jacob-williamson: Good spot thanks, in fact I can delete this test and just use `full_match=True` on the test that already exists which fails if there are other signals in the reading.

PR #2092, opened 2026-06-12
Change oav parameters direction attribute to enum from int

rtuck99:
Changes `OAVParameters` so that `direction` attribute is of type `ShutterState` instead of int.

Required for 
* DiamondLightSource/mx-bluesky#1769

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/oav/oav_parameters.py
+++ b/src/dodal/devices/oav/oav_parameters.py
@@ -7,6 +7,8 @@ from xml.etree.ElementTree import Element
 from daq_config_server import ConfigClient
 from daq_config_server.models import DisplayConfig
 
+from dodal.devices.oav.pin_image_recognition import ScanDirections
+
 # GDA currently assumes this aspect ratio for the OAV window size.
 # For some beamline this doesn't affect anything as the actual OAV aspect ratio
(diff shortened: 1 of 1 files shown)


DominicOram (approved):
Thanks, approving to get this over the line but there's some weird things going on here... 

It feels like this should have been caught by a test in `dodal` but on further investigation `direction` is never used in `dodal`. Instead, [setup_pin_tip_detection](https://github.com/DiamondLightSource/dodal/blob/7916b4390cdf558699efc1868c33457f68dc5336/src/dodal/plans/device_setup_plans/setup_pin_tip_params.py#L10) takes a direction and uses that. However, MX have their own [setup_pin_tip_detection_params](https://github.com/DiamondLightSource/mx-bluesky/blob/32818e37f574b3e904d6512e25223f75c78ba716/src/mx_bluesky/common/device_setup_plans/setup_oav.py#L14), which does use the direction. I think this is either a side-effect of when @noemifrisina split out the pin tip detection to be into `dodal` rather than `mx-bluesky` or when we were trying to get pin tip centring working on aithre. I suggest we try and combine these, I'll make an issue and investigate if I get time

PR #2093, opened 2026-06-15
Add i09-2 stage using StandardMovable

### LINKED ISSUE #2107: Add I09-2 sample stage
Add I09-2 sample stage device to beamline. This one is unique as it is a mix of regular motors and piezoelectric motors. This has a tolerance threshold which can be configured and does not have the extra motor signals e.g upper limits. Only setpoint, readback, threshold and stop.

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/2107

2. Refactor of existing logic is okay
3. Tests pass
4. dodal connect works for i09-2

### Checks for reviewer


src/dodal/devices/movable.py
```diff
@@ -0,0 +1,63 @@
...
+    async def move(self, new_position: float, timeout: float | None) -> None:
+        await self.setpoint.set(new_position, timeout=timeout)
+        await wait_for_value(self.setpoint, new_position, timeout=timeout)
+        # Once setpoint is at new position, we can check for tolerance signal to see if
+        # true now as the within_tolerance window has updated to the new setpoint
+        # position. Now it doesn't matter if motor steps are small or large.
+        await wait_for_value(self.within_tolerance, True, timeout=timeout)
+
+
+def _within_tolerance_read(setpoint: float, readback: float, tolerance: float) -> bool:
```
Relm-Arrowny: It is not really doing read?
```suggestion
def _is_within_tolerance(setpoint: float, readback: float, tolerance: float) -> bool:
```

src/dodal/devices/beamlines/i09_2/i09_2_motors.py
```diff
@@ -0,0 +1,69 @@
...
+
+        self.user_setpoint = epics_signal_rw(float, prefix + ":MOV:RD")
+        self.tolerance = soft_signal_rw(float, initial_value=tolerance)
+        self.motor_stop = epics_signal_w(int, prefix + ":HLT:WR.PROC")
+        super().__init__(
+            tolerance=self.tolerance,
+            setpoint=self.user_setpoint,
+            readback=self.user_readback,
+            name=name,
+        )
```
Relm-Arrowny: This feel really wrong, `PiezoElectricMotor` now has both `user_setpoint(feedback)` and `self._setpoint(feedback)_ref` pointing to the exact same underlying signal. Since the `singalRW` already use standard `setpoint` and `readback`  as names, it’s probably better to just standardise on those names here and eliminate the duplicate references.

src/dodal/devices/beamlines/i09_2/i09_2_motors.py
```diff
@@ -0,0 +1,69 @@
...
+
+from dodal.devices.movable import MovableWithTolerance, MovableWithToleranceLogic
+
+
+@dataclass
+class PiezoElectricMovableLogic(MovableWithToleranceLogic):
+    motor_stop: SignalW[int]
+
+    async def stop(self) -> None:
+        await self.motor_stop.set(1)
```
Relm-Arrowny, line 22: Does epics do the cleanup after motor_stop is call?  If a move is currently in progress, it will be stuck awaiting within_tolerance until timeout as the readback never get to the setpoint.

tests/devices/test_movable.py
```diff
@@ -0,0 +1,129 @@
...
+        movable_with_tolerance.movable_logic.move(new_position=setpoint, timeout=1)
+    ) as move_status:
+        # This prevents a race where the test proceeds before the move coroutine has
+        # had a chance to execute its first steps.
+        await wait_for_value(
+            movable_with_tolerance.movable_logic.setpoint, setpoint, timeout=1
+        )
+        assert (
+            await movable_with_tolerance.movable_logic.setpoint.get_value() == setpoint
+        )
```
Relm-Arrowny: The are doing the same thing here.
```suggestion
        await wait_for_value(
            movable_with_tolerance.movable_logic.setpoint, setpoint, timeout=1
        )
```

Relm-Arrowny (changes requested):
A few minor things.

src/dodal/devices/movable.py
oliwenmandiamond: Updated

src/dodal/devices/beamlines/i09_2/i09_2_motors.py
oliwenmandiamond: I've updated to use a template for the names

src/dodal/devices/beamlines/i09_2/i09_2_motors.py
oliwenmandiamond: I believe it will be okay, StandardMovable should handle the clean up logic. I will add a test for this for my i09-2 class to confirm

oliwenmandiamond: So turns out this isn't covered by `StandardMovable` case. This will be fixed by https://github.com/bluesky/ophyd-async/pull/1320 so when stop is called it will cancel the move so will no longer be waiting forever / for timeout so it doesn't matter if within_threshold = True or not on stop.

src/dodal/devices/movable.py
```diff
@@ -0,0 +1,60 @@
...
+        await self.setpoint.set(new_position, timeout=timeout())
+        await wait_for_value(self.setpoint, new_position, timeout=timeout())
+        # Once setpoint is at new position, we can check for tolerance signal to see if
+        # true now as the within_tolerance window has updated to the new setpoint
+        # position. Now it doesn't matter if motor steps are small or large.
+        await wait_for_value(self.within_tolerance, True, timeout=timeout())
+
+
+def _is_within_tolerance(setpoint: float, readback: float, tolerance: float) -> bool:
+    return abs(setpoint - readback) < abs(tolerance)
```
Relm-Arrowny: With the stop update we no longer need `_within_tolerance signal`, as there a clean way to break the move loop now. We can now match motor logic and remove any within_tolerance derived signal, saving us from setpoint signal read and potential racing condition that comes with it. 
```suggestion
@dataclass
class MovableWithToleranceLogic(MovableLogic[float]):
    tolerance: SignalR[float]

    async def move(self, new_position: float, timeout: TimeoutCalculator) -> None:
        current_tolerance = await self.tolerance.get_value()
        await set_and_wait_for_other_value(
            self.setpoint,
            new_position,
            self.readback,
            lambda v: bool(isclose(v, new_position, atol=current_tolerance)),
            timeout=timeout(),
        )
```

Relm-Arrowny (changes requested):
Just one last thing.

src/dodal/devices/movable.py
oliwenmandiamond: Okay, yes I have removed the within_tolerance signal and simplified this a lot. I have also removed the Device and instead just created the movable_logic that is shared now between our devices as the device never actually saved any code and just restricted the signal names. Please have another look

tests/devices/test_movable.py
```diff
@@ -0,0 +1,127 @@
...
+        current_readback = await movable_with_tolerance.user_readback.get_value()
+        assert is_within_tolerance(
+            setpoint=setpoint, readback=current_readback, tolerance=tolerance
+        )
+
+        # Wait for the underlying move task to complete so no race condition.
+        # This ensures the status has fully processed the tolerance condition
+        # and transitioned to a finished state before asserting `done`.
+        await asyncio.wait_for(move_status.task, timeout=1)
```
Relm-Arrowny: nit: probably should make use of asyncStatus, and we already protected by time out in `move`
```suggestion
        await move_status
```

Relm-Arrowny (approved):
Perfect

tests/devices/test_movable.py
oliwenmandiamond: Good spot! I've applied it

PR #2097, opened 2026-06-17
Added I23ZoomController to account for the differing PV names

### LINKED ISSUE #1771: Change OAV Zoom PVs for I23
I23 has two PVs that are used in other beamlines but the I23 version doesn't exist.
- BL23I-EA-OAV-01:FZOOM:ZOOMPOSCMD
- BL23I-EA-OAV-01:FZOOM:MP:SELECT

Instead, they'll need to use:
- BL23I-DI-OAV-01:ZOOM:ZOOMPOSCMD
- BL23I-DI-OAV-01:CAM:MP:SELECT

adaudon:
Other mx-beamlines use a different PV name for zoom. Created a ZoomController specific to I23.

Fixes diamondlightsource/mx-bluesky#1771

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/dodal/beamlines/i23.py
+++ b/src/dodal/beamlines/i23.py
@@ -10,9 +10,13 @@ from dodal.beamlines.aithre import DISPLAY_CONFIG, ZOOM_PARAMS_FILE
 from dodal.common.beamlines.beamline_utils import set_beamline as set_utils_beamline
 from dodal.common.beamlines.device_helpers import HDF5_SUFFIX
 from dodal.common.visit import LocalDirectoryServiceClient, StaticVisitPathProvider
+from dodal.devices.motors import XYZWrappedOmegaStage
 from dodal.device_manager import DeviceManager
 from dodal.devices.motors import SixAxisGonio
-from dodal.devices.oav.oav_detector import OAVBeamCentreFile
+from dodal.devices.oav.oav_detector import (
+    NullZoomController,
+    OAVBeamCentreFile,
+)
 from dodal.devices.oav.oav_parameters import OAVConfigBeamCentre
 from dodal.devices.oav.pin_image_recognition import PinTipDetection
 from dodal.devices.positioner import Positioner1D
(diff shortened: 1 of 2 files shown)


DominicOram (changes requested):
Thanks, I think that rather than make a new device it might be better to just change the `ZoomController` so you can specify the suffix for the percentage/level. That's what we've done in other places where they have been different. I would also suggest you ask controls to make an alias for `CAM:MP:SELECT` -> `ZOOM:MP:SELECT`, we can then reuse the device with no changes

adaudon:
> Thanks, I think that rather than make a new device it might be better to just change the `ZoomController` so you can specify the suffix for the percentage/level. That's what we've done in other places where they have been different. I would also suggest you ask controls to make an alias for `CAM:MP:SELECT` -> `ZOOM:MP:SELECT`, we can then reuse the device with no changes

That's fair, the initial idea was to minimise changes of common dodal files until I23 could discuss whether they wanted to change their PVs to match other beamlines or to keep it as is. But if it's better longterm then it'd be better to change the ZoomController instead.

adaudon:
Zoom isn't required for optical centering so I23ZoomController was replaced with NullZoomController for now so I23 can decide how they want to proceed with xray centering while having access to optical centering for testing.

src/dodal/beamlines/i23.py
```diff
@@ -75,6 +81,8 @@ def oav(config_client) -> OAVBeamCentreFile:
     return OAVBeamCentreFile(
         prefix=f"{PREFIX.beamline_prefix}-DI-OAV-01:",
         config=OAVConfigBeamCentre(ZOOM_PARAMS_FILE, DISPLAY_CONFIG, config_client),
+        percentage_prefix=f"{PREFIX.beamline_prefix}-DI-OAV-01:ZOOM:",
+        level_prefix=f"{PREFIX.beamline_prefix}-DI-OAV-01:CAM:",
```
DominicOram, line 85: Must: As previously discussed, I don't think this is selecting the zoom level, I think it selects which backlight to use. I think you need to discuss with @JamesOHeaDLS about adding a zoom level enum to i23 like other MX beamlines if you want to use this.

DominicOram (changes requested):
Thanks, one comment in code that I think you need to sort out before merging. Additionally, I think you can do it slightly cleaner by doing something a bit like https://github.com/DiamondLightSource/dodal/commit/b25d7f05a45f888bcaa1d3d651c0d0c6316bd38f. The advantages are:
* You don't need to pass the different prefixes around
* You don't have the if statement (which I think had a bug where if people specified a level_prefix but not a percentage_prefix you would not use it.

In general, I think asking controls to alias these is a good idea, particularly if they can add the level enum at the same time.

adaudon:
Wrote some stuff up in a Confluence page after some more research and to clarify my thoughts.
https://confluence.diamond.ac.uk/x/sxfGFw

PR #2099, opened 2026-06-22
Standardise config client and fix tests dependent on order they are ran

### LINKED ISSUE #2098: Test fails depending on the order they are run
If you attempt to run `dodal/tests/common/beamlines/test_device_instantiation.py::test_device_creation`, it will fail. If you run `dodal/tests/common/beamlines/` so every test for beamlines, it will pass. This means we have test environments depending on each which means each test is not tested independently and causes lots of weirdness depending on the order they are run in. The order should not matter for how tests are run.

This should be investigated to fix and will probably find more tests that depend one each other.

### LINKED ISSUE #2131: i15-1 SafeOrBeamPositioner and Blower devices have no name attribute
[SafeOrBeamPositioner](https://github.com/DiamondLightSource/dodal/blob/b012bbb9d29dc176b287eee4ca3e0d7894a8c839/src/dodal/devices/beamlines/i15_1/safe_or_beam_positioner.py#L30) and [Blower](https://github.com/DiamondLightSource/dodal/blob/b012bbb9d29dc176b287eee4ca3e0d7894a8c839/src/dodal/devices/beamlines/i15_1/blower.py#L43) devices have no name attribute in the `__init__.py` method. Instead, the prefix is passed as name to the parent.

This isn't an immediate issue because the `DeviceManager` within dodal and the `DeviceProcessor` within ophyd-async end up renaming the device anyway to the function name / attribute name so was never detected via tests. However, to keep the API consistent, and to wire in the correct name when the device is first initialised, this should be corrected to add an optional name attribute and passed to the parent classes like all other devices.

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/2098 and Fixes #2131

1. Run a few `dodal/tests/common/beamlines/test_device_instantiation.py::test_device_creation[iXX]` in main branch and this branch. Confirm they now pass and doesn't depend on order.
2. Confirm tests pass
3. Confirm added documentation makes sense

### Checks for reviewer


src/dodal/beamlines/i09_1_shared.py
```diff
@@ -30,10 +29,14 @@
 
 devices = DeviceManager()
 
-set_config_client(ConfigClient())
 LOOK_UPTABLE_FILE = "/dls_sw/i09-1/software/gda/workspace_git/gda-diamond.git/configurations/i09-1-shared/lookupTables/IIDCalibrationTable.txt"
 
 
+@devices.fixture
+def iconfig_client() -> ConfigClient:
+    return ConfigClient()
```
DominicOram: Could: In other beamlines this fixture then calls `set_config_client`. I know it's not strictly required as you're not relying on `get_config_client()` but I would consider doing it here (and the other beamlines you've changed) anyway as you may later expect it there

src/dodal/beamlines/i09_2_shared.py
```diff
@@ -37,6 +36,11 @@
 devices = DeviceManager()
 
 
+@devices.fixture
+def jconfig_client() -> ConfigClient:
+    return ConfigClient(url="https://daq-config.diamond.ac.uk")
```
DominicOram: Could: This is the default URL, why are we specifying it?

system_tests/test_oav_system.py
```diff
@@ -15,8 +15,8 @@
 
 
 @pytest.fixture
-async def oav() -> OAV:
-    oav_config = OAVConfig(TEST_OAV_ZOOM_LEVELS, ConfigClient(""))
+async def oav(mock_config_client: ConfigClient) -> OAV:
```
DominicOram: Should: This fixture isn't actually used anywhere, just remove it.

system_tests/test_oav_system.py
```diff
@@ -35,9 +35,9 @@ def take_snapshot_with_grid(oav: OAV, snapshot_filename, snapshot_directory):
 
 # We need to find a better way of integrating this, see https://github.com/DiamondLightSource/mx-bluesky/issues/183
 @pytest.mark.skip(reason="Don't want to actually take snapshots during testing.")
-def test_grid_overlay(run_engine: RunEngine):
+def test_grid_overlay(run_engine: RunEngine, mock_config_client: ConfigClient):
```
DominicOram, line 29: This test is failing on main anyway - I've made https://github.com/DiamondLightSource/dodal/issues/2108.

src/dodal/testing/fixtures/config_client.py
```diff
@@ -2,25 +2,35 @@
...
 from daq_config_server.models import ConfigModel
 from pydantic import TypeAdapter
 
 T = TypeVar("T", str, dict, ConfigModel)
 
+TModel = TypeVar("TModel", bound=ConfigModel)
 
-def fake_config_server_get_file_contents(
-    filepath: str | Path,
+# Tests can register a specific file path with an assoicated file contents to model
```
DominicOram:
```suggestion
# Tests can register a specific file path with an associated file contents to model
```

tests/devices/beamlines/i09_1_shared/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+)
+
+from dodal.testing.fixtures.config_client import (
+    MOCK_CONFIG_CLIENT_PATH_TO_MODEL_CONVERSION,
+)
+from tests.devices.beamlines.i09_1_shared.test_data import TEST_HARD_UNDULATOR_LUT
+
+
+@pytest.fixture
+def mock_config_client(mock_config_client: ConfigClient):
```
DominicOram: Should: Having this fixture the same name as the one in `config_client` that it then depends on makes it not very obvious which one is being used in any subsequent test.

src/dodal/testing/fixtures/config_client.py
```diff
@@ -2,25 +2,35 @@
...
+# conversion function. This should be moved to daq-config-server.
+# See https://github.com/DiamondLightSource/daq-config-server/issues/194
+MOCK_CONFIG_CLIENT_PATH_TO_MODEL_CONVERSION: dict[
```
DominicOram: Should: I'm not a big fan of having this global variable for all the usual reasons globals are bad, how about https://github.com/DiamondLightSource/dodal/commit/abc563425d87b611ca43c604f42c7bfa05910a83?

docs/how-to/write-tests.md
```diff
@@ -121,6 +121,253 @@ async def test_my_device_read(sim_my_device: MyDevice, run_engine: RunEngine) ->
...
+
+
+@devices.factory()
+def my_device(config_client: ConfigClient) -> MyDevice:
+    return MyDevice(
+        config_client=config_client,
+        path=Path(LOOKUPTABLE_DIR, LOOKUP_FILE_NAME),
+    )
+```
+
```
DominicOram: Should: I like this bit of documentation but it doesn't belong in the "how to write tests" document. It's more about how to use the config server

docs/how-to/write-tests.md
```diff
@@ -121,6 +121,253 @@ async def test_my_device_read(sim_my_device: MyDevice, run_engine: RunEngine) ->
...
+the mock implementation will perform:
+
+```python
+MyConfigModel.model_validate(json.loads(contents))
+```
+
+### Registering a custom parser
+
+Some configuration file formats are not JSON-based and require custom parsing logic. For these cases, tests can register a parser using `MOCK_CONFIG_CLIENT_PATH_TO_MODEL_CONVERSION`.
```
DominicOram: Should: If you accept my suggestion this documentation needs changing a bit

tests/common/beamlines/test_device_instantiation.py
```diff
@@ -1,44 +1,25 @@
...
-    yield
-    clear_config_client()
-
-
 @pytest.mark.parametrize(
     "module_and_devices_for_beamline",
     set(all_beamline_modules()),
     indirect=True,
 )
+# Increase timeout for this specific test as running inividually can take over a second
```
DominicOram:
```suggestion
# Increase timeout for this specific test as running individually can take over a second
```

tests/common/beamlines/test_device_instantiation.py
```diff
@@ -61,6 +42,9 @@ def test_device_creation(module_and_devices_for_beamline):
...
```
DominicOram:
```suggestion
# Increase timeout for this specific test as running individually can take over a second
```

docs/how-to/write-tests.md
```diff
@@ -121,6 +121,253 @@ async def test_my_device_read(sim_my_device: MyDevice, run_engine: RunEngine) ->
...
+```
+
+These files are not available in unit test environments and must not be accessed during tests.
+
+To allow beamline import and device creation tests to run without access to the DLS filesystem, the `module_and_devices_for_beamline` fixture performs two forms of mocking:
+
+1. Replaces `ConfigClient.get_file_contents()` with the `mock_config_client` implementation.
+2. Replaces known module-level file path constants with paths to test data files.
+
+### Mocking `ConfigClient`
```
DominicOram: Could: I'm not sure why we need to document this? Why does someone writing new tests need to know about it?

DominicOram (changes requested):
Thanks, I couldn't reproduce the flaky tests issue but I think this cleans things up a bit anyway. Couple of spelling mistakes, I would suggest installing https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

jacob-williamson (commented):
Have discussed a bit with @DominicOram and I don't think we should care about converters in the mock client as conversion is a server-side process. In dodal we should assume we are receiving converted data from the server, and test data should conform to the post-conversion format.

The exception is where the force-parser option is used, which would force conversion to take place client side, but this is already handled by the existing fixture and should only be a temporary measure wherever it's used.

src/dodal/beamlines/i09_1_shared.py
oliwenmandiamond: I personally want to remove the `set_config_client` as soon as possible when blueapi will allow it. Our group of beamlines don't have the need for it and all use dependency injection rather than global variables and would rather not encourage it if possible

system_tests/test_oav_system.py
oliwenmandiamond: Thanks

docs/how-to/write-tests.md
oliwenmandiamond: Okay, have moved

system_tests/test_oav_system.py
oliwenmandiamond: done

docs/how-to/write-tests.md
oliwenmandiamond: Outdated now

docs/how-to/write-tests.md
oliwenmandiamond: If they add bits to the beamline and use the config_client straight away, it will cause the tests to fail and isn't always obvious why. I've tried to make it as explicit and documented as possible so that it is very easy to fix and patch the bits you need when the test fails

src/dodal/beamlines/i09_2_shared.py
oliwenmandiamond: Removed

src/dodal/beamlines/i09_1_shared.py
```diff
@@ -60,12 +63,14 @@ def ienergy_order() -> UndulatorOrder:
...
-    ienergy_order: UndulatorOrder, iid: UndulatorInMm
+    ienergy_order: UndulatorOrder,
+    iid: UndulatorInMm,
+    iconfig_client: ConfigClient,
 ) -> HardInsertionDeviceEnergy:
     return HardInsertionDeviceEnergy(
         undulator_order=ienergy_order,
         undulator=iid,
-        config_server=get_config_client(),
+        config_server=iconfig_client,
```
Villtord: Nit: could we also rename `config_server` argument here to `config_client `please?

docs/how-to/write-tests.md
```diff
@@ -121,6 +121,137 @@ async def test_my_device_read(sim_my_device: MyDevice, run_engine: RunEngine) ->
...
+@pytest.fixture
+def path_to_mock_data() -> dict:
+    return {
+        TEST_HARD_UNDULATOR_LUT: parse_i09_hu_undulator_energy_gap_lut(
+            Path(TEST_HARD_UNDULATOR_LUT).read_text()
+        )
+    }
+```
+
+The mock `ConfigClient` is then configured with this data:
```
Villtord: Could add that this mock already exist in `tests/conftest.py` and there is no need to add it.

docs/how-to/write-tests.md
```diff
@@ -121,6 +121,137 @@ async def test_my_device_read(sim_my_device: MyDevice, run_engine: RunEngine) ->
...
+from daq_config_server.client import ConfigClient
+from dodal.devices.my_device import MyDevice
+from ophyd_async.core import init_devices
+
+
+@pytest.fixture
+def my_device(mock_config_client: ConfigClient) -> MyDevice:
+    with init_devices(mock=True):
+        return MyDevice(
+            config_client=mock_config_client,
```
Villtord, line 144: would this flag `mock=True` propagate to `mock_config_client` and do "unexpected" things?

tests/conftest.py
```diff
@@ -29,45 +30,80 @@
...
 
+@pytest.fixture
+def path_to_mock_data() -> dict:
+    return {}
+
+
+@pytest.fixture
+def mock_config_client(path_to_mock_data) -> ConfigClient:
+    config_client = ConfigClient()
+    config_client.configure_mock(path_to_mock_data)
```
Villtord: Was it not possible to use this:
```
with init_devices(mock=True):
    config_client = ConfigClient()
```

Villtord (commented):
Sorry I've got a few questions while I was reading it through.

tests/conftest.py
oliwenmandiamond: No, this only works with `Device`s. `ConfigClient` is not a `Device`.

docs/how-to/write-tests.md
oliwenmandiamond: No, init_devices only works on Devices. ConfigClient is not a device so wouldn't be effected. This is outdated now and moved to contructor anyway and not with mock=True/False but you provide a server_response object which default is real but can sub out with Mock one.

src/dodal/beamlines/i09_1_shared.py
oliwenmandiamond: Renamed this. The original thought of doing it this way was because DeviceManager would throw an error because if we define it i09_1_shared and i09_2_shared, DeviceManager throws an error on duplicate devices. However, reading the code duplicate devices are not allowed, duplicate fixtures are okay and can be overwritten so actually causes no issue.

docs/how-to/write-tests.md
oliwenmandiamond: done

DominicOram (approved):
Great, thank you!

PR #2106, opened 2026-06-24
Tidy up circularity testing of reciprocal function pairs

### LINKED ISSUE #2102: Improve conversion tests including circularity functionality checks in dodal / general maths
Split from issue #2101
There are tests in dodal / common / general maths module for 
general purpose quantity conversions and for transmission specific conversion.

Issues with these tests
( whilst mostly well written )
a) the tests feature very similar input - nothing other than powers of ten have been tried out
b) the circularity tests could have used a common circularity harness instead of each test rolling its own and being 
  potential subject to an internal flaw

## Acceptance Criteria
* That unit tests for conversions feature more test cases, more realistic
* That the circularity tests use a common "harness" mechanism for checking any reciprocating pair of functions

CoePaul:
* For unary operator functions acting on float inputs like conversions in both directions between one unit and another

* This replaces separately rolled once per test circularity with use of a common test circularity framework class

Fixes #2102 

1. Confirm CI tests all pass
2. See that You're happy as a reviewer with the separation out of circularity testing into its own class (ancilliary to testing)

### Checks for reviewer

### DIFF
--- a/tests/common/general_maths/test_arithmetic_conversions.py
+++ b/tests/common/general_maths/test_arithmetic_conversions.py
@@ -1,4 +1,5 @@
 import math
+from collections.abc import Callable
 
 import pydantic
 import pytest
@@ -13,6 +14,8 @@ from dodal.common.general_maths.arithmetic_conversions import (
     convert_mm_to_microns,
     convert_percentage_to_factor,
 )
+
+from .operator_inversion_pairing import OperatorInversionPairing
 
 
 # expected success tests (the 'Happy Path'): All numbers here are arbitrary
@@ -56,29 +59,51 @@ def test_conversion_from_microns_to_centimetres(input, result):
     assert convert_microns_to_cm(input) == pytest.approx(result)
 
 
-# Circular tests (all numbers here arbitrary)
+# Circular "sanity check" tests, exercise pairs of reciprocating functions
+# proving the result of applying a function and its inverse results in the original value
 
 
-@pytest.mark.parametrize("input", [0.0, 1.0, 10.0, 100.0])
-def test_circular_cm_to_mm_and_back(input):
-    assert convert_cm_to_mm(convert_mm_to_cm(input)) == pytest.approx(input)
-    assert convert_mm_to_cm(convert_cm_to_mm(input)) == pytest.approx(input)
-
-
-@pytest.mark.parametrize("input", [0.0, 1.0, 10.0, 100.0])
-def test_circular_microns_to_mm_and_back(input):
-    assert convert_microns_to_mm(convert_mm_to_microns(input)) == pytest.approx(input)
-    assert convert_mm_to_microns(convert_microns_to_mm(input)) == pytest.approx(input)
-
-
-@pytest.mark.parametrize("input", [0.0, 1.0, 10.0, 100.0])
-def test_circular_percentage_to_factor_and_back(input):
-    assert convert_percentage_to_factor(
-        convert_factor_to_percentage(input)
-    ) == pytest.approx(input)
-    assert convert_factor_to_percentage(
-        convert_percentage_to_factor(input)
-    ) == pytest.approx(input)
+@pytest.mark.parametrize(
+    "f, g, numerical_args",
+    [
+        (
+            convert_ev_to_kev,
+            lambda k: k * 1000.0,
+            [16.83, 0.0, 0.037, 1.0, 6.208, 18, 12345.6, 28906.4],
+        ),
+        (
+            convert_mm_to_cm,
+            convert_cm_to_mm,
+            [-16.83, 0.0, 0.037, 1.0, 6.208, 18, 102.99],
+        ),
+        (
+            convert_microns_to_cm,
+            lambda x: convert_mm_to_microns(convert_cm_to_mm(x)),
+            [-6.119, 0.0, 0.764, 1.02, 62.45, 12754, 3154.59],
+        ),
+        (
+            convert_microns_to_mm,
+            convert_mm_to_microns,
+            [-12.38, 0.0, 0.307, 1.0, 6.45, 24, 231.089],
+        ),
+        (
+            convert_factor_to_percentage,
+            convert_percentage_to_factor,
+            [0.0, 1.0, 0.5, 0.367, 27.404, 100.0, 99.8, 53.647],
+        ),
+    ],
+)
+def test_reciprocal_function_pairs_nest_consistent_with_identity(
+    f: Callable[[float], float],
+    g: Callable[[float], float],
+    numerical_args: list[float],
+):
+    for op_pair in [
+        OperatorInversionPairing(f, g),
+        OperatorInversionPairing(g, f),
+    ]:
+        for x in numerical_args:
+            assert op_pair.composed_operator_is_consistent_with_identity_operator(x)
 
 
 # The inauspicuous path
--- a/tests/common/general_maths/operator_inversion_pairing.py
+++ b/tests/common/general_maths/operator_inversion_pairing.py
@@ -1,0 +1,49 @@
+from collections.abc import Callable
+from dataclasses import dataclass
+
+import pytest
+
+
+@dataclass(frozen=True)
+class OperatorInversionPairing:
+    """Represents a pair of mutually reciprocal maths functions, with unique mappings of input to output.
+        Useful for sanity check behavioural tests above the unit level but below integration level.
+        A typical use case is to ensure that after two mutually cancelling numerical operations,
+         an original input number is restored.  This helps tests flag whenever one function breaks,
+         for example - owing to a typo or incorrect or inverted multiplication factor.
+         ( Yes there are even numbers of self-cancelling mistakes which can still hide bugs but
+           hopefully the individual functions, functional tests flush those out ).
+
+    Attributes:
+        unary_op: A unary operation mathematical function (specifically with unique one-to-one mapping).
+        inverse_op: The inverse unary operation.
+    """
+
+    unary_op: Callable[[float], float]
+    inverse_op: Callable[[float], float]
+
+    def _composed_operator(self, x: float) -> float:
+        """Applies both the unary operation followed by the inverse operation on a numerical input.
+            On, for example, the happy path of a test, this round-trip can be expected to result in the original value x.
+            Internal method.
+
+        Args:
+            x (float): Any numerical argument suitable for the unary operations under test.
+
+        Returns:
+            float: The result from nested application of first the unary operation and then its inverse on x.
+        """
+        _f_of_x = self.unary_op(x)
+        return self.inverse_op(_f_of_x)
+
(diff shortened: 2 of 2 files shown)


tests/common/general_maths/circularity_functional_verification.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+    def verify_forwards_circulation_returns_to_same_value(self, x: float):
+        conversion = self.f(x)
+        assert self.g(conversion) == pytest.approx(x)
+
+
+    def verify_reverse_circulation_returns_to_same_value(self, x: float):
+        conversion = self.g(x)
+        assert self.f(conversion) == pytest.approx(x)
```
oliwenmandiamond: I think this is overkill. I think a better solution would be to make it a util function

```Python
def assert_roundtrip(forward, reverse, x):
    assert reverse(forward(x)) == pytest.approx(x)
    assert forward(reverse(x)) == pytest.approx(x)
```

Then you could use it like this:

```Python
def test_circular_mm_cm_interconversions():
    for x in [1, 2, ...]: 
        assert_roundtrip(convert_cm_to_mm, convert_mm_to_cm, x)
```

oliwenmandiamond (changes requested):
I think is a good idea, but looks to be overkill. You can simplify this a lot and add it into a `utils.py` function inside `tests/common/general_maths` so more helpful functions to assist with testing could be added as well.

tests/common/general_maths/circularity_functional_verification.py
CoePaul: I don't see why making a class is such a heavy weight cost - there is no over kill. 
If everybody in the group had to pay 10 quid every time a class was used then fair.

CoePaul: also note Your suggested util function is really two functions in one ( but doesn't say that on the tin )

oliwenmandiamond: So your tests are always doing this

```python
t = FunctionalCircularityTester(f, g)
t.verify_forwards_circulation_returns_to_same_value(x)
t.verify_reverse_circulation_returns_to_same_value(x)
```

Why do that when simply do this:
```python
assert_roundtrip(f, g, x)
```
It's a lot more concise and is the point of a util function for testing logic works. 

The name can be whatever you want. Maybe `assert_forward_reverse_func`

CoePaul: I see Your points and counter with these advantages of the OO assuming immutable classes: 

Encapsulation of Behavior: Instead of just grouping data, you are grouping the mathematical contract alongside the functions. The object represents the concept of a "Reciprocal Pair" itself.

    Contextual Clarity: In your test file, writing pair = ReciprocalPair(f, g) explicitly names the relationship between f and g. The code reads as: "This object is a Reciprocal Pair, and I am asking it to verify its integrity at point x."

    Thread Safety / Zero Side Effects: Because it's immutable, you can instantiate it once at the top of a test module and safely share it across dozens of different test conditions without fear of leaks.

Where the Rub Lies (The "Pythonic" Trade-Off)

The reason many Python developers default to the alternative functional approach isn't because immutability is bad—it’s because of Python's language syntax overhead.

In languages like Scala or Kotlin, immutable classes are incredibly lightweight. In Python, creating a class often requires writing boilerplate (__init__, type hints, etc.) or importing helper modules (dataclasses, NamedTuple) just to achieve true immutability.
The Verdict under Immutable OO

If your architectural goal is clean, expressive domain modeling where behaviors belong strictly to the nouns they describe, Version 2 is absolutely the better step. It elevates f and g from two random functions passing through a utility pipe into a cohesive, immutable mathematical domain concept.

CoePaul: I reply to "Why do that when simply do this"

why call "some test line on (x)" <-- one argument 

vs 

do-everything-in-one-kitchen-sink(f,g,x) <-- three arguments and now the reader has to think about each

CoePaul: In a perfect world all methods would only ever take zero or one argument 
code would be a lot longer 
but each line would be very readable

CoePaul:
also the suggested change is tantamount to where we started

tests/common/general_maths/circularity_functional_verification.py
CoePaul: @oliwenmandiamond: I think Your original points did validly highlight the ambiguities inherent in my earlier stab at this.  I've tried to improve the clarity of purpose across the latest update

tests/common/general_maths/test_arithmetic_conversions.py
```diff
@@ -56,29 +58,46 @@ def test_conversion_from_microns_to_centimetres(input, result):
...
+            [0.0, 1.0, 0.5, 0.367, 27.404, 100.0, 99.8, 53.647],
+        ),
+    ],
+)
+def test_sanity_check_reciprocal_function_pairs(
+    function_pair: ReciprocalFunctionPair, numerical_args: list[float]
+):
+    for x in numerical_args:
+        assert x == pytest.approx(function_pair.convert_then_invert(x))
+        assert x == pytest.approx(function_pair.invert_then_convert(x))
```
oliwenmandiamond:
```python
@pytest.mark.parametrize(
    "forward, reverse, numerical_args",
    [
        (
            convert_ev_to_kev,
            lambda k: k * 1000.0,
            [16.83, 0.0, 0.037, 1.0, 6.208, 18, 12345.6, 28906.4],
        ),
        (
            convert_mm_to_cm,
            convert_cm_to_mm,
            [-16.83, 0.0, 0.037, 1.0, 6.208, 18, 102.99],
        ),
        (
            convert_microns_to_cm,
            lambda x: convert_cm_to_mm(convert_mm_to_microns(x)),
            [-6.119, 0.0, 0.764, 1.02, 62.45, 12754, 3154.59],
        ),
        (
            convert_microns_to_mm,
            convert_mm_to_microns,
            [-12.38, 0.0, 0.307, 1.0, 6.45, 24, 231.089],
        ),
        (
            convert_factor_to_percentage,
            convert_percentage_to_factor,
            [0.0, 1.0, 0.5, 0.367, 27.404, 100.0, 99.8, 53.647],
        ),
    ],
)
def test_sanity_check_reciprocal_function_pairs(
    forward: Callable[[float], float], reverse: Callable[[float], float], numerical_args: list[float],
):
    for x in numerical_args:
        assert_forward_reverse(forward, reverse, x)
```
The current ReciprocalFunctionPair implementation stores the two functions and then exposes several wrapper methods that ultimately just call those functions in different orders. Unless we expect to attach additional metadata or behaviour to these pairs in future, I don't see a strong benefit to introducing the extra abstraction.

tests/common/general_maths/reciprocal_function_pair.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        return self.f(x)
+
+    def _invert(self, x: float):
+        """Performs the inverse operation (as defined in the constructor call.
+
+        Args:
+            x: a suitable numerical argument which should be compatible with the functions tested.
+        """
+        return self.g(x)
```
oliwenmandiamond: I am not sure how five methods of a class is better than a simple assert function which achieves the same thing. 

```python
def assert_forward_inverse(forward: Callable[[float], float], inverse: Callable[[float], float], x: float):
    """Assert that two functions are mutual inverses for a given input.

    Verifies that applying the forward function followed by the inverse
    returns the original value, and vice versa, within pytest's floating
    point tolerance.

    Args:
        forward: Function mapping a float to a float.
        inverse: Function expected to be the inverse of ``forward``.
        x: Test value to round-trip through both functions.
    """
    assert reverse(forward(x)) == pytest.approx(x)
    assert forward(inverse(x)) == pytest.approx(x)
```

I can see the value of the class if we expect to add additional metadata or behaviour to these function pairs in future, but based on the current usage it feels unnecessary.

oliwenmandiamond (dismissed):

tests/common/general_maths/reciprocal_function_pair.py
CoePaul: - help me get round my mental block:
I'm struggling to avoid reading the review as 
"You don't have permission to write declarative code that will aid tests - because 
there's no compelling business case for doing so"

oliwenmandiamond: I've requested changes because I agree on the use case but don't agree on the implementation. As this is a general util useful for testing for math function, in theory anyone else should be able to import and use but that isn't possible at the moment. My aim was in the style of ophyd-async [assert functions](https://github.com/bluesky/ophyd-async/blob/0513030afdd76bf2a6a0ccc477ab50de7ea72587/src/ophyd_async/testing/_assert.py#L57) e.g `assert_reading`, `assert_value`, `assert_configuration`, `assert_emitted` etc with some more examples from [dodal](https://github.com/search?q=repo%3ADiamondLightSource%2Fdodal+%22def+assert_%22&type=code).

Thoughts @jacob720 and @rtuck99? I'll let you tip the scales on this PR.

CoePaul: I originally presented a solution that " anyone else should be able to import and use "
because I widely expect that - especially in green field projects like dodal ( but even secretly in legacy code ).

Once the earlier submissions were shot-down for being "unnecessary because we're not going to use it much"
I took it back out the of the way.

A further update sees the universal accessibility restored.

I do fail to see why using a class is objectionable.  OO is a thing - even a good thing.
Use it or lose it.

rtuck99:
Honestly I can't get too excited about any of this. Arguably far too much time has been spent on reviewing this already.

I don't think there's anything particularly wrong about the code as is, if anything I would question the need for any of the conversion functions anyway as I don't see much value in having bespoke functions that multiply or divide by powers of 10 for SI unit conversons. 9 times out of 10 I would prefer someone just have clearly named variables that identify the units, and use concise statements that just multiply or divide by 1000, or whatever, than have to read something littered with clunky convert_x_to_y(blah) calls.

That is, unless there is some later beneficial usage of having this selection of functions, which presumably would require additional machinery to select the correct function. Alternatively, we could go the full hog and use a dedicated special purpose library for this like `pint`. But I have yet to see a convincing argument for using it within `mx-bluesky`.

In short, approved, although I question whether the triviality of the functions being tested outweighs the effort being expended on authoring and testing them as I would be very tempted to inline them if I ever see them being used.

rtuck99 (approved):
Approved

CoePaul:
> I would question the need for any of the conversion functions anyway 

Just for the historical record @rtuck99 - if You saw how often conversions went the wrong way in GDA 
( and how long the code base retained those errors ) You would never ever permit people to "roll their own" locally 
in each bit of code - but would instead - as we are trying to here - insist on a central function call that does the conversion universally.
( It was bad enough in GDA - I'm sure I've even seen roll - your - own conversions - happening inside medical software which is even worse )

PR #2115, opened 2026-07-09
Add i06-1 magnets

### LINKED ISSUE #2120: Create i06 vector magnet device in dodal
To overcome the limitations of legacy GDA software and enable advanced fly-scanning on I06-1, we are creating a dedicated VectorMagnet device within Dodal. Porting this physical magnet device to Bluesky allows us to natively leverage Flyable and Preparable protocols for dynamic ramp-speed adjustments, continuous physical fly scans, and future alternating-energy scans.
### Acceptance criteria : 

- [x] Ophyd Device Implementation: Create a SuperconductingMagnet(VectorMagnet) class inside dodal that maps to the physical magnet controller's EPICS PVs, it should have three independence axis.
- [x] Flyable & Preparable: Each axis of the magnet should be movable and flyable. 
- [x] Pre-Move Limit Check: The device should validates that a requested field is safe within the magnet mode it operating in.
- [x] x y quadrant limit protect: Current epics limit for the x-y quadramt does not reflect the actually limit stated in the manaul.

### Tasks

- [ ]

### Related Resources

see DiamondLightSource/sm-bluesky#335 for detail of Limit and other requirements.

oliwenmandiamond:
Fixes #2120 
This add the basic device layer and control structures for the I06-1 superconducting vector magnet. It implements coordinate transformations, safe coordinated multi-axis moves, and support single axis x, y, z flyscan.

### Changes
- Magnet Controller: Adds SuperConductingMagnetController to execute coordinated step by step movements based on the active MovementStrategy mode.
- Coordinate Systems: Adds MagnetCartesianCoordinates ($X, Y, Z$) and MagnetSphericalCoordinates ($\rho, \theta, \phi$) with automated coordinate conversion.
- Fly Scanning: Implements Flyable and Preparable protocols on MagnetAxis to handle single axis flying.
-  Hardware Mocking: Adds MockSuperConductingMagnetController to simulate IOC ramp state transitions and setpoint behavior during testing.

1. Check implementations make sense
2. Check tests coverage is sufficient
3. Check doc strings for devices make sense
4. Check `dodal connect i06-1`

### Checks for reviewer


src/dodal/beamlines/i06_1.py
```diff
@@ -17,12 +21,12 @@
 
 
 @devices.factory()
-def diff_cooling_temperature_controller() -> Lakeshore336:
+def ls336() -> Lakeshore336:
```
Relm-Arrowny: It help to knowing what physically it is controlling.
```suggestion
def ls336() -> Lakeshore336:
``` dd cooling insert temperature controller ```

```

src/dodal/beamlines/i06_1.py
```diff
@@ -17,12 +21,12 @@
...
     return Lakeshore336(prefix=f"{PREFIX.beamline_prefix}-EA-TCTRL-02:")
 
 
 @devices.factory()
-def diff_heating_temperature_controller() -> Lakeshore336:
+def ls336_2() -> Lakeshore336:
```
Relm-Arrowny: Same as above but for heating.

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
```diff
@@ -0,0 +1,39 @@
...
+    StandardReadableFormat,
+    TimeoutCalculator,
+)
+from ophyd_async.epics.core import epics_signal_rw
+
+
+class RampRateMovableLogic(MovableLogic[float]):
+    async def move(self, new_position: float, timeout: TimeoutCalculator):
+        await self.setpoint.set(new_position, timeout=timeout())
+
```
Relm-Arrowny, line 31: we should have a guard here, assuming you do not have negative ramprate:
```python
async def check_move(self, new_position: float) -> None:
        if new_position == 0.0:
            LOGGER.warning("Ramp rate set to zero. Magnet will not move.")
        current_limit = await self.limit.get_value()
        if new_position > current_limit:
            raise ValueError(
                f"Requested ramp rate {new_position} exceeds the maximum limit of {current_limit}"
            )
```

src/dodal/devices/beamlines/i06_1/magnets/ramp_controller.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class MagnetThreeAxesRampRateController(StandardReadable):
+    def __init__(self, prefix: str, name: str = ""):
+        with self.add_children_as_readables():
+            self.x = MagnetAxisRampRateController(prefix + "-01:")
+            self.y = MagnetAxisRampRateController(prefix + "-02:")
+            self.z = MagnetAxisRampRateController(prefix + "-03:")
+
+        super().__init__(name)
```
Relm-Arrowny: I think is better to have MagnetAxis owning their controller, it make more logical sense as each axis must have its power supply units and its ramp rate. 

I would go further, as these 3 signal is effectively just like MOTOR.max_velocity, MOTOR.velocity, I would just have them as a variable in MagnetAxis.

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
```diff
@@ -0,0 +1,39 @@
...
+
+# Equivalent to GDA SuperconductingMagnetControllerClass
+class MagnetAxisRampRateController(StandardMovable[float], StandardReadable):
+    def __init__(self, prefix: str, name: str = ""):
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
```
Relm-Arrowny, line 49: This is effectively speed for the motor.
```suggestion
        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
```

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,318 @@
...
+        self,
+        prefix: str,
+        axis: str,
+        ramp_rate: MagnetAxisRampRateController,
+        magnet_set_within_boundary: Callable[[], Awaitable[None]],
+        name: str = "",
+    ):
+        # Used in fastfieldscan, need to add fly scan logic.
+        self.ramp_rate_ref = Reference(ramp_rate)
+
```
Relm-Arrowny, line 101: I think This is porbably better?
```suggestion
class MagnetAxis(StandardReadable, StandardMovable[float], Flyable, Preparable):
    def __init__(
        self,
        prefix: str,     
        ramp_prefix: str, 
        axis: str,
        magnet_set_within_boundary: Callable[[], Awaitable[None]],
        name: str = "",
    ):
        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
            self.readback = epics_signal_r(float, prefix + "RBV")
            
        with self.add_children_as_readables():
            self.ramp_rate = MagnetAxisRampRateController(ramp_prefix)

```

Relm-Arrowny: We can go further if we accept ramp_rate is just a signal and make a baseClass where all magnet can use
```python
class BaseMagnetAxis(StandardReadable, StandardMovable[float], Flyable, Preparable):
    def __init__(
        self,
        prefix: str,
        psu_prefix: str,
        magnet_set_within_boundary: Callable[[], Awaitable[None]],
        name: str = "",
    ):
            self.readback = epics_signal_r(float, prefix + "RBV")
        self.demand = epics_signal_rw(float, prefix + "DMD")

            self.ramp_rate = epics_signal_rw(
                float,
                read_pv=psu_prefix + "RBV:FIELDSWEEPRATE",
                write_pv=psu_prefix + "SET:FIELDSWEEPRATE",
            )
        self.ramp_limit = epics_signal_rw(float, psu_prefix + "LIM:RAMPRATE:TPM")# this can be optional
```
The rest is mostly the same and you not longer need the ramp_controller.py

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,318 @@
...
+class MagnetAxisMovableLogic(MovableLogic[float]):
+    axis: str
+    magnet_set_within_boundary: Callable[[], Awaitable[None]]
+
+    async def move(
+        self, new_position: float, timeout: Callable[[], float | None]
+    ) -> None:
+        values = {self.axis: new_position}
+        # ToDo - feed timeout to here?
+        await self.magnet_set_within_boundary(**values)
```
Relm-Arrowny: Passing set_within_boundary down into the child axis creates a circular dependency feels wrong. We shouldn't have an individual axis method controlling or triggering movements across all three axes.
``` python
     await self.setpoint.set(new_position, timeout=timeout())
```

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,318 @@
...
+
+        callback_on_mock_put(device.start_ramp, _set_ramp_status)
+        set_mock_value(device.limit_status, MagnetLimitStatus.OK)
+
+
+@default_mock_class(MockSuperConductingMagnet)
+class SuperConductingMagnet(StandardReadable, Movable[MagnetPosition]):
+    def __init__(
+        self,
+        prefix: str,
```
Relm-Arrowny: So it only flyable per axis?

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,318 @@
...
+        fly_info = error_if_none(
+            self._fly_info,
+            f"{self.name} must be prepared before attempting to kickoff.",
+        )
+        self._fly_status = self.set(fly_info.end_position)
+        # Reset state so prepare must be called with each kickoff again.
+        self._fly_info = None
+
+    def complete(self) -> WatchableAsyncStatus:
+        """Return a ``Status`` and mark it done when acquisition has completed."""
```
Relm-Arrowny: not sure we need that 
```suggestion
```

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,318 @@
...
+            await self._ramp()
+        # Then decrease X
+        if dx < 0:
+            await self.x.demand.set(x)
+            await self._ramp()
+        # Then decrease Y
+        if dy < 0:
+            await self.y.demand.set(y)
+            await self._ramp()
+        # Finally perform all increases together
```
Relm-Arrowny: I understand they do no want to quench the magnet but this feel strange, why we have to do them 1 by 1 when decreasing? this would mean the field direction changes as I decrees the field, but they are happy with increasing all at the same time. 
```python
 await asyncio.gather(

self.x.demand.set(x),

self.y.demand.set(y),

self.z.demand.set(z),

) 
```
We need to talk to them about this part,

Relm-Arrowny (commented):
I gone through it roughly, I have a few questions

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: This is what I've ported from GDA. The original code does this

```python
		# Increase only if one or more need it.
		if dx > 0 or dy > 0 or dz > 0:
			if self.verbose:
				[x0, y0, z0]=self.getMagnetPosition(raw=True); #New starting point
				print "Move X, Y, Z from " + str([x0, y0, z0]) + " to " + str([x1, y1, z1]);
			self.chYi.caput(newY);
			self.chXi.caput(newX);
			self.chZi.caput(newZ);
			self.chGo.caput(self.timeoutLimit, 1);

		if self.verbose:
			print "Waiting for magnet readback to settle down..."
		
		#Due to EPICS flaw, after the callback, still need to wait for a few second for the magnet settle down
		if self._delay is not None:
			sleep(self._delay);
		
		if self.verbose:
			print "Wait completed, checking status..."
		
		self.checkMagnetStatus();

...

	def checkMagnetStatus(self):
		self.getMagnetStatus();
		if self.rampStatus == SuperconductingMagnetClass.SCM_STATUS_DONE and self.limitStatus == SuperconductingMagnetClass.SCM_LIMIT_OK:
			if self.verbose:
				print "The Magnet Ramping is done and the Limit is OK";
			return True;
		else:
			if self.rampStatus == SuperconductingMagnetClass.SCM_STATUS_RAMPING:
				if self.verbose:
					print "The Magnet is still ramping.";
			if self.limitStatus == SuperconductingMagnetClass.SCM_LIMIT_VIOLATION:
				if self.verbose:
					print "Limit violation error!";
			return False;
```
So tried to optimise it where possible. However, am not familiar with the hardware so code might need more tuning. Agreed need to talk to them.

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: yes, the GDA fastfieldscan code specifically only flys one axis

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: I do agree to an extent. However, the GDA code I am porting doesn't make it easy to break this link

src/dodal/devices/beamlines/i06_1/magnets/ramp_controller.py
oliwenmandiamond: So the issue is that the ramp_rate axis has a different PV to the magnet axis. This implies it is a different device + GDA also treat it this way, make ramp_rate axes as their own devices that uses can then interact with. 

I 100% agree that the code would be much nicer if the ramp_rate lived with the magnet axis rather than its own device,  however the PV and GDA code I am porting implies it shouldn't be. Will have to discuss with them on why it is done this way.

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: yep agreed, however as mentioned above the PV's are not the same. 

MagnetAxis x -> BL06J-EA-MAG-01:X:DMD. 
RampRate X -> BL06J-EA-SMC-01:STS:RAMPRATE:TPM

MagnetAxis y -> BL06J-EA-MAG-01:Y:DMD. 
RampRate Y -> BL06J-EA-SMC-02:STS:RAMPRATE:TPM

MagnetAxis z -> BL06J-EA-MAG-01:Z:DMD. 
RampRate Z -> BL06J-EA-SMC-03:STS:RAMPRATE:TPM

which implies they are not the same device and GDA also treats it as its different devices.

src/dodal/beamlines/i06_1.py
oliwenmandiamond: I updated the names to be the same as GDA. I personally think `diff_cooling_temperature_controller` is way too long of a device name for any user to type out. I think a middle ground is to update the device factory methods to include this extra description?

or maybe 
`ls336_cooling`
`ls336_heating`?

src/dodal/devices/beamlines/i06_1/magnets/ramp_controller.py
oliwenmandiamond: The MagnetAxis cannot own the controller. There are 3 MagnetAxis and 1 controller between the 3 so would have 3 parents which is not allowed. We can only give it a reference of it or a the functions to call from the controller

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
oliwenmandiamond: Yes this is an equivalent of a Motor speed, but the difference is we have a readback and a setpoint for it unlike a Motor which has just a single SignalRW. 
This is the readback of the speed so when I read the device e.g ramp_ratex it should be the hinted signal 

assert RE(rd(ramp_ratex)) == await ramp_ratex.readback.get_value()

src/dodal/beamlines/i06_1.py
Relm-Arrowny: sure, I was thinking just add the comment.

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
Relm-Arrowny: I would just make it a sigle SignalRW `epics_signal_rw(float, read_pv = read, write_pv = write)` unless there is a type mismatch. 
But that beside the point, I am just staying we should make speed as `config_signal` because we only need to read it once at scan start and not every point.

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
oliwenmandiamond: Done

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+    return math.degrees(math.atan2(math.hypot(x, z), y))
+
+
+@dataclass(kw_only=True)
+class MagnetPosition:
+    x: float
+    y: float
+    z: float
+
+    def to_spherical(self) -> "MagnetSphericalPosition":
```
Relm-Arrowny: We could make use of annotations here in line with everything else:
```suggestion
    def to_spherical(self) -> MagnetSphericalPosition:
```
with 
```python
from __future__ import annotations
```

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+    theta = math.degrees(math.atan2(-x, z))
+    return theta % 360
+
+
+def read_rho(x: float, y: float, z: float) -> float:
+    return math.hypot(x, y, z)
+
+
+def read_phi(x: float, y: float, z: float) -> float:
+    return math.degrees(math.atan2(math.hypot(x, z), y))
```
Relm-Arrowny: Double nits:
These are cal rather than read?

May be these should live within MagnetPosition, only scm use them no reason to have them in global name space?

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+        if target.y != 0:
+            raise MagnetPositionError(
+                f"Y must remain zero in XZ mode. Requested value was {target.y}"
+            )
+
+        hypot = math.hypot(target.x, target.z)
+        if hypot > self.LIMIT:
+            raise MagnetPositionError(
+                f"Outside XZ operating region. math.hypot(x={target.x}, y={target.y}) = {hypot}. Limit is {self.LIMIT}"
+            )
```
Relm-Arrowny:
```suggestion
        hypot = math.hypot(target.x, target.z)
        if hypot > self.LIMIT:
            raise MagnetPositionError(
                f"Outside XZ operating region. math.hypot(x={target.x}, z={target.z}) = {hypot}. Limit is {self.LIMIT}"
            )
```

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+            ("y", current.y, target.y),
+        ]
+
+        for axis, old, new in decreases:
+            if abs(new) < abs(old):
+                kwargs = {axis: new}
+                steps.append(MagnetStep(**kwargs))
+
+        if any(abs(new) > abs(old) for _, old, new in decreases):
+            steps.append(MagnetStep(x=target.x, y=target.y, z=target.z))
```
Relm-Arrowny: We are still moving all the axis to negative one by one first, I thank we could at least move all negative in one step?
It is worth checking if we actually need to move one by one in negative direction, I think the idea of sphere mode with lower vector limit is to allow async the movement and speed for all 3 axis so that we can keep the field direction. 
We can probably make it in a new issue as spherical is actually out of scope,

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+class MovementStrategy(ABC):
+    @abstractmethod
+    def moves(
+        self, current: MagnetPosition, target: MagnetPosition
+    ) -> list[MagnetStep]:
+        """Return a sequence of safe moves."""
+
+
+class SphericalMovement(MovementStrategy):
+    def moves(
```
Relm-Arrowny: We should check the limit which is R =< 1.75 for sphere mode.

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+
+class PlanarXZMovement(MovementStrategy):
+    LIMIT = 2.0
+
+    def moves(
+        self, current: MagnetPosition, target: MagnetPosition
+    ) -> list[MagnetStep]:
+        if target.y != 0:
+            raise MagnetPositionError(
+                f"Y must remain zero in XZ mode. Requested value was {target.y}"
```
Relm-Arrowny:
```suggestion
                f"Y must remain zero in XZ mode. Requested value was {target.y}"
```

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+        return [MagnetStep(x=target.x, y=target.y, z=target.z)]
+
+
+class PlanarXZMovement(MovementStrategy):
+    LIMIT = 2.0
+
+    def moves(
+        self, current: MagnetPosition, target: MagnetPosition
+    ) -> list[MagnetStep]:
+        if target.y != 0:
```
Relm-Arrowny: Probably should check y is actually zero
```suggestion
        if target.y != 0 or not math.isclose(current.y, 0.0, abs_tol=1e-4):
```

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+    direction through positive quadrant to 2.0 T in the +x direction. At all times
+    | B | <= 2.0 T. Ramp field back to zero in x direction.
+    """
+
+    LIMIT = 2.0
+
+    def moves(
+        self, current: MagnetPosition, target: MagnetPosition
+    ) -> list[MagnetStep]:
+        if target.z != 0:
```
Relm-Arrowny: probably should check current z is zero.

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+            elif value != 0:
+                raise MagnetPositionError(
+                    f"{axis} must remain zero in {self.axis.upper()} mode. Requested value was {value} "
+                )
+
+        return [MagnetStep(**{self.axis: coords[self.axis]})]
+
+
+class QuadrantXYMovement(MovementStrategy):
+    """Start from zero field in all axes. Appy 2.o T in + y direction. Move field
```
Relm-Arrowny:
```suggestion
    """Start from zero field in all axes. Appy 2.0 T in + y direction. Move field
```

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+from dataclasses import dataclass
+
+
+def read_theta(x: float, z: float) -> float:
+    theta = math.degrees(math.atan2(-x, z))
+    return theta % 360
+
+
+def read_rho(x: float, y: float, z: float) -> float:
+    return math.hypot(x, y, z)
```
Villtord: Why do you need to rename a function? Just to satisfy type checking?

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,187 @@
...
+        steps: list[MagnetStep] = []
+
+        # Step 1: remove X component first
+        if current.x != 0:
+            steps.append(MagnetStep(x=0))
+
+        # Step 2: move Y
+        if current.y != target.y:
+            steps.append(MagnetStep(y=target.y))
+
```
Relm-Arrowny: should move y first:

src/dodal/devices/beamlines/i06_1/magnets/ramp_controller.py
```diff
@@ -0,0 +1,76 @@
...
+
+    Exposes the readback ramp rate, demand ramp rate and the maximum
+    permitted ramp rate for one magnet axis.
+    """
+
+    def __init__(self, prefix: str, name: str = ""):
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.readback = epics_signal_rw(float, prefix + "STS:RAMPRATE:TPM")
+        self.demand = epics_signal_rw(float, prefix + "SET:DMD:RAMPRATE:TPM")
+        self.limit = epics_signal_rw(float, prefix + "LIM:RAMPRATE:TPM")
```
Relm-Arrowny: Only demand is writeable I am not sure what would happen if you try to write to a read only epics.
```suggestion
            self.readback = epics_signal_r(float, prefix + "STS:RAMPRATE:TPM")
        self.demand = epics_signal_rw(float, prefix + "SET:DMD:RAMPRATE:TPM")
        self.limit = epics_signal_r(float, prefix + "LIM:RAMPRATE:TPM")
```

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+
+    @AsyncStatus.wrap
+    async def kickoff(self):
+        fly_info = error_if_none(
+            self._fly_info,
+            f"{self.name} must be prepared before attempting to kickoff.",
+        )
+        self._fly_status = self.set(fly_info.end_position)
+        self._fly_info = None
+
```
Relm-Arrowny, line 134: A small issue here, we may need to add a ramp up time otherwise we will measure quite a few points at the starting position. e.g. the time to heat up the switch to inject current. 
This is more of an annoyance than a problem, may be an issue for an other day.

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            self.mode = epics_signal_rw(MagnetModes, prefix + "MODE")
+
+        self.ramp_status = epics_signal_rw(MagnetRampStatus, prefix + "RAMPSTATUS")
+        self.limit_status = epics_signal_rw(MagnetLimitStatus, prefix + "LIMITSTATUS")
+        self.start_ramp = epics_signal_x(prefix + "STARTRAMP.PROC")
+        self.timeout = 600
+
+        super().__init__(name)
```
Relm-Arrowny: This may seem long but ramp rate can be as low as 0.22T/mins for just 1 axis from 0 to 2T will already take ~550s.
I think we know the ramp rate, target and how it move, we should be calculating the timeout.

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+                prefix, ramp_controllers, self.set_within_boundary
+            )
+            self.sph = MagnetSphericalCoordinates(self.cart, self.set_within_boundary)
+
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            self.mode = epics_signal_rw(MagnetModes, prefix + "MODE")
+
+        self.ramp_status = epics_signal_rw(MagnetRampStatus, prefix + "RAMPSTATUS")
+        self.limit_status = epics_signal_rw(MagnetLimitStatus, prefix + "LIMITSTATUS")
+        self.start_ramp = epics_signal_x(prefix + "STARTRAMP.PROC")
```
Relm-Arrowny: deprecated?

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+        super().__init__(name)
+
+    async def _ramp(self):
+        # Setting invalid demand values should put the limit status in violation state.
+        # Block ramp if in violation state.
+        limit_status = await self.limit_status.get_value()
+        if limit_status == MagnetLimitStatus.VIOLTATION:
+            raise MagnetPositionError(f"{self.limit_status.name} is at {limit_status}")
+        self.log.info("About to start ramping the magnet.")
+        await self.start_ramp.trigger(timeout=self.timeout)
```
Relm-Arrowny: We do not want a 600s timeout on triggering.
```suggestion
        await self.start_ramp.trigger()
```

src/dodal/devices/beamlines/i06_1/magnet/movement.py
Relm-Arrowny, line 1: I have a few questions:
-  Each mode specifies limits for both field strength B and ramp rate B/Min. Should we validate ramp speeds, or can we rely on EPICS to handle mode-based limits automatically?
- EPICS populates limits when a mode is selected, it might be safer to read these limits directly from the hardware rather than hardcoding them here.
- I think GDA  ignored modes, hence the multi-step movement sequence(CHECK). For hard modes like SPHERICAL or QUADRANT_XY, breaking moves into manual intermediate steps may be unhelpful as you can achieve that with box mode with higher B field. May be we should leave those mode for a dedicated follow-up issue.

Relm-Arrowny: P.S. I am assuming the manual is right about how the magnet behave.

src/dodal/devices/beamlines/i06_1/magnets/movement.py
oliwenmandiamond: Yes, ophyd-async requires the function to be fully typed for derived signals. I have just tried to use math.hypot directly and can see we get the following errors:
```
ERROR tests/devices/beamlines/i06_1/magnets/test_superconducting_magnet.py::test_scmc_magnet_mode_to_movement_strategy_configuration[UNIAXIAL_Z-UniaxialMovement] - TypeError: <built-in function hypot> does not have a type hint for it's return value
ERROR tests/devices/beamlines/i06_1/magnets/test_superconducting_magnet.py::test_scmc_magnet_mode_to_uniaxial_movement_strategy_configuration - TypeError: <built-in function hypot> does not have a type hint for it's return value
ERROR tests/devices/beamlines/i06_1/magnets/test_superconducting_magnet.py::test_scmc_executes_movement_stragey_and_ramp_at_each_step - TypeError: <built-in function hypot> does not have a type hint for it's return value
======================== 44 passed, 46 errors in 2.97s =========================
```

src/dodal/devices/beamlines/i06_1/magnets/enums.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class MagnetRampStatus(StrictEnum):
+    RAMP_MADE = "RAMP MADE"
+    RAMPING = "RAMPING"
+
+
+class MagnetLimitStatus(StrictEnum):
+    OK = "OK"
+    VIOLTATION = "VIOLATION"
```
Relm-Arrowny: Typo
```suggestion
    VIOLATION = "VIOLATION"
```

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+        self._fly_status = self.set(fly_info.end_position)
+        self._fly_info = None
+
+    def complete(self) -> WatchableAsyncStatus:
+        fly_status = error_if_none(self._fly_status, f"{self.name} kickoff not called.")
+        self._fly_status = None
+        return fly_status
+
+
+class MagnetCartesianCoorindates(StandardReadable, Movable[MagnetPosition]):
```
Relm-Arrowny: typo
```suggestion
class MagnetCartesianCoordinates(StandardReadable, Movable[MagnetPosition]):
```

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
Relm-Arrowny, line 1: I am not 100% but reviewing it I have a burning question in my head:
Should we make MagnetAxis a StandardReadable where it only held signal for a axis.  Leaving all Movable functionality to th controller.
- Moving any single axis affects vector field, so every move requires all axis validation and this is best be done at the controller-level. 
- Implementing Flyable on individual axes works for the uniaxial mode but if I want to fly in B or in angle it can only be done with controller co-ordination so we may as well move the flying into the controller. where it mode has its own MovableLogic to slot in and out.
- Having .set() on individual axes invites someone to move those axis at the same time which would result in a race via set_within_boundary().

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+
+        Any coordinates left as ``None`` retain their current values. The requested
+        target position is validated against the current magnet operating mode and
+        converted into a sequence of movement steps by the corresponding movement
+        strategy. Each step is then applied sequentially until the requested target
+        position is reached.
+        """
+        if x is None and y is None and z is None:
+            raise MagnetPositionError("x, y, and z cannot all be None.")
+        current = MagnetPosition(
```
Relm-Arrowny: could read together like you did a few line up?
```python
  x, y, z = await asyncio.gather(
            self._cart.x.readback.get_value(),
            self._car.y.readback.get_value(),
            self._car.z.readback.get_value(),
        )
```

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+            x=await self.cart.x.readback.get_value(),
+            y=await self.cart.y.readback.get_value(),
+            z=await self.cart.z.readback.get_value(),
+        )
+        target = MagnetPosition(
+            x=current.x if x is None else x,
+            y=current.y if y is None else y,
+            z=current.z if z is None else z,
+        )
```
Relm-Arrowny: This is not ideal, we are effectively setting our target to our readback:
- This cause unnecessary movement.
- Setpoint creep, if the readback is ever so slightly offset, e.g. setpoint = 1T readback is 0.9999, we will keep creeping to lower field.
- We could also violate limit because the readback is not exactly 0 and we trying to an axis with limit of 0.

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+            self._cart_ref().x.readback.get_value(),
+            self._cart_ref().y.readback.get_value(),
+            self._cart_ref().z.readback.get_value(),
+        )
+        current = MagnetPosition(x=x, y=y, z=z).to_spherical()
+        spherical_pos = MagnetSphericalPosition(
+            rho=current.rho if rho is None else rho,
+            theta=current.theta if theta is None else theta,
+            phi=current.phi if phi is None else phi,
+        )
```
Relm-Arrowny: Same creeping issue here, we are setting setpoint to readback.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+    def __init__(
+        self,
+        cart: MagnetCartesianCoorindates,
+        set_mag_within_boundary: MagnetMoveWithinBoundary,
+        name: str = "",
+    ):
+        with self.add_children_as_readables():
+            self.theta = derived_signal_rw(
+                read_theta, self._set_theta, x=cart.x, z=cart.z
+            )
```
Relm-Arrowny, line 208: These allow you to set the angle of the field with rho = 0, which has no really meaning and defaulting to zero, we should raise exception when user try, so they do not try set the angle first and raise the field, 
Or may be we should cache it so they can do that.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -0,0 +1,396 @@
...
+            self.rho = derived_signal_rw(
+                read_rho, self._set_rho, x=cart.x, y=cart.y, z=cart.z
+            )
+            self.phi = derived_signal_rw(
```
Relm-Arrowny, line 212: same issue as theta

Relm-Arrowny (changes requested):
Overall this is looking super solid—I just have one  question regarding moving StandardMovable to the controller level, along with a couple of small edge-case notes.

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: Done

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: The current GDA code does this at the moment. If I move x, it will use the current positions of y and z and set the values to them again. Agreed maybe not ideal so I'll see if I can improve it

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
oliwenmandiamond: > I am not 100% but reviewing it I have a burning question in my head: Should we make MagnetAxis a StandardReadable where it only held signal for a axis. Leaving all Movable functionality to th controller.
> 
>     * Moving any single axis affects vector field, so every move requires all axis validation and this is best be done at the controller-level.
> 

So the issue with having every move be on the controller is that we cannot have any type checking for the move because it might be a single value for `x, y, z, theta, rho, phi` or a `MagnetPosition` or a `MagnetSphericalPositon`. This will be really hard to manage.

I think this would also be confusing to the user if sometimes scmc accepts a group of positions and other time times a single value. Exposing multiple sets via sub devices keeps a reliable API which will always accept the same thing but will only work depending on the mode I think is a better solution. We could make this even easier by creating some `plans` that help with this too. 

>     * Implementing Flyable on individual axes works for the uniaxial mode but if I want to fly in B or in angle it can only be done with controller co-ordination so we may as well move the flying into the controller. where it mode has its own MovableLogic to slot in and out.

I think we can do both. We can make it easy to fly scan using a certain axis and we can also fly scan using a coordinate move of multiple axes via the cart or sph devices. However, this is outside our current scope so I think should be done later.

>     * Having .set() on individual axes invites someone to move those axis at the same time which would result in a race via set_within_boundary().

That is a good point and come up with a solution. I've added a blocker on set_within_boundary() if two moves happen at the same time. 

This is okay:
```python
    run_engine(mv(scmc.mode, MagnetModes.CUBIC))
    # Coordinated parallel move of axes submitted together is okay.
    run_engine(mv(scmc.cart, MagnetPosition(x=1, y=1, z=1)))
```
The below fails trying to move axes in parallel but not submitted together
```python
run_engine(mv(scmc.cart.x, 0.5, scmc.cart.y, 0.5))
```
Added a test to cover for this

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
oliwenmandiamond: Issue for another time

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: The issue is we then to calculate it for each axis and each step to get a valid timeout. I think I will just remove the timoout for now and then we improve this later as the change is too big

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: yep, updated

src/dodal/devices/beamlines/i06_1/magnets/movement.py
oliwenmandiamond: Updated

src/dodal/devices/beamlines/i06_1/magnets/movement.py
oliwenmandiamond: Done

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: I have updated this. We now submit partial positions (Magnet|Request / MagnetSphericalRequest) and if we to check the field limit, we combine the position we want to move to any that are None with current positions to calculate it

src/dodal/devices/beamlines/i06_1/magnets/superconducting_magnet.py
oliwenmandiamond: Should be fixed now

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+
+    @classmethod
+    def axis_must_be_zero(cls, mode: MagnetMode, axis: str, value: float):
+        return cls(
+            f"Axis {axis} must remain zero for mode {mode}. Requested value was {value}."
+        )
+
+
+class MovementStrategy:
+    def check_within_limits(
```
Villtord, line 50: NIT: You could include call to `check_within_limits` into `move_steps` as it sounds logical to be there, also can make `check_within_limits` abstract method as well here.

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,207 @@
...
+
+class MagnetPositionError(Exception):
+    @classmethod
+    def total_field_mag_outside_limit(
+        cls, mode: MagnetMode, limit: float, pos: MagnetPosition
+    ):
+        return cls(
+            f"Target field magnitude of {pos.field_magnitude} T exceeds "
+            f"limit of {limit} for mode {mode}. Requested position: {pos}."
```
Relm-Arrowny: Nit
```suggestion
            f"Target field magnitude of {pos.field_magnitude}T exceeds "
            f"limit of {limit} for mode {mode}. Requested position: {pos}."
```

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+    ) -> list[MagnetRequest]:
+
+        steps: list[MagnetRequest] = []
+        decreases = [
+            ("z", current_readback.z, target.z),
+            ("x", current_readback.x, target.x),
+            ("y", current_readback.y, target.y),
+        ]
+        # Do all decreasing of axes first as one step. This ensures we don't
+        # quench the magnet.
```
Relm-Arrowny, line 95: Should: Could we make a ticket and add to this comment to revisit this, the negative first moving restriction is not in the manual and we tried single step move and asked Dirk about it and we believe it should be safe to move everything at once.
```suggestion
        # Do all decreasing of axes first as one step. This ensures we don't
        # quench the magnet.
        # To do optimise magnet movement #issue. 
```

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+        increase_kwargs = {}
+        for axis, old, new in decreases:
+            if new is not None and abs(new) > abs(old):
+                increase_kwargs[axis] = new
+
+        if increase_kwargs:
+            steps.append(MagnetRequest(**increase_kwargs))
+
+        return steps
+
```
Relm-Arrowny, line 114: Should: There is no reason to loop x, y, z twice.


```suggestion
        decrease_kwargs = {}
        increase_kwargs = {}
        for axis, old, new in decreases:
            if new is not None:
                if abs(new) < abs(old):
                   decrease_kwargs[axis] = new
                if abs(new) > abs(old):
                   increase_kwargs[axis] = new

        if decrease_kwargs:
            steps.append(MagnetRequest(**decrease_kwargs))
        if increase_kwargs:
            steps.append(MagnetRequest(**increase_kwargs))

        return steps
```

src/dodal/devices/beamlines/i06_1/magnets/movement.py
```diff
@@ -0,0 +1,207 @@
...
+        self, current_readback: MagnetPosition, target: MagnetRequest
+    ) -> list[MagnetRequest]:
+        """Return a sequence of safe moves."""
+
+
+class SphericalMovement(MovementStrategy):
+    LIMIT = 1.75
+    MODE = MagnetMode.SPHERICAL
+
+    def check_within_limit(
```
Relm-Arrowny: Must: Matching the base class?
```suggestion
    def check_within_limits
```

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+            )
+
+    def move_steps(
+        self, current_readback: MagnetPosition, target: MagnetRequest
+    ) -> list[MagnetRequest]:
+        return [MagnetRequest(x=target.x, z=target.z)]
+
+
+@dataclass
+class UniaxialMovement(MovementStrategy):
```
Relm-Arrowny, line 161: could/maybe:  There is no real reason to check the other inactive axis, since you can only get here if you are in the correct mode so we could just.
```python
@dataclass
class UniaxialMovement(MovementStrategy):
    mode: MagnetMode
    limit: float

    def check_within_limits(
        self, current_readback: MagnetPosition, target: MagnetRequest
    ) -> None:
        axis = self.mode.axis_alias
        resolved_value = getattr(target.resolve_pos(current_readback), axis)

        if resolved_value is not None and abs(resolved_value) > self.limit:
            raise MagnetPositionError.axis_outside_limit(
                self.mode, self.limit, resolved_value, axis
            )

    def move_steps(
        self, current_readback: MagnetPosition, target: MagnetRequest
    ) -> list[MagnetRequest]:
        axis = self.mode.axis_alias
        return [MagnetRequest(**{axis: getattr(target, axis)})]
```

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+
+class PlanarXZMovement(MovementStrategy):
+    LIMIT = 2.0
+    MODE = MagnetMode.PLANAR_XZ
+
+    def check_within_limits(
+        self, current_readback: MagnetPosition, target: MagnetRequest
+    ) -> None:
+        if target.y is not None and target.y != 0:
+            raise MagnetPositionError.axis_must_be_zero(self.MODE, "y", target.y)
```
Relm-Arrowny, line 147: could/maybe: if you in xy mode y can not be anything else so may be we do not need to check.

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+    """
+
+    LIMIT = 2.0
+    MODE = MagnetMode.QUADRANT_XY
+
+    def check_within_limits(
+        self, current_readback: MagnetPosition, target: MagnetRequest
+    ) -> None:
+        if target.z is not None and target.z != 0:
+            raise MagnetPositionError.axis_must_be_zero(self.MODE, "z", target.z)
```
Relm-Arrowny: same as above if mode is correct no need to check inactive.

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+            steps.append(MagnetRequest(x=0))
+
+        # Step 2: move Y
+        if target.y is not None and current_readback.y != target.y:
+            steps.append(MagnetRequest(y=target.y))
+
+        # Step 3: move X to its final value
+        if target.x is not None and target.x != 0:
+            steps.append(MagnetRequest(x=target.x))
+
```
Relm-Arrowny: Should: There is no real reason to ramp x back to 0 first, that part is done when we switch mode. you can definitely move x-y together as long as you are within 2.0T and the ramp rate limit.

src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -0,0 +1,207 @@
...
+        if target.x is not None and current_readback.x != 0:
+            steps.append(MagnetRequest(x=0))
+
+        # Step 2: move Y
+        if target.y is not None and current_readback.y != target.y:
+            steps.append(MagnetRequest(y=target.y))
+
+        # Step 3: move X to its final value
+        if target.x is not None and target.x != 0:
+            steps.append(MagnetRequest(x=target.x))
```
Relm-Arrowny: Must: Missing checks to enforce positive quadrant boundaries . We should check x >= 0 and y >= 0 before allowing the move, we should also check H < 2.0T. If we do not do speed matching we must also do decrease first to stop it going beyond 2.0T during move.

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
```diff
@@ -0,0 +1,77 @@
...
+    Exposes the readback ramp rate, demand ramp rate and the maximum
+    permitted ramp rate for one magnet axis.
+    """
+
+    def __init__(self, prefix: str, name: str = ""):
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.readback = epics_signal_r(float, prefix + "STS:RAMPRATE:TPM")
+        self.demand = epics_signal_rw(float, prefix + "SET:DMD:RAMPRATE:TPM")
+        self.ramp_limit = epics_signal_r(float, prefix + "LIM:RAMPRATE:TPM")
+        self.axis_limit = epics_signal_r(float, prefix + "LIM:FIELD:NOW")
```
Relm-Arrowny: Should: Axis_limit belong to MagnetAxis.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -0,0 +1,394 @@
...
+
+    async def move(self, new_position: float, timeout: TimeoutCalculator) -> None:
+        values = {self.axis_mode.axis_alias: new_position}
+        await self.magnet_set_within_boundary(MagnetRequest(**values))
+
+
+@default_mock_class(DeviceMock)
+# Don't want to sync readback and setpoint. IOC behaviour will move readback to demand
+# once start_ramp is triggered. This logic lives in the SuperConductingMagnetController.
+class MagnetAxis(StandardReadable, StandardMovable[float], Flyable, Preparable):
```
Relm-Arrowny, line 81: Should: I still think there is no good reason to make `MagnetAxis` moveable, as we should never allow anyone to move an axis bypass the controller. 

If we can do this: 
```python
yield from bps.scan(
    [detector], 
    scmc.cart, 
    MagnetRequest(x=0.0), 
    MagnetRequest(x=5.0), 
    num=100
)
```
Then I think every uniaxisal move should go via `MagnetCartesianCoordinates`,
otherwise an uniaxial device is probably better:
```python
class MagnetUniaxialAxis(
    StandardReadable, StandardMovable[float], Flyable, Preparable
):
    def __init__(self, controller: "SuperConductingMagnetController", name: str = ""):
        self._controller = controller
   @AsyncStatus.wrap
    async def set(self, value: float):
        mode = await self._controller.mode.get_value()
        if mode not in (MagnetMode.UNIAXIAL_X, MagnetMode.UNIAXIAL_Y, MagnetMode.UNIAXIAL_Z):
            raise RuntimeError("no")
        axis = mode.axis_alias  
        await self._controller.set_within_boundary(MagnetRequest(**{axis: value}))
```

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -0,0 +1,394 @@
...
+        with self.add_children_as_readables():
+            self.x = MagnetAxis(
+                prefix + "X:", MagnetMode.UNIAXIAL_X, ramp_rate.x, mag_within_boundary
+            )
+            self.y = MagnetAxis(
+                prefix + "Y:", MagnetMode.UNIAXIAL_Y, ramp_rate.y, mag_within_boundary
+            )
+            self.z = MagnetAxis(
+                prefix + "Z:", MagnetMode.UNIAXIAL_Z, ramp_rate.z, mag_within_boundary
+            )
```
Relm-Arrowny, line 169: we should make these private and no longer need to pass mag_within_boundary into the axis if we accepte axis should just be device readable.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -0,0 +1,394 @@
...
+        await self.set(value.start_position)
+
+    @AsyncStatus.wrap
+    async def kickoff(self):
+        fly_info = error_if_none(
+            self._fly_info,
+            f"{self.name} must be prepared before attempting to kickoff.",
+        )
+        self._fly_status = self.set(fly_info.end_position)
+        self._fly_info = None
```
Relm-Arrowny, line 133: Should: flyable should be move to the controller or it's subclass so mode is check before we try to fly we do not want to try to fly x when in y mode:

```python
@AsyncStatus.wrap
async def prepare(self, value: FlyMagnetInfo):
    mode = await self.mode.get_value()
    check what is our current mode
    do the prepare logic depend on mode.
```

Relm-Arrowny (dismissed):
The core logic for movement strategies and mode boundary enforcement looks great! I still feel we should make MagnetAxis a pure StandardReadable so we do not bypass the controller's safety logic and eliminates the need to inject mag_within_boundary callbacks into MagnetAxis.

I know you might not be fully convinced t about keeping MagnetAxis read-only and I probably did not explain it well. If you're busy or would prefer, I'm happy to take a stab at the branch to refactor MagnetAxis, and the controller to show you what I had in mind!

src/dodal/devices/beamlines/i06_1/magnets/movement.py
oliwenmandiamond: Nice spot!

src/dodal/devices/beamlines/i06_1/magnets/movement.py
oliwenmandiamond: Done

src/dodal/devices/beamlines/i06_1/magnet/movement.py
oliwenmandiamond: Thanks for optimising, done

src/dodal/devices/beamlines/i06_1/magnet/movement.py
oliwenmandiamond: The IOC probably does do that, but if with the off hand it slipped through probably best to double check it?

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
oliwenmandiamond: In epics it belongs to the power supply and is reflected with the PV. My goal is to model how it appears in epics / hardware so it refeclts it as much as possible. We shouldn't be providing two completely different PV's to a class. We should create two classes that resemble the PV structure and then combine them together with a composite device of the two classes.

Also as it is not used anywhere in the code, I will just remove it. If I did add it back, I would change this to be a power supply class which has the ramp rate class inside and then the additional pv signals such as the limit

src/dodal/devices/beamlines/i06_1/magnet/movement.py
oliwenmandiamond: Same as above

src/dodal/devices/beamlines/i06_1/magnet/movement.py
oliwenmandiamond: Same as above. I will just add though this is independent of the hardware so I think makes sense to leave it in as it doesn't make sense to use with Z being above zero.

src/dodal/devices/beamlines/i06_1/magnet/movement.py
oliwenmandiamond: I did this originally but then it means I have to remember to do this for every single strategy. Instead, it is an API and is now used by the magnet controller to call it once

src/dodal/devices/beamlines/i06_1/magnet/movement.py
oliwenmandiamond: Done

src/dodal/devices/beamlines/i06_1/magnet/movement.py
oliwenmandiamond: Done

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
oliwenmandiamond: Removed

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
oliwenmandiamond: I disagree on this. The current API the beamline expect is that they can use float values to move a single axis. Your suggestion will remove this. I think once we speak to beamline staff more, we can maybe do this in the future. Also you're scan above wouldn't work because bluesky has no idea how to generate points from 1 to 5 using `MagnetRequest` as it is not a float so would then also need to create lots of wrapper plans which is more effort than it is worth.

If a user really wanted to, they could bypass the controller no matter the design. They can set their own demand values and then ramp directly using the signal. If we make the API as user friendly as possible they wouldn't do this.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
oliwenmandiamond: As mentioned above, I disagree on this. We can revisit this in a future PR if needed.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
oliwenmandiamond: Same as above. This is too big and can done in another PR if needed.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
Relm-Arrowny: > I disagree on this. The current API the beamline expect is that they can use float values to move a single axis. Your suggestion will remove this. I think once we speak to beamline staff more, we can maybe do this in the future. Also you're scan above wouldn't work because bluesky has no idea how to generate points from 1 to 5 using `MagnetRequest` as it is not a float so would then also need to create lots of wrapper plans which is more effort than it is worth.

I think you missed the otherwise option in my previous comment! That is exactly why I suggested MagnetUniaxialAxis as a separate device class. It accepts standard float inputs and works out-of-the-box with bps.scan and bps.mv (and option 1 can work with scanspec too).

You are still missing my broader point about avoiding tight coupling. Passing controller callbacks down into individual MagnetAxis and MagnetCartesianCoordinates constructors tightly couples  child devices with controller orchestration. A  axis should not need to know about modes or controller safety rules, its sole job is to represent a  signal.

src/dodal/devices/beamlines/i06_1/magnet/ramp_controller.py
Relm-Arrowny: Removing the PV signal only treats the symptom, not the root cause.
We should use IOC live hardware limits via PVs like LIM:FIELD:NOW.And  basing boundary calculations on live limit signals ensures our movement strategies dynamically reflect the real state of the power supply and magnet hardware. The only hard code limit should be H(r) which are not available in epics.

oliwenmandiamond: I've created this issue https://github.com/DiamondLightSource/sm-bluesky/issues/351 so I will do this next in new PR

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
oliwenmandiamond: The controller doesn't know about modes of controller safety rules. It delegates to the controller for this

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
oliwenmandiamond: Another look at this, this actually done already. It uses the set method which delegates to the controller which will raise an error if in wrong mode e.g fly x but in mode y. I can add test for this to make it clearer.

src/dodal/devices/beamlines/i06_1/magnet/movement.py
Relm-Arrowny: add https://github.com/DiamondLightSource/sm-bluesky/issues/356

src/dodal/devices/beamlines/i06_1/magnet/movement.py
Relm-Arrowny: Only xy mode is not proected

Relm-Arrowny (approved):
Looks great! Thanks for pushing this update—the multi-axis movement strategy delegation and fly-scan protocols look super clean and well-structured.

Before hitting merge, I went ahead and added a quick minor safety tweaks:
    Default Timeout on _trigger_ramp(): Added a fallback safety timeout (timeout=500.0) when waiting for RAMP_MADE so Bluesky plans won't hang indefinitely if the IOC stalls.

Relm-Arrowny:
https://github.com/DiamondLightSource/sm-bluesky/issues/356
https://github.com/DiamondLightSource/dodal/issues/2166
#2166 is more important than the other.

Issue #1381, opened 2025-07-24
Clean up tests area


DominicOram:
Currently we have (at least) two overlapping structures for tests e.g:

* Tests for `/src/dodal/devices/i03/dcm.py` live in `tests/devices/unit_tests/i03/test_dcm.py`
* Tests for `/src/dodal/devices/i04/murko_results.py` live in `/tests/devices/i04/test_murko_results.py`

The original idea was that the `tests` directory structure should match the structure of the `src` directory and so not need random `unit_tests` folders inside. If we change this that's fine but it should be consistent.

## Acceptance Criteria
* Directory structure is consistent
* The structure is documented

PR #2121, opened 2026-07-16
add i10-1 and i06-1magnet temperature controller

### LINKED ISSUE #2113: I10-1 high field magnetic controllers
### Goals
- [x]  The three missing i10-1 temperature controllers are fully integrated and available for users to perform Bluesky operations.
- [x] Users can set target temperatures/PID and ramp rates via Bluesky plan commands.

### Tasks
- [x] Determine EPICS PV for missing devices.
- [x] Add those devices with test to i10-1.
- [ ] Reuse, PID in lakeshore if possible,

### Related Resources
should wait for: #2107 as it would be nice to have _within_tolerance.

### LINKED ISSUE #2124: Add support for I06-1 Magnet Temperature Controller
The I06-1 superconducting magnet (SCM) temperature controller is currently not implemented in DODAL, limiting access to temperature monitoring and control through via Bluesky.
## Acceptance Criteria
- [ ] A new SCMTemperatureController device exists in DODAL.

Relm-Arrowny:
Fixes #2113, #2124 

Requires:[ ophyd-async(#1381)](https://github.com/bluesky/ophyd-async/issues/1381). #2155 

## added a new generic temperature_controller:
- TemperatureController base class
- BaseTemperatureSensor and BaseHeater
- Shared PID component
- Temperature movable logic with configurable tolerance and stop behaviour
-  Support for multiple sensor readbacks
- Support switchable readback control sensor.
## Added support for the I06-1 SCM Cryocon M32 temperature controller:
- Heater configuration, ramp settings and PID controls

## Added support for the I10-1 High Field Magnet temperature controllers:
- New HighFieldMagnetTemperatureController device

## Fixed historical typo in module naming:
- temperture_controller → temperature_controller

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/beamlines/i06_1/magnet/temperature_controller.py
```diff
@@ -0,0 +1,108 @@
...
+
+        super().__init__(name=name)
+
+    def _create_config_signals(self, prefix: str, attr_prefix: str) -> None:
+        """Helper to generate all the config signals."""
+        suffixes = ["MIN", "MAX", "SLOPE", "OFFSET"]
+        for sfx in suffixes:
+            attr_name = f"{attr_prefix}_{sfx.lower()}"
+            signal = epics_signal_r(float, f"{prefix}:{sfx}")
+            setattr(self, attr_name, signal)
```
oliwenmandiamond: This is a good use case of a helper function, but this causes issues with static typing. Now for this device, you won't be able to find these signals on the class even though they exist which is something we should avoid. I think a much better solution to this is to create a this into another device you can reuse


```python
class CryoconM32Sensor(StandardReadable):

            self.sensor = epics_signal_r(float, prefix + "STS:T1")

             min = epics_signal_r(float, prefix + "MIN")
             max = epics_signal_r(float, prefix + "MAX")
             slope = epics_signal_r(float, prefix + "SLOPE")
             offset = epics_signal_r(float, prefix + "OFFSET")
        super().__init__(name)

```

Now you create sensor1 and sensor2. If you need to group them together, you can.

src/dodal/devices/beamlines/i10_1/high_field_magnet/temperature_controller.py
```diff
@@ -0,0 +1,77 @@
...
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            for sfx in config_suffixes:
+                signal_name = f"config_{sfx}"
+                if hasattr(self, signal_name):
+                    raise AttributeError(
+                        f"Cannot add configuration signal '{signal_name}': attribute already exists."
+                    )
+                signal = epics_signal_r(float, prefix + suffix + sfx)
+                setattr(self, signal_name, signal)
+        super().__init__(name=name)
```
oliwenmandiamond: Same as above, I don't think the loss of static type checking is worth it here and instead should be made from composition of devices

oliwenmandiamond (changes requested):
Thanks, looks good but have feedback to improve. As @Villtord mentioned in standup, could we please also make sure to add doc strings to the devices. Thanks!

src/dodal/devices/beamlines/i06_1/magnet/temperature_controller.py
```diff
@@ -0,0 +1,158 @@
...
+    ):
+
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.sensor = epics_signal_r(float, prefix + "STS:T1")
+            self.sensor2 = epics_signal_r(float, prefix + "STS:T2")
+        with self.add_children_as_readables():
+            self.sensor_config = CryoconM32SensorConfig(prefix + "STS:T1")
+            self.sensor2_config = CryoconM32SensorConfig(prefix + "STS:T2")
+
+        super().__init__(name=name)
```
oliwenmandiamond: At the moment you have the sensor and the sensor_config separate and makes it awkward. I think if we did it like this

```python
class CryoconM32Sensor(StandardReadable):
            self.sensor = epics_signal_r(float, prefix + "STS:T1")

            self.min = epics_signal_r(float, prefix + "MIN")
            self.max = epics_signal_r(float, prefix + "MAX")
            self.slope = epics_signal_r(float, prefix + "SLOPE")
            self.offset = epics_signal_r(float, prefix + "OFFSET")
        super().__init__(name)

class CryoconM32DualSensor(BaseTemperatureSensor):
    """Cryocon M32 dual-channel temperature sensor.
    ...
    """
            self.sensor1 = CryoconM32Sensor(prefix + "STS:T1")
            self.sensor2 = CryoconM32Sensor(prefix + "STS:T2")
```
It groups much more logically.

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,192 @@
...
+                    ", ".join(f"'{s}'" for s in available) if available else "none"
+                )
+                raise ValueError(
+                    f"Invalid readback target '{sensor_name}'. "
+                    f"Target must be exactly 'sensor' or 'sensor' followed by an integer (e.g., 'sensor2'). "
+                    f"Available sensors on {self.name}: [{available_sensors}]"
+                )
+            LOGGER.info(f"Setting active sensor on {self.name} to: '{sensor_name}'")
+
+        self._active_sensor_name = sensor_name
```
oliwenmandiamond: So this looks to be exactly the same use case i09 had where we have two "things" and then you want a device to coordinate which "thing" you have selected and give give read and set values to be dynamically set depending on which one selected. So I think we should reuse this here

We create `sensor1` and `sensor2` independently and then create the `dual_temperature_sensor` device which holds reference to `sensor1` and `sensor2`. You then use derived signals to wire up the logic to coordinate the set/readback if `sensor1` selected or `sensor2`. The selected signal can live inside the class if its only place it is needed but if its needed multiple places for coordination, then it must be an external device with reference (this is i09 case as if we select dcm source, we need to also select the dcm source shutter and vise versa with pgm).

Can you please take a look at https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/beamlines/i09.py at devices `source_selector`, `dual_energy_source`, and `dual_fast_shutter`and their correspond logic to see what I mean?

oliwenmandiamond (changes requested):

src/dodal/devices/beamlines/i06_1/magnet/temperature_controller.py
Villtord: Btw there seem to be too many 
`(prefix + "STS:T1")`

src/dodal/devices/temperature_controller/base_temperature_controller.py
Relm-Arrowny: I have kept the sensor selection self-contained for now since the selection should only happen via the controller/plan, avoiding the overhead of managing soft/derived signals. 
It also makes `BaseTemperatureSensor` much more flexible we aren't limited to just two sensors, so we can dynamically scale to as many channels as we need (sensor, sensor2, sensor3, etc.) without updating enums or derived signal logic.

src/dodal/beamlines/i06_1.py
```diff
@@ -34,3 +37,8 @@ def xabs() -> XYThetaStage:
 @devices.factory()
 def dd() -> DiffractionDichroism:
     return DiffractionDichroism(f"{PREFIX.beamline_prefix}-EA-DDIFF-01:")
+
+
+@devices.factory()
+def scm_temperature_controller() -> SCMTemperatureController:
+    return SCMTemperatureController(prefix=f"{PREFIX.beamline_prefix}-EA-TCTRL-01:")
```
oliwenmandiamond:
```suggestion
@devices.factory()
def scm_temp_controller() -> SCMTemperatureController:
    return SCMTemperatureController(prefix=f"{PREFIX.beamline_prefix}-EA-TCTRL-01:")
```
Make it a little shorter so easier to type

src/dodal/devices/beamlines/i06_1/magnet/temperature_controller.py
```diff
@@ -0,0 +1,169 @@
...
+            self.slope = epics_signal_r(float, prefix + ":SLOPE")
+            self.offset = epics_signal_r(float, prefix + ":OFFSET")
+        super().__init__(name)
+
+    def set_name(self, name: str, *, child_name_separator: str | None = None) -> None:
+        super().set_name(name, child_name_separator=child_name_separator)
+        self.sensor.set_name(name)
+
+
+class SCMSensor(BaseTemperatureSensor):
```
oliwenmandiamond: SCM doesn't mean anything unless you are familiar with the acronyms. I would prefer 
`SuperConductingMagnetTemperatureSensor` so it is explicit what it is and then when creating instances for the beamline, it is then shortened using the acronyms 

e.g 
```python
scm_temp_sensor = SuperConductingMagnetTemperatureSensor(...)
```

src/dodal/devices/beamlines/i06_1/magnet/temperature_controller.py
```diff
@@ -0,0 +1,169 @@
...
+            )
+            self.mode = epics_signal_rw(HeaterMode, f"{prefix}DMD:{infix}LOOPTYPE")
+
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.output = epics_signal_r(float, f"{prefix}STS:{infix}HTRREAD")
+
+        super().__init__(name=name)
+
+
+class SCMTemperatureController(TemperatureController):
```
oliwenmandiamond: Same again here.

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,192 @@
...
+    async def stop(self):
+        """Stop any active temperature move by setting the setpoint to the current readback value."""
+        current_val = await self.readback.get_value()
+        LOGGER.info(
+            f"Stopping temperature move: setting setpoint to current readback value ({current_val})"
+        )
+        await self.setpoint.set(current_val)
+
+
+class PID(StandardReadable):
```
oliwenmandiamond, line 44: Again with the above, I think this should be 

```python
class ProportionalIntegralDerivative(StandardReadable):
```

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,192 @@
...
+        super().__init__(name=name)
+
+    @property
+    def movable_logic(self) -> TemperatureMovableLogic:  # type: ignore[override]
+        """Readback needed to be the active sensor, hence not cached_property."""
+        return TemperatureMovableLogic(
+            setpoint=self.user_setpoint,
+            readback=self.sensor.active_sensor,
+            tolerance=self.tolerance,
+        )
```
oliwenmandiamond: Must: `active_sensor` is type `SignalR[float]`. Therefore this should be a derived_signal and then it reads all of the possible sensors and then returns the one that is selected / active. Then we don't create a new instance of movable logic every single time it is ever called.

src/dodal/devices/beamlines/i10_1/high_field_magnet/temperature_controller.py
```diff
@@ -0,0 +1,97 @@
...
+            self.setpoint = epics_signal_rw(float, prefix + "MANV:SET")
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.output = epics_signal_r(float, prefix + "HEATERP")
+
+        super().__init__(name=name)
+
+
+class HFMTemperatureController(TemperatureController):
+    """Temperature controller for the High-Field Magnet.
+
```
oliwenmandiamond, line 55: Again as above, would prefer explicit 
```python
class HighFieldMagnetTemperatureController(TemperatureController):
```

Then have the instance be shortened name. It also keeps it consistent with naming of the classes above. Makes much easier for someone new or unfamiliar to understand the beamline than being overloaded with new acronyms everywhere.

src/dodal/devices/temperature_controller/base_temperature_controller.py
oliwenmandiamond: Yep, okay makes sense. I have one issue with how it is done with `movable_logic` though which needs to be addressed.

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self._active_sensor_name = sensor_name
+
+
+SensorT = TypeVar("SensorT", bound=BaseTemperatureSensor)
+HeaterT = TypeVar("HeaterT", bound=BaseHeater)
+
+
+class TemperatureController(
+    StandardReadable, StandardMovable, Generic[SensorT, HeaterT]
```
oliwenmandiamond:
```suggestion
    StandardReadable, StandardMovable[float], Generic[SensorT, HeaterT]
```

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,192 @@
...
+    Attributes:
+        setpoint: Signal controlling the heater power setpoint.
+        output: Read-only signal indicating current heater output level.
+    """
+
+    setpoint: SignalRW[float]
+    output: SignalR[float]
+
+
+class BaseTemperatureSensor(StandardReadable, Movable):
```
oliwenmandiamond:
```suggestion
class BaseTemperatureSensor(StandardReadable, Movable[float]):
```

oliwenmandiamond (changes requested):
Looks good, thanks. Some comments on class naming which I think needs to be updated to make support easier long term and `movable_logic` needs to be addressed.

src/dodal/devices/temperature_controller/base_temperature_controller.py
Relm-Arrowny: I think this one may be too far, PID is industry standard term in hardware control and beamline systems. Expanding it to ProportionalIntegralDerivative feels overly verbose without adding anything.

src/dodal/devices/beamlines/i06_1/magnet/temperature_controller.py
```diff
@@ -0,0 +1,180 @@
...
+        name: Name of the controller instance. Defaults to "".
+    """
+
+    def __init__(
+        self,
+        prefix: str,
+        infix: str = "LOOP1:",
+        name: str = "",
+    ):
+
```
oliwenmandiamond:
```suggestion
    ):
```

src/dodal/devices/beamlines/i10_1/high_field_magnet/temperature_controller.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    def __init__(
+        self,
+        prefix: str,
+        suffix: str = "STEMP",
+        name: str = "",
+    ):
+
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
```
oliwenmandiamond:
```suggestion
    ):
        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
```

src/dodal/devices/beamlines/i10_1/high_field_magnet/temperature_controller.py
```diff
@@ -0,0 +1,115 @@
...
+            self.sensor = epics_signal_r(float, f"{prefix}{suffix}")
+            self.sensor2 = epics_signal_r(float, f"{prefix}{suffix}2")
+            self.sensor3 = epics_signal_r(float, f"{prefix}{suffix}3")
```
oliwenmandiamond: I think this a perfect use case for a `DeviceVector`

You could then do

```python
self.sensor = DeviceVector(
    {
         1: epics_signal_r(float, f"{prefix}{suffix}",
         2: epics_signal_r(float, f"{prefix}{suffix}2",
         3: epics_signal_r(float, f"{prefix}{suffix}3",
    }
)
```

You can then very naturally pick which sensor you want as the active one. You can even make this purely driven by configuration too by making the DeviceVector sensor as an argument to the device and then the configuration drives how many sensors there are rather than hard coded. 

This is very similar to my use case with ScalerCard

https://github.com/DiamondLightSource/dodal/pull/2132/changes#diff-43f09dab391c98d08ff8f96f761264b5cb55864596578e9ee5d0454f8b067727R80

https://github.com/DiamondLightSource/dodal/pull/2132/changes#diff-514a943a12602afd75fca7073e88e3db60dea915dda9e99cda3a9ef252bb1bd3R40

src/dodal/devices/beamlines/i10_1/high_field_magnet/temperature_controller.py
```diff
@@ -0,0 +1,115 @@
...
+    def _select_sensor(
+        self, active_sensor_name: str, sensor: float, sensor2: float, sensor3: float
+    ) -> float:
+        match active_sensor_name:
+            case "sensor2":
+                return sensor2
+            case "sensor3":
+                return sensor3
+            case _:
+                return sensor
```
oliwenmandiamond: However, I think using a DeviceVector adds complication. I remember I tried doing something like this before:

```python
        self.active_sensor = derived_signal_r(
            raw_to_derived=self._select_sensor,
            active_sensor_name=self._active_sensor_name,
           # Need function to map device name to the device and use these as the args for function
           **{
                  self.sensor[0].name: self.sensor,
                  ...
             }
        )

    def _select_sensor(
        self, active_sensor_name: str, **kwargs: dict[str, float],
    ) -> float:
              return kwargs[active_sensor_name]
```
But I think derived signal doesn't like **kwargs. I might explore this and raise an issue in ophyd-async as would be handy if it was a little less restrictive. Can you just double check you can't do something like this?

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,182 @@
...
+    """Base interface for temperature sensors supporting dynamic readback targeting."""
+
+    sensor: SignalR[float]
+    active_sensor: SignalR[float]
+
+    def __init__(self, name: str = ""):
+
+        self._active_sensor_name = soft_signal_rw(str, initial_value="sensor")
+
+        super().__init__(name=name)
```
oliwenmandiamond:
```suggestion
    def __init__(self, name: str = ""):
        self._active_sensor_name = soft_signal_rw(str, initial_value="sensor")
        super().__init__(name=name)
```

tests/devices/beamlines/i10_1/high_field_magnet/test_temperature_controller.py
```diff
@@ -0,0 +1,57 @@
...
+    hfm_temp_controller: HighFieldMagnetTemperatureController,
+):
+
+    await assert_reading(
+        hfm_temp_controller,
+        {
+            "hfm_temp_controller-sensor-sensor": partial_reading(0.0),
+            "hfm_temp_controller-sensor-sensor2": partial_reading(0.0),
+            "hfm_temp_controller-sensor-sensor3": partial_reading(0.0),
+        },
```
oliwenmandiamond, line 37: If we could get `DeviceVector` working, it would flow a little more naturally  

```python
    await assert_reading(
        hfm_temp_controller,
        {
            "hfm_temp_controller-sensor-1": partial_reading(0.0),
            "hfm_temp_controller-sensor-2": partial_reading(0.0),
            "hfm_temp_controller-sensor-3": partial_reading(0.0),
        },
```

tests/devices/temperature_controller/test_base_temperature_controller.py
```diff
@@ -0,0 +1,157 @@
...
+    BaseTemperatureSensor,
+    TemperatureController,
+)
+
+
+class MinimalMockSensor(BaseTemperatureSensor):
+    def __init__(self, name: str = ""):
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.sensor = epics_signal_r(float, "prefix + suffix")
+            self.sensor2 = epics_signal_r(float, "prefix + suffix2")
```
oliwenmandiamond:
```suggestion
            self.sensor = epics_signal_r(float, "PV1")
            self.sensor2 = epics_signal_r(float, "PV2")
```
I would prefer something like this as it looks a little odd currently as I originally thought it was a typo

oliwenmandiamond (commented):
Looks good, couple comments. Can we test to see if DeviceVector would work with it? If it doesn't work, happy to approve. It probably won't work but is a very natural fit for this and if find any issues with it we can raise with ophyd-async and then revisit this later

oliwenmandiamond:
(We also need to up test coverage)

src/dodal/devices/beamlines/i10_1/high_field_magnet/temperature_controller.py
```diff
@@ -0,0 +1,105 @@
...
+            sensor_map = {"sensor1": ""}
+            sensor_map.update({f"sensor{i}": str(i) for i in range(2, 4)})
+        sensor = TemperatureSensor[HighFieldMagnetSensor](
+            DeviceMap(
+                {
+                    name_key: HighFieldMagnetSensor(f"{prefix}STEMP{pv_suffix}")
+                    for name_key, pv_suffix in sensor_map.items()
+                }
+            )
+        )
```
oliwenmandiamond, line 82: I personally dislike this sensor_map. It looks quite awkward when you could just construct `TemperatureSensor` class outside this class and pass it directly in with the DeviceMap already configured.

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,189 @@
...
+        setpoint: Signal controlling the heater power setpoint.
+        output: Read-only signal indicating current heater output level.
+    """
+
+    setpoint: SignalRW[float]
+    output: SignalR[float]
+
+
+class BaseTemperatureSensor(StandardReadable):
+    sensor: SignalR[float]
```
oliwenmandiamond: I think this should go further. For each class that uses it, you do the following:

```python
            self.sensor = epics_signal_r(float, pv)
```
So why not do that here in the base class?

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __init__(self, channel: DeviceMap[SensorDevice], name: str = ""):
+        if not channel:
+            raise ValueError(
+                f"Cannot initialize {self.__class__.__name__} with an empty DeviceMap."
+            )
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            self.active_sensor_name = soft_signal_rw(
+                str, initial_value=list(channel.keys())[0]
+            )
```
oliwenmandiamond, line 113: With latest ophyd-async version, you can validation now to soft signals https://github.com/bluesky/ophyd-async/issues/1119. I think a good use case here is to make it so it can only ever be set to keys of the configured channels. So it's basically an enum without using an enum.

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __init__(
+        self,
+        setpoint: SignalRW[float],
+        sensor: TemperatureSensor[SensorDevice],
+        heater: HeaterT,
+        pid: PID,
+        name: str = "",
+    ):
+
```
oliwenmandiamond:
```suggestion
    ):
```

oliwenmandiamond (commented):
Thanks, this looks much better now. I have couple more comments and then once ophyd-async blocker added, looks great. Thank you!

src/dodal/devices/beamlines/i10_1/high_field_magnet/temperature_controller.py
Relm-Arrowny: This is the i10 sub class, we are efficiently doing what you are saying, think of it as a helper function so we do not have to do this x times in i10 config:
```python
sensor = TemperatureSensor[HighFieldMagnetSensor](
    DeviceMap({
        "sensor1": BaseTemperatureSensor("BL01I-EA-HFM-01:STEMP"),
        "sensor2": BaseTemperatureSensor("BL01I-EA-HFM-01:STEMP2"),
        "sensor3": BaseTemperatureSensor("BL01I-EA-HFM-01:STEMP3"),
    })
)
hfm_temp = HighFieldMagnetTemperatureController(prefix="BL01I-EA-HFM-01:", sensor=sensor)
```

src/dodal/devices/temperature_controller/base_temperature_controller.py
```diff
@@ -0,0 +1,201 @@
...
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            self.tolerance = soft_signal_rw(float, initial_value=0.1)
+            self.user_setpoint = setpoint
+
+        super().__init__(name=name)
+
+    @cached_property
+    def movable_logic(self) -> TemperatureMovableLogic:
+        """Readback needed to be the active sensor, hence not cached_property."""
```
oliwenmandiamond: Doc string is out of date

```suggestion
```

oliwenmandiamond (approved):
LGTM, thanks for doing this

PR #2122, opened 2026-07-17
Update to ophyd-async v0.20.1

oliwenmandiamond:
Update to latest ophyd-async version v0.20.1.

### Summary of changes
- Replaces Bluesky protected attribute assignment in tests with `set_mock_attr`
- Fixes tests that failed due to signals not being named or devices not connected.

1.Check tests pass
2. Check AI conversions of setting mock bluesky verbs done correctly

### Checks for reviewer


coretl:
## ophyd-async 0.20.1 `set_mock_attr` fixes

Bumped the `ophyd-async[ca,pva]` pin to `>=0.20.1` (now resolved in `uv.lock`) and fixed the remaining reserved-attr test failures using the new `set_mock_attr` helper (imported from `ophyd_async.core` — it isn't re-exported from `ophyd_async.testing` in this release).

**Tests fixed with `set_mock_attr`** (direct `signal.set = mock`/`.read =`/`.trigger =`/etc. assignments replaced with `set_mock_attr(signal, "attr", mock)`):
- `tests/devices/beamlines/i03/test_beamsize.py`, `i03/test_undulator_dcm.py`, `i04/test_beamsize.py`
- `tests/devices/beamlines/i06_shared/test_i06_apple2controller.py`, `i09_2/test_i09_2_motors.py`, `i10/test_i10_apple2.py`, `i15_1/test_hexapod.py`, `i17/test_i17_apple2.py`, `i21/test_toolpoint_motion.py`
- `tests/devices/electron_analyser/base/test_detector_logic.py`
- `tests/devices/insertion_device/test_apple2_controller.py`, `test_apple2_undulator.py`, `test_energy.py`, `test_polarisation.py`
- `tests/devices/oav/conftest.py`, `tests/devices/oav/test_oav.py`
- `tests/devices/test_attenuator.py`, `test_bart_robot.py`, `test_common_mirrors.py`, `test_focusing_mirror.py`, `test_gridscan.py`, `test_smargon.py`, `test_zebra.py`, `test_zocalo_results.py`

Also removed the temporary `OPHYD_ASYNC_ALLOW_RESERVED_ATTRS=YES` autouse fixture from `tests/conftest.py` added in a previous commit — the `set_mock_attr` fixes above cover every case it was masking, confirmed by re-running the full suite with the mask removed (identical failure set before/after).

**Remaining failures left for a human** (all pre-existing, none addressable by `set_mock_attr`):
- 56 `tests/devices/beamlines/{i09_2_shared,i10}/test_i09_apple2.py` / `test_i10_apple2.py` cases — `TypeError: mock_config_client.<locals>.my_side_effect() missing 1 required positional argument: 'desired_return_type'`. This is a **second, still-broken duplicate `mock_config_client` fixture** in `src/dodal/testing/fixtures/devices/hard_undulator.py` (positional-only `my_side_effect(file_path, desired_return_type, reset_cached_result)`) — the "Fix ConfigClient issue" commit only fixed the copy in `src/dodal/testing/fixtures/devices/apple2.py`. Likely just needs the same default-args treatment applied there.
- 1 `tests/devices/beamlines/i15_1/test_robot.py::test_puck_program_loaded_before_position_selected` — EPICS/IOC timeout, infra-related.

Full suite: 3268 passed, 58 failed (all the `mock_config_client` cases above), 1 skipped.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
Fixed a pyright regression from the `set_mock_attr` change: `set_mock_attr(device, "attr", mock)` doesn't let pyright narrow `device.attr` to the mock's type, so later `device.attr.assert_*(...)` calls were failing type-checking with `reportAttributeAccessIssue`. Fixed by capturing `set_mock_attr`'s return value (it returns the mock unchanged, per its own docstring) and asserting on that instead, across the 14 affected test files. `pyright` and the full test suite are clean again (same 58 pre-existing `mock_config_client` failures as before, nothing new).

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
CI is fully green now — the "Standardise config client" merge (#2099) replaced the duplicate `mock_config_client` fixture mechanism and fixed the remaining 56/58 failures I'd flagged as out-of-scope. Confirmed locally on `6b424b6`: 3375 passed, 1 skipped, 0 failed; pre-commit, ruff, and pyright all clean. All GitHub Actions checks (lint, tests × 4 Python versions, docs, dist, codecov) are passing.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/dodal/devices/insertion_device/apple2_controller.py
```diff
@@ -132,11 +132,10 @@ def __init__(
             btm_outer = phase.btm_outer.user_readback
         else:
             # If locked phase axes make the locked phase 0.
-            top_inner = btm_outer = soft_signal_rw(float, initial_value=0.0)
+            top_inner = btm_outer = 0.0
```
DominicOram, line 137: Should: Why are we doing this now? It seems like a bad pattern for these to be different types depending on what type `phase` is. It means that any downstream code trying to read these values has to handle the fact that they may be different types.

tests/devices/beamlines/i10_1/high_field_magnet/test_high_field_magnet.py
```diff
@@ -36,11 +37,9 @@ async def test_locate(high_field_magnet: HighFieldMagnet):
 
 
 async def test_stop_success(high_field_magnet: HighFieldMagnet):
-    set_mock_value(high_field_magnet.user_readback, 7.5)
-    set_mock_value(high_field_magnet.user_readback, 1.5)
+    initial_val = await high_field_magnet.user_readback.get_value()
     await high_field_magnet.stop()
-    assert high_field_magnet._set_success is False
-    assert await high_field_magnet.user_setpoint.get_value() == 1.5
+    get_mock_put(high_field_magnet.user_setpoint).assert_awaited_once_with(initial_val)
```
Relm-Arrowny: I think the intention of this test is to verify that calling stop() resets the setpoint to the current readback value. However, both user_readback and user_setpoint default to 0 initially so current this test will always pass.

```suggestion
    set_mock_value(high_field_magnet.user_setpoint, 7.5)
    set_mock_value(high_field_magnet.user_readback, 1.5)
    await high_field_magnet.stop()
    get_mock_put(high_field_magnet.user_setpoint).assert_awaited_once_with(1.5)
```

DominicOram (changes requested):
Thanks, generally looks good. Just the one query

src/dodal/devices/insertion_device/apple2_controller.py
oliwenmandiamond: We need `top_inner` and `btm_outer` as inputs to the derived signal polarisation. For unlocked phase axes, these are real signals (`SignalR`) that need to be read by the derived signal.

For locked phase axes, `top_inner` and `btm_outer` don't exist as hardware signals, and their values are always zero for the polarisation calculation. In that case, we don't need to create soft signals just to represent a constant value. `derived_signal_rw` accepts both signals and constants as inputs, so using 0.0 here is intentional. The downstream code here is only the derived signal which handles constants and signals.

The soft signals can't simply be added as children of this class because, when the phase axes are unlocked, those signals already belong to the phase device and therefore already have a parent. Creating another signal would also be unnecessary duplication.

This is now an issue for derived signals because it now raise error if they are not named due to `merged_gathered_dicts` and creating soft signals that are not part of the class is what caused the issue as they won't be automatically named.

I'll add a comment to make this distinction clearer.

Relm-Arrowny (commented):
Looks good overall, just one small thing to address before merging!

Could you also add a note to the PR description outlining the changes? 
Something like:
    "Replaces Bluesky protected attribute assignment in tests with set_mock_attr"

coretl:
Addressed both review notes:
- Added a "Summary of changes" section to the PR description per @Relm-Arrowny's request.
- Fixed `test_stop_success` in `tests/devices/beamlines/i10_1/high_field_magnet/test_high_field_magnet.py` per @Relm-Arrowny's suggestion — `user_readback` was never set away from its default 0, so the test passed trivially regardless of `stop()`'s behaviour. Now sets `user_setpoint`/`user_readback` to distinct values so the assertion actually exercises the reset logic.

---
_Generated by [Claude Code](https://claude.ai/code)_

coretl:
FYI: `test (ubuntu-latest, 3.11)` failed once on `93b717b` with `test_given_program_not_running_and_pin_unmounting_but_new_pin_not_mounting_when_load_pin_then_timeout` (`AssertionError: Expected 'mock' to have been called once. Called 0 times.`), while 3.12/3.13/3.14 passed on the same commit and it passes consistently locally (5 isolated runs + full suite). Looks like pre-existing timing flakiness from this test's tight timeouts (`_set_fast_robot_timeouts`) rather than a real regression — not something to chase from this PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/dodal/devices/insertion_device/apple2_controller.py
DominicOram: > The downstream code here is only the derived signal which handles constants and signals.

Ok, in which case can you make them private?

> I'll add a comment to make this distinction clearer

I think if they're private there's no need for a comment, but up to you. Thanks!

DominicOram (approved):

src/dodal/devices/insertion_device/apple2_controller.py
oliwenmandiamond: These are not class attributes, so how can they be made private?

Relm-Arrowny: I think the point is that they aren't exposed to downstream code—they're just internal constants for an undulator without the extra phase axis. We previously used a derived_signal for these constants simply because we couldn't use a raw constant value before.
This is also probably not in scope.

DominicOram: Sorry, I'm being stupid, ignore me

tests/devices/beamlines/i10_1/high_field_magnet/test_high_field_magnet.py
oliwenmandiamond: I've improved the test to make it clearer

Issue #2123, opened 2026-07-17
Create I19 attenuation sub-systems


CoePaul:
Addresses #1007 
Creating the dodal device which serves as the entire transmission system

This splits naturally ( under the bonnet / hood ) into two sub-systems 
  as per the business logic requirements requested by the scientist.

https://confluence.diamond.ac.uk/x/4Iq4Fw

  Use either the wheel / aluminium wedge sub-system 
  or the resin wedge system ( or neither ).

Creating therefore each sub-system with these capabilities:

A sub-system which can:

 a) report what the sub-system attenuation would be for any input x-ray energy, motor configuration
  -- this should **not ignore invalid** options from business logic - but instead just calculate the attenuation
  -- it will then be able to report the live attenuation if it finds 
        the motors in a non-canonical state
  -- this same functionality is then perfectly primed to answer questions about 
       the current attenuation at the current energy and current motor positions
        because You just feed those into the function arguments
 
b) predict what the fastest-closest match attenuation (would be hypothetically), given
 -- a set of motor positions
 -- an x-ray energy
 - the energy and motor positions need not be the current ones - **the calculator should just calculate**

c) report back motor positions needed to achieve a given attenuation request

d) answer thresholding questions needed for the business logic as it decides which sub-systems to use

N.B. Given that the lingua franca of the attenuation business logic is the logarithmic Bn unit, 
  all attenuation calculation results and requests will be made in these units.  
 [ ( This is the whole reason that unit was invented in the first place )](https://confluence.diamond.ac.uk/x/U4c0EQ)
 

## Acceptance Criteria
* That an attenuation sub-system class exists with the above listed capabilities as demonstrated in tests

PR #2125, opened 2026-07-20
Tighten up I19 Attenuator related definitions

CoePaul:
Contributes to #2123

* Rename the attenuator motor position dataclass to make it clear it can be used for more than just setting demands

* Fix associated test which was brittle on the class name

* Replace looser PositiveInt type ( which coerces ) with the stricter Annotated StrictInt type for wheel index positions

* Adjust tests of Attenuator positions to catch up with these changes -- widen tests of invalid definitions using parametized testing

1. Confirm renaming of production code and corresponding test classes makes coherent sense
2. Ensure all CI tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
+++ b/src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
@@ -1,8 +1,15 @@
-from typing import Annotated, Any, Self
+from typing import Annotated, Self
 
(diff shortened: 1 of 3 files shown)


rtuck99 (approved):
Looks fine, I don't have particular opinons other than it feels like set() should call model_dump() instead of validate_complete_damnd() otherwise you're only getting half the benefit of using pydantic models. But that would require a change in the logic at both ends, so I'm not going to insist on it.

PR #2130, opened 2026-07-21
Add system absorber types

CoePaul:
Contributes to #2123 

1. Ensure changes to production code are properly documented with docstrings
2. Ensure changes to production code are properly covered in unit tests
3. Ensure variable and method names make sense and that the logic seems consistent
4. Confirm all tests pass in CI

### Checks for reviewer

### DIFF
--- a/src/dodal/common/general_maths/absorbers.py
+++ b/src/dodal/common/general_maths/absorbers.py
@@ -1,0 +1,141 @@
+from typing import Annotated, Protocol, runtime_checkable
+
+from pydantic import ConfigDict, Field, StrictFloat, validate_call
+from pydantic.dataclasses import dataclass
+
+from dodal.common.general_maths.absorber_geometry import (
+    TaperedGeometryProvider,
+    ThicknessProvider,
+)
+from dodal.common.general_maths.material_absorption_maths import AbsorptionCalculator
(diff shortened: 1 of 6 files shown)


src/dodal/common/general_maths/absorbers.py
```diff
@@ -0,0 +1,136 @@
+from typing import Annotated, Protocol, runtime_checkable
+
+from pydantic import ConfigDict, Field, StrictFloat, validate_call
+from pydantic.dataclasses import dataclass
+
+from .absorber_geometry import TaperedGeometryProvider, ThicknessProvider
```
noemifrisina: Should: use absolute instead of relative imports, makes things clearer in the long run and anyway we should keep it consisten with the rest of the repo

noemifrisina (changes requested):
Thanks, one comment in code.
Also please reword the PR title to something that explains what it changes

src/dodal/common/general_maths/absorbers.py
CoePaul: OK - I had tried to be "pythonic" but frankly, always prefer the full absolute path

noemifrisina (approved):
lgtm

PR #2132, opened 2026-07-22
I06 magnet scaler cards and detectors

### LINKED ISSUE #339: Add i06-1 scaler card channels used to measure during a fastfieldscan
In preparation for the flyscanning fastfieldscan on i06-1, we need the detectors that measure through out the flyscan. These are a scaler card device with multiple channels. I09 and i10 also use scaler cards so a solution should be devised that should (hopefully) be generic enough that each beamline can reuse.

# Acceptance criteria
- A controller that allows you update the collection time
- Multiple channels to read from 
- A controller that is triggerable that then starts the acquisition. The real IOC would then update the channel reads to be a new value from what they measured

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/sm-bluesky/issues/339

2. Check tests pass
3. Check dodal connect

### Checks for reviewer


oliwenmandiamond:
Currently blocked by https://github.com/bluesky/ophyd-async/pull/1382

src/dodal/devices/scaler_card.py
```diff
@@ -0,0 +1,116 @@
...
+    async def trigger(self):
+        """Start a scaler count and wait for it to complete."""
+        await self.start_count.set(True)
+        await wait_for_value(self.start_count, False, None)
+
+
+DeviceT = TypeVar("DeviceT", bound=Device)
+
+
+class ScalerCardChannels(
```
Relm-Arrowny: nit: I would call scalerCard, it is not just channels.
```suggestion
class ScalerCard(
```

src/dodal/devices/scaler_card.py
```diff
@@ -0,0 +1,116 @@
...
+    @AsyncStatus.wrap
+    async def set(self, value: float):
+        """Set the scaler integration time."""
+        await self.integration_time.set(value)
+
+    @AsyncStatus.wrap
+    async def trigger(self):
+        """Start a scaler count and wait for it to complete."""
+        await self.start_count.set(True)
+        await wait_for_value(self.start_count, False, None)
```
Relm-Arrowny: Should: Assuming `start_count`, goes to `False` when counting is done we should add a time out, if not we need to make sure we do not read before counting is done?
```suggestion
        timeout = await self.integration_time.get_value() + <some_default_overhead>
        await self.start_count.set(True)
        await wait_for_value(self.start_count, False, timeout)
```

src/dodal/devices/scaler_card.py
```diff
@@ -0,0 +1,116 @@
...
+class ScalerCardChannels(
+    StandardReadable, Triggerable, Movable[float], Generic[DeviceT]
+):
+    """Expose a logical group of channels from a scaler card.
+
+    The channels are exposed as a readable device and are associated with a
+    :class:`ScalerCardController` that controls the underlying scaler card. Multiple
+    instances of this class can share the same controller, allowing different groups of
+    scaler channels to be exposed as separate logical devices while being acquired by
+    the same physical scaler card. The channels are also added as readbales with the
```
Relm-Arrowny:
```suggestion
    the same physical scaler card. The channels are also added as readables with the
```

Relm-Arrowny (changes requested):

src/dodal/devices/scaler_card.py
oliwenmandiamond: The GDA has this

```
         private static final String START = ".CNT";
         ...
	@Override
	public int getStatus() throws DeviceException {
		if (epicsDevice != null) {
			return !waitingForPutToStart && (Short)epicsDevice.getValue(ReturnType.DBR_NATIVE, RECORD, START) == 0 ? Detector.IDLE
					: Detector.BUSY;
		}
		return Detector.IDLE;
```

So the status / detector is busy if the  start_count equals value 1 and is idle if this 0 so it does do this behaviour. I've added the default timeout which should be sufficient, we do not need to check that trigger is done before read

src/dodal/devices/scaler_card.py
oliwenmandiamond: Done

src/dodal/devices/scaler_card.py
```diff
@@ -64,25 +65,24 @@ async def set(self, value: float):
     async def trigger(self):
         """Start a scaler count and wait for it to complete."""
         await self.start_count.set(True)
-        await wait_for_value(self.start_count, False, None)
+        timeout = await self.integration_time.get_value() + DEFAULT_TIMEOUT
+        await wait_for_value(self.start_count, False, timeout)
```
Relm-Arrowny, line 69: Nit: default_timeout probably is overkill at 10,

Relm-Arrowny (approved):

PR #2138, opened 2026-07-28
Allow for all scan directions in the OAV parameters

noemifrisina:
Temporarily reverts #2092 and changes the OAVParameters direction in order to allow values different from 0 and 1 from the OAVCentring file, eg. I19-1 has a direction value of "2", which is "up".

This issues should be fixed properly in a separate PR.

1. Run tests

### Checks for reviewer


rtuck99 (approved):
Approved

PR #2151, opened 2026-07-31
Move shutter state to plan

oliwenmandiamond:
First part in fixing https://github.com/DiamondLightSource/sm-bluesky/issues/349

De tangle shutters and sequence file states from the electron analysers and move this logic to the analyserscan plans https://github.com/DiamondLightSource/sm-bluesky/pull/350

This makes the detectors much simpler and stateless, now the plan coordinates shutter and configuring the sequence file the detector per step per region. 

### Checks for reviewer


src/dodal/devices/selectable_source.py
```diff
@@ -23,19 +32,56 @@ def get_obj_from_selected_source(selected_source: SelectedSource, s1: T, s2: T)
...
+    def __init__(
+        self,
+        source1: SignalR[float],
+        source2: SignalR[float],
+        selected_source: SignalRW[SelectedSource],
+        name: str = "",
+    ):
+        self.selected_source_ref = Reference(selected_source)
+        self.source1_ref = Reference(source1)
+        self.source2_ref = Reference(source2)
```
Villtord: I am not sure why you need those references - they don't seem to be used anywhere in this class

Villtord (changes requested):
Looks fine for me except one comment.
Just to clarify you renamed/moved some classes and deleted some logic but it looks like it's a preparation step for a bigger change?

src/dodal/devices/selectable_source.py
oliwenmandiamond: Yep, good point. Have removed

oliwenmandiamond:
> Looks fine for me except one comment. Just to clarify you renamed/moved some classes and deleted some logic but it looks like it's a preparation step for a bigger change?

So I simplified the classes and moving a lot of the logic to the plan level. I also moved DualEnergySource out from the electron analyser module as it is not a hard dependency for the electron analyser

tests/devices/test_selectable_source.py
```diff
@@ -1,44 +1,129 @@
...
 from dodal.devices.selectable_source import (
+    DualEnergySource,
     SelectedSource,
-    SourceSelector,
     get_obj_from_selected_source,
 )
 
 
+def test_get_obj_from_selected_source_get() -> None:
+    obj1, obj2 = 1, 2
```
Villtord: I think it's better to use real objects rather than integers -

`obj1, obj2 = object(), object()`

 that will assure real object is passed in ` get_obj_from_selected_source` function rather than test integer value passed. For example this amended tested function below will pass your test, but clearly it's not correct:
```
def get_obj_from_selected_source(source, obj1, obj2):
    # BUG: ignores obj1/obj2, returns fresh literals instead
    if source == SelectedSource.SOURCE1:
        return 1
    if source == SelectedSource.SOURCE2:
        return 2
```

src/dodal/devices/selectable_source.py
```diff
@@ -23,19 +31,53 @@ def get_obj_from_selected_source(selected_source: SelectedSource, s1: T, s2: T)
...
+                "eV",
+                selected_source=selected_source,
+                s1=source1,
+                s2=source2,
             )
+
+        self._validate_config_signal([source1, source2])
+        with self.add_children_as_readables(StandardReadableFormat.CONFIG_SIGNAL):
+            self.source1, _ = soft_signal_r_and_setter(str, initial_value=source1.name)
+            self.source2, _ = soft_signal_r_and_setter(str, initial_value=source2.name)
```
Villtord: Nit: do you perhaps want to rename those to `self.source1_name` and `self.source2_name` as you already have source1 and source2 as SignalR[float] or is it intentional that they have same names? I am not sure why you need them at all since your source1 and source2 both have names?

Villtord (changes requested):
Thanks, just a few more suggestions.

tests/devices/test_selectable_source.py
oliwenmandiamond: Done

src/dodal/devices/selectable_source.py
oliwenmandiamond: I added them as configuration so that if you later use the device and then want to check how the device was configured, e.g is source dcm or pgm, is source2 pgm or dcm etc, you have a record during data collection. I've added to use *_name

Villtord (approved):

Issue #2148, opened 2026-07-30
Correct type usage in maths module: XrayEnergy type to be removed


CoePaul:
When #2130 was added to dodal, various aspects were done badly

## Acceptance Criteria
* The any type hints using XRayEnergy and the XRayEnergy definition be removed from the codebase

* dataclass usage for create-once, read only classes be replaced by pydantics BaseModel

* that ingoing arguments to maths functions in that script be strictly keyword only

CoePaul:
This has been covered by work done in #2152  against #2150

PR #2152, opened 2026-07-31
Add MaterialAbsorptionSpectrum

### LINKED ISSUE #2150: AbsorptionSpectrum: Pair off AbsorptionCalculator x-ray energy range where it is valid
** Background **
AbsorptionCalculator ( amongst other of its input args) accepts and needs an x-ray energy in keV. 
but its calculation results are only valid for a specific range of energies ( the higher energies beyond a specific absorption line )

A given material can be modelled across different energy ranges using multiple AbsorptionCalculator instances.
So each calculator ought to be paired off against the valid range of energies.

Example from beamline scientist given below ( attached plot )

** Note **
Because there is more than one type of calculator and because one type is a CompoundCalculator which combines results from 
internal Calculator instances - the application of an EnergyRange directly into AbsorptionCalculator would unleash a bit of a mess.
It is better - in the code architecture - to pair off the energy range against the calculator ( external to the calculator ).

## Acceptance Criteria
* That a new AbsorptionSpectrum class is made pairing off absorption calculator and a valid range of energies.
* That the tests of AbsorptionCalculator are updated to push x-ray energy validation onto the new class

Contributes to #2123 

Image

CoePaul:
* Addresses #2150, #2148 

* Add new types for mathematical interval in all its closed/open ended variants

* Use this interval to adjust some of the internal validation in existing general_maths classes

* Address issues with inconsistent types across absorption related classes old code had incoherent blends of dataclass, pydantic BaseModel and Protocol which the IDE flagged - or showed up in debugging tests that failed midway through the updates here

* Also tidied other inconsitencies and room for optimisation -- unbalanced mocking in behaviour tests was defeating internal validation so that has been addressed -- repeated and unnecessarily short lists of invalid test inputs have been extended and made a constant in the test suite allowing them to be maintained in one location

* Other similar small tweaks to existing code or tests such as explicitly adding missing type hints, for example -> None to the return signature from tests

Addresses #2150 

1. Check the added classes for Interval types and MaterialAbsorptionSpectrum
2. Check the tests added for these new classes cover all the sensible test cases (note the caveats in test comments for Intervals)
3. Check other updates necessitated by either consistency / getting tests to now pass in light of the update
4. Confirm the CI tests and codecoverage and linters are all happy

### Checks for reviewer

### DIFF
--- a/src/dodal/common/general_maths/absorber_geometry.py
+++ b/src/dodal/common/general_maths/absorber_geometry.py
@@ -1,8 +1,10 @@
 from collections.abc import Callable
-from typing import Annotated, Literal, Protocol, Self, runtime_checkable
+from functools import cached_property
+from typing import Annotated, Literal, Protocol, runtime_checkable
 
 from pydantic import (
     BaseModel,
+    ConfigDict,
     Field,
     PrivateAttr,
     StrictFloat,
@@ -15,6 +17,10 @@ from dodal.common.general_maths.arithmetic_conversions import (
     convert_microns_to_cm,
     convert_mm_to_cm,
     get_straight_line_y,
+)
+from dodal.common.general_maths.interval import (
+    FloatInterval,
+    OpenInterval,
 )
 
 SupportedThicknessUnits = Annotated[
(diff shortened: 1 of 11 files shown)


CoePaul:
Split the changes into two separate commits - for easier reviewing.

CoePaul:
> Hold off on reviewing - am in the middle of a rebase / refactor

Cross that out now - it is ready for review - _once I can work out how to take it out of draft mode_

src/dodal/common/general_maths/interval.py
```diff
@@ -0,0 +1,76 @@
+from pydantic import BaseModel, ConfigDict, StrictFloat, model_validator
+
+
+class FloatInterval(BaseModel):
```
rtuck99, line 4: Does this really need to be a Pydantic model?
Pydantic models do add quite a bit of overhead to startup times and test execution

src/dodal/common/general_maths/absorber_geometry.py
```diff
@@ -92,21 +98,23 @@ class WedgeGeometry(BaseModel):
...
-        bound = 5.0  # round number cotangent limit for thin wedge
-        if -bound < value < bound:
-            msg: str = f"taper_contangent must correspond to thin wedge, of magnitude {bound} or larger"
-            raise ValueError(msg)
+    def validate_taper_is_thin(cls, value: float, _magnitude=5) -> float:
+        _invalid_taper_range: FloatInterval = OpenInterval(
+            lower=-_magnitude, upper=_magnitude
+        )  # round number cotangent limit for thin wedge
+        if value in _invalid_taper_range:
+            _msg: str = f"taper_contangent must correspond to thin wedge, of magnitude {_magnitude} or larger"
```
rtuck99, line 111:
```suggestion
            _msg: str = f"taper_cotangent must correspond to thin wedge, of magnitude {_magnitude} or larger"
```

src/dodal/common/general_maths/interval.py
rtuck99, line 4: My other comment would be, _do we really need this_? I mean, it's very nice and all to have type-safe representation of range types, but in this specific application, are we ever likely to encounter all these different types of ranges? If your energy is at the exact limit of the absorption curve, just how valid are the numbers you are getting out?

I suspect in practice, such ranges are defined with sufficient margins that being 1eV outside the range specified is neither here nor there.

src/dodal/common/general_maths/material_absorption_maths.py
```diff
@@ -90,17 +108,102 @@ class SingleRollOffAbsorptionCalculator(AbsorptionCalculator):
...
+            Real materials the final absorption will be positive.
+
+        Args:
+            energy_kev (StrictFloat): The individual energy per photon in kilo-electronvolts
+
+        Returns:
+            (float): Model adjustment of attenuation per cm of absorber material depth.
+        """
+        # Iterate sequentially through our modelled energy intervals
+        for absorption_model in self.intervals:
```
rtuck99, line 197: Perhaps we need to assert somewhere that our intervals are non-overlapping. Or specify that if they are overlapping, how they are resolved.

rtuck99 (approved):
Approved - some observations, up to you to decide if you want to make any changes

PR #2153, opened 2026-08-03
Independent magnet axis

### LINKED ISSUE #352: Decouple MagnetAxis and Use Live EPICS Limits
### Goals
- [x] As a beamline scientist, I can run standard single-axis scans using bps.scan and bps.mv with float inputs via MagnetUniaxialAxis.
- [x] As a developer/maintainer, I can construct MagnetAxis and MagnetCartesianCoordinates instances without passing controller callbacks (mag_within_boundary, ramp_rate) down into child constructors.
- [x] As a beamline safety check, movement strategies dynamically validate boundaries using live EPICS PV signals (e.g., LIM:FIELD:NOW) rather than hardcoded Python limit values.

### Tasks
- [x] Refactor MagnetAxis into a pure StandardReadable device and strip controller callback parameters from child device constructors.
- [x] ] Update movement validation logic to evaluate boundary checks against dynamic EPICS limit signals instead of hardcoded numbers (preserving hardcoded bounds only for $H(r)$ curves).
- [x] Replace manual _moving boolean flags with asyncio.Lock primitives to ensure async concurrency safety.
- [ ]  Write unit and integration tests covering single-axis bps.scan movements and dynamic limit validations.

### Related Resources
https://github.com/DiamondLightSource/dodal/pull/2115

Relm-Arrowny:
Fixes [sm-blusky#352](https://github.com/DiamondLightSource/sm-bluesky/issues/352)

@oliwenmandiamond  Not clean yet, but this shows the general idea. Running the dodal connect tests helped me see why you wanted to break the axis into two devices, but to me they represent the field and its limits, so they belong together.

## Changes
- Replaced  boolean `self_moving` with `asyncio.Lock()` and updated `set_within_boundary`.
- Removed ramp_controller and integrated it into magnet axis as simple signal.
- Made magnet axis standardReadable and without controller's `set_within_boundary` injection making it impossible to make a field movement without accessing `controller._start_ramp`
- Added dynamic hardware limit checking so `MovementStrategy` checks live limits rather than hardcoded limits.
- Move flyable into controller allow flyscan in cube and uniaxial mode.
- Added dynamic step timeout calculations (calculate_timeout_per_step)
- Added SuperConductingMagnet wrapper to preserve the user interface.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/dodal/devices/beamlines/i06_1/magnet/movement.py
```diff
@@ -73,16 +88,23 @@ class SphericalMovement(MovementStrategy):
...
-        self, current_readback: MagnetPosition, target: MagnetRequest
+        self,
+        current_readback: MagnetPosition,
+        target: MagnetRequest,
+        limits: MagnetPosition,
     ) -> None:
+
         mag_pos_after_move = target.resolve_pos(current_readback)
+        self._check_epics_hardware_limits(mag_pos_after_move, limits)
         if mag_pos_after_move.field_magnitude > self.LIMIT:
```
oliwenmandiamond, line 99: This `check_within_limits` is done for every step that is produced by `move_steps`. 

`_check_epics_hardware_limits` should be done once and independent of the these limits. The movement strategy is only checking if the total field or other bits that are mandated by the mode that the IOC does not check. 

You have also duplicated the entire API calls. If this either done outside of this entirely, it is done once or if it put in a new method to the base `MovementStrategy` and then have the API call it once rather making every movement strategy implementation check it (which might be forgotton).

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -92,152 +84,25 @@ class MagnetAxis(StandardReadable, StandardMovable[float], Flyable, Preparable):
...
-            self.rho = derived_signal_rw(
-                MagnetPosition.calc_rho, self._set_rho, x=cart.x, y=cart.y, z=cart.z
+            self.readback = epics_signal_r(float, f"{prefix}-MAG-01:{axis}:RBV")
+        self.demand = epics_signal_rw(float, f"{prefix}-MAG-01:{axis}:DMD")
+        self.limit = epics_signal_r(float, f"{prefix}-SMC-0{axis_number}:LIM:FIELD:NOW")
+        with self.add_children_as_readables(StandardReadableFormat.HINTED_SIGNAL):
+            self.ramp_rate = epics_signal_rw(
+                float,
+                read_pv=f"{prefix}-SMC-0{axis_number}:STS:RAMPRATE:TPM",
+                write_pv=f"{prefix}-SMC-0{axis_number}:SET:DMD:RAMPRATE:TPM",
```
oliwenmandiamond, line 99: As mentioned before, this should not be done. You are combining lots of different base PV's to represent one device when this does not reflect the hardware. 

Each magnet axis has a power supply associated with it which then defines the limits and ramp rates. This should be it's own device using `-SMC-0X:` and then use pass it in as reference to the axis of PV` -MAG-01:`

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -390,14 +270,327 @@ async def set_within_boundary(self, value: MagnetRequest):
...
+
+        else:
+            raise ValueError(
+                f"Cannot prepare fly scan in mode {mode}. Only uniaxial and cubic modes are supported for now."
+            )
+        start_position = MagnetRequest(**{value.fly_axis: value.start_position})
+        await self.set_within_boundary(start_position)
+        fly_axis = getattr(self, value.fly_axis)
+        await fly_axis.ramp_rate.set(value.ramp_rate)
+
```
oliwenmandiamond, line 403: I don't agree with this approach.

If i go to fly the magnet controller and I forget to change mode, I might fly scan something completely unexpected.

e.g fly scan scmc, I don't know what I am actually fly scanning when running the plan, I have to have made sure I am in the right mode first which you can very easily not do.

Where as if I fly scan scmc.cart.x, I know for a fact I am always fly scanning x. If i fly scan scmc.cart.z, I am always fly scanning z. It just might throw error if in the wrong mode which is safer. The API is consistent and and never changes where as this one is dynamic and will cause problems down the road.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
```diff
@@ -390,14 +270,327 @@ async def set_within_boundary(self, value: MagnetRequest):
...
+                raw_to_derived=self._read_position,
+                set_derived=self._set_y,
+                val=self._controller_ref().y.readback,
+            )
+            self.z = derived_signal_rw(
+                raw_to_derived=self._read_position,
+                set_derived=self._set_z,
+                val=self._controller_ref().z.readback,
+            )
+
```
oliwenmandiamond, line 464: I don't agree with this, we are adding more complication and layers to an already existing complicated device. 

This still doesn't stop the original problem when trying to move two axes at a time independently so I don't see the point. 

We also lose all the benefits of using `StandardMovable`, now we don't get progress updates during a move of an independent axis. 

This also does the exact same thing the original implementation did of delegating the logic to the controller but with extra steps

oliwenmandiamond (changes requested):
Please see my comments.

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
Relm-Arrowny: Check FlyVectorMagnetInfo, you tell the controller which axis you want to fly, this make it easy to add the other modes and axises. 
```python
class FlyVectorMagnetInfo(ConfinedModel):
    fly_axis: str = Field(frozen=True)
    """Axis to fly along, one of 'x', 'y' or 'z'. Todo theta, phi or rho."""
    start_position: float = Field(frozen=True)
    """Start position of the magnet move. in Tesla"""

    end_position: float = Field(frozen=True)
    """End position of the magnet move, in Tesla."""

    ramp_rate: float = Field(frozen=True, gt=0)
    """Ramp rate of the magnet move, in Tesla/s."""

```

src/dodal/devices/beamlines/i06_1/magnet/superconducting_magnet.py
Relm-Arrowny: You cannot move two axis at a time, there is a asynio.lock on `set_within_boundary`  This only here to presevert the interface.  This work just like what you had before but without the circular dependencies.

oliwenmandiamond: I think you are misinterpretating what circular means here in this context. The approach before did not have circular dependencies.

This is circular
`
Controller -> owns Axis -> owns Controller`

The approach before did this

```
SuperConductingMagnetController
    ├── cart
    │     ├── x
    │     ├── y
    │     └── z
    └── sph

Axis
    └── callback -> controller.set_within_boundary()
```

The axis doesn't know about the controller type. It only knows about a callable (`MagnetMoveWithinBoundary`). That's dependency inversion, not a circular dependency.

The important thing is that `MagnetAxis` doesn't know what object implements `MagnetMoveWithinBoundary`.

The dependency is this 
```
Controller
    owns Axis

Axis
    depends on Protocol

Controller
    implements Protocol
```
There isn't an object reference back to the controller, only an injected callable.

`StandardMovable` does the exact same thing. The device creates the signals, it then creates a movable logic and then passes the signals to the movement logic. The device then then calls movement logic methods which then use the signals from the device. It is not circular in the sense you are describing because there are no hard dependency on the signals, the signals are abstract and can pass in anything. The same applies here with the Callable. If we passed in the entire controller, then yes that is circular but we don't.

This implementation does the exact same thing as before, only now you are passing in the full controller when it uses nothing but the callable function anyway while adding the additional complexity of trying to separate it further but leads to the same result...

```python
await self._controller_ref().set_within_boundary(value)
```

plus we lose the benefit of `StandardMovable`

Relm-Arrowny: You are right.
image

oliwenmandiamond:
Closing this as no longer needed, replaced by https://github.com/DiamondLightSource/dodal/pull/2115

PR #2154, opened 2026-08-03
Update `save-panda` to use DeviceManager

### LINKED ISSUE #2144: Save-panda unable to find device factory
Running `save-panda` currently doesn't work and just prints the error

```
Couldn't create device panda: Unable to find factory for panda
```

Still looking into this but the problem seems to be in the fact that `make_devices` may not work with DeviceManager.

## Acceptance Criteria
* We are once again able to save panda configurations as yaml files

noemifrisina:
Fixes #2144 

As all beamlines with a panda have been moved to use `DeviceManager`, the old way of looking for factories has been removed from this script.

1. Check tests pass
2. Try running save-panda on a beamline network and verify it can save the configuration file correctly. Run it again on main and verify it fails there.

### Checks for reviewer


src/dodal/plans/save_panda.py
```diff
@@ -70,19 +72,34 @@ def main(argv: list[str] | None = None):
...
+        module = import_module(module)
+
+    if (manager := getattr(module, "devices", None)) and isinstance(
+        manager, DeviceManager
+    ):
+        factories = manager.get_all_factories()
+    else:
+        raise ValueError(f"No device manager found in {module}")
+    device = manager.build_devices(factories[device_name])
+    return device
```
tpoliaw: If you're only building the one device, you can call build on it directly.

```
if panda := getattr(module, device_name, None):
    return panda.build()
else:
    raise ValueError(...)
```

If you want to connect it at the same time instead of calling connect later, you can use `.build(connect_immediately=True)`.

src/dodal/plans/save_panda.py
```diff
@@ -70,19 +72,34 @@ def main(argv: list[str] | None = None):
...
+        # Build and connect device
+        device = build_device(f"dodal.beamlines.{module_name}", device_name)
+        device.connect()
+        print("Panda device built and connected")
     except Exception as error:
         sys.stderr.write(f"Couldn't create device {device_name}: {error}\n")
         sys.exit(1)
 
-    panda = devices[device_name]
+    panda = device.devices[device_name]
```
tpoliaw: ...and this would no longer be needed. If you only build the one device, you'll get the device directly instead of getting a map back that you have to look the device up in.

src/dodal/plans/save_panda.py
```diff
@@ -70,19 +72,34 @@ def main(argv: list[str] | None = None):
...
     run_engine = RunEngine()
     print("Creating devices...")
     module_name = module_name_for_beamline(beamline)
     try:
-        devices = make_device(
-            f"dodal.beamlines.{module_name}", device_name, connect_immediately=True
-        )
+        # Build and connect device
+        device = build_device(f"dodal.beamlines.{module_name}", device_name)
+        device.connect()
```
tpoliaw: This would then become `panda = build_device(...)`

tpoliaw (approved):

PR #2155, opened 2026-08-04
Update to ophyd-async 0.21.1

oliwenmandiamond:
Update to latest ophyd-async release. This has updated Motor to now use StandardFlyable so have updated ID logic to reflect this

1. Check tests pass

### Checks for reviewer


tests/devices/insertion_device/test_energy.py
```diff
@@ -102,7 +101,7 @@ async def test_insertion_device_energy_prepare_success(
     mock_set.assert_awaited_once_with(energy=750)
     get_mock_put(
         mock_id_controller.apple2().gap().user_setpoint
-    ).assert_awaited_once_with(str(ramp_up_start))
+    ).assert_awaited_once_with(ramp_up_start)
```
Relm-Arrowny: If it is not putting str, we have broken something. 
in UndulatorGap
```python
        self.user_setpoint = epics_signal_rw(str, prefix + "BLGSET")
```

oliwenmandiamond: The issue is that we trying to rewrite Motor signals which is now causing inconsistently with the movable and fly logic changes so the signal hasn't been swapped out correctly. I think we need to apply the correct MovableLogic and use derived signals to convert between int and str.

Relm-Arrowny: I am not sure if we broke anything last time we change it to movablelogic, I think we should just revert it back to its original, there isnt anything to gain by forcing a very unique single action non-stoppable device into a double action movablelogic, It introduces a lot of unnecessary boilerplate for a device that was already working perfectly.

oliwenmandiamond:
Blocked by https://github.com/DiamondLightSource/dodal/pull/2161

Relm-Arrowny (approved):

PR #2161, opened 2026-08-05
Update id logic to use standard movable

### LINKED ISSUE #2158: Update insertion device devices to user StandardMovable and StandardFlyer
All of the devices used within dodal/devices/insertion_device were created before ophyd-async introduced `StandardMovable` and `StandardFlyer`. They have clashing and overriding methods that these classes implement e.g `GapSafeMotorNoStop` overrides the `prepare` method so will leave the class in an inconsistent state as it should now be injecting in FlyableLogic instead. 

As a result, many of the classes and methods can be refactored into these which will help to keep the code base consistent and modernised.

oliwenmandiamond:
Fixes https://github.com/DiamondLightSource/dodal/issues/2158

This fixes the insertion device logic to be compatible with `StandardMovable`, specifically the `UndulatorGap`. This has also restructured the ID logic so that we now have these new files:
- apple2_undulator_phases.py
- apple2_undulator_gap.py
- apple2_undulator_base.py
- apple2_undulaltor_motors.py
- apple2_controller.py

Other notable change is `MotorStringSetpoint` which allows us to still use `Motor` which have setpoints as str still take  floats using a derived signal, allowing the ID logic to not break when using `MotorMovableLogic` (soon to be `MotorFlyableMoveLogic`).

This has made the logic much more modular and easier to understand. Please note I have not altered the actual logic or the tests so this hasn't changed the ID behaviour. This was necessary because `Motor` has changed to use `StandardMovable` (and in ophyd-async 0.21.1 `StandardFlyable`) which made the current API conflicting as it was overriding methods it shouldn't of. This will make the migration to ophyd-async 0.21.1 much easier to manage. I've also added more documentation to classes to make it easier for someone else to maintain in the future.

1. Check refactor makes sense
2. Tests pass
3. dodal connect works

### Checks for reviewer


src/dodal/devices/insertion_device/apple2_undulator_base.py
```diff
@@ -0,0 +1,102 @@
...
+        raise RuntimeError(f"{status.name} is DISABLED and cannot move.")
+    if gate_val is UndulatorGateStatus.OPEN:
+        raise RuntimeError(f"{gate.name} is already in motion.")
+
+
+async def set_move_and_wait_for_gate(
+    gate: SignalR[UndulatorGateStatus], set_move: SignalW[int], timeout: float | None
+):
+    await set_move.set(value=1, timeout=timeout)
+    await wait_for_value(gate, UndulatorGateStatus.CLOSE, timeout=timeout)
```
Relm-Arrowny, line 45: We probably have not test it with the real system enough, should add to avoid racing condition. :

```suggestion
    await set_move.set(value=1, timeout=timeout)
    await wait_for_value(gate, UndulatorGateStatus.OPEN, timeout=5)
    await wait_for_value(gate, UndulatorGateStatus.CLOSE, timeout=timeout)
```

src/dodal/devices/insertion_device/apple2_undulator_gap.py
```diff
@@ -0,0 +1,170 @@
...
+    Supports standard moves as well as fly scanning.
+
+    Args:
+        prefix: Beamline-specific PV prefix.
+        name: Device name.
+    """
+
+    def __init__(self, prefix: str, name: str = ""):
+        super().__init__(prefix=prefix, name=name)
+        self.set_move = epics_signal_rw(int, prefix + "BLGSETP")
```
Relm-Arrowny: nit: we made set move SignalW through out so I am not sure which we want to change. I probably does not matter in practice. 
```suggestion
        self.set_move = epics_signal_w(int, prefix + "BLGSETP")
```

src/dodal/devices/insertion_device/apple2_undulator_phase_axes.py
```diff
@@ -0,0 +1,177 @@
...
+    """Two phase Motor to make up the locked id phase motion."""
+
+    def __init__(
+        self,
+        prefix: str,
+        top_outer: str,
+        btm_inner: str,
+        name: str = "",
+    ):
+        # Gap demand set point and readback
```
Relm-Arrowny: Could:Not gap
```suggestion

```

src/dodal/devices/insertion_device/apple2_undulator_phase_axes.py
```diff
@@ -0,0 +1,177 @@
...
+        with self.add_children_as_readables():
+            self.top_outer = UndulatorPhaseMotor(prefix=f"{prefix}BL{top_outer}")
+            self.btm_inner = UndulatorPhaseMotor(prefix=f"{prefix}BL{btm_inner}")
+        # Nothing move until this is set to 1 and it will return to 0 when done.
+        self.set_move = epics_signal_rw(int, f"{prefix}BL{top_outer}" + "MOVE")
```
Relm-Arrowny: nit:?
```suggestion
        self.set_move = epics_signal_w(int, f"{prefix}BL{top_outer}" + "MOVE")
```

src/dodal/devices/insertion_device/apple2_undulator_phase_axes.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Args:
+        prefix (str): The setting prefix PV.
+        name (str, optional): Name of the Id phase device.
+    """
+
+    def __init__(self, prefix: str, name: str = ""):
+        motor_pv = f"{prefix}MTR"
+        super().__init__(motor_pv, prefix + "SET", name=name)
+        del self.motor_stop
```
Relm-Arrowny, line 51: should: matching gap and its safer.
```suggestion
        self.motor_stop = None
```

src/dodal/devices/insertion_device/apple2_undulator_gap.py
```diff
@@ -0,0 +1,170 @@
...
+        Stores fly info for later use in kickoff.
+        """
+        await self.motor.prepare(value)
+
+    @AsyncStatus.wrap
+    async def kickoff(self):
+        await self.motor.kickoff()
+
+    def complete(self) -> WatchableAsyncStatus:
+        return self.motor.complete()
```
Relm-Arrowny: I think the safe `SafeUndulatorMoverBase` is already built in the `UndulatorGapMotor`
this: #2167

Relm-Arrowny (changes requested):
look good thanks for doing all the work.

src/dodal/devices/insertion_device/apple2_undulator_phase_axes.py
oliwenmandiamond: Done

src/dodal/devices/insertion_device/apple2_undulator_phase_axes.py
oliwenmandiamond: Removed, done

src/dodal/devices/insertion_device/apple2_undulator_base.py
oliwenmandiamond: The issue is we then need to create sophisticated mock logic for the id to make sure whenever set_move is called, we then open and close the gate. I agree this should be done, but need to do another issue. I tried to this in a change I've now reverted as it was getting too big and complicated for this change. I will create an issue for this (https://github.com/DiamondLightSource/sm-bluesky/issues/359)

src/dodal/devices/insertion_device/apple2_undulator_gap.py
oliwenmandiamond: done

src/dodal/devices/insertion_device/apple2_undulator_phase_axes.py
oliwenmandiamond: done

src/dodal/devices/insertion_device/apple2_undulator_base.py
Relm-Arrowny: I am prettry sure we are already cloved it open and then close in test:https://github.com/DiamondLightSource/dodal/blob/3a2d507e248ac044b3b0fd2eda8902e26c7a8829/tests/devices/insertion_device/test_apple2_undulator.py#L110

src/dodal/devices/insertion_device/apple2_undulator_base.py
```diff
@@ -0,0 +1,77 @@
...
+    timeouts, while this class provides the common sequence of:
+
+    * checking that motion is permitted,
+    * writing demand positions,
+    * triggering the controller move,
+    * waiting for motion to complete.
+
+    Attributes:
+        set_move: Signal used to trigger motion after demands have been written.
+    """
```
Relm-Arrowny: Should: DocString need updating, This is no longer SafeUndulator, it no longer provides     ` checking that motion is permitted,
    or waiting for motion to complete.`
```suggestion
class UndulatorBase(abc.ABC, Generic[T]):
    """Base class for Apple2 undulator devices that use gated motion.

    Subclasses implement writing demand positions and estimating move
    timeouts, while this class provides the common sequence of:

   
    * writing demand positions,
    * triggering the controller move,

    Attributes:
        set_move: Signal used to trigger motion after demands have been written.
    """
```

src/dodal/devices/insertion_device/apple2_controller.py
```diff
@@ -29,6 +35,59 @@
...
+
+    @AsyncStatus.wrap
+    async def set(self, id_motor_values: Apple2Val) -> None:
+        """Check ID is in a movable state and set all the demand value before moving
+        them all at the same time.
+        """
+        gap = self.gap_ref()
+        phase = self.phase_ref()
+        # Only need to check gap as the phase motors share both status and gate with gap.
+        await undulator_check_move(gap.status, gap.gate)
```
Relm-Arrowny: Maybe: I think this probably fine. We probably could have just done this without refactoring safeUndulator.
```python
 await gap.movable_logic.check_move(new_position=float(id_motor_values.gap))
 await phase.movable_logic.check_move(new_position=float(id_motor_values.phase)) x4?
```
There is an argument we should do the check early and let it fail before trying to set_demand. In fact  thinking about it, we probably should as we are only protected by the ioc limit here on set.

Relm-Arrowny (changes requested):

src/dodal/devices/insertion_device/apple2_controller.py
oliwenmandiamond: So gap and phase both use the exact same PV's for gate and status. So this should be a single device that both gap and phase use to check the move is valid first. This way, we only do only check it once. I tried to do this here https://github.com/DiamondLightSource/dodal/pull/2174 and merged into this but it got too complicated and large so I decided to revert and will do this in another change as outside the scope of this issue

oliwenmandiamond: Sorry, I missunderstood this. Yes I can add that

oliwenmandiamond: Okay, done this now. Added Checkable protocol so now we use this

Relm-Arrowny (approved):
Thanks LGTM, love the pr note.

PR #2164, opened 2026-08-05
class AttenuatorMotorPositions: cleanse obsolete naming

### LINKED ISSUE #2163: Rename method in AttenuatorMotorPositions
** Motivation **
When AttenuatorMotorPositions was first created it was thought of only as a demand / request 
for a change in motor positions within the i19 attenuation system.

The implementation is basically a dict wrapped in some validation.

Later ( #2125 ) it was realised that the same dict can store both demands and reports back regarding the 
present location of each system motor - a valuable resource for answering questions like :

"What is the present attenuation of the system at x-ray energy <think of a number> keV ?"

When that update renamed AttenuatorMotorPositionDemands -> AttenuatorMotorPositions
it failed to update the now stranded obsolete name of the method 
"validated_complete_demand" 

** Task **
a) update the method name to match the move towards more generalised use of the class

b) make the method lean on a cached property - as we don't need to recalculate the result of a 
smush together of the two input dict fields 
every time the call is made for the combined dict

## Acceptance Criteria
* That the validate method name changes to reflect the broader use of the class
* That a corresponding fix has been raised in i19-bluesky - as this will be a breaking change
* That the method leans on a cached property
* That linters and CI tests etc are all placated as per usual

CoePaul:
* Internal naming conventions within the Attenuator motor positions class had been left in prior state before recent rename ( where the rename relfected a widened scope from purely demanding motor moves to also capture live positions of motors )

   The resulting obsolescene meant names of methods and internal variables
   did not reflect the wider usage scope - that is hereby fixed

* Also make internal merge of discrete indexed positions and continuous motor axis positions into a cached property saves the merge being recalculated - and makes its evaluation lazy

* Update tests to reflect renaming

* Improve tests with parametrized test cases

* Fix quirk in axis naming convention which used to (inadvertently) permit leading integer in name/tag of wheel or axial motor

* Tighten test cases to capture this and other oversights in original test suite

Fixes #2163

1. Confirm name updates make logical sense to any reader unfamiliar with what the code does
2. Confirm You don't need to read the code in detail to understand the names - if that proves to be necessary, fail the CR
3. Confirm linters etc are happy

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
+++ b/src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
@@ -1,3 +1,4 @@
+from functools import cached_property
 from typing import Annotated, Self
 
 from ophyd_async.core import AsyncStatus
@@ -18,22 +19,44 @@ from dodal.devices.beamlines.i19.access_controlled.hutch_access import (
     ACCESS_DEVICE_NAME,
 )
 
-PermittedKeyStr = Annotated[str, StringConstraints(pattern="^[A-Za-z0-9-_]+$")]
+PermittedKeyStr = Annotated[
+    str, StringConstraints(pattern=r"^[_A-Za-z][A-Za-z0-9-_]*$")
+]
 
 
 class AttenuatorMotorPositions(BaseModel):
+    """Motor positions for attentuators in the attenuation system, be they indices on a discrete steps motor or continuous positions on an axis, or axes.
+
+    This is a validated dict which pairs off naming "tags" for each motor (e.g. "x" for a lateral x-axis motor or "w" for a filter wheel),
+    with the corresponding int or float motor position ( or position demand / request ) as appropriate.
+
+    Attributes:
+        continuous_positions: dict contributions pairing names of motors against the discrete motor position (likely to be in mm)/
+        discrete_indices: dict contributions pairing names of index stepping positioners against the discrete position index (e.g. for a filter wheel).
+
+    Examples:
+        ( Given that i19 mounted a resin wedge on their x-axis motor,
+            an aluminium wedge on their y-axis motor and their first filter wheel we will name "w" ... )
+
+        Request for motor changes:
+            {"x": 20.4} and {"w":3} might indicate a request to move x-axis to 20.4 mm and the wheel round to slot 3.
+
+        Reporting of system motor positions (on say i19, where wheel index 1 is reserved for an EMPTY slot):
+        {"x": 28.62, "y": 5.0} and {"w": 1} might indicate the resin wedge is at 28.62 mm, aluminium wedge is OUT, filter wheel at EMPTY (OUT)."
+    """
+
     model_config = ConfigDict(frozen=True)
 
-    continuous_demands: dict[PermittedKeyStr, StrictFloat | StrictInt] = Field(
+    continuous_positions: dict[PermittedKeyStr, StrictFloat | StrictInt] = Field(
         default_factory=dict, kw_only=True
     )
-    indexed_demands: dict[PermittedKeyStr, Annotated[StrictInt, Field(gt=0)]] = Field(
+    discrete_indices: dict[PermittedKeyStr, Annotated[StrictInt, Field(gt=0)]] = Field(
         default_factory=dict, kw_only=True
     )
 
     @model_validator(mode="after")
     def no_keys_clash(self) -> Self:
-        common_keys = set(self.continuous_demands).intersection(self.indexed_demands)
+        common_keys = set(self.continuous_positions).intersection(self.discrete_indices)
         common_key_count = sum(1 for _ in common_keys)
         if common_key_count < 1:
             return self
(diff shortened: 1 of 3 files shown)


src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
```diff
@@ -18,22 +19,41 @@
...
 
 class AttenuatorMotorPositions(BaseModel):
+    """Motor positions for attentuators in the attenuation system, be they indices on a discrete steps motor or continuous positions on an axis, or axes.
+
+    This is a validated dict which pairs off naming "tags" for each motor (e.g. "x" for a lateral x-axis motor or "w" for a filter wheel),
+    with the corresponding int or float motor position ( or position demand / request ) as appropriate.
+
+    Attributes:
+        continuous_positions: dict contributions pairing names of motors against the discrete motor position (likely to be in mm)/
+        discrete_indicecs: dict contributions pairing names of index stepping positioners against the discrete poisition index (e.g. for a filter wheel).
```
DominicOram: Should: 

```suggestion
        discrete_indices: dict contributions pairing names of index stepping positioners against the discrete position index (e.g. for a filter wheel).
```
Do you have a spellcheck plugin installed in vscode? I can help you add it if not

src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
```diff
@@ -18,22 +19,41 @@
...
+
+    Examples:
+        Request for motor changes:
+            {"x": 20.4} and {"w":3} might indicate a request to move x-axis to 20.4 mm and the wheel round to slot 3.
+
+        Reporting of system motor positions (on say i19, where wheel index 1 is reserved for an EMPTY slot):
+        {"x": 28.62, "y": 5.0} and {"w": 1} might indicate the resin wedge is at 28.62 mm, aluminium wedge is OUT, filter wheel at EMPTY (OUT)."
```
DominicOram, line 45: Nit: The mention of "resin"/"aluminium" is jarring here as they have not yet been mentioned. It seems like this is just talking about an "x" and a "y"

src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
```diff
@@ -42,10 +62,26 @@ def no_keys_clash(self) -> Self:
             error_msg = f"Common {ks} found in distinct motor demands: {common_keys}"
             raise ValueError(error_msg)
 
-    def validated_complete_demand(
+    @cached_property
+    def _cached_merged_validated_dict(
```
DominicOram, line 69: Should: Why do we have the extra layer of `_cached_merged_validated_dict` here rather than just have this as `validated_complete_demand`? I don't see how we benefit from it

tests/devices/beamlines/i19/access_controlled/test_attenuator_motor_positions.py
```diff
@@ -7,214 +8,318 @@
...
-        continuous_demands=standard_wedge_position_demand,
-        indexed_demands=standard_wheel_position_demand,
+        continuous_positions=standard_wedge_position_demand,
+        discrete_indices=standard_wheel_position_demand,
     )
     assert position_demand is not None
 
 
-def test_that_attenuator_position_demand_triplet_provides_expected_rest_format():
+def test_that_attenuator_motor_positions_triplet_provides_expected_rest_format() -> (
```
DominicOram, line 179: Nit: In general a return type of `None` is implied by just giving no return type, particularly in tests where returning anything other than `None` doesn't make much sense. But it's not hurting anything if you want to keep it in

tests/devices/beamlines/i19/access_controlled/test_attenuator_motor_positions.py
```diff
@@ -7,214 +8,318 @@
...
-def test_that_attenuator_position_demand_can_be_created_with_only_one_wheel():
+@pytest.mark.parametrize(
+    "wheel_name",
+    ["w", "V", "_w1", "_W_one", "w4", "spare_wheel"],
+)
+@pytest.mark.parametrize(
+    "trial_index",
+    TRIAL_WHEEL_INDICES,
+)
+def test_that_attenuator_motor_positions_can_be_created_for_only_one_wheel(
```
DominicOram: Nit: The two sets of parametrize will do a cross product, so this is actually running 42 tests, which seems a little overkill. But they're quick so there's no massive downside, happy to keep in

DominicOram (changes requested):
Thanks, mostly just suggestions other than the spelling

tests/devices/beamlines/i19/access_controlled/test_attenuator_motor_positions.py
CoePaul: I asked which was more pythonic ( to state None expressly or not bother mentioning it )
I learn -> None is allegedly strictly pythonic... so I did that

( The allegation was that - generally - in the absence of a stated return type people might be tempted to assume You forgot )

tests/devices/beamlines/i19/access_controlled/test_attenuator_motor_positions.py
CoePaul: it would be nice if it could do half the matrix up to the diagonal, I wanted cases 1 to 6 as they feature on the real ( first ) filter wheel on the beamline plus one index we don't presently have 
and then it doesn't make sense to only test them all against one letter

I could pair them off "w" with 1 
"V" with 5 etc but then as you say the time is not much - tests should I think err on overkill 
until people complain about CI run time

src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
CoePaul: the benefits are 
a) if we never call the "API" method validated_and_complete ( used by calling classes ) then no time is ever wasted constructing the merged dict
b) if the merged dict is ever required, the CPU cost of computing it is only ever paid once
c) the downside is the memory cost - but I think it's worth it

src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
CoePaul: fair point - it makes sense inside i19 - but not outside

src/dodal/devices/beamlines/i19/access_controlled/attenuator_motor_squad.py
CoePaul: once You surrender to using a spellchecker - you will never spell again

CoePaul: but if there was a terminal command which sweeps a whole directory ...

CoePaul: I have now incorprorated hunspell into a bash function

DominicOram (approved):
Thanks!

PR #2169, opened 2026-08-11
I15-1: Add a summing tetramm device

DominicOram:
Fixes https://github.com/DiamondLightSource/crystallography-bluesky/issues/44

1. Confirm tests pass
2. Confirm explanation of logic makes sense
3. Confirm we can `dodal connect i15-1`

### Checks for reviewer


DominicOram:
I have asked i22 if they would also like this but they don't. @iain-hall, would i18 like this behaviour?

iain-hall:
Not sure if this would be useful for I18 (or i20 or b18). They all have several Tetramms each, but the plugin chains seem to be a little different to i15-1 (e.g. there are no ROI or 'process' plugins), and the 'raw' averages per channel are normally provide what they are interesting in measuring.

jacob-williamson (approved):
Looks good!

PR #2171, opened 2026-08-12
i15-1: Add more intelligent temperature setting for blower

DominicOram:
Needed for https://github.com/DiamondLightSource/crystallography-bluesky/issues/92

1. Confirm tests pass

### Checks for reviewer

### DIFF
--- a/src/dodal/devices/beamlines/i15_1/blower.py
+++ b/src/dodal/devices/beamlines/i15_1/blower.py
@@ -96,6 +96,8 @@ class Blower(SafeOrBeamPositioner):
     the given pneumatic is open to allow airflow.
     """
 
+    DEFAULT_SETTLE_TIME = 180
+
     def __init__(
         self,
         prefix: str,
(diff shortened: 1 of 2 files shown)


DominicOram:
@oliwenmandiamond, this is my first time using `StandardMovable` in anger, have I used it correctly?

src/dodal/devices/beamlines/i15_1/blower.py
```diff
@@ -26,35 +105,18 @@ def __init__(
...
-                self._set_temperature,
-                temperature_rbv=self._temperature_rbv,
-            )
-
-        self._pneumatic = epics_signal_r(ValveState, pneumatic_pv)
         self.ramp_rate_c_per_sec = epics_signal_rw(
             float, f"{prefix}RR", f"{prefix}RR:RBV"
         )
 
-        super().__init__(motion_pv, config_client, xpdf_parameters_path, name)
```
jacob-williamson, line 44: What would happen if we tried to set a temperature before setting the settle time?

jacob-williamson (approved):
Looks good, one question in the code

src/dodal/devices/beamlines/i15_1/blower.py
DominicOram: I _think_ the soft signal defaults to 0. In the ULIMS sechma the default is 180 though so maybe we should use that here

Issue #2172, opened 2026-08-13
Pydantic BaseModels to capture I19 transmission system JSON specifications


CoePaul:
**Background**
A canonical JSON file will be beamline-scientist editable configuration to specify 
- the material parameters of absorbers in the system
- the shapes and mounting of the absorbers
- the prioritisation of absorber uses ( **EDIT** - this last criterion has been dropped - as it is not hardware configuration )

## Acceptance Criteria
* That BaseModel classes exist in harmony with the JSON file structure at 
   _/dls_sw/i19-1/software/daq_configuration/json-shared/TransmissionSystemSpecification.json_

* that test suites exist to ensure the validation behaves appropriately,
  accepting valid input and rejecting invalid input

CoePaul:
<img width="1600" height="974" alt="Image" src="https://github.com/user-attachments/assets/d0dd1800-aee4-4439-bcbb-df9a9be301bc" />

PR #2173, opened 2026-08-13
Issue2172

CoePaul:
Contributes to #2172 

1. Check BaseModel classes against JSON in file mentioned from i19 config in #2172 description
2. Check tests cover reasonable expectations for JSON validation
3. Ensure CI passes the tests and that the python matches house style ( etc )

### Checks for reviewer


src/dodal/devices/beamlines/i19/transmission/spec_from_config/energy_interval_spec.py
```diff
@@ -0,0 +1,32 @@
...
+    Note:
+        The range of energies is a closed interval, so inclusive of end values.
+
+    Attributes:
+        units: At present only 'keV' (or 'kiloelectronvolts') are supported.
+        lower: The lower end of the energy interval (range) for specific absorption fit curve.
+        upper: The upper end of the energy interval (range) for specific absorption fit curve.
+    """
+
+    units: str = Field(..., pattern=r"^(keV|kiloelectronvolts)$")
```
DominicOram: Should: 
```suggestion
    units: Literal["keV", "kiloelectronvolts"]
```
is more concise and readable. It also means the type system will enforce it later if we ever use the value

src/dodal/devices/beamlines/i19/transmission/spec_from_config/energy_interval_spec.py
```diff
@@ -0,0 +1,32 @@
...
+    lower: StrictFloat
```
DominicOram: Could: 
```suggestion
    lower: StrictFloat = Field(gt=0)
```
Would do the check that it's greater than 0 for you so you do not need to do it in the validator

src/dodal/devices/beamlines/i19/transmission/spec_from_config/fitted_absorption_curve_spec.py
```diff
@@ -0,0 +1,46 @@
...
+    Attributes:
+        photon_absorption:
+            Material characteristic scaling constant which defines absorption curve.
+        roll_off:
+            Power-law exponent for variation with energy, usually something like -2.75 ∓ 0.24.
+        residuals_polynomial_coeffs:
+            Optional residuals correction, zeroth order parameter first.
+    """
+
+    photon_absorption: StrictFloat
```
DominicOram: Could: As above,
```suggestion
    photon_absorption: StrictFloat= Field(gt=0)
```

src/dodal/devices/beamlines/i19/transmission/spec_from_config/fitted_absorption_curve_spec.py
```diff
@@ -0,0 +1,46 @@
...
+    roll_off: StrictFloat
```
DominicOram: Could: 
```suggestion
    roll_off: StrictFloat = Field(lt=-2, gt=-4)
```
(probably with -2 and -4 pulled into constants). Let's pydantic do all the checking for you. With this and above you can remove the `validate_attributes` entirely

src/dodal/devices/beamlines/i19/transmission/spec_from_config/fitted_absorption_curve_spec.py
```diff
@@ -0,0 +1,46 @@
...
+    def validate_attributes(self) -> "FittedAbsorptionCurveSpec":
+        if self.photon_absorption > 0.0 and self.roll_off in PHYSICAL_ROLL_OFF:
+            return self
+        if not self.photon_absorption > 0:
+            raise ValueError(
+                f"Photon absorption constant {self.photon_absorption} is invalid."
+            )
+        _msg = (
+            f"Absorption roll off {self.roll_off} does not seem likely on physics grounds."
+            if self.roll_off < 0
```
DominicOram: Should: Why have a separate check on the positive case? Then have a cryptic message that doesn't tell the user how to fix the issue? I think better to remove this case and just check that it's around -3

src/dodal/devices/beamlines/i19/transmission/spec_from_config/foil_spec.py
```diff
@@ -0,0 +1,48 @@
...
+    Attributes:
+        units: The length unit used to specify the foil thickness.
+        value: The foil thickness in the specified unit.
+    """
+
+    units: str
+    value: StrictFloat
+
+    # Base Model internal setting to make this class immutable
+    model_config = ConfigDict(frozen=True, extra="forbid")
```
DominicOram, line 22: Could: It's not clear how you're going to use this later but if the user is going to specify different units and you want them normalised we could do it here, something like:
```
from pydantic import BaseModel, model_validator
from pint import UnitRegistry

ureg = UnitRegistry()

class FoilThicknessSpec(BaseModel):
    units: str
    value: float

    @property
    def quantity(self):
        return self.value * ureg(self.units)

    @property
    def thickness_um(self):
        return self.quantity.to("micrometer").magnitude
```
or if we don't want to normalise we can at least make sure it's a length:
```
from pydantic import BaseModel, field_validator
from pint import UnitRegistry

ureg = UnitRegistry()

class FoilThicknessSpec(BaseModel):
    units: str
    value: float

    @field_validator("units")
    @classmethod
    def validate_length_unit(cls, v: str) -> str:
        try:
            unit = ureg.Unit(v)
        except Exception:
            raise ValueError(f"Unknown unit '{v}'")

        if unit.dimensionality != ureg.meter.dimensionality:
            raise ValueError(
                f"Unit '{v}' is not a unit of length"
            )

        return v
```

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+
+class LateralMotorSpec(BaseModel):
+    """The positions scale for an axial wedge motor (as specified by configuration, typically **JSON**).
+
+    Notes:
+        - The (potentially counterintuitive) class' attribute names match beamline scientists' "domain jargon".
+
+        - Each lateral motor drives one absorber wedge sideways across the x-ray beam:
+            - expected sideways motions are pure horizontal or pure vertical
+                - but that's a detail, azimuthal orientation around the beam should not matter.
```
DominicOram: Nit: Then why mention it? This is already complicated

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+            - The position of the scale zero relative to the wedge location is hereby nailed down.
```
DominicOram: Nit: I'm not sure what this line means?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Attributes:
+        units: At present only "mm" are supported.
+        out: Motor position consistent with having fully retracted the wedge from the x-ray beam.
+        threshold: Motor position for minimal absorption in the permitted active absorbing position range.
+        max: Motor position for minimal absorption in the permitted active range.
+        tolerance: Motor position accepted margin for readout error.
+    """
+
+    units: str = Field(default="mm")
```
DominicOram: Could: You should enforce this, otherwise people could put something in the JSON that's wrong
```suggestion
    units: Literal["mm"] = Field(default="mm")
```

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+    out: StrictFloat
```
DominicOram, line 50: Nit: `out_position` would make more sense to me

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+                - and then a range of "active" positions, ( where the mathematical assumption of a linear taper is reasonable ).
+
+        - Although the units here will default to **mm** when unspecified (i.e. when omitted),
+        the units setting is still likely to appear in the JSON, as a handy reminder to human readers of the configuration file.
+
+    Attributes:
+        units: At present only "mm" are supported.
+        out: Motor position consistent with having fully retracted the wedge from the x-ray beam.
+        threshold: Motor position for minimal absorption in the permitted active absorbing position range.
+        max: Motor position for minimal absorption in the permitted active range.
```
DominicOram: Should: should this be 
```suggestion
        max: Motor position for maximum absorption in the permitted active range.
```

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+    tolerance: StrictFloat = Field(default=5.0e-3)
+
+    # Base Model internal setting to make this class immutable
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    @model_validator(mode="after")
+    def validate_attributes(self) -> "LateralMotorSpec":
+        # counter-intuitive naming, but depending on
+        # motor scale orientation relative to wedge orientation
+        # max (thickness) position could greater or smaller than threshold
```
DominicOram: Nit:
```suggestion
        # max (thickness) position could be greater or smaller than threshold
```

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+        _lower = min(self.out, self.max)
+        _upper = max(self.out, self.max)
+        _interval = ClosedInterval(lower=_lower, upper=_upper)
+        if self.threshold in _interval:
+            return self
+        msg: str = r"Inconsistent wedge geometry: Threshold not between max and out."
+        raise ValueError(msg)
```
DominicOram, line 69: Nit: This feels more readable on one line:
```suggestion
        raise ValueError("Inconsistent wedge geometry: Threshold not between max and out.")
```
Should: It doesn't need to be a raw string.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @model_validator(mode="after")
+    def validate_attributes(self) -> "LateralMotorSpec":
+        # counter-intuitive naming, but depending on
+        # motor scale orientation relative to wedge orientation
+        # max (thickness) position could greater or smaller than threshold
+        _lower = min(self.out, self.max)
+        _upper = max(self.out, self.max)
+        _interval = ClosedInterval(lower=_lower, upper=_upper)
+        if self.threshold in _interval:
```
DominicOram, line 66: Should: 
```suggestion
        if _lower <= self.threshold <= _upper:
```
is a lot more readable as it's standard python, compared to a class that I have to go and look up the definition of.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
```diff
@@ -0,0 +1,92 @@
...
+        AxisNameValidation.validate_axis_name(axis_name=key_name)
+
+    @classmethod
+    def extract_motors_specifications(
+        cls,
+        *,
+        system_configuration: SystemConfiguration,
+        motor_identifier: str,
+    ) -> LateralMotorSpec:
+        """Extracts wheel specification from configuration of the transmission system."""
```
DominicOram: Should: These aren't wheels
```suggestion
        """Extracts lateral motor specification from configuration of the transmission system."""
```

src/dodal/devices/beamlines/i19/transmission/spec_from_config/system_aspect_base_parser.py
```diff
@@ -0,0 +1,85 @@
...
+
+from dodal.devices.beamlines.i19.transmission.spec_from_config.system_configuration import (
+    SystemConfiguration,
+)
+
+# 1. Define a local TypeVar for generic aspect specs
+TxSubSpecT = TypeVar("TxSubSpecT")
+
+
+class SystemAspectBaseParser(RootModel[dict[str, TxSubSpecT]], Generic[TxSubSpecT]):
```
DominicOram, line 17: Must: This structure adds loads of complexity and boilerplate that pydantic can just give you for free. See https://github.com/DiamondLightSource/dodal/compare/issue2172...2172_without_SystemAspectBaseParser?expand=1 where I have removed the need for `LateralMotorsConfig`. I'm pretty sure we can do this for all the other `SystemAspectBaseParser` children too?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/system_configuration.py
```diff
@@ -0,0 +1,34 @@
+from functools import cached_property
+from typing import Any
+
+from pydantic import BaseModel, ConfigDict
+
+
+class SystemConfiguration(BaseModel):
```
DominicOram, line 7: Should: As above I think following through with removing `SystemAspectBaseParser` will also remove this

src/dodal/devices/beamlines/i19/transmission/spec_from_config/material_absorption_spectrum_spec.py
```diff
@@ -0,0 +1,127 @@
...
+        - The parameters of a fitted absorption curve.
+
+    Attributes:
+        valid_energies: Specifies x-ray energy *domain* valid for the fitted curve.
+        fit_parameters: Specifies parameters needed to calculate absorption values on the fitted curve.
+
+    *See also, these closely related classes:*
+        **FittedAbsorptionCurveSpec**: Fitted curve specification
+        **EnergyIntervalSpec**: Energy range (interval) specification
+        **ClosedInterval**: General maths class underpinning interval definition
```
DominicOram, line 43: Nit: Not sure why this is referenced here?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/material_absorption_spectrum_spec.py
```diff
@@ -0,0 +1,127 @@
...
+        if isinstance(data, list):
+            return {"absorption_curves": data}
+        if isinstance(data, dict):
+            return data
+        raise ValueError(
+            "Absorption spectrum data should be a list or dict of fitted curves."
+        )
+
+
+class MaterialAbsorptionSpectralConfig(
```
DominicOram, line 82: Must: See general comment on `SystemAspectBaseParser`

src/dodal/devices/beamlines/i19/transmission/spec_from_config/name_validation.py
```diff
@@ -0,0 +1,109 @@
+import re
+from functools import cached_property
+from re import Pattern
+from typing import Final
+
+
+class NameValidator:
```
DominicOram, line 7: Could: This seems a bit overkill given that pydantic can just match the regexes on the fields for you but I understand this gives better error messages

src/dodal/devices/beamlines/i19/transmission/spec_from_config/name_validation.py
```diff
@@ -0,0 +1,109 @@
...
+        """Ensures motor axis names follow axis naming rules.
+
+        Note:
+            The axis name must start either with a letter, or an underscore followed by a letter,
+            and subsequently optional characters can only be numbers or letters:  Total length up to nine characters.
+        """
+        cls._VALIDATOR.validate_name(name_to_check=axis_name)
+
+
+class MaterialNameValidation:
```
DominicOram, line 54: Could: Why does it matter what someone wants to name their material?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wedges_spec.py
```diff
@@ -0,0 +1,116 @@
...
+class WedgeSpec(BaseModel):
+    """The specification for a wedge motor (as specified by configuration, typically **JSON**).
+
+    Notes:
+        - Each absorber wedge has a motor to drive it sideways across the x-ray beam:
+            - expected sideways motions are pure horizontal or pure vertical
+                - but that's a detail, azimuthal orientation around the beam should not matter.
+            - The wedge motor scale has
+                - an **out** position
+                - and then a range of "active" positions, ( where the mathematical assumption of a linear taper is reasonable ).
```
DominicOram, line 78: Should: None of this is actually matched by anything in the Spec so it's not clear why it's here and I need to think about it at this point?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wheels_spec.py
```diff
@@ -0,0 +1,71 @@
...
+    Notes:
+        - Each filter wheel has a motor to rotate the wheel into and out of the x-ray beam:
+
+    Attributes:
+        foils: Occupancy dict mapping slot numbers to foils.
+        out: The slot number to use for removing all absorbers from the x-ray beam.
+        permissions: Those slot numbers which are permitted to be in use.
+    """
+
+    foils: dict[str, FoilSpec]
```
DominicOram, line 32: Nit: These are always numbers, right?
```suggestion
    foils: dict[int, FoilSpec]
```

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wheels_spec.py
```diff
@@ -0,0 +1,71 @@
...
+    @model_validator(mode="after")
+    def validate_out_in_permissions(self) -> "WheelSpec":
+        """Ensures the 'out' slot position is included in the 'permissions' list."""
+        if self.out in self.permissions:
+            return self
+        _msg: str = f"The permitted wheel slot indices {self.permissions} must include the 'out' slot position ({self.out} - but do not!"
+        raise ValueError(_msg)
+
+
+class WheelsConfig(SystemAspectBaseParser[WheelSpec]):
```
DominicOram, line 48: Must: As above re `SystemAspectBaseParser`

DominicOram (commented):
Great, I think the main question is around why we have `SystemAspectBaseParser` when we could use standard pydantic tools instead?

Additionally, are the wedges mounted on the `lateral_motors`? In which case why are they separate things in the json structure? Why not have something like:
```
  "wedges": {
    "y": {
      "material" : "aluminium",
      "geometry": {
        "taper_cotangent" : 9.3985,
        "tip" : 5.06,
        "voids": []
        "motion": {
          "units": "mm",
          "out" : 5.0,
          "threshold": 8.9,
          "max": 98.0,
          "tolerance": 5.0e-3
        }
      }
    },
}
```
This would remove a bunch of edge cases about one not being defined when the other is or the names not matching

I haven't looked at the tests yet, I think once we resolve the discussion on `SystemAspectBaseParser` they might change a bit anyway.

DominicOram (changes requested):
Sorry, meant the above as a request changes, not a comment

src/dodal/devices/beamlines/i19/transmission/spec_from_config/material_absorption_spectrum_spec.py
CoePaul: ClosedInterval [ CI ] is what the EnergyIntervalSpec [ EIS ] will metamorph into / be backed by.
The role of CI in context would not be at all obvious - whereas EIS is a clearer label.
EIS is a CI in all but name - so go read CI to understand the do-s and don't-s.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/name_validation.py
CoePaul: exactly

src/dodal/devices/beamlines/i19/transmission/spec_from_config/name_validation.py
CoePaul: It doesn't - but there's every advantage in picking up typos from edits to JSON.
You can't catch them all - but does that really mean You should catch no categories at all?

CoePaul: I just don't want material 1234_iron. - Why would I allow ints at the start?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/system_aspect_base_parser.py
CoePaul: The aim of having each child type is because the sub-dict key name in the JSON matches the field name of the corresponding child class.

I don't want a giant dict of everything 
for the next stage (e.g. building a wedge geometry model thing )
I want the sub-dict 

otherwise everything has to learn how to digest the whole ( JSON ) 
and is then brittle to changes in any part of the whole, even when that changes is irrelevant.

Think of this a compartmentalised factory of model T fords

I don't want the tyres dept to care what the steering wheel looks like.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/system_configuration.py
CoePaul: I think You've basically removed the entire point of the exercise if You do that - 
why not just reject the entire PR in that case.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wedges_spec.py
CoePaul: Notes are "You can read this - but You don't have to"

that's why it's a note.
I depends whether the poor developer reading this is trying to understand the whole system or not.

Image

CoePaul: Looking at it from the "I understand the transmission system" end of the telescope - the reader WILL be wondering where's the "move to the absorber OUT the way" gubbins?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wheels_spec.py
CoePaul: JSON keys are only strings - I didn't invent JSON

Image
Image

src/dodal/devices/beamlines/i19/transmission/spec_from_config/energy_interval_spec.py
CoePaul: I didn't know You could enforce a choice of two strings by means of a Literal list...

src/dodal/devices/beamlines/i19/transmission/spec_from_config/energy_interval_spec.py
CoePaul: yeah maybe - but since I already have a validator with custom explanations - why leave half the unhappy path to go unlabelled / poorly explained?

If it shaved off the entire validator that would be different

src/dodal/devices/beamlines/i19/transmission/spec_from_config/fitted_absorption_curve_spec.py
CoePaul: Here it probably makes sense

src/dodal/devices/beamlines/i19/transmission/spec_from_config/fitted_absorption_curve_spec.py
CoePaul: a) I wasn't aware Field would work with constants - although it seems reasonable that it can

b) there's no good name for the bounds as they are arbitrary and probably too wide and too "3-centric" 
I just don't want +2.67 to be accepted 
and accepting -4.2 would imply accepting a typo 

but yes it could be better explained - how it is somewhat vague and that vagueness is reflected in the code

src/dodal/devices/beamlines/i19/transmission/spec_from_config/fitted_absorption_curve_spec.py
CoePaul: I'll look again - I'm sure bits of it made sense as I was tweaking it

src/dodal/devices/beamlines/i19/transmission/spec_from_config/foil_spec.py
CoePaul: the conversions already committed to general_maths module 
have set the ground work for which units are accepted and which are not 
( they - after all - came in the furniture removal van from my Jython partial implementation of i19 attenuation )

this code here is "just" dovetailing to that existing convertor which will do cm to unit and and unit to cm 
for a few of the likely units 

I don't think I want to permit people to specify a foil thickness in furlongs, (nor barleycorns) even if it is a valid length unit

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: because there's no reason why anything would stop working ( in the mechanical / system side ) 
if I had motors moving along diagonal lines like the arms of a capital X 

it's "just" that right now I've been lazy and only coded options for the present hardware as it is 

if the code were really well written I would have permitted anything that's physically possible 
to equally simple to do in code / json 

but I was lazy and went with the "we aren't using it now - and the pain of doing it later will be left as expensive" and leaning heavily on the "nobody is going to do it differently or need diagonal motion"

but I should - at the very least - comment that I thought about it

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: You have two definitions of "along the wedge"

You have a wedge centric frame which exists for a "platonic" wedge floating in a universe where only the wedge exists.  It can define an axis ( let's call it lambda ) and count from lambda = 0 at the tip ( say )
to slices further along the wedge as it thickens - going "up" the taper.

Then You have a motor axis - which is just controls team in epics and the motor encode 
agreeing on how they define zero. ( and might be labelled something like x )

In order for the wedge / motor duo to become a useful entity - somebody somewhere has to nail the relative offset of the lambda scale and the x-scale. 

Note
( All this silently ignores the fact that there's no guarantee the lambda and x axes align in space 
and don't in fact have a small angle - which could be anywhere on a cone of misalignment ...

... Fortunately to some extent that potential conical "misalignment error" is fudged away by the calibration measurements made when the motor is moved and readings of absorption are made 

but it's one thing to add a factor in and then cancel it ( which further assumes nothing moves, 
ie the cone doesn't sag, flop, drift nor roll around since You last calibrated )

and another to have never had two almost cancelling aspects in a hypothetical system with zero 
such degrees of freedom.

CoePaul: It is all very complicated - because it's real hardware - there's no carpet to bury it under 
just because a poor dev doesn't wanna know about it - much better IMHO to rub the developers face in all of it and let them slowly recover - than to pat them on the head and tell them ( inaccurately ) there's nothing to worry about.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: yerrrssss

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: That does help - I didn't know it worked like that ( on the pydantic side )
but I thought I could let people type nothing at all as well ( and have mm filled in for them )

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: I know - but in all the beamline side the scientists already call it "out" 
The JSON starts on their page ( and all our fields here are in their language ) 
and then eventually we can - having absorbed all the values ) 
"transliterate" names in deeper parts of the code base - should we so choose

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: guilty m'laud

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: yes but having created the bloody interval thing I'm going to use it

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: It was an f string when I was reporting back max and out - but then I stopped doing that 
it can be literal yes as you suggest

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: You have a point

src/dodal/devices/beamlines/i19/transmission/spec_from_config/system_aspect_base_parser.py
CoePaul: The parent class removes a load of boilerplate that the children were repeating four times over.

Getting to have instances of the specific child types ( starting from JSON ) 
is what all this code is supposed to be doing. 

I don't want a giant dict which is just "what the JSON said" - otherwise I could just use json.loads 
and be done

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wedges_spec.py
CoePaul: Every attenuating element has a "range" of positions which are absorbing 
( sometimes - ie for a foil, range = 1 place only )
and a singular location for OUT 
( in principle - even if not in practice - lateral motor +/- margin of readout error is roughly a singular location )

OUT is an almost canonical addendum to the "range" of permitted positions for an absorber. 

It's a very inhomogenous logical smush together of range + OUT but I couldn't see any way round it

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wheels_spec.py
CoePaul: so no - they are never integers and always "integers"

src/dodal/devices/beamlines/i19/transmission/spec_from_config/fitted_absorption_curve_spec.py
DominicOram: They could just be named something like `REALISTIC_PHYSICAL_LOWER_BOUND`?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/foil_spec.py
DominicOram: Sure, then maybe we should import `SupportedThicknessUnits` and use that here?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/energy_interval_spec.py
DominicOram: Sure, happy to leave as is

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
DominicOram: I agree that there's no point generalising it to diagonal motion. Happy to leave the comment in if you feel strongly

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
DominicOram: Thanks, I understand and agree that it's worth a comment but I think the phrasing is ambiguous, `Nailed down` doesn't actually tell you how the two co-ordinate systems relate. Do you mean `The position of zero on the motor is assumed to be configured to match the position of zero on the wedge`? In which case saying that seems more accurate

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
DominicOram: I believe this does both i.e. you must either specify `"mm"` or leave it out, in which case it defaults to `"mm"`. Though a test on that would be good

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
DominicOram: If the names in the JSON are fixed I believe you can use aliases to convert them on load e.g.
```suggestion
    out_position: StrictFloat = Field(alias="out")
```
Will load in `out` and put it in `out_position`. Though I appreciate this has the potential to confuse if comparing the JSON with the code. Feel free to take it or leave it.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
DominicOram: I guess maybe that's an argument that shouldn't have it at all then. But I don't want to rehash things that are out of scope of the PR. In this specific instance the above is more readable and more pythonic.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/system_aspect_base_parser.py
DominicOram: I agree you don't want to pass the whole structure around, but you don't need to, you could just pull out e.g. `spec.lateral_motors` and pass that to the thing that needs it. The general model of how pydantic is expected to work is you do ingest and validation once as early as possible and now you have a nice data structure that you can portion out to things later on if you wish.

>I don't want a giant dict which is just "what the JSON said" - otherwise I could just use json.loads
and be done

But that's basically what pydantic is, it's a `json.loads` that converts into nice well typed data that you can now assume is correctly validated when you use it in the rest of the codebase.

src/dodal/devices/beamlines/i19/transmission/spec_from_config/material_absorption_spectrum_spec.py
DominicOram: Sure, happy to leave in

src/dodal/devices/beamlines/i19/transmission/spec_from_config/name_validation.py
DominicOram: Ok, can we at least add a comment to the affect that it's actually just convention/preference? Otherwise I worry that someone will ask for it in 5 years for a legitimate reason and we'll be worried we can't change it as it might break something

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wedges_spec.py
DominicOram: >Looking at it from the "I understand the transmission system" end of the telescope - the reader WILL be wondering where's the "move to the absorber OUT the way" gubbins?

Yes, exactly and they're not here. Ok, so happy to leave the description here if we have a note to `see LateralMotorSpec`?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/wheels_spec.py
DominicOram: Sure, the JSON dict key must be a string but the python dict key doesn't have to be so you could do the conversion here to enforce it to an integer and stop the user putting something in that isn't an integer

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
DominicOram: From discussion, I've realised that `tip` is where this offset lays. Can we instead change the comment to say that?

src/dodal/devices/beamlines/i19/transmission/spec_from_config/lateral_motor_spec.py
CoePaul: another source of confusion is that wheel would call its "out_index" 
( and later in the business logic out_position and out_index will be doing the same job ).

PR #2177, opened 2026-08-14
Add constant for filepath

jacob-williamson:
Needed for https://github.com/DiamondLightSource/crystallography-bluesky/pull/124

1. Check path is correct

### Checks for reviewer


src/dodal/beamlines/i15_1.py
```diff
@@ -42,6 +42,9 @@
 XPDF_CRYSTAL_LUT_FILEPATH = (
     "/dls_sw/i15-1/software/daq_configuration/xpdf_crystal_lut.txt"
 )
+TTH_ANGLE_TO_COLLECTION_TIME_FILEPATH = (
+    "/dls_sw/i15-1/software/daq_configuration/tth_angle_to_collection_time.txt"
+)
```
oliwenmandiamond, line 47: This is not used in dodal, so why is it needed in dodal layer rather than plan layer?

jacob-williamson: It just means all config file paths are in the same place, will move it to the plan though

